PR #2, opened 2018-09-01
Add mautrix-telegram to bridge with telegram services

izissise:
It add an option `matrix_mautrix_telegram_enabled` to install [mautrix-telegram](https://github.com/tulir/mautrix-telegram) a telegram bridge


roles/matrix-server/tasks/setup_synapse_ext_mautrix_telegram.yml
```diff
@@ -0,0 +1,65 @@
...
+      {{ ["{{ matrix_synapse_app_service_config_file_mautrix_telegram }}"] | to_nice_json  }}
+  when: "matrix_mautrix_telegram_enabled"
+
+#
+# Tasks related to getting rid of matrix-mautrix-telegram (if it was previously enabled)
+#
+
+- name: Ensure matrix-mautrix-telegram.service doesn't exist
+  file:
+    path: "{{ role_path }}/templates/systemd/matrix-mautrix-telegram.service.j2"
```
spantaleev: You probably meant `path: "/etc/systemd/system/matrix-mautrix-telegram.service"`?

examples/host-vars.yml
```diff
@@ -24,4 +24,11 @@ matrix_coturn_turn_static_auth_secret: ""
 
 # A secret used to protect access keys issued by the server.
 # You can put any string here, but generating a strong one is preferred (e.g. `pwgen -s 64 1`).
-matrix_synapse_macaroon_secret_key: ""
\ No newline at end of file
+matrix_synapse_macaroon_secret_key: ""
+
+# Mautrix telegram
```
spantaleev: My goal was to have the `examples/host-vars.yml` file contain only the absolutely required configuration for having the playbook run.

Most other additional services and extensions require additional configuration, which can be discovered from the role's `defaults/main.yml` file or, even better, from the the docs.

Would you get rid of these here and create some new documentation file for this bridge integration in `docs/` (possibly `docs/configuring-playbook-bridge-mautrix-telegram.md`?
We can then link to that file from within `docs/configuring-playbook.md`.

roles/matrix-server/defaults/main.yml
```diff
@@ -74,6 +74,11 @@ matrix_synapse_container_additional_volumes: []
 # Contains definition objects like this: `{"name": "..", "level": "DEBUG"}
 matrix_synapse_additional_loggers: []
 
+# A list of service config files
+# This list gets populated dynamically based on Synapse extensions that have been enabled.
+# Conatains fs paths
```
spantaleev: Small typo ;)

roles/matrix-server/defaults/main.yml
```diff
@@ -203,6 +210,10 @@ matrix_riot_web_enabled: true
 matrix_riot_web_default_identity_server_url: "https://{{ matrix_synapse_trusted_third_party_id_servers[0] }}"
 
 
+# Matrix mautrix is a Matrix <-> Telegram bridge
+matrix_mautrix_telegram_enabled: false
```
spantaleev: Maybe we should add some default values for `matrix_mautrix_telegram_api_id` and `matrix_mautrix_telegram_api_hash` here (like the ones you have in `examples/host-vars.yml` now).

These defaults won't be useful unless changed, but it's a good way to document all variables that this integration requires in a single place, making it discoverable (in addition to being discoverable by reading the docs file).

This way, if someone runs the playbook without redefining the variables, at least it won't explode with errors.

Right now, this would cause trouble for existing users of the playbook, unless they merge `examples/host-vars.yml` with what they already have. Most users should not need to care about defining any ``matrix_mautrix_telegram_*` variables, unless they want to use that integration.

examples/host-vars.yml
izissise: I'll write some doc :)

izissise: Don't hesitate to tell me, if it is too light

roles/matrix-server/tasks/setup_synapse_ext_mautrix_telegram.yml
```diff
@@ -60,6 +63,6 @@
 
 - name: Ensure matrix-mautrix-telegram.service doesn't exist
   file:
-    path: "{{ role_path }}/templates/systemd/matrix-mautrix-telegram.service.j2"
+    path: "{{ role_path }}/templates/systemd/matrix-mautrix-telegram.service"
```
spantaleev: I think this is still wrong.
Isn't it supposed to delete the remote file on the server, and not the local template file from the playbook?

docs/configuring-playbook-bridge-mautrix-telegram.md
```diff
@@ -0,0 +1,5 @@
+# Setting up Mautrix Telegram
+
+The playbook can install and configure [mautrix-telegram](https://github.com/tulir/mautrix-telegram) for you.
+
+See that project's [documentation](https://github.com/tulir/mautrix-telegram/wiki#usage) to learn what it does and why it might be useful to you.
```
spantaleev: That's nice!

Maybe it can also have some snippet with what variables need to be defined to get the playbook to install the bridge?

Like:

> To get it installed, you'd need to obtain API keys from https://my.telegram.org/apps and then use the following playbook configuration:
> 
> ```yaml
> matrix_mautrix_telegram_enabled: true
> matrix_mautrix_telegram_api_id: YOUR_TELEGRAM_APP_ID
> matrix_mautrix_telegram_api_hash: YOUR_TELEGRAM_API_HASH
> ```

spantaleev:
Also, because of this:
```yaml
- name: Ensure Matrix Mautrix telegram config installed
  ....
  when: "matrix_mautrix_telegram_enabled and mautrix_config_file.stat.exists == False"
```

re-running the playbook may not update the configuration and would rather preserve any manual user edits to it.

It's not that it's bad.. and it's probably the easiest way to allow configuration customizations.. But it may lead to unexpected results sometimes. For example, if we ever introduce a configuration fix of some sort, it won't get applied at all for existing users, because they'd already have a configuration created.

We used to have it be something like that for the `homserver.yaml` configuration before, but have stopped doing it, as it makes it more predictable - re-running the playbook always gives you the same result.

Still, you can keep it like that for mautrix-telegram's configuration if you think that's best for it.
Just wanted to make sure it's done intentionally.

izissise:
The problem is in the config.yml there two variables:
```
  as_token: "This value is generated when generating the registration"
  hs_token: "This value is generated when generating the registration"
```
are generated the first time and are share with registration.yml file.
It would be possible to generate the `registration.yml` from ansible and populate {a,h}s_token accordingly.

spantaleev:
I see.. It's a little complicated then. Maybe we should leave it as it is, unless you can think of some other way to create those secrets beforehand and put them in the configuration.

Would you fix the remaining issue with the `Ensure matrix-mautrix-telegram.service doesn't exist` task? It still tries to delete something from the role directory and not the actual file on the server.

roles/matrix-server/tasks/setup_synapse_ext_mautrix_telegram.yml
izissise: Oops I went to fast there 💃

izissise: Actually I copied this line from `setup_corporal.yml`, maybe it should be changed there too :)

spantaleev: Thanks for letting me know! I've fixed `setup_corporal.yml`.

Unless you can think of something else, we can probably merge this pull request when this file removal thing gets fixed.

izissise: Fixed

spantaleev:
Merged! Thank you @izissise! :+1: 

This playbook improvement is also about to be mentioned in the next [This Week in Matrix](https://matrix.org/blog/category/general/this-week-in-matrix/) blog post. I'm guessing a lot of people would appreciate your work! Hopefully, this also paves the way for adding support for more bridges to the playbook.

izissise:
Awesome! thanks you! I wouldn't add got into matrix if it wasn't for your really well documented playbook!

PR #14, opened 2018-10-18
Fix link to using your own webserver

aaronraimist:


spantaleev:
Thank you, @aaronraimist! :+1:

PR #15, opened 2018-10-19
Add missing tag to task

anadahz:
The setup-all should include the setup_well_known task as well


spantaleev:
Thank you for finding and fixing this! :+1:

PR #17, opened 2018-10-21
Add DNS records table

anadahz:


spantaleev:
I've changed it to `CNAME` in d5655b75273.

PR #19, opened 2018-10-21
Expose mautrix-telegram public endpoint through nginx

izissise:
It used to allow user to logging with their own account to the bot
see: https://github.com/tulir/mautrix-telegram/wiki/Authentication#replacing-telegram-accounts-matrix-puppet-with-matrix-account for more informations

The endpoint is an uuid hash generated from the matrix_synapse_macaroon_secret_key so it is not easily guessable by a third party


spantaleev:
I've merged this. Thank you! :+1: 

Do you think we should update `docs/configuring-playbook-bridge-mautrix-telegram.md` as well, with some additional usage details (including information about this public endpoint)?

PR #20, opened 2018-10-24
Mautrix whatsapp

izissise:
Add mautrix-whatsapp

Tested for two days, and it works well :)


spantaleev:
Wow, that's great to see!

Let's wait for the mautrix-whatsapp PR to get merged and let's merge this!

By the way, recently I've been thinking about restructuring the default variables file (`defaults/main.yml`).
As we add more features, it's starting to get longer and some related variables are too far apart or prefixed differently.

For example, the currently-named `matrix_docker_image_mautrix_whatsapp` would probably be better named as `matrix_whatsapp_docker_image` (or something like that), bringing it all under the common (`matrix_whatsapp_` prefix).. As well as bringing all these variables closer together in the file.

I know we have to redo this for many other things as well, so that will be some future work (no need to water-down this PR's patch-set with such stuff).

izissise:
@rubo77 I don't know, that's the name of the project :D

I agree on the variable renaming 👍

izissise:
Good for merge !

spantaleev:
Merged and added a documentation link! Thanks @izissise! :+1:

PR #21, opened 2018-10-30
Allow specifying which parts of the role run from playbooks

Cadair:
I am trying to use this as a role from my own playbook. This is the simplest (and most minimal) change I could think of to allow me to select which bits I want to run.

### DIFF
--- a/roles/matrix-server/defaults/main.yml
+++ b/roles/matrix-server/defaults/main.yml
@@ -284,3 +284,12 @@ matrix_ssl_use_staging: false
 
 # Specifies when to attempt to retrieve new SSL certificates from Let's Encrypt.
 matrix_ssl_renew_cron_time_definition: "15 4 */5 * *"
+
+# Variables to Control which parts of the role run.
+run_setup: true
+run_upgrade_postgres: true
+run_start: true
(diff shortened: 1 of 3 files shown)


aaronraimist:
Can’t you already do this by running something like `ansible-playbook -i inventory/hosts setup.yml --tags=self-check`?

Cadair:
I am trying to use this from an existing playbook which I don't want to have to specify tags from the command line. To the best of my knowledge you can't say "run these tags in this role" from inside a playbook?

spantaleev:
I guess with this we can have the best of both worlds:
- cloning the project gives you a full playbook you can easily run
- yet people with more advanced scenarios can use just the role part of it, so it's easy to plug into their own other playbooks.

Thank you for figuring out a nice way to do it, and for implementing it! :+1: 

It would probably be nice to document this to tell others how to use it.

Maybe a new document (`docs/configuring-playbook-use-in-another-role.md`) and a link to that from **Other configuration options** in `docs/configuring-playbook.md`?

I guess we can then also update the main `README.md` to mention that the playbook's role is easily reusable in other playbooks.

PR #27, opened 2018-11-02
Disable TLS 1.0

aaronraimist:
Feel free to close if you think TLS 1.0 comparability is important but this is how I have my server setup. (Actually on mine I also disabled 1.1)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# 2018-11-03
+
+## SSL protocols used to serve Riot and Synapse
+
+There's now a new `matrix_nginx_proxy_ssl_protocols` playbook variable, which controls the SSL protocols used to serve Riot and Synapse. It's default value is `TLSv1.1 TLSv1.2`. This playbook previously used `TLSv1 TLSv1.1 TLSv1.2` to serve Riot and Synapse.
+
+You may wish to reenable TLSv1 if you need to access Riot in older browsers.
+
+Note: Currently the dockerized nginx doesn't support TLSv1.3. See https://github.com/nginxinc/docker-nginx/issues/190 for more details.
+
+
 # 2018-11-01
 
 ## Postgres 11 support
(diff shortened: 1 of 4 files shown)


spantaleev:
I also wanted to enable TLS v1.3 some couple of weeks ago and have found the following problems:

- the nginx Docker image does not have OpenSSL 1.1.1, so TLS v1.3 doesn't work. It's tracked in  nginxinc/docker-nginx#190. Enabling TLS v1.3 in the config doesn't lead to any errors on a new nginx version (like the one we use via the Docker image), but it nevertheless doesn't affect anything (it doesn't really get enabled in the end).

- people who disable the nginx proxy installed by this playbook (`matrix_nginx_proxy_enabled: false`), are likely to be reusing these nginx configuration files (we generate them a little differently if that option is toggled, and then they work outside of a container setup). Such people likely run nginx outside of Docker and install it via their OS's package manager. For this reason, it's likely an old nginx version, which doesn't know what `TLS v1.3`, leading to the following nginx error: `nginx: [emerg] invalid value "TLSv1.3" in /matrix/nginx-proxy/conf.d/matrix-synapse.conf:30`

So I think:
- we'd better let the protocol list be configurable (a new `matrix_nginx_proxy_ssl_protocols` variable, which defaults to something like `TLSv1.1 TLSv1.2 TLSv1.3` or whatever). This way, people like you can disable TLSv1.1 as well, and people like me (using an external old nginx) can disable the problematic TLSv1.3
- we should document this in `docs/configuring-playbook-own-webserver.md`. It's most likely folks who are "using their own webserver" that care about it, so it makes sense that we put it in there.. People who use the integrated nginx proxy are probably OK with going with whatever (sane) defaults we have for them.
- there's probably not much point in rushing to do this, because, as mentioned before, TLS v1.3 will not really kick-in, even if enabled.

spantaleev:
Thank you! :+1: 

It's nice that it's configurable now!
I was just hoping that it's not too early to turn off TLSv1 by default, but let's give it a try. At least people can change it manually, if necessary.

aaronraimist:
As far as I know the only browsers that still get used a bit and dont support TLSv1 are like IE versions older than 10. I dont even think Riot works in IE 11 so I dont think its going to affect anyone.

spantaleev:
True! Looks like Firefox on Windows XP is fine, so that's good as well.
Some old Android I found could also use the Riot app, so maybe nothing would break for no one.

PR #38, opened 2018-11-18
Add more configured options for riot-web and homeserver.

anadahz:


examples/host-vars.yml
```diff
@@ -26,3 +26,15 @@ matrix_coturn_turn_static_auth_secret: ""
 # You can put any string here, but generating a strong one is preferred (e.g. `pwgen -s 64 1`).
 matrix_synapse_macaroon_secret_key: ""
 
+# Set autojoin room(s)
```
spantaleev: All default variables are meant to be specified in `roles/matrix-server/defaults/main.yml`.

We have `examples/host-vars.yml` to serve as a bare-minimum example of what people **must** define in order to get this playbook working. Besides that, people can override any of the many defaults from `roles/matrix-server/defaults/main.yml`.

So all of these ought to be moved to  `roles/matrix-server/defaults/main.yml`.
Not only that, but they'd better have sensible default values, so that the majority of people would not have to change them. Keeping backward compatibility is nice too. Not that we couldn't break it, if it makes sense.

I guess `riot_roomdir_servers` can be defined as `['matrix.org']` to preserve compatibility with what we have now and to provide a nice default.

`matrix_autojoin_rooms` can be defined as an empty list (`[]`) for the same reason.

roles/matrix-server/templates/riot-web/config.json.j2
```diff
@@ -10,10 +10,24 @@
     "integrations_jitsi_widget_url": "{{ matrix_riot_web_integrations_jitsi_widget_url }}",
     "bug_report_endpoint_url": "https://riot.im/bugreports/submit",
     "enableLabs": true,
+{% if riot_roomdir_servers is defined %}
```
spantaleev: If we provide nice defaults in `roles/matrix-server/defaults/main.yml` as discussed, we wouldn't need to have an `is defined` check at all, which will make the code nicer and cleaner

roles/matrix-server/templates/riot-web/config.json.j2
```diff
@@ -10,10 +10,24 @@
...
+{% for server in riot_roomdir_servers %}
+    "roomDirectory": {
+        "servers": [
+            "{{ server }}"
+        ]
+    },
+{% endfor %}
+{% else %}
```
spantaleev: We won't need such `else` clauses if we had defined `riot_roomdir_servers` with a nice default of `['matrix.org']`.

roles/matrix-server/templates/riot-web/config.json.j2
```diff
@@ -10,10 +10,24 @@
...
     "roomDirectory": {
         "servers": [
             "matrix.org"
         ]
+{% endif %}
+{% if riot_welcome_bot|default(true)|bool %}
```
spantaleev: This could also be simplified as `{% if riot_welcome_bot %}` if we have a `riot_welcome_bot` default variable specified in `roles/matrix-server/defaults/main.yml`.

By the way, naming that `matrix_riot_web_welcome_user_id` would be preferable, because:
- it groups it into the `matrix_riot_web_` "namespace", along with all other riot-web configuration
- it follows the actual riot-web configuration variable name more closely (`welcomeUserId`)

roles/matrix-server/templates/synapse/homeserver.yaml.j2
```diff
@@ -432,7 +432,11 @@ turn_allow_guests: False
 ## Registration ##
 
 # Enable registration for new users.
+{% if matrix_enable_reg|default(false)|bool %}
```
spantaleev: We don't need an `if` here
We can use something like this:
```yaml
enable_registration: {{ matrix_synapse_enable_registration|to_yaml }}
```

roles/matrix-server/templates/synapse/homeserver.yaml.j2
```diff
@@ -476,9 +480,16 @@ trusted_third_party_id_servers:
 
 # Users who register on this homeserver will automatically be joined
 # to these rooms
+{% if matrix_autojoin_rooms is defined %}
```
spantaleev: Same here. With a sensible default (empty list), we can skip the `is defined` check and just do something like:

```yaml
{% if matrix_autojoin_rooms|length > 0 %}
auto_join_rooms: {{ matrix_autojoin_rooms|to_yaml }}
```

Well, it's similar..

anadahz:
@spantaleev thanks for your comments!
I created a fix for this PR here: https://github.com/anadahz/matrix-docker-ansible-deploy/tree/fix/add/config
Please let me know if there is anything missing.

anadahz:
Thanks for reviewing the patch @spantaleev .

> @anadahz, normally you can keep adding commits and pushing to the same branch that you opened this PR with.

I was not waiting for the PR to be merged (I should have mentioned it in the discussion) and that's why I created another branch to show you my idea, in any case now it's merged and you have fixed/updated portions of my patch, thank you!

> One thing I'm wondering is, whether matrix_riot_web_welcome_user_id should default to @riot-bot:matrix.org or not. It's probably you that I've discussed it with in the #matrix-docker-ansible-deploy:devture.com room, but we found out that the bot won't respond to users on other homeservers for some reason..

> So maybe it's better if we actually disable it (setting the matrix_riot_web_welcome_user_id default to null (~ in YAML)).

I agree the riot bot should be disabled by default or at least we should find out how we can enable it and use it before enabling it by default. This was one of the reasons this patch had some inconsistent variables as I was using a number of non default variables in my host_vars but still I should have tested it better ;)

PR #46, opened 2018-11-28
Fix caddy .well-known config

aaronraimist:
I assume you'll also need to change this for nginx and apache. Without this, Riot (support has only landed on [riot.im/develop](https://riot.im/develop) so far) will tell you "Cannot find homeserver".

### DIFF
--- a/docs/configuring-well-known.md
+++ b/docs/configuring-well-known.md
@@ -71,6 +71,7 @@ server {
 
 ```caddy
 proxy /.well-known/matrix https://matrix.DOMAIN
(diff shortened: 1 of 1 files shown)


spantaleev:
Is there a client which makes use well-known now?

Perhaps riot-web, judging by this.

spantaleev:
Thanks for bringing this up!

I believe 9dad4c7c2dfae068 is a better way to go about it.
Can you verify that it works for you? We can then close this, I guess.

aaronraimist:
Yep that's a better solution, works great. Thanks!

PR #47, opened 2018-11-30
Set default mautrix-telegram image to 0.4.0

izissise:
tulir has released a new mautrix-telegram version !

Tested locally and it works correctly


spantaleev:
Thanks for keeping us up to date! :bowing_man:

PR #48, opened 2018-12-01
Comment that secrets mustn't contain quotes/backticks

haslersn:
Putting a backtick into the variable `matrix_synapse_macaroon_secret_key`
leads to an error when starting `matrix-synapse.service`, so the service
can't start. Error Message:

```
[...]
[...] docker[19227]: yaml.scanner.ScannerError: while scanning a double-quoted scalar
[...] docker[19227]:   in /data/homeserver.yaml, line 451, column 29
[...] docker[19227]: found unknown escape character '`'
[...] docker[19227]:   in /data/homeserver.yaml, line 451, column 37
[...] docker[19227]: Traceback (most recent call last):
[...] docker[19227]:   File /start.py, line 66, in <module>
[...]
```

Therefore, added comments that the secrets mustn't contain quotes or
backticks.

### DIFF
--- a/examples/host-vars.yml
+++ b/examples/host-vars.yml
@@ -20,8 +20,10 @@ host_specific_hostname_identity: YOUR_BARE_DOMAIN_NAME_HERE
 
 # A shared secret (between Synapse and Coturn) used for authentication.
 # You can put any string here, but generating a strong one is preferred (e.g. `pwgen -s 64 1`).
+# It must NOT contain any quotes or backticks.
 matrix_coturn_turn_static_auth_secret: ""
 
 # A secret used to protect access keys issued by the server.
 # You can put any string here, but generating a strong one is preferred (e.g. `pwgen -s 64 1`).
+# It must NOT contain any quotes or backticks.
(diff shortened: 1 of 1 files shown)


spantaleev:
I think it's rather a problem of how we populate this secret inside of `homeserver.yaml` (see `roles/matrix-server/templates/synapse/homeserver.yaml.j2`):

```jinja
macaroon_secret_key: "{{ matrix_synapse_macaroon_secret_key }}"
```

Perhaps if that were something like this, quotes would not be a problem:

```jinja
macaroon_secret_key: {{ matrix_synapse_macaroon_secret_key|to_yaml }}
```

Besides `macaroon_secret_key` and `turn_shared_secret`, other things are also affected by the same problem, I think. Things like `registration_shared_secret` and `form_secret`, the way we specify `matrix_synapse_password_config_pepper`, the way we specify `matrix_postgres_connection_password`.

I'm on the go now, so I can't make and test the correct adjustments though. I can probably make a fix on Monday, unless you want to do it earlier - by doing/testing it and making a pull request.

PR #49, opened 2018-12-01
Fix some bugs due to variables containing quotes/backticks

haslersn:
**I only tested this by verifying that `matrix-synapse.service` starts successfully using the default configuration.**

Put variables that can contain arbitrary strings into a block scalar. This way,
no escaping of quotes or backticks is needed.

This fixes some bugs. For example, putting a backtick into the variable
`matrix_synapse_macaroon_secret_key` previously led to an error when starting
`matrix-synapse.service`, so the service couldn't start. The error message was:

```
[...]
[...] docker[19227]: yaml.scanner.ScannerError: while scanning a double-quoted scalar
[...] docker[19227]:   in /data/homeserver.yaml, line 451, column 29
[...] docker[19227]: found unknown escape character '`'
[...] docker[19227]:   in /data/homeserver.yaml, line 451, column 37
[...] docker[19227]: Traceback (most recent call last):
[...] docker[19227]:   File /start.py, line 66, in <module>
[...]
```

### DIFF
--- a/roles/matrix-server/templates/synapse/homeserver.yaml.j2
+++ b/roles/matrix-server/templates/synapse/homeserver.yaml.j2
@@ -216,7 +216,8 @@ database:
   name: "psycopg2"
   args:
     user: "{{ matrix_postgres_connection_username }}"
-    password: "{{ matrix_postgres_connection_password }}"
+    password: >-
+      {{ matrix_postgres_connection_password }}
     database: "{{ matrix_postgres_db_name }}"
     host: "{{ matrix_postgres_connection_hostname }}"
     cp_min: 5
@@ -411,7 +412,8 @@ recaptcha_siteverify_api: "https://www.google.com/recaptcha/api/siteverify"
 turn_uris: ["turn:{{ hostname_matrix }}:3478?transport=udp", "turn:{{ hostname_matrix }}:3478?transport=tcp"]
 
 # The shared secret used to compute passwords for the TURN server
-turn_shared_secret: "{{ matrix_coturn_turn_static_auth_secret }}"
+turn_shared_secret: >-
+  {{ matrix_coturn_turn_static_auth_secret }}
 
 # The Username and password if the TURN server needs them and
 # does not use a token
@@ -453,7 +455,8 @@ enable_registration: {{ matrix_synapse_enable_registration }}
 
 # If set, allows registration by anyone who also has the shared
 # secret, even if registration is otherwise disabled.
-registration_shared_secret: "{{ matrix_synapse_registration_shared_secret }}"
+registration_shared_secret: >-
+  {{ matrix_synapse_registration_shared_secret }}
 
 # Set the number of bcrypt rounds used to generate password hash.
 # Larger numbers increase the work factor needed to generate the hash.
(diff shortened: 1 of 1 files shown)


spantaleev:
Thanks so much for taking the time and making a fix and pull request! 🙇‍♂️

I don't mean to be ungrateful, but I'm wondering if there is any advantage to using block scalars compared to `to_yaml` when the intention is to put a variable in a YAML in a friendly and safe way?

To me, it would seem more advantageous to use `to_yaml` (well, or `to_nice_yaml` in some other non-scalar instances) for 2 reasons:

- it makes things shorter (one line shorter for each instance) and simpler - the configuration is kept as close as possible to the original `homeserver.yaml` generated by Synapse, which makes it easier to diff with it later to pull in new changes
- using `to_yaml` reveals our intentions of representing something in a proper/safe way in YAML better than (ab)using block strings. Future contibutors can also look at these instances and follow that, instead of thinking "ah, embedding a string in YAML, I can do simpler" and then repeating our initial mistake.

haslersn:
I tried `to_yaml` and it didn't make any difference to doing nothing.

spantaleev:
Have you tried doing `to_yaml` in combination with removing the quotes around the string?

That is to say:

```jinja
macaroon_secret_key: "{{ matrix_synapse_macaroon_secret_key }}"
```

becoming:

```jinja
macaroon_secret_key: {{ matrix_synapse_macaroon_secret_key|to_yaml }}
```

Note how there are no quotes before `{{` and after `}}` in the latter.

haslersn:
That works.

haslersn:
Using `|to_yaml` sometimes places a new line containing `...` after the secret. That's stange.

haslersn:
> I haven't had the chance to test it yet, but looking at the [PyYAML Documentation](https://pyyaml.org/wiki/PyYAMLDocumentation), it looks like we might play with something like `explicit_end=False` (just a guess) and see if it changes anything about that `...` suffix.

I just tried and didn't succeed.

haslersn:
[Ansible doesn't consider this a bug](https://github.com/ansible/ansible/issues/17095), however they don't provide a solution.

spantaleev:
Nice find of an obscure and awful bug!

I guess we'd need to resort to hacks then.. Or using `|to_json`, which we've already used in a few other places. Using `to_json` has always left a bit of a bad taste in my mouth though..

Why would one be passing variables through a JSON filter when trying to generate YAML?
Especially if you've seen `to_yaml` being used elsewhere and are wondering why it's not used here..

Well, [YAML](https://en.wikipedia.org/wiki/YAML) (at least as of 1.2) is supposedly a superset of JSON, so any JSON should also be valid YAML, I guess.
In that case, using `to_json` makes some sense..

I guess if `to_json` works well and we need to choose between using that and doing block string scalars (with their downsides listed above), `to_json` might be the lesser evil.

What do you think?

haslersn:
> Well, [YAML](https://en.wikipedia.org/wiki/YAML) (at least as of 1.2) is supposedly a superset of JSON, so any JSON should also be valid YAML, I guess.

I didn't know about that relationship between JSON and YAML. In case they are compatible regarding their escaping, I'd prefer using `|to_json`.

spantaleev:
I guess they should be compatible. We can give it a try.

PR #51, opened 2018-12-04
Default matrix-synapse-rest-auth endpoint to matrix-mxisd

haslersn:
matrix-synapse-rest-auth is commonly used together with mxisd. The uninformed might either

* not know that `matrix_synapse_ext_password_provider_rest_auth_endpoint` has to be set or
* set it wrongly to `http://localhost:8090` which doesn't work because matrix-synapse doesn't run on the localhost, but inside a docker container.

The correct value for that common use case is `http://matrix-mxisd:8090` and is a better default than the empty string.

### DIFF
--- a/roles/matrix-server/defaults/main.yml
+++ b/roles/matrix-server/defaults/main.yml
@@ -141,7 +141,7 @@ matrix_synapse_password_providers_enabled: false
 # See: https://github.com/kamax-io/matrix-synapse-rest-auth
 matrix_synapse_ext_password_provider_rest_auth_enabled: false
 matrix_synapse_ext_password_provider_rest_auth_download_url: "https://raw.githubusercontent.com/kamax-io/matrix-synapse-rest-auth/v0.1.1/rest_auth_provider.py"
-matrix_synapse_ext_password_provider_rest_auth_endpoint: ""
+matrix_synapse_ext_password_provider_rest_auth_endpoint: http://matrix-mxisd:8090
 matrix_synapse_ext_password_provider_rest_auth_registration_enforce_lowercase: false
 matrix_synapse_ext_password_provider_rest_auth_registration_profile_name_autofill: true
 matrix_synapse_ext_password_provider_rest_auth_login_profile_name_autofill: false


spantaleev:
I'm also a proponent of sensible defaults, but I'm wondering if having a default that points to mxisd is a good idea, because:

- it somewhat suggests that mxisd is configured to handle it by default. This is not true -- mxisd requires custom configuration to be your authentication backend, including the replacement of the default config template `matrix_mxisd_template_config`. It's a non-straightforward thing to set up.

- [docs/configuring-playbook-rest-auth.md](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-rest-auth.md) already mentions `matrix_synapse_ext_password_provider_rest_auth_endpoint` explicitly and suggests that you need to change it

Example from `docs/configuring-playbook-rest-auth.md`:

```yaml
matrix_synapse_ext_password_provider_rest_auth_endpoint: "http://change.me.example.com:12345"
```

Maybe instead, we should:

- keep the default of empty string, because there's no other sensible default we can provide

- keep the mention of the `matrix_synapse_ext_password_provider_rest_auth_endpoint` variable on `docs/configuring-playbook-rest-auth.md`

- add another mention/note about mxisd integration (especially when mxisd runs in a container through the playbook) on `docs/configuring-playbook-rest-auth.md`

- add a check in `setup_synapse_ext_rest_auth.yml`, so that it will fail if `matrix_synapse_ext_password_provider_rest_auth_endpoint` is not configured:

```yaml
- name: Fail if REST Auth endpoint not set
  fail:
    msg: "You need to set matrix_synapse_ext_password_provider_rest_auth_endpoint"
  when: "matrix_synapse_ext_password_provider_rest_auth_enabled and matrix_synapse_ext_password_provider_rest_auth_endpoint == ''"
```

I think the REST auth documentation should be kept somewhat neutral too. Mentioning a common use case (mxisd integration) is good, but there will also be plenty of people who use the REST Auth provider without mxisd, but instead pointing it to their own public (or private) authentication endpoint. Maybe we should have a note about both. Something like:

> Note: the endpoint that you use needs to be accessible from within the Synapse container (this is where the REST auth module operates). The endpoint either needs to be a publicly-accessible URL or a URL pointing to a Docker service in the same network where Synapse runs (by default this is the `matrix` network). Services running on localhost are not visible from the container, so they cannot be used.

-------

Thinking about the above note and the defensive `fail` check above, we might add one more `fail` check, for people who haven't read the note. After all, [people don't read](https://www.joelonsoftware.com/2000/04/26/designing-for-people-who-have-better-things-to-do-with-their-lives/).

To be extra-friendly, in addition to the above `fail` check, we might add one more `fail` check like this:

```yaml
- name: Fail if REST Auth endpoint set incorrectly
  fail:
    msg: "You need to set matrix_synapse_ext_password_provider_rest_auth_endpoint to something which points to a public URL or a service within the `{{ matrix_docker_network }}` Docker network. See docs/configuring-playbook-rest-auth.md"
  when: "matrix_synapse_ext_password_provider_rest_auth_enabled and (matrix_synapse_ext_password_provider_rest_auth_endpoint.startswith('http://localhost') or matrix_synapse_ext_password_provider_rest_auth_endpoint.startswith('http://127.0.0.1'))"
```

haslersn:
> * it somewhat suggests that mxisd is configured to handle it by default. This is not true -- mxisd requires custom configuration to be your authentication backend, including the replacement of the default config template `matrix_mxisd_template_config`. It's a non-straightforward thing to set up.
Not anymore if #53 gets merged.

> * [docs/configuring-playbook-rest-auth.md](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-rest-auth.md) already mentions `matrix_synapse_ext_password_provider_rest_auth_endpoint` explicitly and suggests that you need to change it
If people read they will do that and probably try setting it to `http://localhost:8090` (like I did at first).

> * keep the default of empty string, because there's no other sensible default we can provide
[matrix-synapse-rest-auth](https://github.com/kamax-matrix/matrix-synapse-rest-auth) and [mxisd](https://github.com/kamax-matrix/mxisd) were developed by the same [person](https://github.com/kamax-matrix), so it seems like a sensible default to use them together.

> * add another mention/note about mxisd integration (especially when mxisd runs in a container through the playbook) on `docs/configuring-playbook-rest-auth.md`
or just enhance its automatic integration.

> * add a check in `setup_synapse_ext_rest_auth.yml`, so that it will fail if `matrix_synapse_ext_password_provider_rest_auth_endpoint` is not configured:
The check could as well test if mxisd has been disabled and the default endpoint not overwritten. (The check should fail in that case.)

> I think the REST auth documentation should be kept somewhat neutral too. Mentioning a common use case (mxisd integration) is good, but there will also be plenty of people who use the REST Auth provider without mxisd, but instead pointing it to their own public (or private) authentication endpoint. Maybe we should have a note about both. Something like:
Doing this is already more advanced, than using the matrix-mxisd container that comes with this role.
The default should be to simplify the less advanced option.

> Thinking about the above note and the defensive `fail` check above, we might add one more `fail` check, for people who haven't read the note. After all, [people don't read](https://www.joelonsoftware.com/2000/04/26/designing-for-people-who-have-better-things-to-do-with-their-lives/).
> 
> To be extra-friendly, in addition to the above `fail` check, we might add one more `fail` check like this:
> 
> ```yaml
> - name: Fail if REST Auth endpoint set incorrectly
>   fail:
>     msg: "You need to set matrix_synapse_ext_password_provider_rest_auth_endpoint to something which points to a public URL or a service within the `{{ matrix_docker_network }}` Docker network. See docs/configuring-playbook-rest-auth.md"
>   when: "matrix_synapse_ext_password_provider_rest_auth_enabled and (matrix_synapse_ext_password_provider_rest_auth_endpoint.startswith('http://localhost') or matrix_synapse_ext_password_provider_rest_auth_endpoint.startswith('http://127.0.0.1'))"
> ```
Can still be done.

PR #52, opened 2018-12-04
Add utf8 encoding option to log config

aaronraimist:
This is the now recommended practice as of v0.34.0rc1
https://github.com/matrix-org/synapse/blob/develop/UPGRADE.rst#upgrading-to-v0340

Should work on Python 2 and 3 although I've only tested on Python 3


spantaleev:
Thank you! :+1: 
You're so quick to keep us up-to-date with new developments!

I've just tested this on Python 2 (Synapse 0.33.9) and it seems like it doesn't cause a problem.

PR #53, opened 2018-12-04
Support configuring mxisd's identity stores (two of them)

haslersn:
_mxisd_ supports several identity stores. Add support to configure two of them through role variables:

* synapseSql (storing identities directly in Synapse's database)
* LDAP

This removed the need to copy `mxisd.yaml.j2` to the inventory in case one wants to use LDAP as identity store. Note that the previous solution (copying `mxisd.yaml.j2`) was poor because of two reasons:

* The copy remains outdated in case the original is updated in future versions of this repo.
* The role's configuration should be in one place (configured only through role variables) instead of in multiple.

Configuring more identity stores through role variables can be supported in the future.

### DIFF
--- a/roles/matrix-server/defaults/main.yml
+++ b/roles/matrix-server/defaults/main.yml
@@ -237,6 +237,39 @@ matrix_mxisd_data_path: "{{ matrix_mxisd_base_path }}/data"
 # Enabling this is discouraged. Learn more here: https://github.com/kamax-io/mxisd/blob/master/docs/features/identity.md#lookups
 matrix_mxisd_matrixorg_forwarding_enabled: false
 
+# mxisd has serveral supported identity stores.
+# One of them is storing identities directly in Synapse's database.
+# Learn more here: https://github.com/kamax-matrix/mxisd/blob/master/docs/stores/synapse.md
+matrix_mxisd_synapsesql_enabled: true
+matrix_mxisd_synapsesql_type: postgresql
+matrix_mxisd_synapsesql_connection: //{{ matrix_postgres_connection_hostname }}/{{ matrix_postgres_db_name }}?user={{ matrix_postgres_connection_username }}&password={{ matrix_postgres_connection_password }}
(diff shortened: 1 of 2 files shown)


spantaleev:
It's nice we can support the LDAP use-case without having to override the template file!

It's also nice that we no longer hardcode-enable the Synapse identity store, but rather have it be configurable.

Thank you for covering so many of mxisd's options! :+1: 

--------

I agree that overriding the template and keeping your own template around is annoying and prone to getting outdated quickly.
It's done, because mxisd supports so many features that adding full/exhaustive support for all those would mean spending a lot of time and introducing many additional playbook variables.

This PR gets us closer to full support, but even with that, we still fail to cover even half of what mxisd is capable of. That thing is huge! As of right now, mxisd supports 7 [identity stores](https://github.com/kamax-matrix/mxisd/blob/master/docs/stores/README.md). Not only that, but it supports various other [bridging options](https://github.com/kamax-matrix/mxisd/blob/master/docs/features/bridge-integration.md), [federation options](https://github.com/kamax-matrix/mxisd/blob/master/docs/features/federation.md), etc.

We'll need a lot more work to cover all mxisd use cases, such that getting rid of support for overriding the file will be feasible. I'm not sure we want to spend too much time doing that, but perhaps little by little (just like you did with this PR), we'll eventually reach such a state.

haslersn:
> We'll need a lot more work to cover all mxisd use cases, such that getting rid of support for overriding the file will be feasible. I'm not sure we want to spend too much time doing that, but perhaps little by little (just like you did with this PR), we'll eventually reach such a state.

I did it this way in order to stay consistent with that every option is a single key in camel case.  But perhaps something like #54 would be better. It covers every option with no effort.

spantaleev:
Yes, that's quite nice and consistent with what we're doing everywhere!

#54 is very interesting too, as it will let us support everything mxisd supports in a much more flexible manner.

PR #54, opened 2018-12-05
Make every mxisd configuration option accessible

haslersn:
22523c0e42 added role variables that wrap certain mxisd configuration keys. This has the obvious drawback that not all keys are accessible but only the wrapped.

This change replaces the previously added wrapper role variables by a single dictionary variable `matrix_mxisd_extra_config` in which arbirary mxisd configuration keys can be added.

There's still the same default mxisd configuration. Keys of that default are overwritten individually when added to `matrix_mxisd_extra_config`.

### DIFF
--- a/roles/matrix-server/defaults/main.yml
+++ b/roles/matrix-server/defaults/main.yml
@@ -239,37 +239,19 @@ matrix_mxisd_data_path: "{{ matrix_mxisd_base_path }}/data"
 # Enabling this is discouraged. Learn more here: https://github.com/kamax-io/mxisd/blob/master/docs/features/identity.md#lookups
 matrix_mxisd_matrixorg_forwarding_enabled: false
 
-# mxisd has serveral supported identity stores.
-# One of them is storing identities directly in Synapse's database.
-# Learn more here: https://github.com/kamax-matrix/mxisd/blob/master/docs/stores/synapse.md
-matrix_mxisd_synapsesql_enabled: true
-matrix_mxisd_synapsesql_type: postgresql
-matrix_mxisd_synapsesql_connection: //{{ matrix_postgres_connection_hostname }}/{{ matrix_postgres_db_name }}?user={{ matrix_postgres_connection_username }}&password={{ matrix_postgres_connection_password }}
 
-# LDAP is another identity store that's supported by mxisd.
-# Learn more here: https://github.com/kamax-matrix/mxisd/blob/master/docs/stores/ldap.md
-matrix_mxisd_ldap_enabled: false
-matrix_mxisd_ldap_connection_host: ldapHostnameOrIp
-matrix_mxisd_ldap_connection_tls: false
-matrix_mxisd_ldap_connection_port: 389
-matrix_mxisd_ldap_connection_baseDns: ['OU=Users,DC=example,DC=org']
-matrix_mxisd_ldap_connection_bindDn: CN=My Mxisd User,OU=Users,DC=example,DC=org
-matrix_mxisd_ldap_connection_bindPassword: TheUserPassword
-# The following keys are optional:
-# matrix_mxisd_ldap_filter: ""
-# matrix_mxisd_ldap_attribute_uid_type: uid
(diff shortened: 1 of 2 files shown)


spantaleev:
Shouldn't `matrix_mxisd_extra_config` contain some default value? It's a little weird how it just ends at it now.

It probably also needs to be completely flat and not have nested structures, otherwise this won't work.

For example, an extra config like this:

```yaml
matrix_mxisd_extra_config:
  matrix:
     domain: something here
```

as opposed to this:

```yaml
matrix_mxisd_extra_config:
    matrix.domain: something here
```

will likely fail the check for `'matrix.domain' in matrix_mxisd_extra_config`. Or will it be flattened by Ansible automatically?

spantaleev:
I wonder if we can flatten it somehow..

.. or put up a warning that it must be flat.

And then add some `fail` checks to assist the people who have not read the warning.

haslersn:
> For example, an extra config like this:
> 
> ```yaml
> matrix_mxisd_extra_config:
>   matrix:
>      domain: something here
> ```

Are nested keys even supported by _mxisd_?

spantaleev:
Looking at mxid's [documentation about Configuration Syntax](https://github.com/kamax-matrix/mxisd/blob/master/docs/configure.md#syntax), it does seem like it supports both ways of doing things.

I think mxisd flattens the configuration at some point and doesn't care after that.

I've tried looking for a way to flatten a dictionary object in Ansible, but I haven't found one so far.
It should be possible to do by [developing a custom Ansible module](https://docs.ansible.com/ansible/2.7/dev_guide/developing_modules.html), but that seems a bit extreme.

Since it's hard to do, we should probably stay away from flattening dictionaries and just:
- recommend/warn that `matrix_mxisd_extra_config` needs to be flat (single-level dictionary)
- see if we can detect if it's not flat (more than 1 level) and `fail`

We may be able to detect multiple levels by something like this (code is untested):

```yaml
# Note: the string below contains 4 spaces, corresponding to at least 2 levels of nesting
# due to each level being represented by 2 spaces (`indent=2`).
- set_fact:
    is_matrix_mxisd_extra_config_flat: "{{ '    ' not in matrix_mxisd_extra_config|to_nice_yaml(indent=2) }}
```

spantaleev:
This pull request is obsoleted by 9a9b738 .

We have a default configuration (`matrix_mxisd_configuration_yaml`) and an override/extension configuration (`matrix_mxisd_configuration_extension_yaml`).
People can also override the default directly as well.

We have a check to ensure people that use one of the many deprecated variables will be made aware that they need to change. The list of deprecated variables is:
- `matrix_mxisd_ldap_enabled`
- `matrix_mxisd_ldap_connection_host`
- `matrix_mxisd_ldap_connection_tls`
- `matrix_mxisd_ldap_connection_port`
- `matrix_mxisd_ldap_connection_baseDn`
- `matrix_mxisd_ldap_connection_baseDns`
- `matrix_mxisd_ldap_connection_bindDn`
- `matrix_mxisd_ldap_connection_bindDn`
- `matrix_mxisd_ldap_connection_bindPassword`
- `matrix_mxisd_ldap_filter`
- `matrix_mxisd_ldap_attribute_uid_type`
- `matrix_mxisd_ldap_attribute_uid_value`
- `matrix_mxisd_ldap_connection_bindPassword`
- `matrix_mxisd_ldap_attribute_name`
- `matrix_mxisd_ldap_attribute_threepid_email`
- `matrix_mxisd_ldap_attribute_threepid_msisdn`
- `matrix_mxisd_ldap_identity_filter`
- `matrix_mxisd_ldap_identity_medium`
- `matrix_mxisd_ldap_auth_filter`
- `matrix_mxisd_ldap_directory_filter`
- `matrix_mxisd_template_config`

We also have an automated check to ensure people do not use property-style configuration keys (dot-notation), as this will [not be supported by the upcoming mxisd 1.3.0](https://github.com/kamax-matrix/mxisd/wiki/Upgrade#v130).

In any case, thank you @haslersn for leading the way towards a better way to configure mxisd via this playbook! :+1:

PR #55, opened 2018-12-08
Enable support for custom HTML in riot-web homepage

anadahz:
* Add default template file for homepage HTML
* Add default riot-web config options for homepage

### DIFF
--- a/roles/matrix-server/tasks/setup/setup_riot_web.yml
+++ b/roles/matrix-server/tasks/setup/setup_riot_web.yml
@@ -20,14 +20,15 @@
 
 - name: Ensure Matrix riot-web configured
   template:
-    src: "{{ role_path }}/templates/riot-web/{{ item }}.j2"
-    dest: "{{ matrix_riot_web_data_path }}/{{ item }}"
+    src: "{{ item.src }}"
+    dest: "{{ matrix_riot_web_data_path }}/{{ item.name }}"
     mode: 0644
     owner: "{{ matrix_user_username }}"
     group: "{{ matrix_user_username }}"
   with_items:
-    - "riot.im.conf"
-    - "config.json"
+    - {src: "{{ role_path }}/templates/riot-web/riot.im.conf.j2", name: "riot.im.conf"}
+    - {src: "{{ role_path }}/templates/riot-web/config.json.j2", name: "config.json"}
+    - {src: "{{ matrix_riot_web_homepage_template }}", name: "home.html"}
   when: matrix_riot_web_enabled
 
 - name: Ensure matrix-riot-web.service installed
--- a/roles/matrix-server/templates/riot-web/home.html.j2
+++ b/roles/matrix-server/templates/riot-web/home.html.j2
@@ -1,0 +1,308 @@
+<style type="text/css">
+
+/* we deliberately inline style here to avoid flash-of-CSS problems, and to avoid
+ * voodoo where we have to set display: none by default
+ */
+
+.mx_HomePage_header h1 {
+    margin-left: 0px;
+    margin-bottom: 0px;
+    margin-top: 20px;
+    margin-right: 20px;
+    color: #454545;
+}
+
+.mx_HomePage_header h2 {
+    margin-left: 0px;
+    margin-top: 5px;
+    margin-bottom: 20px;
+    margin-right: 20px;
+    color: #454545;
+}
+
+.mx_HomePage_header h1 a {
+    color: #454545;
+}
+
+.mx_HomePage h3 {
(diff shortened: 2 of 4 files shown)


anadahz:
> Calling the variable something like matrix_riot_web_homepage_template (instead of matrix_riot_web_config) is probably better.

Indeed it makes more sense to use this variable instead, I changed the variables accordingly.

>I also wonder whether using and maintaining our own copy of riot-web's home.html is something we want to take on. It seems like something that may easily get out of date.

According to the commits history of [home.html](https://github.com/vector-im/riot-web/commits/master/res/home.html) it seems that is not getting updated so often.

> On the other hand it's nice we can point it to our own "Riot Bot" (or rather, not show that at all).

This is being handled by the variable [`matrix_riot_web_welcome_user_id`](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/roles/matrix-server/defaults/main.yml#L323) if it's undefined the bot section [will not be displayed](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/55/files#diff-bf8bfad4be66c0c6f136def536bd4cb8R140).

> We may also be able to inject an additional item in the "Running Matrix services" list, which points to the support channel for this playbook (#matrix-docker-ansible-deploy:devture.com).

Good point added the support channel of this playbook.

roles/matrix-server/tasks/setup/setup_riot_web.yml
```diff
@@ -28,6 +28,7 @@
   with_items:
     - "riot.im.conf"
     - "config.json"
+    - "{{ matrix_riot_web_homepage_template }}"
```
spantaleev: Thinking about it now..

A few lines above, we have such a thing:

```yaml
src: "{{ role_path }}/templates/riot-web/{{ item }}.j2"
```

.. meaning that it will try to look for the file within the `{{ role_path }}/templates/riot-web/` directory.

This means that people can't override `matrix_riot_web_homepage_template` in a useful way.
I imagine people may want to point it to another file (one that lives outside that directory), but it's not possible with the current code.

---------------------

It's probably best if the whole code is rewritten like this (note: untested!):

```yaml
- name: Ensure Matrix riot-web configured
  template:
    src: "{{ item.src }}"
    dest: "{{ matrix_riot_web_data_path }}/{{ item.name }}"
    mode: 0644
    group: "{{ matrix_user_username }}"
  with_items:
    - {src: "{{ role_path }}/templates/riot-web/riot.im.conf.j2", name: "riot.im.conf"}
    - {src: "{{ role_path }}/templates/riot-web/config.json.j2", name: "config.json"}
    - {src: "{{ matrix_riot_web_homepage_template }}", name: "home.html"}
when: matrix_riot_web_enabled
```

With that, we can change the default for `matrix_riot_web_homepage_template` (in `roles/matrix-server/defaults/main.yml`) like this:

```yaml
matrix_riot_web_homepage_template: "{{ role_path }}/templates/riot-web/home.html"
```

.. and everything should work as expected.

roles/matrix-server/templates/riot-web/home.html.j2
```diff
@@ -0,0 +1,308 @@
...
+{% if matrix_riot_web_homepage_template_technical %}
+
+    <h3>_t("Matrix technical discussions")</h3>
+    <h4>_t("Running Matrix services")</h4>
+
+    <div class="mx_HomePage_row">
+        <div class="mx_HomePage_room">
+            <a href="#/room/#matrix-docker-ansible-deploy:devture.com">
+                <img class="mx_HomePage_icon" src="home/rooms/matrix.png">
+                <span class="mx_HomePage_name">Synapse Support Community</span>
```
spantaleev: This should be changed to "Matrix Docker Ansible Deploy".

anadahz: I missed that thanks.

roles/matrix-server/tasks/setup/setup_riot_web.yml
anadahz: Makes sense, make an almost similar change.

anadahz:
Ready for review.

spantaleev:
I've merged this and it seems to work!

I've only had to remove the `_t()` (translation) call for our custom button.
Otherwise the "Support for Matrix Docker Ansible role" label was being prefixed by some other "missing translation" label. Maybe this only happened when another locale was used (besides the default "en"), but..

Thank you for your work on this! :+1:

anadahz:
> I've only had to remove the _t() (translation) call for our custom button.

Good catch, thanks!

PR #59, opened 2018-12-17
Apache webserver sample configuration

ampoz:
Sample configuration file and instructions for using Apache instead of Nginx as the webserver/proxy.

### DIFF
--- a/docs/configuring-playbook-own-webserver.md
+++ b/docs/configuring-playbook-own-webserver.md
@@ -28,9 +28,9 @@ matrix_nginx_proxy_enabled: false
 
 - ensure that you restart/reload your webserver once in a while, so that renewed SSL certificates would take effect (once a month should be enough)
 
-**Apache2 sample configuration files**
+**Apache2 sample configuration file**
 
-1. Create a new apache configuration file named 000-matrix-ssl.conf and enable it.
+1. Create a new apache configuration file named 000-matrix-ssl.conf and enable it. Make certain to replace DOMAIN/SSL values with the correct ones for your server.
 
        # Auto redirect http to https
        <VirtualHost *:80>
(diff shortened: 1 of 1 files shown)


docs/configuring-playbook-own-webserver.md
```diff
@@ -26,4 +26,75 @@ matrix_nginx_proxy_enabled: false
...
+       <VirtualHost *:80>
+           ServerName matrix.DOMAIN
+           Redirect permanent / https://matrix.DOMAIN/
+       </VirtualHost>
+
+       <VirtualHost *:443>
+           ServerName matrix.DOMAIN
+
+           SSLEngine On
+           SSLCertificateFile /etc/letsencrypt/live/DOMAIN/fullchain.pem
```
spantaleev: Maybe this can be set to:

```
SSLCertificateFile /matrix/ssl/config/live/matrix.DOMAIN/fullchain.pem
SSLCertificateKeyFile /matrix/ssl/config/live/matrix.DOMAIN/privkey.pem
```

for a better default.

docs/configuring-playbook-own-webserver.md
```diff
@@ -26,4 +26,75 @@ matrix_nginx_proxy_enabled: false
...
+1. Create a new apache configuration file named 000-matrix-ssl.conf and enable it. Make certain to replace DOMAIN/SSL values with the correct ones for your server.
+
+       # Auto redirect http to https
+       <VirtualHost *:80>
+           ServerName matrix.DOMAIN
+           Redirect permanent / https://matrix.DOMAIN/
+       </VirtualHost>
+
+       <VirtualHost *:443>
+           ServerName matrix.DOMAIN
```
spantaleev: Do you think we should add some `riot.DOMAIN` configuration as well?

Which makes me think.. maybe this documentation page will become quite long if we do.

It might be better if we put these sample configuration files in something like `examples/own-web-server/apache/..`.
You can then make multiple files, one for each vhost (currently just `matrix.` and `riot.`, but we may have `dimension.` too in the future).

spantaleev:
Closing this. I've included a modification of the proposed configuration in the `examples/apache` directory.

Thank you for providing a working Apache configuration, @ampoz! :+1:

PR #66, opened 2019-01-07
Use yaml syntax instead of key=value syntax consistently 

### LINKED ISSUE #62: Use yaml syntax instead of key=val consistently
`[..]/tasks/import/import_media_store.yml` for example has a mix of both formats which hurts readability IMO

jcgruenhage:


spantaleev:
Thank you! :+1:

Issue #65, opened 2019-01-07
Audit results


jcgruenhage:
I'll put small things that I notice doing my audit of this here, which aren't big/important enough to get their own issue.

 - [x] `self_check_dns.yml`: maybe use the dig lookup plugin instead of a command, and if you use a command, use a command and not a shell.
 - [ ] There are a lot of set_fact tasks, that don't depend on anything that is only available at runtime, moving those into defaults or vars (depending on what they are exactly) would make sense IMO.
 - [x] There should not be a default UID/GID for the matrix user (if this is even a user that should exist, having dedicated users for components would also be a possibility). When creating a user using ansible it has the UID/GID in the result and then you can use that.
 - [ ] The whole letsencrypt business is a bit weird, I really don't like certbot. Have you thought about maybe using dehydrated here? I think it integrates with ansible+containers a lot better than certbot
 - [ ] `setup/ssl/setup_ssl_self_signed_obtain_for_domain.yml`:  You could generate the ssl cert manually only on hosts which don't provide up to date pyopenssl and do the proper way on other hosts
 - [x] periodic nginx restarts could be done with the crontab module instead of templating a cron.d file
 - [x] the start tasks thingy could (and should) be replaced with a loop. also, daemon_reload is not needed there. We also don't need daemon_reload in a lot of other cases, that should be handled with a handler and not a task.
 - [x] going from yaml truth values to json truth values via "true if var else false" is ugly, you can use the |to_json filter for that
 - [ ] if the docker_container module of ansible is used for launching the containers, the env file templates are unnecessary, the vars can just be taken from vars directly

spantaleev:
Thank for your detailed report!

Some of these are already fixed, while others are in progress. I'll go through all of them below, in order.

-------------

> `self_check_dns.yml`: maybe use the dig lookup plugin instead of a command, and if you use a command, use a command and not a shell.

Fixed in f92c4d5a27d775.

During that, I've found out that we don't do a similar self-check for our [mxisd](https://github.com/kamax-io/mxisd) install, so I've added support for that in ef2dc3745aab2d5.

-------------

> There are a lot of set_fact tasks, that don't depend on anything that is only available at runtime, moving those into defaults or vars (depending on what they are exactly) would make sense IMO.

Can you give a specific example?

Maybe this is one?

```yaml
- set_fact:
    matrix_synapse_app_service_config_file_mautrix_whatsapp: '/app-registration/mautrix-whatsapp.yml'
[...]

PR #68, opened 2019-01-08
Switch to managing cronjobs with the Ansible cron module

spantaleev:
As suggested in #65 (Github issue), this patch switches
cronjob management from using templates to using Ansible's `cron` module.

It also moves the management of the nginx-reload cronjob to `setup_ssl_lets_encrypt.yml`,
which is a more fitting place for it (given that this cronjob is only required when
Let's Encrypt is used).

Pros:
- using a module is more Ansible-ish than templating our own files in
special directories

- more reliable: will fail early (during playbook execution) if `/usr/bin/crontab`
is not available, which is more of a guarantee that cron is working fine
(idea: we should probably install some cron package using the playbook)

Cons:
- invocation schedule is no longer configurable, unless we define individual
variables for everything or do something smart (splitting on ' ', etc.).
Likely not necessary, however.

- requires us to deprecate and clean-up after the old way of managing cronjobs,
because it's not compatible (using the same file as before means appending
additional jobs to it)


spantaleev:
Thanks for testing! Merged now.

PR #72, opened 2019-01-11
Update README to link to bubuntux/riot-web

aaronraimist:
This new image is awesome 🎉


spantaleev:
It really is! Thanks for finding it!

PR #73, opened 2019-01-12
 Split playbook into multiple roles

spantaleev:
As suggested in #63 (Github issue), splitting the
playbook's logic into multiple roles will be beneficial for
maintainability.

This patch realizes this split. Still, some components
affect others, so the roles are not really independent of one
another. For example:
- disabling mxisd (`matrix_mxisd_enabled: false`), causes Synapse
and riot-web to reconfigure themselves with other (public)
Identity servers.

- enabling matrix-corporal (`matrix_corporal_enabled: true`) affects
how reverse-proxying (by `matrix-nginx-proxy`) is done, in order to
put matrix-corporal's gateway server in front of Synapse

We may be able to move away from such dependencies in the future,
at the expense of a more complicated manual configuration, but
it's probably not worth sacrificing the convenience we have now.

As part of this work, the way we do "start components" has been
redone now to use a loop, as suggested in #65 (Github issue).
This should make restarting faster and more reliable.


jcgruenhage:
> disabling mxisd (matrix_mxisd_enabled: false), causes Synapse
and riot-web to reconfigure themselves with other (public)
Identity servers.

I don't really see this as a dependency, because this just means that the  synapse/riot roles should have configurable identity servers, with the default going to vector.im inside the role. The override to use the local mxisd should then happen via vars in the playbook, making them nicely separated.

------------------------------------

> enabling matrix-corporal (matrix_corporal_enabled: true) affects
how reverse-proxying (by matrix-nginx-proxy) is done, in order to
put matrix-corporal's gateway server in front of Synapse

This is similar IMO, having the upstream homeserver endpoint configurable seems nice for other things than matrix-corporal. Put a default inside the role, and then build the override in the playbook.

jcgruenhage:
Apart from that I think this should be possible to be nicely separated, I haven't had time to look at it in-depth.

spantaleev:
We can have defaults in each role, and then have one role redefine the default of another.

We do something similar with corporal already.
The nginx proxy configuration defaults are:

```yaml
# The addresses where the Matrix Client API is.
# Certain extensions (like matrix-corporal) may override this in order to capture all traffic.
matrix_nginx_proxy_matrix_client_api_addr_with_proxy_container: "matrix-synapse:8008"
matrix_nginx_proxy_matrix_client_api_addr_sans_proxy_container: "localhost:8008"
```

which `roles/matrix-corporal/tasks/init.yml` overrides:

```yaml
- name: Override configuration specifying where the Matrix Client API is
  set_fact:
    matrix_nginx_proxy_matrix_client_api_addr_with_proxy_container: "matrix-corporal:41080"
    matrix_nginx_proxy_matrix_client_api_addr_sans_proxy_container: "localhost:41080"
  when: "matrix_corporal_enabled"
```

This works, but with the following downsides:

- order of role execution is important. If the `matrix-corporal` is after the `matrix-nginx-proxy` role in the playbook, the override will not happen in time and `matrix-nginx-proxy`'s configuration will be incorrect

- variables that other components override (like `matrix_nginx_proxy_matrix_client_api_addr_with_proxy_container` and `matrix_nginx_proxy_matrix_client_api_addr_sans_proxy_container` in this case) cannot be redefined by the playbook user. In the above example, even if you define your own values, if matrix-corporal is enabled, it will take over.

-------------

It will be cleanest (code-wise) if there are no such cross-role changes at all.

If so, enabling something like corporal would need be done with more than just `matrix_corporal_enabled: true` and a few more `matrix_corporal_` definitions. You you'd need to hook it to matrix-nginx-proxy by fiddling with its configuraiton as well.

Likewise for enabling/disabling mxisd. We need to assume a default state for the playbook (mxisd enabled, I guess) and set the defaults of the other components assuming that. If someone then disables mxisd, it's their own responsibility to provide working identity servers to Synapse and riot-web.

While I like how we'll greatly reduce code complexity, I worry if this will make the playbook more complicated. The above use-cases (enabling corporal; disabling mxisd) are kind of advanced scenarios though, so whoever is smart enough to handle those, surely is smart enough to define a few more variables to make the other components play along well (?).

jcgruenhage:
> order of role execution is important. If the matrix-corporal is after the matrix-nginx-proxy role in the playbook, the override will not happen in time and matrix-nginx-proxy's configuration will be incorrect

That only happens if you put the override into the role, but the thing glueing together roles are playbooks, so it should land in the playbook vars instead. Having it in the playbook vars with jinja is also a bit faster than having it be a set_fact based task.

--------------------------------

> variables that other components override (like matrix_nginx_proxy_matrix_client_api_addr_with_proxy_container and matrix_nginx_proxy_matrix_client_api_addr_sans_proxy_container in this case) cannot be redefined by the playbook user. In the above example, even if you define your own values, if matrix-corporal is enabled, it will take over.

If the glue vars are defined in the playbook and not in the roles, users can override them.

spantaleev:
I've merged this, as it appears to work well and isn't really worse than what we had before.

The glue-vars-in-playbook idea seems to have some merit. I'm exploring it separately and will push it to another branch when it has progressed.

Issue #80, opened 2019-01-16
Publish to Ansible galaxy


TheLastProject:
Could you please submit the roles to https://galaxy.ansible.com/ so that they can be added to other playbooks more easily? :)

spantaleev:
I'm currently working on making the roles more independent of one another, which should bring us one step closer to being able to publish them separately.

samyak-jain:
@spantaleev Since #81 is merged, is this issue possible to fix?

spantaleev:
Closing this as part of a backlog review, in line with the related #947 (restructuring into an Ansible collection).

Publishing to Ansible Galaxy never happened in the seven years since this was filed, and the playbook has meanwhile settled on an architecture that removes the motivation for it: the reusable roles are extracted to standalone repositories (mostly under the [mother-of-all-self-hosting](https://github.com/mother-of-all-self-hosting) organization) and are consumed via ansible-galaxy with git-pinned versions in requirements.yml, without Galaxy needing to host anything. The roles remaining in this repository are playbook-specific by design (tightly coupled to the playbook's wiring) and would not be usable standalone, so publishing them would not benefit anyone, while maintaining a Galaxy namespace and release publishing would add ongoing upkeep with no functional gain.

PR #74, opened 2019-01-13
Add missing vars to roles where they are used

Cadair:

PR #81, opened 2019-01-16
Make roles more independent of one another

spantaleev:
This is a continuation of the role-split done by #73.

Hopefully, having the roles more independent of one another makes them easier to use in other playbooks (obsoleting the need for #74).

It probably also brings us closer to being able to publish the roles on [Ansible Galaxy](https://galaxy.ansible.com/), as discussed in #80.

---

With this change, the following roles are now only dependent on the minimal `matrix-base` role:
- `matrix-corporal`
- `matrix-coturn`
- `matrix-mailer`
- `matrix-mxisd`
- `matrix-postgres`
- `matrix-riot-web`
- `matrix-synapse`

The `matrix-nginx-proxy` role still does too much and remains dependent on the others.

Wiring up the various (now-independent) roles happens via a glue variables file (`group_vars/matrix-servers`).
It's triggered for all hosts in the `matrix-servers` group.

According to Ansible's rules of priority, we have the following chain of inclusion/overriding now:
- role defaults (mostly empty or good for independent usage)
- playbook glue variables (`group_vars/matrix-servers`)
- inventory host variables (`inventory/host_vars/matrix.<your-domain>`)

All roles default to enabling their main component
(e.g. `matrix_mxisd_enabled: true`, `matrix_riot_web_enabled: true`).
Reasoning: if a role is included in a playbook (especially separately, in another playbook), it should "work" by default.

Our playbook disables some of those if they are not generally useful (e.g. `matrix_corporal_enabled: false`).


spantaleev:
I think this is good enough now and there aren't many problems for backward compatibility.
Merged, so we can proceed forward.

Still would be glad to hear about any issues that may arise for people who actually use the roles in their own separate playbook (@Cadair).

PR #84, opened 2019-01-27
Make all containers start as non-root and without capabilities

spantaleev:
This makes all containers (except mautrix-telegram and mautrix-whatsapp), start as a non-root user.

We do this, because we don't trust some of the images.
In any case, we'd rather not trust ALL images and avoid giving `root` access at all.
We can't be sure they would drop privileges or what they might do before they do it.

Because Postfix doesn't support running as non-root, it had to be replaced by an Exim mail server.

The matrix-nginx-proxy nginx container image is patched up (by replacing its main configuration) so that it can work as non-root.
It seems like there's no other good image that we can use and that is up-to-date (https://hub.docker.com/r/nginxinc/nginx-unprivileged is outdated).

Likewise for riot-web (https://hub.docker.com/r/bubuntux/riot-web/), we patch it up ourselves when starting (replacing the main nginx configuration).
Ideally, it would be fixed upstream so we can simplify.


efraimbart:
If adding a new mautrix bridge role, should I assume it can run without root access and include `--read-only`?

spantaleev:
I can see that some roles (like `matrix-mautrix-facebook`) use `--user={{ matrix_user_uid }}:{{ matrix_user_gid }}` and `--cap-drop=ALL`. Whether `--read-only` is possible, I'm not sure. It probably is, but we should test that and apply it to our bridges.

PR #87, opened 2019-01-31
Add support for matrix-appservice-irc

Plailect:
I have tested this and everything works as expected on my personal server.

This uses [this](https://hub.docker.com/r/tedomum/matrix-appservice-irc/) docker image, and by extension [this](https://github.com/TeDomum/matrix-appservice-irc) fork of matrix-appservice-irc.

### DIFF
--- a/docs/configuring-playbook-bridge-appservice-irc.md
+++ b/docs/configuring-playbook-bridge-appservice-irc.md
@@ -16,8 +16,10 @@ matrix_appservice_irc_configuration_extension_yaml: |
   #
   # If you need something more special, you can take full control by
   # completely redefining `matrix_appservice_irc_configuration_yaml`.
+  # 
+  # For a full example configuration with comments, see `roles/matrix-synapse/defaults/main.yml`
   #
-  # Example configuration extension follows:
+  # A simple example configuration extension follows:
   #
   ircService:
     databaseUri: "nedb://data" # does not typically need modification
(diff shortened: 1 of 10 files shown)


docs/configuring-playbook-bridge-appservice-irc.md
```diff
@@ -0,0 +1,435 @@
...
+
+The playbook can install and configure [matrix-appservice-irc](https://github.com/TeDomum/matrix-appservice-irc) for you.
+
+See the project's [documentation](https://github.com/TeDomum/matrix-appservice-irc/blob/master/HOWTO.md) to learn what it does and why it might be useful to you.
+
+You'll need to use the following playbook configuration:
+
+```yaml
+matrix_matrix_appservice_irc_enabled: true
+matrix_appservice_irc_configuration_extension_yaml: |
```
spantaleev: Maybe posting the full contents of `matrix_appservice_irc_configuration_extension_yaml` makes it overly long (and hard to keep in sync with the other example of it that we have in `roles/matrix-synapse/defaults/main.yml`).

Maybe it would be better if we have a minimal example and point people to the ``roles/matrix-synapse/defaults/main.yml`` instead?

That way, they can see the default configuration in `matrix_appservice_irc_configuration_yaml` too.

docs/configuring-playbook-bridge-appservice-irc.md
```diff
@@ -0,0 +1,435 @@
+# Setting up Appservice IRC (optional)
+
+The playbook can install and configure [matrix-appservice-irc](https://github.com/TeDomum/matrix-appservice-irc) for you.
+
+See the project's [documentation](https://github.com/TeDomum/matrix-appservice-irc/blob/master/HOWTO.md) to learn what it does and why it might be useful to you.
+
+You'll need to use the following playbook configuration:
+
+```yaml
+matrix_matrix_appservice_irc_enabled: true
```
spantaleev: The `roles/matrix-synapse/defaults/main.yml` file actually uses `matrix_appservice_irc_enabled`, not `matrix_matrix_appservice_irc_enabled`.

I am also in favor of dropping the double `matrix` prefix. We just need to fix this documentation.

roles/matrix-synapse/templates/ext/appservice-irc/systemd/matrix-appservice-irc.service.j2
```diff
@@ -0,0 +1,26 @@
...
+
+[Service]
+Type=simple
+ExecStartPre=-/usr/bin/docker kill matrix-appservice-irc
+ExecStartPre=-/usr/bin/docker rm matrix-appservice-irc
+ExecStart=/usr/bin/docker run --rm --name matrix-appservice-irc \
+			--log-driver=none \
+			-e "UID={{ matrix_user_uid }}" -e "GID={{ matrix_user_gid }}" \
+			--network={{ matrix_docker_network }} \
+			-p 127.0.0.1:9999:9999 \
```
spantaleev: Do we really need to expose port `9999` to the local host?

You can make this conditional (off by default) by doing a few things.

First, introducing a new variable in `roles/matrix-synapse/defaults/main.yml` called something like `matrix_app_service_container_expose_port` (default value of `false`). See `matrix_synapse_container_expose_client_server_api_port` at the beginning of that same defaults file for reference.

Then, wire this variable to whether matrix-nginx-proxy is enabled or not. We do this in the glue file (`group_vars/matrix-servers`).

```
matrix_app_service_container_expose_port: "{{ not matrix_nginx_proxy_enabled }}"
```

And finally, you can edit the `matrix-appservice-irc.service.j2` file to make it use it:

```

			{% if matrix_app_service_container_expose_port %}
			-p 127.0.0.1:9999:9999 \
			{% endif %}
```

roles/matrix-synapse/templates/ext/appservice-irc/systemd/matrix-appservice-irc.service.j2
```diff
@@ -0,0 +1,26 @@
...
+Requires=matrix-synapse.service
+After=matrix-synapse.service
+
+[Service]
+Type=simple
+ExecStartPre=-/usr/bin/docker kill matrix-appservice-irc
+ExecStartPre=-/usr/bin/docker rm matrix-appservice-irc
+ExecStart=/usr/bin/docker run --rm --name matrix-appservice-irc \
+			--log-driver=none \
+			-e "UID={{ matrix_user_uid }}" -e "GID={{ matrix_user_gid }}" \
```
spantaleev: We could probably do with the rootless stuff and dropping capabilities in another PR later.

roles/matrix-synapse/tasks/ext/appservice-irc/setup.yml
```diff
@@ -0,0 +1,73 @@
...
+    owner: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_username }}"
+  when: "matrix_appservice_irc_enabled"
+
+- stat:
+    path: "{{ matrix_appservice_irc_base_path }}/passkey.pem"
+  register: irc_passkey_file
+
+- name: Generate matrix-appservice-irc passkey if it doesn't exist
+  shell: /usr/bin/openssl genpkey -out {{ matrix_appservice_irc_base_path }}/passkey.pem -outform PEM -algorithm RSA -pkeyopt rsa_keygen_bits:2048
```
spantaleev: We can probably do this by not shelling out to the `openssl` binary, but rather by using Ansible's [openssl_privatekey](https://docs.ansible.com/ansible/latest/modules/openssl_privatekey_module.html) module.

I haven't tested it, but it should be something like this, I imagine:

```
- name: Generate matrix-appservice-irc passkey if it doesn't exist
  openssl_privatekey:
    path: "{{ matrix_appservice_irc_base_path }}/passkey.pem"
    size: 2048
    force: false
  when: matrix_appservice_irc_enabled
```
Note: it doesn't even need a `irc_passkey_file.stat` stuff, because the module only regenerates the file if it doesn't exist. So the above `stat` call for `irc_passkey_file` can probably be removed.

However, I'm not sure if using the `openssl_privatekey` is an improvement.
It requires the python-pyOpenSSL library, which I'm not sure if it's available by default.

I guess it may be safer to go with the `openssl` binary and keep the existing code. If you'd like to experiment and find out if some major distros can execute `openssl_privatekey` without explicitly installing additional packages, you could.. If it's too much work, it's good to leave it as it is.

roles/matrix-synapse/templates/ext/appservice-irc/systemd/matrix-appservice-irc.service.j2
Plailect: Sounds good.

roles/matrix-synapse/tasks/ext/appservice-irc/setup.yml
Plailect: I'd say that it's a better idea to use the built-in module instead of calling out to binaries. I tried it on my machine (Arch) and it worked just fine (I already had the python-pyopenssl package install). I'll add a note to the doc page that says it's required to use the IRC bridge.

spantaleev:
Merged now! Thanks for you great work! :+1:

-------------

On a related note, I gave it a try (on Arch Linux) and it indeed failed with an error like this:

> the python pyOpenSSL module is required

Installing `python-pyopenssl` fixes the problem, but it's additional hassle for people that want to use the IRC bridge.

I was wondering whether we should stay away from the Ansible way, and follow the more practical way -- using `openssl` and making it easier for people (no need to install additional dependencies). We can even make sure `openssl` is installed on the server (if not already) by listing it in the `matrix-server-base` role.

Plailect:
Either way is fine, feel free to revert commit 0dac5ea if you prefer that.

spantaleev:
My previous comment is inaccurate.. It leads you to believe that installing pyOpenSSL on the local machine resolves the problem, which is not true.. pyOpenSSL needs to be installed on the server. Doing that is even more annoying.

I've reverted 0dac5ea and now we have an IRC bridge that's easy to install by anyone!

Thanks again! :+1:

Plailect:
Huh, my Ubuntu server must have had it installed by default then? Weird. Either way, good to hear :)

PR #88, opened 2019-01-31
Adding .editorconfig.

dhose:
https://editorconfig.org/

Only a suggestion, change as you would see fit.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,24 @@
+# This file is the top-most EditorConfig file
+root = true
+
+# All Files
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = tab
+indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+#########################
+# File Extension Settings
+#########################
+
+# YAML Files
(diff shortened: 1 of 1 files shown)


spantaleev:
I gave it a try and it works well (tested using VS Code).

I found the following things:

- having a default tab size of `4` seems more standard than `2`

- for YAML files, we should include a few more types. I've found the following to work well:`[*.{yml,yaml,log.config.j2,yaml.j2}]`

- I wonder why we wouldn't trim whitespace for `[*.md]` files (`trim_trailing_whitespace = false`)?

- We don't have any of the following types of files, so I'm not sure it makes sense adding them: `[*.{ini,py,py.tpl,rst}]`.

dhose:
I have updated the PR now.

Regarding markdown and trimming trailing whitespace, it breaks markdown - https://github.com/Microsoft/vscode/issues/1679

spantaleev:
That's great!

I've added a comment about the "two spaces makes a new line" thing to prevent other people wondering about it in the future.

Thank you, @dhose! :+1:

PR #91, opened 2019-02-04
Turn off IPv6 when using your own Nginx server

aaronraimist:


aaronraimist:
Also the bit in `homeserver.yaml` fixes these warning messages in the log. My understanding is [::] will listen on both IPv6 and IPv4 on Linux.

```
2019-02-02 16:10:54,450 - synapse.app - 47 - WARNING -  - Failed to listen on 0.0.0.0, continuing because listening on [::]
2019-02-02 16:10:54,482 - synapse.app - 47 - WARNING -  - Failed to listen on 0.0.0.0, continuing because listening on [::]
```

spantaleev:
Thank you! :+1:

PR #99, opened 2019-02-19
Python dns/dig dependancies on debian

sigio:


spantaleev:
Hi and thank you for your contribution!

I can't seem to understand why we'd need to install `dig` and `python-dns` on the server though?

If you're trying to make the self-check (`--tags=self-check`) work (namely the DNS SRV record check in `roles/matrix-base/tasks/self_check_dns_srv.yml`), it should be noted that the check executes on the local machine, and not on the server. Thus, it's the local machine that needs to have either `dig` or `python-dns` installed, not the server.

If you happen to run the playbook on the Matrix server, targeting itself, then it is a special case which makes them both the same machine. Usually, it's not the case.

sigio:
Yeah... I was running the playbook on the target system, so that's why I encountered this issue.
Might not be needed then ;)

sigio:
Feel free to close/reject this one then ... or maybe update the readme to also install this besides ansible.
Also... maybe make a note of the minimal ansible version (2.4 I guess). I initiially tried with the version in debian stable, which was too old, so used the version from pip-install (2.7.x) which did work.

spantaleev:
We were already mentioning the minimum version requirement for Ansible on the `docs/ansible.md` page, linked from Prerequisites (`docs/prerequisites.md`).

In 747574ab560ad47, I've put another more explicit link to the Ansible version requirements.

Additionally, `dig` / `python-dns` are now also mentioned on the Prerequisites page.

Thanks for the input! :+1:

PR #102, opened 2019-02-28
Don't to_json the matrix_synapse_database_password

TheLastProject:
It's good policy to use ansible vault to define passwords. However, due to an Ansible bug, to_json will not cause the password to be decrypted: https://github.com/ansible/ansible/issues/24425

### DIFF
--- a/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
+++ b/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
@@ -327,7 +327,7 @@ database:
   name: "psycopg2"
   args:
     user: {{ matrix_synapse_database_user|to_json }}
-    password: {{ matrix_synapse_database_password }}
+    password: {{ matrix_synapse_database_password|string|to_json }}
     database: "{{ matrix_synapse_database_database }}"
     host: "{{ matrix_synapse_database_host }}"
     cp_min: 5
@@ -522,7 +522,7 @@ recaptcha_siteverify_api: "https://www.google.com/recaptcha/api/siteverify"
 turn_uris: {{ matrix_synapse_turn_uris|to_json }}
 
 # The shared secret used to compute passwords for the TURN server
-turn_shared_secret: {{ matrix_synapse_turn_shared_secret|to_json }}
+turn_shared_secret: {{ matrix_synapse_turn_shared_secret|string|to_json }}
 
 # The Username and password if the TURN server needs them and
 # does not use a token
@@ -569,7 +569,7 @@ enable_registration: {{ matrix_synapse_enable_registration|to_json }}
 
 # If set, allows registration by anyone who also has the shared
 # secret, even if registration is otherwise disabled.
-registration_shared_secret: {{ matrix_synapse_registration_shared_secret|to_json }}
+registration_shared_secret: {{ matrix_synapse_registration_shared_secret|string|to_json }}
 
 # Set the number of bcrypt rounds used to generate password hash.
 # Larger numbers increase the work factor needed to generate the hash.
@@ -643,11 +643,10 @@ app_service_config_files: {{ matrix_synapse_app_service_config_files }}
 # enables MAU tracking for application service users.
 track_appservice_user_ips: False
 
-
 # a secret which is used to sign access tokens. If none is specified,
 # the registration_shared_secret is used, if one is given; otherwise,
 # a secret key is derived from the signing key.
-macaroon_secret_key: {{ matrix_synapse_macaroon_secret_key|to_json }}
+macaroon_secret_key: {{ matrix_synapse_macaroon_secret_key|string|to_json }}
 
 # Used to enable access token expiration.
 expire_access_token: False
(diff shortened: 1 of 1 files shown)


spantaleev:
People may have double-quotes and other weird things in their passwords, which would make the whole YAML file invalid, unless we do `|to_json`.

For other strings, it may be relatively safe to go without it, but probably not for passwords. See #49.

--------

It doesn't seem like I can reproduce the problem on Ansible 2.7.8 (running on Python 3.7.2).

I've encrypted the password using `ansible-vault encrypt-string password-here`.
I then use the following configuration in my vars file:

```
# Value is "password-here", encrypted with a password of "test"
matrix_postgres_connection_password: !vault |
          $ANSIBLE_VAULT;1.1;AES256
          65623734663461663832323535666135306435333732633364336566386464376335383965323964
          3766643237306264336235353563333431366430343863330a393530356437303765366637363931
          64646134383830656432633231336638316139356532623039383164303635646130633334646365
          6265616435326335320a623736666361346336376366393636373337343630303663343664666232
          3530
```

After that, I've re-ran the playbook like this: `ansible-playbook -i inventory/hosts --ask-vault-pass setup.yml --tags=setup-synapse`, supplying a password of "test".

`grep password /matrix/synapse/config/homeserver.yaml` shows that the password is there, as expected.

Maybe the Ansible bug has been fixed already...? Or are you doing things differently?

TheLastProject:
Hmm, odd, it really is broken for me, even after upgrading from Ansible 2.7.0 to 2.7.8. Maybe I specify the variables in another place? They're in my group-vars for matrix-app which is a child of the matrix group (containing the default values).

Anyway, adding |string before the |to_json seems to fix it for me and I think it probably won't break anything even with weird strings.

TheLastProject:
Hmm, strangely enough the matrix_synapse_macaroon_secret_key stays a vault  variable, yet the others are correctly translated with |string| in between. Weird stuff...

spantaleev:
Could it be because the others (e.g. `matrix_synapse_database_password`) are specified by referring to other variables (e.g. `matrix_synapse_database_password: "{{ matrix_postgres_connection_password }}"`), and it's those other variables (`matrix_postgres_connection_password`) that are the vault value?

.. whereas `matrix_synapse_macaroon_secret_key` is defined directly as a vault variable?

TheLastProject:
I do define `matrix_synapse_database_password` directly, just like `matrix_synapse_macaroon_secret_key`, but I notice that `matrix_synapse_macaroon_secret_key` has no default value in the defaults yaml I use from this repo (which I saved as `group_vars/matrix.yml`).

Even If I define `matrix_synapse_macaroon_secret_key: "{{ matrix_synapse_macaroon_secret_key_test }}"` in the group_vars/matrix.yml and set `matrix_synapse_macaroon_secret_key` to a vault encrypted secret in `inventories/production/group_vars/matrix-app.yml` to override it (which my playbook uses to call the rolls of this GitHub repo with), it still throws the vault var into the `homeserver.yml`.

Weirdly enough, setting matrix_synapse_macaroon_secret_key_test makes the check for matrix_synapse_macaroon_secret_key being defined fail.

Even when I completely remove `group_vars/matrix.yml`, the issue remains. Removing the `matrix_synapse_macaroon_secret_key` variable from `defaults/main.yml` in the `matrix-synapse` role also doesn't fix anything.

`matrix_synapse_form_secret`, however, does decrypt properly.

The only fix I seem to have is setting my matrix-app.yml group_vars up as such:
```
matrix_synapse_macaroon_secret_key_test: !vault |
            $ANSIBLE_VAULT;1.1;AES256
            [hidden]
matrix_synapse_macaroon_secret_key: "{{ matrix_synapse_macaroon_secret_key_test }}"
```

In fact, I don't even need the |string| cast then anymore.

I suppose I'll close this pull request because it doesn't really properly fix it, but indeed it seems necessary that every variable is "wrapped" by another variable if they're vault encrypted, or it'll break, which this role just happens to mostly do already. I just need to be careful to set the "wrapped" variables, and I suppose matrix-synapse should have a wrapper for the macaroon thing, I will create another pull request for that.

TheLastProject:
Actually, no new pull request, setting `matrix_synapse_macaroon_secret_key_value: "{{ matrix_synapse_macaroon_secret_key }}"` in `matrix-synapse/defaults/main.yml` doesn't work either.

spantaleev:
So there is no proper way to work around it so far?

Back to needing `|string`?
From what I gather, doing that also didn't fix all occurrences either?

TheLastProject:
Sadly, I haven't found a reliable way to fix this. If I ever do, I'll make sure to report back.

TheLastProject:
You know when I said it didn't work? I was editing the wrong file. Just passing it through |string| before the |to_json| does seem to work reliably. Which variables should we do that with though, because anything that could accept a list would likely break.

spantaleev:
I guess people might use vault for these:

- `matrix_synapse_registration_shared_secret`
- `matrix_synapse_database_user`
- `matrix_synapse_database_password`
- `matrix_synapse_turn_shared_secret`
- `matrix_postgres_connection_username`
- `matrix_postgres_connection_password`
- `matrix_coturn_turn_static_auth_secret`
- `matrix_corporal_matrix_auth_shared_secret`
- `matrix_corporal_matrix_registration_shared_secret`

We probably only need to do it wherever `|to_json` or some other such filter is used (and now when it's just `{{ var }}`), but we'd probably better do it for all instances anyway. Who knows, maybe something doesn't have `|to_json` now but would have it tomorrow..

PR #104, opened 2019-03-01
allow exposing mautrix_telegram port

paulbdavis:
This allows the out of matrix links to work when not using the built in nginx proxy

### DIFF
--- a/roles/matrix-synapse/defaults/main.yml
+++ b/roles/matrix-synapse/defaults/main.yml
@@ -210,7 +210,7 @@ matrix_mautrix_telegram_api_hash: YOUR_TELEGRAM_API_HASH
 # Use an uuid so it's not easily discoverable
 matrix_mautrix_telegram_public_endpoint: "/{{ matrix_synapse_macaroon_secret_key | password_hash('sha512', 'telegram') | to_uuid }}"
 # Set this to a port number to expose on the host when not using the nginx proxy
-matrix_mautrix_telegram_container_expose_port: false
+matrix_mautrix_telegram_container_exposed_port_number: ~
 
 # Matrix mautrix is a Matrix <-> Whatsapp bridge
 # Enable whatsapp bridge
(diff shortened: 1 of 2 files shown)


roles/matrix-synapse/defaults/main.yml
```diff
@@ -209,7 +209,8 @@ matrix_mautrix_telegram_api_hash: YOUR_TELEGRAM_API_HASH
 # Mautrix telegram public endpoint to log in to telegram
 # Use an uuid so it's not easily discoverable
 matrix_mautrix_telegram_public_endpoint: "/{{ matrix_synapse_macaroon_secret_key | password_hash('sha512', 'telegram') | to_uuid }}"
-
+# Set this to a port number to expose on the host when not using the nginx proxy
+matrix_mautrix_telegram_container_expose_port: false
```
spantaleev: If this is meant to be a port number (integer), having a default value which is a boolean is a bit strange.

We'd better use a default value of null (`~` in YAML). This way, it's either an integer or null.

In any case, both what you have now and what I'm proposing as an alternative lead to the same - `matrix_mautrix_telegram_container_expose_port` being different than all other `_expose_port` variables we have (those are true/false values only and don't specify a port). However, it's probably a good idea to break our old consistency and move toward this new model ("port number or null").

However, I'd probaly rename the variable to something that has less of a "this is a boolean" feel.
Perhaps instead of `matrix_mautrix_telegram_container_expose_port`, we could have it be named `matrix_mautrix_telegram_container_exposed_port_number`.

Once this bridge leads the way, we can redo our other `_expose_port` instances to follow suit.

roles/matrix-synapse/templates/ext/mautrix-telegram/systemd/matrix-mautrix-telegram.service.j2
```diff
@@ -21,6 +21,9 @@ ExecStart=/usr/bin/docker run --rm --name matrix-mautrix-telegram \
 			--user={{ matrix_user_uid }}:{{ matrix_user_gid }} \
 			--cap-drop=ALL \
 			--network={{ matrix_docker_network }} \
+			{% if matrix_mautrix_telegram_container_expose_port %}
```
spantaleev: If going with the `null` default value, it would be better if the check is more explicit (e.g. `{% if matrix_mautrix_telegram_container_exposed_port_number is not none %}`

paulbdavis:
Makes sense. I’ll get those changes made and update the PR (some time later this week or weekend)

spantaleev:
Thank you! :+1:

PR #105, opened 2019-03-03
Add Support for matrix-appservice-discord 

Lionstiger:
Adds the ability to bridge between Discord channels and Matrix rooms.


Lionstiger:
Credit:
Most of this is taken from https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/87
by https://github.com/Plailect

matrix-servers
```diff
@@ -0,0 +1,352 @@
+---
```
spantaleev: I'm assuming you unintentionally copied this file from `group_vars/matrix-servers` into the base directory.
It can be deleted.

docs/configuring-playbook-bridge-appservice-discord.md
```diff
@@ -0,0 +1,20 @@
...
+
+See the project's [documentation](https://github.com/Half-Shot/matrix-appservice-discord/blob/master/README.md) to learn what it does and why it might be useful to you.
+
+Setup Instructions: 
+
+loosely based on [this](https://github.com/Half-Shot/matrix-appservice-discord#setting-up)
+
+1. Create a Discord Application [here](https://discordapp.com/developers/applications/me/create).
+2. Retrieve Client ID and Bot token from this Application.
+3. Enable the bridge with "matrix_appservice_discord_enabled: true" in your vars.yml and provide id and token.
```
spantaleev: Wrapping code like this in code blocks (using backticks, instead of double-quotes) would be better.

roles/matrix-synapse/defaults/main.yml
```diff
@@ -641,3 +644,124 @@ matrix_appservice_irc_configuration_extension_yaml: |
...
+
+# Matrix Appservice Discord is a Matrix <-> Discord bridge
+# Enable Discord bridge
+matrix_appservice_discord_enabled: false
+
+matrix_appservice_discord_docker_image: "halfshot/matrix-appservice-discord:latest"
+
+matrix_appservice_discord_base_path: "{{ matrix_base_data_path }}/appservice-discord"
+
+matrix_appservice_discord_client_ID: "YOUR DISCORD APP CLIENT ID"
```
spantaleev: Since all other variables used by the playbook are in all-lowercase, it's probably better if we change this to lowercase as well (`matrix_appservice_discord_client_id`).

roles/matrix-synapse/templates/ext/appservice-discord/systemd/matrix-appservice-discord.service.j2
```diff
@@ -0,0 +1,29 @@
...
+Requires=matrix-synapse.service
+After=matrix-synapse.service
+
+[Service]
+Type=simple
+ExecStartPre=-/usr/bin/docker kill matrix-appservice-discord
+ExecStartPre=-/usr/bin/docker rm matrix-appservice-discord
+ExecStart=/usr/bin/docker run --rm --name matrix-appservice-discord \
+			--log-driver=none \
+			-e "UID={{ matrix_user_uid }}" -e "GID={{ matrix_user_gid }}" \
```
spantaleev: Since this is based on the IRC bridge, it lacks some security improvements we've added to the other bridges (telegram, whatsapp):

- running containers a a regular user from the very beginning, instead of using `root` and letting the container drop privileges later (e.g. `--user={{ matrix_user_uid }}:{{ matrix_user_gid }}`, instead of `-e "UID={{ matrix_user_uid }}" -e "GID={{ matrix_user_gid }}"`)

- dropping capabilities (`--cap-drop=ALL`)

Looks like the Discord bridge is written in TypeScript.
This is different than the Telegram bridge (written in Python) and the Whatsapp bridge (written in Go).
It's also written by a different person.. So it may be that special tweaks are necessary to enable it (besides using `--user={{ matrix_user_uid }}:{{ matrix_user_gid }}` and `--cap-drop=ALL`), compared to what we've done for the other bridges.

We can do these after this pull request gets merged!

roles/matrix-synapse/tasks/ext/appservice-discord/setup.yml
```diff
@@ -0,0 +1,73 @@
...
+    mode: 0644
+    owner: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_username }}"
+  when: "matrix_appservice_discord_enabled"
+
+- name: Ensure matrix-appservice-discord.service installed
+  template:
+    src: "{{ role_path }}/templates/ext/appservice-discord/systemd/matrix-appservice-discord.service.j2"
+    dest: "/etc/systemd/system/matrix-appservice-discord.service"
+    mode: 0644
```
spantaleev: Adding `register: matrix_appservice_discord_systemd_service_result` here will be helpful later, for making systemd automatically get reloaded when the `.service` file changes (see other review comment below).

Note: `result:` should be indented to the same level as `template:`, not nested within it.

roles/matrix-synapse/tasks/ext/appservice-discord/setup.yml
```diff
@@ -0,0 +1,73 @@
...
+  when: "matrix_appservice_discord_enabled"
+
```
spantaleev: Adding a new task here would be useful:

```yaml
- name: Ensure systemd reloaded after matrix-appservice-discord.service installation
  service:
    daemon_reload: yes
  when: "matrix_appservice_discord_enabled and matrix_appservice_discord_systemd_service_result.changed"
```

This brings it up to date with what we've done yesterday in 45618679f519408b59.
That is, it makes systemd automatically get reloaded when the `.service` file gets updated.

matrix-servers
Lionstiger: deleted. was indeed unintentional.
https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/105/commits/e54f3889663a53f0721e5760d3ee853230e0ac1b

docs/configuring-playbook-bridge-appservice-discord.md
Lionstiger: changed. https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/105/commits/b5ae0254fd4c8b24106ba9cb61a3fc2846923392

roles/matrix-synapse/defaults/main.yml
Lionstiger: changed. https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/105/commits/2d78c5f89d1e20f11e42ed4e6df44a1323b26ceb

roles/matrix-synapse/tasks/ext/appservice-discord/setup.yml
Lionstiger: added this together with the addition of the systemd reload below.
https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/105/commits/278484656b75b70e3cac70fc432ffc6b3672ce7f

roles/matrix-synapse/templates/ext/appservice-discord/systemd/matrix-appservice-discord.service.j2
Lionstiger: made these changes. because the files in the matrix/appservice-discord/ directory were previously created as root via the old service, i had to change the ownership of those files in my setup. 
i believe this wont be an issue for new installs.
learned something about git along the way, to explain the mess of force-pushes below.
https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/105/commits/c2834d222611db86ef9b3c1e4725ac07da0f35a3

Lionstiger:
All changes made. Tested in my setup and still works.

spantaleev:
Looks great!

Once again, thank you for all the hard work! :+1:

PR #107, opened 2019-03-07
Implement self-hosted Dimension server

NullIsNot0:
!NB: Code needs some work before merging!

I would like to suggest implementing self-hosted Dimension server, so Integrations start working in Riot web. I found nice project https://github.com/turt2live/matrix-dimension and tried to add to this project, but I'm not good with configuring Docker images and "matrix-dimension" container never gets started. Maybe someone can take a look at this problem, so everyone can benefit from hosting their own Dimension servers.

Config for `host-vars.yml` is like this (it also needs to be documented):
```
# Self hosted Dimension server:
matrix_dimension_enabled: true
# For information on how to acquire an access token, visit https://t2bot.io/docs/access_tokens
matrix_dimension_access_token: MDAxYmxvY2F0aW9u...obh5qKpghHYYmk7f3O9cvCg
# Users in form: ['@user1:domain.com', '@user2:domain.com']
matrix_dimension_admin: ['@myusername1:example.com', '@myusername2:example.com']

# Point Integrations to self hosted Dimension server:
matrix_riot_web_integrations_ui_url: "{{ matrix_dimension_integrations_ui_url }}"
matrix_riot_web_integrations_rest_url: "{{ matrix_dimension_integrations_rest_url }}"
matrix_riot_web_integrations_widgets_urls: "{{ matrix_dimension_integrations_widgets_urls }}"
matrix_riot_web_integrations_jitsi_widget_url: "{{ matrix_dimension_integrations_jitsi_widget_url }}"
```
Also documentation needs to be updated for users to create another DNS record: `dimension.DOMAIN.COM`


spantaleev:
Thanks for getting started on this very important addition! :+1: 

I'll try it out soon and try to find what the problem is.

roles/matrix-dimension/templates/systemd/matrix-dimension.service.j2
```diff
@@ -0,0 +1,23 @@
...
+After=docker.service
+Requires=docker.service
+
+[Service]
+Type=simple
+ExecStartPre=-/usr/bin/docker kill matrix-dimension
+ExecStartPre=-/usr/bin/docker rm matrix-dimension
+ExecStart=/usr/bin/docker run --rm --name matrix-dimension \
+			--log-driver=none \
+			--cap-drop=ALL \
```
spantaleev: Dropping capabilities is good, but we'd better also not start the container as `root`, but as a regular user.
It's done by passing a `--user` flag to the `docker run` command.

For other services we use the `matrix` user (e.g. `--user={{ matrix_user_uid }}:{{ matrix_user_gid }}`).

For Dimension, however, it seems like that won't work, as most other things are using the container's `node` user (uid=1000, gid=1000).

The Docker image can probably be improved to not be so picky, but for now, we can try to accommodate it.

I propose adding these additional variables to `roles/matrix-dimension/defaults/main.yml`:

```yaml
# The user and group id correspond to the node user in the `turt2live/matrix-dimension` image.
matrix_dimension_user_uid: 1000
matrix_dimension_user_gid: 1000
```

You can then adjust `matrix-dimension.service.j2` to contain a line like this:

```
			--user={{ matrix_dimension_user_uid }}:{{ matrix_dimension_user_gid }} \
```

For the image to be able to start though, it would need to access its `/data` files (at least `/data/config.yaml`).
We need to adjust permission for this too (see my other review comments for this).

roles/matrix-dimension/templates/config.yaml.j2
```diff
@@ -0,0 +1,80 @@
...
+
+# Homeserver configuration
+homeserver:
+  # The domain name of the homeserver. This is used in many places, such as with go-neb
+  # setups, to identify the homeserver.
+  name: "{{ matrix_domain }}"
+
+  # The URL that Dimension, go-neb, and other services provisioned by Dimension should
+  # use to access the homeserver with.
+  clientServerUrl: "https://{{ matrix_domain }}"
```
spantaleev: `clientServerUrl` should use `matrix_server_fqn_matrix`, instead of `matrix_domain`. Otherwise it will generate incorrect URLs (an example page that fails because of this is the "Sticker Packs" page).

group_vars/matrix-servers
```diff
@@ -222,6 +223,8 @@ matrix_ssl_domains_to_obtain_certificates_for: |
     ([matrix_server_fqn_matrix])
     +
     ([matrix_server_fqn_riot] if matrix_riot_web_enabled else [])
+    +
+    ([matrix_server_fqn_dimension] if matrix_dimension_enabled else [])
   }}
 
 ######################################################################
```
spantaleev: Let's add a `matrix-dimension` configuration section to this `group_vars/matrix-servers` file.
It should be added after the `matrix-coturn` section (because the sections are ordered alphabetically).

We can set `matrix_dimension_enabled: false`, because enabling it by default from the get-go seems hard (it requires a working server, a user and a working admin access token).

`false` is the default value, so the above doesn't make any difference, but:
- it explicitly sets it, which is better for getting an overview of how the playbook wires the roles
- we may add other variables to that section in the future as well

Thinking of which, we **do** need to add one more variable from the very beginning (I'm discussing it in another review comment):

```yaml
# Normally, matrix-nginx-proxy is enabled and nginx can reach Dimension over the container network.
# If matrix-nginx-proxy is not enabled, or you otherwise have a need for it, you can expose
# the Dimension HTTP port to the local host (`127.0.0.1:8184`).
matrix_dimension_container_expose_port: {{ not matrix_nginx_proxy_enabled }}
```

roles/matrix-dimension/tasks/setup_dimension.yml
```diff
@@ -0,0 +1,79 @@
...
+  when: "matrix_dimension_enabled and matrix_dimension_systemd_service_result.changed"
+
+#
+# Tasks related to getting rid of the dimension (if it was previously enabled)
+#
+
+- name: Check existence of matrix-dimension service
+  stat:
+    path: "/etc/systemd/system/matrix-dimension.service"
+  register: matrix_dimension_service_stat
```
spantaleev: Adding ` when: not matrix_dimension_enabled` to this task would be better.
We don't need to execute it, unless Dimension is disabled.

roles/matrix-dimension/tasks/setup_dimension.yml
```diff
@@ -0,0 +1,79 @@
...
+# Tasks related to setting up the dimension
+#
+
+- name: Ensure dimension base path exists
+  file:
+    path: "{{ matrix_dimension_base_path }}"
+    state: directory
+    mode: 0750
+    owner: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_username }}"
```
spantaleev: For the Dimension container to be able to access the directory (and write to it), we should use `matrix_dimension_user_gid` for `group` and a mode of `0770`.

roles/matrix-dimension/tasks/setup_dimension.yml
```diff
@@ -0,0 +1,79 @@
...
+  when: matrix_dimension_enabled
+
+- name: Ensure dimension environment variables file created
+  template:
+    src: "{{ role_path }}/templates/config.yaml.j2"
+    dest: "{{ matrix_dimension_base_path }}/config.yaml"
+    mode: 0644
```
spantaleev: We can do the following to make sure the file is readable by the Dimension container:

```
    mode: 0640
    group: "{{ matrix_dimension_user_gid }}"
```

spantaleev:
I think we can also add these additional variables to the `riot-web` section in `group_vars/matrix-servers`:

```yaml
# Use Dimension if enabled, otherwise fall back to Scalar
matrix_riot_web_integrations_ui_url: "{{ matrix_dimension_integrations_ui_url if matrix_dimension_enabled else 'https://scalar.vector.im/' }}"
matrix_riot_web_integrations_rest_url: "{{ matrix_dimension_integrations_rest_url if matrix_dimension_enabled else 'https://scalar.vector.im/api' }}"
matrix_riot_web_integrations_widgets_urls: "{{ matrix_dimension_integrations_widgets_urls if matrix_dimension_enabled else ['https://scalar.vector.im/api'] }}"
matrix_riot_web_integrations_jitsi_widget_url: "{{ matrix_dimension_integrations_jitsi_widget_url if matrix_dimension_enabled else 'https://scalar.vector.im/api/widgets/jitsi.html' }}"
```

----------

In the future, we should even consider changing the default to be privacy-first (no integration manager).

We could have a documentation page which tells people what their options are as far as Integration Managers are concerned:
- installing Dimension using the playbook
- using the publicly-hosted Dimension service - https://dimension.t2bot.io/ 
- using the proprietary Scalar service (requires explicitly setting `matrix_riot_web_` values to Scalar URLs)

I don't know if we should do it right now, given that:
- Riot on Android/iOS only use Scalar
- Dimension fails on mobile devices anyway

--------------

I also haven't tested your changes yet. I'll do it later.

I guess what's left to do is add some documentation about it.
If you can't find the time to do it, I can help you out with that.

NullIsNot0:
I'll try to deal with documentation during the weekend

NullIsNot0:
I managed to run matrix-dimension containes successfully, but clicking on Integrations button gives the blank white window. After some time I get errors in browser's console about CORS policy:
`Access to XMLHttpRequest at 'https://dimension.EXAMPLE.COM/api/v1/scalar/register' from origin 'https://riot.EXAMPLE.COM' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource`.
[See error screenshot](https://imgur.com/MO2ZPQa)
I have tried to add `add_header Access-Control-Allow-Origin *;` in `matrix-dimension.conf.j2`, but then I get error about double Access-Control-Allow-Origin. Any ideas?

roles/matrix-dimension/templates/config.yaml.j2
```diff
@@ -0,0 +1,80 @@
...
+
+# Settings for interacting with Telegram. Currently only applies for importing
+# sticker packs from Telegram.
+telegram:
+  # Talk to @BotFather on Telegram to get a token
+  botToken: "YourTokenHere"
+
+# Settings for controlling how logging works
+logging:
+  file: logs/dimension.log
```
spantaleev: A relative path like this makes Dimension write logs to `/home/node/matrix-dimension/logs/dimension.log`.

I guess we either want to:
- write logs where they would be accessible (`file: /data/logs/dimension.log`, so that they'd appear outside of the container)
- not write logs at all (`file: /dev/null` seems to work), as logging already happens via systemd/journald anyway (`journalctl -u matrix-dimension`)

Going with the latter is probably better. We'd better set `consoleLevel: verbose` as well.

roles/matrix-dimension/templates/config.yaml.j2
```diff
@@ -17,7 +17,7 @@ homeserver:
   # The URL that Dimension should use when trying to communicate with federated APIs on
   # the homeserver. If not supplied or left empty Dimension will try to resolve the address
   # through the normal federation process.
-  #federationUrl: "https://t2bot.io:8448"
+  federationUrl: "https://matrix-synapse:8448"
```
NullIsNot0: My setup does not expose matrix to other servers, so I can't use https://{{ matrix_server_fqn_matrix }}:8448 here. And I think it would me more universal solution to use Docker's network, not external one.
But this configuration still gives me error:
```
Mar 08 21:49:11 srv-matrix-riot docker[31882]: Mar-8-2019 19:49:11.005 +00:00 info [Webserver] Incoming request: POST /api/v1/scalar/register
Mar 08 21:49:11 srv-matrix-riot docker[31882]: Mar-8-2019 19:49:11.006 +00:00 verbose [matrix] Cached federation URL for EXAMPLE.COM is https://matrix-synapse:8448
Mar 08 21:49:11 srv-matrix-riot docker[31882]: Mar-8-2019 19:49:11.007 +00:00 info [matrix] Doing federated API call: https://matrix-synapse:8448/_matrix/federation/v1/openid/userinfo
Mar 08 21:49:11 srv-matrix-riot docker[31882]: Mar-8-2019 19:49:11.011 +00:00 error [matrix] Error calling /_matrix/federation/v1/openid/userinfo
Mar 08 21:49:11 srv-matrix-riot docker[31882]: Mar-8-2019 19:49:11.012 +00:00 error [matrix] connect ECONNREFUSED 172.18.0.4:8448
```

spantaleev: I agree, using the local URL is preferable (unless it causes some other trouble). Hopefully, as the comment says, `federationUrl` is only used by Dimension internally.

------------

Unless you have `matrix_synapse_federation_enabled: false` to completely disable federation, the `federation` listener is available on `http://matrix-synapse:8048` (note that this is `http`).

`https://matrix-synapse:8448` usually doesn't work, because Synapse usually does not do TLS termination by itself anymore (it's matrix-nginx-proxy that handles `matrix.DOMAIN:8448` and forwards it to `http://matrix-synapse:8048`). It's only if `matrix_synapse_tls_federation_listener_enabled: true` that `matrix-synapse:8448` is available.

In any case, if a server is federating, `http://matrix-synapse:8048` is always exposed, so that's the one we can bet on.

spantaleev:
I'm guessing your `Access-Control-Allow-Origin` troubles are because the server responded with a 500 error.
Since the response got answered by matrix-nginx-proxy (which generates the 500 error response due to inability to reach Dimension), that response doesn't contain `Access-Control-Allow-Origin`. The browser doesn't like that.

If the request were to get forwarded to Dimension, it would answer with a response having `Access-Control-Allow-Origin: *`, as expected. All is fine in that case.

We'd better ignore this.

roles/matrix-dimension/templates/config.yaml.j2
NullIsNot0: Actually I had `matrix_synapse_federation_enabled: false`. Added validation rule for this.
`http://matrix-synapse:8048` works.

spantaleev: That's a good idea!

spantaleev:
The only other thing I could think of is: switching from using a template file for the Dimension configuration to a variable.

See how we do it for mxisd in `roles/matrix-mxisd/defaults/main.yml`.
We have 3 variables:
- `matrix_mxisd_configuration_yaml` (default configuration)
- `matrix_mxisd_configuration_extension_yaml` (custom configuration override - empty by default)
- `matrix_mxisd_configuration` (final configuration value, defined as a merge of the 2 other things above)

Later on, instead of using the `template` Ansible module to generate the file, we do it like this:

```yaml
- name: Ensure mxisd config installed
  copy:
    content: "{{ matrix_mxisd_configuration|to_nice_yaml }}"
    dest: "{{ matrix_mxisd_config_path }}/mxisd.yaml"
    mode: 0644
    group: "{{ matrix_user_username }}"
  when: matrix_mxisd_enabled
```

The benefit it that people that want to have a more custom configuration (additional Dimension options or some overrides) can easily do so, merely by changing their own `vars.yml` file. We won't even need to add additional variables for tweaking all the other stuff (say someone wants to tweak the telegram `botToken`.. they can just use the new `matrix_dimension_configuration_extension_yaml` variable and define 2 lines for that override).

NullIsNot0:
OK, I've made config file to generate the same way it's done for mxisd. Dimension works fine, except for Sticker packs. It does not show images. The same problem is described in https://github.com/turt2live/matrix-dimension/issues/240 but I see that our config is ok: `name` and `clientServerUrl` are already pointed to `{{ matrix_server_fqn_matrix }}`. Have no idea what's wrong.
[See error screenshot](https://imgur.com/a/EvoL6jp)

spantaleev:
Stickers seem to work fine for me. I'm not sure what could be wrong.

Does the log for Dimension contain "Error: Can't set headers after they are sent" as described in that issue?

------------

It does seem like the Dimension integration is feature complete now.
I'm really happy with how nicely it has shaped up! Great work! :+1: 

The only thing that's left to do is to document it a bit.
If you wish, I could wait for you to add some documentation, so that we merge it all together.. or I could merge this now and the documentation part could be left for later.

NullIsNot0:
I found these logs on synapse server:
```
Mar 09 20:10:37 srv-matrix-riot docker[25311]: 2019-03-09 18:10:37,149 - synapse.access.http.8008 - 233 - INFO - GET-8055 - 10.0.1.83 - 8008 - Received request: GET /_matrix/media/r0/thumbnail/t2bot.io/9b74aaaec56ebb8b4d483fc1840827be?width=48&height=48&method=scale&animated=true
Mar 09 20:10:37 srv-matrix-riot docker[25311]: 2019-03-09 18:10:37,153 - synapse.rest.media.v1.media_repository - 393 - WARNING - GET-8055 - Not retrying destination 't2bot.io'
Mar 09 20:10:37 srv-matrix-riot docker[25311]: 2019-03-09 18:10:37,154 - synapse.http.server - 85 - INFO - GET-8055 - <XForwardedForRequest at 0x7fd186b397b8 method='GET' uri='/_matrix/media/r0/thumbnail/t2bot.io/9b74aaaec56ebb8b4d483fc1840827be?width=48&height=48&method=scale&animated=true' clientproto='HTTP/1.0' site=8008> SynapseError: 502 - Failed to fetch remote media
Mar 09 20:10:37 srv-matrix-riot docker[25311]: 2019-03-09 18:10:37,155 - synapse.access.http.8008 - 302 - INFO - GET-8055 - 10.0.1.83 - 8008 - {None} Processed request: 0.006sec/0.000sec (0.004sec, 0.000sec) (0.001sec/0.001sec/1) 62B 502 "GET /_matrix/media/r0/thumbnail/t2bot.io/9b74aaaec56ebb8b4d483fc1840827be?width=48&height=48&method=scale&animated=true HTTP/1.0" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.119 Safari/537.36" [0 dbevts]
```
But I can't see any outgoing connection in firewall to `t2bot.io`.

---
Please feel free to commit theese changes and I'll make another pull request for documentation.

spantaleev:
Merged!
Thank you for doing a great job on this! :+1: 

--------------

As for the stickers, it seems like your server can't retrieve them from t2bot.io for some reason.

I wonder if it's a configuration problem or due to your server not federating correctly. As far as I know, media-fetching should not be affected by federation, but.. I'm not sure. Something else might be going on.

In any case, it doesn't seem like a Dimension problem. Still, it's interesting to figure out..

PR #108, opened 2019-03-07
Add support for custom MXISD templates

TheLastProject:
The template can be set straight in the variable:

```
matrix_mxisd_threepid_medium_email_custom_invite_template: |
  This is an
  example template
```

### DIFF
--- a/roles/matrix-mxisd/defaults/main.yml
+++ b/roles/matrix-mxisd/defaults/main.yml
@@ -43,6 +43,16 @@ matrix_mxisd_dns_overwrite_enabled: false
 matrix_mxisd_dns_overwrite_homeserver_client_name: "{{ matrix_server_fqn_matrix }}"
 matrix_mxisd_dns_overwrite_homeserver_client_value: "http://matrix-synapse:8008"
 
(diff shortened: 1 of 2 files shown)


roles/matrix-mxisd/defaults/main.yml
```diff
@@ -43,6 +43,16 @@ matrix_mxisd_dns_overwrite_enabled: false
 matrix_mxisd_dns_overwrite_homeserver_client_name: "{{ matrix_server_fqn_matrix }}"
 matrix_mxisd_dns_overwrite_homeserver_client_value: "http://matrix-synapse:8008"
 
+# Override the default email templates
```
spantaleev: Maybe we should have a comment that says that these variables are meant to contain actual template contents (and not a file-path)?

Reading the actual task logic makes it clear, but only looking at the variables may be confusing.

TheLastProject: Agreed and changed

spantaleev:
Thank you! :+1: 

I'm sure this will come in helpful, especially for some in-production scenarios!

PR #110, opened 2019-03-08
Fix indenting of generators value

TheLastProject:
I made an indenting mistake in https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/108, causing generators to be put as a subkey of password. Oops.


spantaleev:
Oh, glad you noticed it quickly!

Thankfully breakage only happens when those variables are used, so I doubt anyone was affected.

PR #111, opened 2019-03-10
Provide Dimension documentation

NullIsNot0:
I have written short documentation about Dimension

### DIFF
--- a/docs/configuring-dns.md
+++ b/docs/configuring-dns.md
@@ -41,7 +41,7 @@ As the table above illustrates, you need to create 2 subdomains (`matrix.<your-d
 The `riot.<your-domain>` subdomain is necessary, because this playbook installs the Riot web client for you.
 If you'd rather instruct the playbook not to install Riot (`matrix_riot_web_enabled: false` when [Configuring the playbook](configuring-playbook.md) later), feel free to skip the `riot.<your-domain>` DNS record.
 
-The `dimension.<your-domain>` subdomain is necessary, because this playbook installs the Dimension integrations manager for you. Dimension installation is disabled by default and if you explicitly do not enable it's installation (`matrix_dimension_enabled: true` when [Configuring the playbook](configuring-playbook.md) later), feel free to skip the `dimension.<your-domain>` DNS record.
+The `dimension.<your-domain>` subdomain may be necessary, because this playbook could install the [Dimension integrations manager](http://dimension.t2bot.io/) for you. Dimension installation is disabled by default, because it's only possible to install it after the other Matrix services are working (see [Setting up Dimension](docs/configuring-playbook-dimension.md) later). If you do not wish to set up Dimension, feel free to skip the `dimension.<your-domain>` DNS record.
 
 
 ## `_matrix._tcp` SRV record setup (temporary requirement)
--- a/docs/configuring-dns.md
+++ b/docs/configuring-dns.md
@@ -26,6 +26,7 @@ It's a temporary requirement during the Synapse v0.99/v1.0 transition.
 | ----- | ----------------------- | -------- | ------ | ---- | ---------------------- |
 | A     | `matrix`                | -        | -      | -    | `matrix-server-IP`     |
 | CNAME | `riot`                  | -        | -      | -    | `matrix.<your-domain>` |
+| CNAME | `dimension`             | -        | -      | -    | `matrix.<your-domain>` |
 | SRV   | `_matrix._tcp`          | 10       | 0      | 8448 | `matrix.<your-domain>` |
 | SRV   | `_matrix-identity._tcp` | 10       | 0      | 443  | `matrix.<your-domain>` |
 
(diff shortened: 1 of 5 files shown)


docs/configuring-dns.md
```diff
@@ -40,6 +41,8 @@ As the table above illustrates, you need to create 2 subdomains (`matrix.<your-d
 The `riot.<your-domain>` subdomain is necessary, because this playbook installs the Riot web client for you.
 If you'd rather instruct the playbook not to install Riot (`matrix_riot_web_enabled: false` when [Configuring the playbook](configuring-playbook.md) later), feel free to skip the `riot.<your-domain>` DNS record.
 
+The `dimension.<your-domain>` subdomain is necessary, because this playbook installs the Dimension integrations manager for you. Dimension installation is disabled by default and if you explicitly do not enable it's installation (`matrix_dimension_enabled: true` when [Configuring the playbook](configuring-playbook.md) later), feel free to skip the `dimension.<your-domain>` DNS record.
```
spantaleev: There's a typo (`it's` -> `its`).

I'm also wondering if we should use some better wording to stress on the fact that Dimension is optional.
For example, instead of "is necessary" we could say "may be necessary" (if you enable it..).

We could also link to the "Setting up Dimension" page, instead of saying this: "(`matrix_dimension_enabled: true` when [Configuring the playbook](configuring-playbook.md) later)", because setting up Dimension is more involved than just `matrix_dimension_enabled: true`. It does require an access token, etc.

Example of a reworded message:

```
The `dimension.<your-domain>` subdomain may necessary, because this playbook could install the [Dimension integrations manager](http://dimension.t2bot.io/) for you. Dimension installation is disabled by default, because it's only possible to install it after the other Matrix services are working (see [Setting up Dimension](docs/configuring-playbook-dimension.md) later). If you do not wish to set up Dimension, feel free to skip the `dimension.<your-domain>` DNS record.
```

docs/configuring-playbook-dimension.md
```diff
@@ -0,0 +1,48 @@
+# Setting up Dimension (optional)
+
+Dimension integrations manager installation is disabled by default. You can enable it in your configuration file (`inventory/matrix.<your-domain>/vars.yml`):
+
+```yaml
+matrix_dimension_enabled: true
+```
+Your Matrix also needs federation enabled for Dimension to work, so if you have disabled it by setting `matrix_synapse_federation_enabled` to `false`, please remove or comment this variable in your configuration file. Or set it to:
```
spantaleev: Since federation being enabled (`matrix_synapse_federation_enabled: true`) is the default option and people need to go out of their way to disable it, I wonder if we should pay that much attention to it here.

For most people, it does water down the message of this documentation entry with needless information.

Because we have a nice `validate_config.yml`, those that enable it will be made aware that federation is necessary.
It might still make sense to have a tooltip on this page though, to prevent someone wasting some effort on Dimension, only to find out that they can't use it do to federation being disabled. So we might mention it, but maybe with less text/codeblocks. It might also be a good idea to say that federation being enabled is the default, so that we don't throw in most people into wondering about what their settings are.

Example:

For an Integration Manager like Dimension to work, your server needs to have federation enabled (`matrix_synapse_federation_enabled: true`). This is the default for this playbook, so unless you've explicitly disabled federation, you're okay.

docs/configuring-playbook-dimension.md
```diff
@@ -0,0 +1,48 @@
+# Setting up Dimension (optional)
+
+Dimension integrations manager installation is disabled by default. You can enable it in your configuration file (`inventory/matrix.<your-domain>/vars.yml`):
+
```
spantaleev: An important prerequisite for Dimension working is the `dimension.<your-domain>` DNS record being set up correctly.

Maybe it makes sense to mention it again here (and link to `configuring-dns.md`), for those that have ignored setting up the record until now.

docs/configuring-playbook-dimension.md
```diff
@@ -0,0 +1,48 @@
+# Setting up Dimension (optional)
+
+Dimension integrations manager installation is disabled by default. You can enable it in your configuration file (`inventory/matrix.<your-domain>/vars.yml`):
```
spantaleev: Maybe "Dimension integrations manager" could link to https://dimension.t2bot.io/, so that people who don't know what it is, can find out.

docs/configuring-playbook-dimension.md
```diff
@@ -0,0 +1,48 @@
...
+
+```yaml
+matrix_dimension_enabled: true
+```
+Your Matrix also needs federation enabled for Dimension to work, so if you have disabled it by setting `matrix_synapse_federation_enabled` to `false`, please remove or comment this variable in your configuration file. Or set it to:
+
+```yaml
+matrix_synapse_federation_enabled: true
+```
+
```
spantaleev: It's probably also a good idea to mention something like this in the foreword:

> **Dimension can only be installed after Matrix services are installed and running.**
> If you're just installing Matrix services for the first time, please continue with the Configuration/Installation flow and come back here later.

docs/configuring-playbook.md
```diff
@@ -54,3 +54,5 @@ When you're done with all the configuration you'd like to do, continue with [Ins
 - [Setting up Appservice IRC bridging](configuring-playbook-bridge-appservice-irc.md) (optional)
 
 - [Setting up Appservice Discord bridging](configuring-playbook-bridge-appservice-discord.md) (optional)
+
+- [Setting up Dimension (optional)](configuring-playbook-dimension.md) (optional)
```
spantaleev: Besides this (which is okay), we can also link to the same documentation page from `installing.md`.

I guess that after `installing.md`'s:

```markdown
- or [create your first user account](registering-users.md)
```

line, we could add another one:

```markdown
- or [set up the Dimension Integrations Manager](configuring-playbook-dimension.md)
```

This is because Dimension cannot be configured by going through the normal flow for the first time (clone the playbook, read all  "Configuring" pages, adjust your configuration, run Install and you're done). Setting up Dimension requires that someone sets up the other Matrix services and then comes back to set up Dimension. So we should remind people of that after they've managed to create their first Matrix user.

NullIsNot0:
Thank you for documentation improvement suggestions. I have made changes.

spantaleev:
Great! Thank you very much! :+1:

PR #112, opened 2019-03-10
Fix Dimension documentation

NullIsNot0:
1. Fix incorrect link in DNS configuration page
2. Make "Getting access token" in Dimension documentation more descriptive


spantaleev:
Good catch! Thank you! :+1:

PR #113, opened 2019-03-11
Fix Matrix homeserver name in Dimension configuration

NullIsNot0:
Today I lost few hours trying to figure out why my Dimension server does not connect to Matrix Federation server Docker container `matrix-synapse` as configured: `federationUrl: "http://matrix-synapse:8048"` The cause was `homeserver.name` in config: `matrix.HOSTNAME.COM`, but it needs to be just `HOSTNAME.COM`.
After changing name from `{{ matrix_server_fqn_matrix }}` to `{{ matrix_domain }}` Dimension started to work again for me.

### DIFF
--- a/roles/matrix-dimension/defaults/main.yml
+++ b/roles/matrix-dimension/defaults/main.yml
@@ -38,7 +38,7 @@ matrix_dimension_configuration_yaml: |
   homeserver:
     # The domain name of the homeserver. This is used in many places, such as with go-neb
     # setups, to identify the homeserver.
-    name: "{{ matrix_server_fqn_matrix }}"
+    name: "{{ matrix_domain }}"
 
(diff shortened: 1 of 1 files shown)


spantaleev:
Good catch!

I thought your initial code (9735a2f6) was doing exactly that.

Maybe after I suggested fixing `clientServerUrl` to use `matrix_server_fqn_matrix`, the `name` value also got changed to that unintentionally?

In any case, this seems like the correct thing to do now! Thank you! :+1:

NullIsNot0:
Yes, you are totally right - I changed from `{{ matrix_domain }}` to `{{ matrix_server_fqn_matrix }}` almost immediately after you suggested to change `clientServerUrl`. Thought if was correct, but it wasn't.

PR #114, opened 2019-03-12
Add option to enable Dimension widgets serve sites with self signed certs

NullIsNot0:
I tried to add custom widget with website where SSL certificate was self signed. Dimension did not allow me to do that saying `error [DimensionWidgetService] unable to verify the first certificate`.
Found out that Node needs global parameter `NODE_TLS_REJECT_UNAUTHORIZED` set to `0` to stop checking SSL certificates validity. I know - it's a security issue, but if I really need to embed such websites, I can take this risk.


spantaleev:
It doesn't hurt to have it. Thank you! :+1:

NullIsNot0:
Thank you for accepting this PR!

PR #115, opened 2019-03-12
Start appservice-irc as non-root

Plailect:
I've tested this on my personal homeserver and it works just fine.


spantaleev:
Thank you! :+1: 

With this, all containers (except Goofys) now run as non-root and with capabilities dropped!

PR #116, opened 2019-03-14
Federation self-check respects matrix_synapse_federation_enabled

lorrin:
When federation was intentionally disabled (via `matrix_synapse_federation_enabled: false`), self-check `ansible-playbook -i inventory/hosts setup.yml --tags=self-check` was failing when it got to the federation check. This PR modifies the self-check to test for working federation when `matrix_synapse_federation_enabled: true` and to test for no federation when `matrix_synapse_federation_enabled: false`.


spantaleev:
Thank you! :+1:

PR #117, opened 2019-03-16
Allow configuring push content for matrix-synapse

verb:
This allows overriding the default value for `include_content`. Setting this to false allows homeserver admins to ensure that message content isn't sent in the clear through third party servers.


spantaleev:
Merged! Thank you! :+1: 

I've added `|to_json` in 2f1662626e9a86f for consistency. For all other boolean options we already do the same.

PR #118, opened 2019-03-16
Use common path for systemctl in lets encrypt cron

verb:
Currently the nginx reload cron fails on Debian 9 because the path to systemctl is /bin/systemctl rather than /usr/bin/systemctl.

CentOS 7 places systemctl in both /bin and /usr/bin, so we can just use /bin/systemctl as the full path.

### DIFF
--- a/roles/matrix-nginx-proxy/tasks/ssl/setup_ssl_lets_encrypt.yml
+++ b/roles/matrix-nginx-proxy/tasks/ssl/setup_ssl_lets_encrypt.yml
@@ -92,7 +92,7 @@
       hour: 4
       minute: 20
       day: "*/5"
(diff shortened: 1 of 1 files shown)


aaronraimist:
It doesn't seem like Ubuntu has it there

verb:
> It doesn't seem like Ubuntu has it in `/bin`

What makes you say that? Ubuntu uses `/bin/systemctl` in 16.04 LTS, 18.04 LTS and 18.10, but I didn't check every in-between release because I expect it to be the same as Debian.

aaronraimist:
Hmm never mind. I could have sworn it wasn't there when I checked yesterday.

spantaleev:
Thanks for catching this! And thanks for verifying it for Ubuntu! :+1:

PR #119, opened 2019-03-16
Update homeserver.yaml with some new options we could enable

aaronraimist:
The sentry integration looks like it could be interesting


spantaleev:
Thanks for keeping up! :+1:

PR #120, opened 2019-03-18
Add config options to turnserver.conf

Cadair:


Cadair:
If we can think of a reliable way of getting the local ip for the host we could default the denied lists to all the standard private ranges? At the very least should we default the denied list to the private range used by docker?

spantaleev:
Hmm.. I thought that there is no need for any Docker container to talk to `matrix-coturn`?

Cadair:
Yeah I think that's true. So does that mean that container just doesn't have a docker ip range?

spantaleev:
Ah, it's just now that I've read about what `denied-peer-ip` does.. It doesn't deny access to Coturn, but rather denies requests from Coturn to other IPs.

This would be useful for preventing attacks on other private Docker services we run.

I would have said "we'd better isolate matrix-coturn into its own Docker network to guarantee it not being able to talk to anything", but looking at it now, we already do this. Coturn uses the default Docker network, not the `matrix` network that our other services use. We do it merely because Coturn doesn't need to talk to anything and nothing needs to talk to it.

Still, because this is so dangerous (for containers in the default Docker network -- none by the playbook, but there may be some others running on the host), we may wish to isolate Coturn into another network (e.g. `matrix-coturn`).

If we isolate it like this, we probably don't need to worry about using `denied-peer-ip` at all.

----------

Using `denied-peer-ip` seems complicated, by the way.

If we deny too many local ranges, people accessing their Matrix services from the local network (like you) may find that Coturn doesn't want to relay for such local addresses.

We can deny just the range that the container's network (e.g. the `matrix-coturn` network's range) uses.
But that needs to be determined at runtime (`docker network inspect --format="{{(index .IPAM.Config 0).Subnet}}" matrix-coturn`).

It also seems like we need to exclude the container's own address from the block range to support client->TURN->TURN->client flows.

Maybe it's best to avoid using `denied-peer-ip`/`allowed-peer-ip` for blocking private addresses. Those should be nicely blocked thanks to Docker's network isolation. These options would still be useful for blocking additional addresses though. I've merged this! :+1:

PR #126, opened 2019-03-22
Specify that cron is likely required on the server (fixes #125)

### LINKED ISSUE #125: /etc/cron.d is required with Let's Encrypt but not created if missing
As per the optional Let's Encrypt integration in the nginx reverse proxy (in `roles/matrix-nginx-proxy/tasks/ssl/setup_ssl_lets_encrypt.yml`), the playbook tries to install a recurrent certificate renewal through scheduled tasks (using the `ansible-cron` module. Though it is not verified that `cron` is available on the server, or maybe it is just that the drop-in directory is not available at first (`/etc/cron.d`).

I found a line referencing this issue back in #68 :
> @spantaleev 
>     * more reliable: will fail early (during playbook execution) if `/usr/bin/crontab`
>       is not available, which is more of a guarantee that cron is working fine
>       (**idea: we should probably install some cron package using the playbook**)

And indeed the playbook failed for me today on a bare VM (Debian 9.2) because the directory `/etc/cron.d` was missing. Because `cron` was not installed (and [it should be](https://docs.ansible.com/ansible/latest/modules/cron_module.html#requirements) for the `ansible-cron` module to work correctly).

After manually installing `cron` and replaying the playbook, all is fine.

Is the solution as simple as dropping a new requirement in `roles/matrix-base/tasks/setup_server_base.yml`? But the Let's Encrypt task is optional.
In my opinion, the sanest place to write this requirement down would be on the prerequisites for this playbook (`docs/prerequisites.md`) even though it is not automatic.

bibz:
When using Let's Encrypt SSL certificates, a cronjob is set up to
automatically renew them. Though it does require a `cron`-compatible
program on the server.

This fixes the error that is caused by the `/etc/cron.d` directory
not existing and the `ansible-cron` module trying to write out a
file there -- without checking if the directory exists first.

---

See #125: Since the `cron` program might be required (or not if the user has their own SSL certificates), I offer to first warn people off about it in the documentation.

### DIFF
--- a/docs/prerequisites.md
+++ b/docs/prerequisites.md
@@ -4,6 +4,8 @@
 
 - [Python](https://www.python.org/) being installed on the server. Most distributions install Python by default, but some don't (e.g. Ubuntu 18.04) and require manual installation (something like `apt-get install python`).
 
+- a `cron`-like tool installed on the server such as `cron` or `anacron` to automatically schedule the Let's Encrypt SSL certificates. *This can be ignored if you use your own SSL certificates.*
+
 - the [Ansible](http://ansible.com/) program being installed on your own computer. It's used to run this playbook and configures your server for you. Take a look at [our guide about Ansible](ansible.md) for more information, as well as [version requirements](ansible.md#supported-ansible-versions) and alternative ways to run Ansible.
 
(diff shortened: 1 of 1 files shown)


spantaleev:
Thank you! :+1: 

In most cases (non-bare distro installs), a cron program would be available, so most users can probably ignore this.

PR #132, opened 2019-04-05
Goofys Custom Endpoint Support

console-cowboy:
Goofys can work with more than just Amazon S3. _Hopefully_ this is all it takes to enable configuring a custom endpoint for custom endpoints like Minio.

### DIFF
--- a/roles/matrix-synapse/defaults/main.yml
+++ b/roles/matrix-synapse/defaults/main.yml
@@ -218,7 +218,9 @@ matrix_synapse_ext_password_provider_ldap_filter: ""
 
 
 matrix_s3_media_store_enabled: false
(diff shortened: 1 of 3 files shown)


spantaleev:
Thank you! :+1:

spantaleev:
@wabuMike, thanks for catching this! :+1: 

It was indeed missing an `if` there. I've fixed it up.

wabuMike:
Another thing that didn't work for me was `if matrix_s3_media_store_custom_endpoint_enabled is false`. Had to put `if matrix_s3_media_store_custom_endpoint_enabled == false` to make it work.

spantaleev:
I've changed that to `{% if not matrix_s3_media_store_custom_endpoint_enabled %}` in the same commit, so it should be okay now.

PR #134, opened 2019-04-14
Make Dimension communicate to Synapse through Docker network

NullIsNot0:
Media is pulled from client side, so we specify external Matrix DNS name as mediaUrl


spantaleev:
Thank you! :+1:

PR #136, opened 2019-04-16
Update prerequisites.md

brettnreno:
I discovered setup would not run if only python 3 was installed on the server.  Maybe helpful with future distros?


jdreichmann:
Out of interest: What packages were missing that were python2 only and on what distro is your server? Because IIRC i was only having trouble with python3 & it's firewalld module, and that has been resolved (Fedora 27 or 28)

spantaleev:
Since there's no feedback and this doesn't seem to be a problem for us, I'm closing it.

If anyone experiences this issue in the future, let's reopen it.

PR #139, opened 2019-04-17
add ngnix-status to config

ugs9fsy71cb:


spantaleev:
This seems useful, thank you! :+1: 

I'm also thinking it might be useful to provide one more variable - a `matrix_nginx_proxy_nginx_status_allowed_addresses` list, which lets people configure the nginx `allow` directive.

For example:

```yaml
matrix_nginx_proxy_nginx_status_enabled: true

matrix_nginx_proxy_nginx_status_allowed_addresses:
- 8.8.8.8
- 1.1.1.1
```

This new variable could default to `matrix_nginx_proxy_nginx_status_allowed_addresses: ['{{ ansible_default_ipv4.address }}']` or just to an empty list. I think both are okay.

The docs should mention an example showing `matrix_nginx_proxy_nginx_status_allowed_addresses` though.

---------

If you'd rather not work on this, I can just merge this as is or help out by doing it myself.

ugs9fsy71cb:
I added your suggests.

spantaleev:
Thank you!

I was just about to merge this, but I've noticed that you've made the changes to `matrix-domain.conf.j2`.

This configuration file is only installed when [serving the base domain](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-base-domain-serving.md) - it needs to be enabled explicitly via `matrix_nginx_proxy_base_domain_serving_enabled: true` or it will not be used.

This means that most users (those who don't serve the base domain via the playbook) can't benefit from the nginx-status feature.

------

I guess a better way to do it is to restore `matrix-domain.conf.j2` as it was and make similar changes to `matrix-synapse.conf.j2` instead.

It's the only one that is guaranteed to be available (the Riot domain may be disabled; the Dimension domain is not available by default, etc.).

If serving it from the `matirx.` domain, renaming the settings to be prefixed with `matrix_nginx_proxy_proxy_matrix_` may make sense, because all settings pertaining to the `matrix.` domain are prefixed this way.

I guess it would be something like this:

```yaml
matrix_nginx_proxy_proxy_matrix_nginx_status_enabled: true

matrix_nginx_proxy_proxy_matrix_nginx_status_allowed_addresses:
- 8.8.8.8
- 1.1.1.1
```

There's a lot of ugly word repetitions (proxy; nginx), but it's consistent with the other variables.

-----------

Besides this, the documentation page should probably mention where (on which domain) the status would be accessible. By default, there's `matrix.DOMAIN`, `riot.DOMAIN` and sometimes `dimension.DOMAIN` for now. Certain people also serve the base `DOMAIN` via the same reverse proxy configured by the playbook (as mentioned above).

In such cases, it might not be entirely obvious that it's the `matrix.DOMAIN` which serves the nginx status route.

spantaleev:
Looks like this pull request is dead, so I'll close this.

PR #144, opened 2019-04-22
make welcome.html customizable

dhoffend:


roles/matrix-riot-web/defaults/main.yml
```diff
@@ -28,3 +28,6 @@ matrix_riot_web_embedded_pages_home_path: ~
 
 # Controls whether the self-check feature should validate SSL certificates.
 matrix_riot_web_self_check_validate_certificates: true
+
+# cross use of variable defined in matrix-synapse role
+matrix_synapse_enable_registration: false
```
spantaleev: Defining variables from another "namespace" is discouraged.

A better way to do this would be to define a new `matrix_riot_web_..` variable (e.g. `matrix_riot_web_registration_enabled: false` in `matrix-riot-web/defaults/main.yml`). This is just some sane default provided by the `matrix-riot-web` role.

For ease of use, the playbook can then override this role variable by wiring it to `matrix_synapse_enable_registration`. It's possible to accomplish this by defining something like this in the **riot-web** section of the playbook's `group_vars/matrix-servers` file:

```yaml
matrix_riot_web_registration_enabled: "{{ matrix_synapse_enable_registration }}"
```

The benefits of doing it this way are that:

- Each role can be used independently of the other roles because none of them cross-reference variables (some people use the roles separately, not as part of this integrated playbook)

- thanks to the playbook's wiring (in the `group_vars/matrix-servers` file), changing `matrix_synapse_enable_registration` automatically reflects to `matrix_riot_web_registration_enabled` - a convenience feature for users of the playbook

- if necessary, `matrix_riot_web_registration_enabled` can also be controlled separately -- people can override it and set an explicit value (one that breaks the connection to `matrix_synapse_enable_registration`)

----------

You would probably notice the inconsistency in naming that I'm suggesting (`matrix_riot_web_registration_enabled`, instead of `matrix_riot_web_enable_registration`).

Generally, we aim to go for naming like `matrix_riot_web_registration_enabled`, so my proposal is consistent with that.
A lot of roles use variables named following this pattern.

The Synapse `matrix_synapse_enable_registration` variable is an outlier, because it's built like this:
- obligatory `matrix_synapse_` prefix, because it resides in the `matrix-synapse` role
- `enable_registration` suffix which is identical to the variable it controls in `homeserver.yaml`

dhoffend:
Okay I've take your comments and renamed the variable (which makes sense). I wasn't used to much logic in the defaults vars file.

roles/matrix-riot-web/defaults/main.yml
```diff
@@ -28,3 +28,6 @@ matrix_riot_web_embedded_pages_home_path: ~
 
 # Controls whether the self-check feature should validate SSL certificates.
 matrix_riot_web_self_check_validate_certificates: true
+
+# cross use of variable defined in matrix-synapse role
+matrix_riot_web_registration_enabled: "{{ matrix_synapse_enable_registration }}"
```
spantaleev: This still references a variable from another role (`matrix-synapse`) right from within the `matrix-riot-web` role.

The way we do this same thing for other roles is:

- we put a sane default value here (`matrix_riot_web_registration_enabled: false`), which doesn't reference any external variables which may not be defined. People who only use the `matrix-riot-web` role (and not the `matrix-synapse` role) outside of this playbook can still have the role function correctly, because each role can be used in isolation.

- do the same cross-referencing in another file, specific to the playbook. This is the `group_vars/matrix-servers` file for us. By adding `matrix_riot_web_registration_enabled: "{{ matrix_synapse_enable_registration }}"` to the **riot-web** section of that file, we wire things the way we want for the playbook and override the sane default value defined in the role.

roles/matrix-riot-web/defaults/main.yml
```diff
@@ -28,3 +28,6 @@ matrix_riot_web_embedded_pages_home_path: ~
 
 # Controls whether the self-check feature should validate SSL certificates.
 matrix_riot_web_self_check_validate_certificates: true
+
+# cross use of variable defined in matrix-synapse role
```
spantaleev: I guess we should adjust this comment to say what the setting does.

Mentioning the matrix-synapse role is not necessary, especially given that each role can be used independently (especially in the future - see #80)

dhoffend:
You're right, guess it was a bit late ... fixed it.

Currently there's a registration link and recovery link in the login section. But this seems to be ReactJS Code and in the riot-web code there's no config variable to disable registration at all or links to recovery function.

spantaleev:
Thank you! :+1: 

Yes, I guess a better solution would be to introduce such a "disable registration" option to Riot.
Maybe in the future, as such an option becomes available, we could switch to it.

eMPee584:
Mh.. this is good, but once you actually change the template, you have to git maintain a forked file.. 🤔

PR #145, opened 2019-04-22
disable password change api, when password providers are used

dhoffend:
This pull request is a workaround until https://github.com/matrix-org/synapse/pull/5092 lands upstream. It's really a problem when you use password_providers and users are allowed to change their password but it will only updated locally. It's not only confusing but also a security risk because you can login using 2 different passwords (local and the one from password_provider)

### DIFF
--- a/roles/matrix-nginx-proxy/defaults/main.yml
+++ b/roles/matrix-nginx-proxy/defaults/main.yml
@@ -76,6 +76,9 @@ matrix_nginx_proxy_proxy_synapse_metrics: false
 matrix_nginx_proxy_proxy_synapse_metrics_basic_auth_enabled: false
 matrix_nginx_proxy_proxy_synapse_metrics_basic_auth_key: ""
 
+# if you want to disable password change (when using external password providers)
+matrix_nginx_proxy_proxy_matrix_password_change_disabled: false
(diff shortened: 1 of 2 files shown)


aaronraimist:
It probably doesn’t make sense to include a workaround like this in the playbook. You can just run this change until your PR lands and then make a PR for the actual feature.

dhoffend:
That's okay for me. I just wanted to documented things I stumpled across.

spantaleev:
I'd say, let's see what happens with the original PR for a proper way to disable password-changing regardless of the reverse-proxy being used (some people use Apache or Caddy, etc.).

As of right now, what you're proposing can also be done with configuration like this:

```yaml
matrix_nginx_proxy_proxy_matrix_additional_server_configuration_blocks:
- "location /_matrix/client/r0/account/password { deny all; }"
```

Of course, if the original PR doesn't land, it may be nice to have the convenience variable that you're proposing.

spantaleev:
Looks like this is not necessary anymore, thanks to your Synapse pull request becoming part of v1.1.0.

In da6edc9, I've added an option for controlling the new `localdb_enabled` setting and updated the documentation for all password providers.

PR #147, opened 2019-04-25
Set Riot's enable_presence_by_hs_url to false if presence is disabled

aaronraimist:
This is kind of a hack put in place until homeservers have a way of advertising that they have disabled presence and makes Riot hide all the presence related info.

### DIFF
--- a/roles/matrix-riot-web/templates/config.json.j2
+++ b/roles/matrix-riot-web/templates/config.json.j2
@@ -14,10 +14,8 @@
 		"servers": {{ matrix_riot_web_roomdir_servers|to_json }}
 	},
 	"welcomeUserId": {{ matrix_riot_web_welcome_user_id|to_json }},
-	{% if matrix_synapse_use_presence|to_json %}
-	"enable_presence_by_hs_url": {
-		{{ matrix_riot_web_default_hs_url|to_json }}: false
-	},
+	{% if matrix_riot_web_enable_presence_by_hs_url is not none %}
+		"enable_presence_by_hs_url": {{ matrix_riot_web_enable_presence_by_hs_url|to_json }},
 	{% endif %}
 	"embeddedPages": {
 		"homeUrl": {{ matrix_riot_web_embedded_pages_home_url|to_json }}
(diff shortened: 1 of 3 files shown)


roles/matrix-riot-web/templates/config.json.j2
```diff
@@ -14,6 +14,11 @@
 		"servers": {{ matrix_riot_web_roomdir_servers|to_json }}
 	},
 	"welcomeUserId": {{ matrix_riot_web_welcome_user_id|to_json }},
+	{% if matrix_synapse_use_presence|to_json %}
```
spantaleev: Accessing variables from another namespace is discouraged.
The `matrix-riot-web` role should only access `matrix_riot_web_` variables.

This way, people can use each role by itself, without having to guess and define variables that they can't see in the role's defaults file (`roles/matrix-riot-web/defaults/main.yml`).

The way we keep roles independent like that, while still being able to wire them together is by doing 2 things:
- defining some `matrix_riot_web_` variable with a sane default value
- re-defining that same `matrix_riot_web_` variable outside of the role, in a playbook file (`group_vars/matrix-servers`). When re-defining it in that file, we can define it in terms of some other playbook variable (like `matrix_synapse_use_presence` for example).

---------

In this case, it seems most useful if we define an `matrix_riot_web_enable_presence_by_hs_url: ~` variable as a default in the `roles/matrix-riot-web/defaults/main.yml` file.

This is a sane default for people using the `matrix-riot-web` role by itself, without the rest of the playbook.

Then `roles/matrix-riot-web/templates/config.json.j2` could be like this:

```jinja2
{% if matrix_riot_web_enable_presence_by_hs_url is not none %}
"enable_presence_by_hs_url": {{ matrix_riot_web_enable_presence_by_hs_url|to_json }},
{% endif %}
```

-------

And to make the in-playbook experience more integrated and automated, we could redefine `matrix_riot_web_enable_presence_by_hs_url` in the **riot-web** section of `group_vars/matrix-servers` like this:

```jinja2
matrix_riot_web_enable_presence_by_hs_url: |
  {{
    none
    if matrix_synapse_use_presence
    else {matrix_riot_web_default_hs_url: false}
  }}
```

This way, people who use the playbook (and not roles individually) also benefit from increased integration between components.

aaronraimist: Brilliant! Thanks for writing this PR for me :)

spantaleev:
Awesome! Thank you for enabling this feature! :+1:

PR #154, opened 2019-04-27
Update mxisd (1.3.1 -> 1.4.1)

aaronraimist:


spantaleev:
Thank you! :+1: 

I've confirmed basic functionality (associating a 3pid with an account) still works and hasn't experienced any regression.

PR #157, opened 2019-04-30
Add variable to disable homeserver url preview

TheLastProject:
Self-explanatory I suppose :)

### DIFF
--- a/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
+++ b/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
@@ -525,7 +525,7 @@ max_upload_size: "{{ matrix_synapse_max_upload_size_mb }}M"
 # an explicit url_preview_ip_range_blacklist of IPs that the spider is
 # denied from accessing.
 #
-url_preview_enabled: true
+url_preview_enabled: {{ matrix_synapse_url_preview_enabled|to_json }}
 
 # List of IP address CIDR ranges that the URL preview spider is denied
 # from accessing.  There are no defaults: you must explicitly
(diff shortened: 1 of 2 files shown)


roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -525,7 +525,7 @@ max_upload_size: "{{ matrix_synapse_max_upload_size_mb }}M"
 # an explicit url_preview_ip_range_blacklist of IPs that the spider is
 # denied from accessing.
 #
-url_preview_enabled: true
+url_preview_enabled: "{{ matrix_synapse_url_preview_enabled }}"
```
spantaleev: Why add it as a string? Does Synapse understand it correctly if so?

I guess something like this would be ideal:

```
url_preview_enabled: {{ matrix_synapse_url_preview_enabled|to_json }}
```

It's what we do for other boolean variables as well.

TheLastProject: Was a bit too uncareful. Good find and fixed :)

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -525,7 +525,7 @@ max_upload_size: "{{ matrix_synapse_max_upload_size_mb }}M"
 # an explicit url_preview_ip_range_blacklist of IPs that the spider is
 # denied from accessing.
 #
-url_preview_enabled: true
+url_preview_enabled: "{{ matrix_synapse_url_preview_enabled|to_json }}"
```
spantaleev: Still wrapped in double-quotes?

TheLastProject: Wow, I am not awake this morning. Fixed for real now.

spantaleev:
Awesome! Thank you! :+1:

PR #158, opened 2019-04-30
Add the ability to update user passwords with ansible

lpopov:


Cadair:
Love the idea of this, but why did you choose to edit the database rather than use the synapse admin password reset API?

lpopov:
Because I didn't know about it :laughing:. I was looking at https://github.com/matrix-org/synapse#password-reset and decided to implement it.

Looking at the password reset API, it seems to me that it would be better for this to be implemented in synapse, similar to the `register_new_matrix_user` python script. This will allow us to not care if matrix-postgres container is used or not. After that, we can implement this functionality using the new synapse python script.

I'll make a pull request there, with an `update_matrix_user_password` script, but for now this may help someone.

Cadair:
I am confused by why that's in the readme, I was referring to this: https://github.com/matrix-org/synapse/blob/master/docs/admin_api/user_admin_api.rst#reset-password we could just call the API from ansible?

roles/matrix-synapse/tasks/update_user_password.yml
```diff
@@ -0,0 +1,48 @@
...
+    seconds: 7
+  when: start_result.changed
+
+- name: Wait a while, so that Matrix Postgres can manage to start
+  pause:
+    seconds: 7
+  when: postgres_start_result.changed
+
+- name: Generate password hash
+  shell: "/usr/local/bin/matrix-synapse-generate-password-hash {{ password }}"
```
spantaleev: Wouldn't it be possible to replace this custom script with `/usr/bin/docker exec -it matrix-synapse /usr/local/bin/hash_password -c /data/homeserver.yaml -p {{ password }}`

It's very cool how it uses `expect` to help us doing it non-interactively, but if we can, it'd be nicer to avoid it (less code & no need to install `expect`).

-------

We may even be able to have just one single script (called `matrix-postgres-set-user-password`) which does everything all at once (generates the hash and runs the SQL). I can see how keeping it separate may have some value too, but it may also be causing confusion (too many scripts in `/usr/local/bin/`).

spantaleev:
Using the API would require an admin Matrix account's access token though, so it's not as easy to use compared to just calling a script or running the playbook command.

lpopov:
@Cadair @spantaleev maybe we can read the `registration_shared_secret` from the synapse homeserver.yaml file, but it would be better for this to be in synapse, than in ansible.

spantaleev:
From what I know, `registration_shared_secret` can't be used to call any Synapse API though, except for the user registration one.

roles/matrix-synapse/tasks/update_user_password.yml
lpopov: I have missed the existence of the `-p` option... I'll change it and will implement it with one script. There is no point of having a `matrix-synapse-generate-password-hash` when there is a docker command for that.

spantaleev: Great!

lpopov:
Now that I read the API docs again, it says that it needs an `access_token` of a server admin, for resetting user passwords. It makes no sense to ask for it or to get it from the database, if we can directly write to the database.

spantaleev:
The benefit of using the password-change API is that:
- it's straightforward, documented and likely to work well even if something changes with regard to password hashing
- it doesn't require reaching into multiple containers which may not be there (like `matrix-postgres`)
- (**false**) it would work for other Matrix servers that implement the spec. So if we support something other than Synapse in the future, we can just call the API and have it work.

Now that last one would be a huge benefit, but it's not true. Even though the API endpoint is `/_matrix/client/r0/admin/reset_password/<user_id>`, it's not really part of the Matrix Client-Server API. See: https://github.com/matrix-org/synapse/issues/4850

Even if we could have all these benefits (which we don't), we would still need to obtain an admin's access token somehow. While possible (running some `SELECT` over the `access_tokens` table to find one for ANY admin user), it's hacky and may not work if there are no currently-logged-in admin users.

I guess it's easiest if we stick to doing it by reaching into the Postgres database.

roles/matrix-synapse/tasks/update_user_password.yml
lpopov: The command now uses the '-p' option to generate the password hash. `matrix-synapse-generate-password-hash` script and `expect` requirements have been removed.

spantaleev:
Awesome! Merged! :+1:

PR #159, opened 2019-04-30
Add the possibility to pass extra flags to the docker container

TheLastProject:
Untested, but submitting it to get your two cents.

I've been having to deal with a network setup where the external domain names cannot be resolved internally. Thus, I've been adding --add-host flags to the systemd files.

`--add-host` however, seems very much a very specific edge case, so I figured I should instead make it possible to add support for any extra docker arguments, so that any weird edge case can be worked around by users.

How does this seem?


spantaleev:
It seems like a great feature to have! :+1:

Let's test it and get it merged!

TheLastProject:
Can confirm it works for me :)

spantaleev:
I've also tested it just now and couldn't find any regressions.

Thank you! :+1:

PR #160, opened 2019-05-03
Fix template indentation

danbob:
This yaml template wouldn't compile until I changed all the flow control lines to two leading spaces.

### DIFF
--- a/roles/matrix-dimension/defaults/main.yml
+++ b/roles/matrix-dimension/defaults/main.yml
@@ -31,6 +31,7 @@ matrix_dimension_integrations_jitsi_widget_url: "https://{{ matrix_server_fqn_di
 # For a more advanced customization, you can extend the default (see `matrix_dimension_configuration_extension_yaml`)
 # or completely replace this variable with your own template.
 matrix_dimension_configuration_yaml: |
+  #jinja2: lstrip_blocks: True
   # The web settings for the service (API and UI).
   # It is best to have this run on localhost and use a reverse proxy to access Dimension.
   web:
@@ -125,5 +126,3 @@ matrix_dimension_configuration_extension_yaml: |
 # Holds the final Dimension configuration (a combination of the default and its extension).
 # You most likely don't need to touch this variable. Instead, see `matrix_dimension_configuration_yaml`.
 matrix_dimension_configuration: "{{ matrix_dimension_configuration_yaml|from_yaml|combine(matrix_dimension_configuration_extension, recursive=True) }}"
-
-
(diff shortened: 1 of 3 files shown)


spantaleev:
I cannot reproduce this problem with Ansible 2.7.10 and Jinja 2.10.1.
The YAML template compiles even as it is.

What is your setup like? Let's figure out what necessitates such kind of fixes.

danbob:
I have the same versions. Do you have `matrix_mxisd_threepid_medium_email_custom_templates_enabled: true`? I didn't have trouble until I set that.

I dug into it a bit more and this is sufficient to fix the problem:

```yaml
# roles/matrix-mxisd/defaults/main.yml

matrix_mxisd_configuration_yaml: |
  #jinja2: lstrip_blocks: True
  matrix: ...
```

spantaleev:
Yes.

To test it out, I've used the following configuration:

```yaml
matrix_mxisd_threepid_medium_email_custom_templates_enabled: true
matrix_mxisd_threepid_medium_email_custom_invite_template: test
```

.. and it created the configuration and template file successfully.

I wonder what `#jinja2: lstrip_blocks: True` has to do with fix this particular issue. Hmm..

danbob:
It was harder than I expected to come up with a minimum fail condition, but here it is: the problem only occurs if either of the *other* templates are set:
```yaml
matrix_mxisd_threepid_medium_email_custom_templates_enabled: true

matrix_mxisd_threepid_medium_email_custom_session_validation_template: test
# and/or
matrix_mxisd_threepid_medium_email_custom_matrixid_template: test
```
I think jinja by default only strips out the block tags themselves and not the leading whitespace unless you set that option. So the resulting yaml ends up with these blank lines with just whitespace in them, and the yaml parser is choking on some of them.

spantaleev:
Wow, great find! :+1: 

Looks like the `#jinja2: lstrip_blocks: True` fix is better (preserves the nice indentation we have and is not fragile to future additions).

We have a few more variables that this fix can be applied to:

- `matrix_dimension_configuration_yaml`
- `matrix_appservice_irc_configuration_yaml`

These don't contain any `{% if %}` logic, so they don't suffer from the problem at the moment, but it's nice if we future-proof them.

Would you update your PR to apply `#jinja2: lstrip_blocks: True` to all of these and to not change the indentation of `matrix_mxisd_configuration_yaml`?

danbob:
Done!

spantaleev:
Thanks for finding and fixing this weird issue with #108! :+1:

PR #164, opened 2019-05-06
string before to_json when string value is expected

TheLastProject:
This prevents Ansible from sometimes failing to decrypt vault variables.

I've tried to be careful and only cast string values to string before to_json, not lists or booleans, yet also as much as possible so people can use vault variables as much as possible (who knows, usernames may be vault-worthy for some). Please double check if I didn't accidentally cast too much. Proper fix for #102.

### DIFF
--- a/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
+++ b/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
@@ -255,11 +255,11 @@ listeners:
 # instance, if using certbot, use `fullchain.pem` as your certificate,
 # not `cert.pem`).
 #
-tls_certificate_path: {{ matrix_synapse_tls_certificate_path|string|to_json }}
+tls_certificate_path: {{ matrix_synapse_tls_certificate_path|to_json }}
 
 # PEM-encoded private key for TLS
 #
-tls_private_key_path: {{ matrix_synapse_tls_private_key_path|string|to_json }}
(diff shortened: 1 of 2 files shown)


roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -914,15 +914,15 @@ password_config:
 {% if matrix_synapse_email_enabled %}
 email:
    enable_notifs: true
-   smtp_host: {{ matrix_synapse_email_smtp_host|to_json }}
-   smtp_port: {{ matrix_synapse_email_smtp_port|to_json }}
+   smtp_host: {{ matrix_synapse_email_smtp_host|string|to_json }}
+   smtp_port: {{ matrix_synapse_email_smtp_port|string|to_json }}
```
spantaleev: `matrix_synapse_email_smtp_port` is a number (`587` by default in the role, and overriden to `8025` in `group_vars/matrix-servers`).

I'm not sure if Synapse casts it automatically (it probably does), but it's most likely better if we just leave it without `|string`. I don't see much reason for obfuscating a port number by encrypting the variable.

Using `|int` yields the correct numeric value, but I'm not sure if it's a working workaround for your encryption problem or if that requires passing it through `|string`. Or maybe `|string|int`?

I guess we better remove the `|string` part completely though.

spantaleev:
Looks good! :+1: 
Merged!

PR #165, opened 2019-05-06
Allow changing Riot bug report URL

TheLastProject:
Just want to make sure users of our hosted instance won't bug the Matrix team ;)


TheLastProject:
Actually, wait, this is not actually the bug report URL. I think this is where debug logs may be sent? Hmm...

aaronraimist:
That’s the same thing. Anyway unless you are customizing like the actual features of Riot for yourself, you are running the same Riot as anyone else so the bug reports are just as valid.

TheLastProject:
Well sure they are valid, but I worry corporate end-users will see the link and report "help I can't login" there, you know? Non-Riot issues. Better to be able to filter those first.

aaronraimist:
Sure. (for that issue specifically though, you can only report bugs once you are logged in)

PR #166, opened 2019-05-08
Mautrix facebook

izissise:
This add tulir's mautrix-facebook bridge to facebook messenger.

It is basically a copy paste of mautrix-telegram playbooks with slight modification :)


spantaleev:
Moar bridges! Wonderful! Thank you! :+1:

PR #167, opened 2019-05-08
Add registrations_require_3pid synapse option

danbob:


spantaleev:
Works well! Thanks! :+1:

PR #170, opened 2019-05-09
Fix matrix-synapse example for Caddy

babolivier:
The instruction that tells Caddy to make exceptions when proxying is `except`, instead of `without`, which trims the given prefix from the proxied URIs, as per https://caddyserver.com/docs/proxy

I also took the liberty to expand the examples' comments about TLS, including instructions about Caddy's built-in LE support.


spantaleev:
Thank you! :+1:

PR #175, opened 2019-05-10
Mention the Synapse User Admin API on updating-users-passwords.md

aaronraimist:


spantaleev:
Merged! :+1: 

I like how the options are separated now, but I'm not sure Option 1 should be so prominent.
I imagine that to most people "use the API" is complicated, because:
- they don't know what an access token is and how to figure out what their one is (especially if they don't use a client that shows the token)
- the [API description](https://github.com/matrix-org/synapse/blob/master/docs/admin_api/user_admin_api.rst#reset-password) is too technical, so a regular person cannot translate the description into a call like this:

```
curl -H 'Authorization: Bearer ACCESS_TOKEN_HERE' \
-XPOST \
--data '{"new_password": "PASSWORD_HERE"}' \
'https://matrix.DOMAIN/_matrix/client/r0/admin/reset_password/@USERNAME:DOMAIN'
```

Without reading the code, I'm not even sure if `<user_id>` is supposed to be a full user id (`@username:domain`) or a bare one (`username`). My example assumes a full user id is required.

----

I suppose that the [API description](https://github.com/matrix-org/synapse/blob/master/docs/admin_api/user_admin_api.rst#reset-password) can be updated to be more helpful or that we could reference it on our docs (like we already do), but also provide an example usage command.

Even if we do this, I guess that coming up with a working access token and running a command like this is definitely much more complicated than just calling the playbook (with a consistent and simple command) and letting it do all the work for you. For this reason, it's probably best if we make the playbook method the first and/or preferred choice.. We can still leave it as option number 2, but improve option 1 to:
- provide an example call (using `curl`)
- mention option 2 as a much easier alternative

aaronraimist:
Yeah that makes sense. I'll create another PR with an example and also move it down to be option 3. Personally for me the API is easier and more flexible but I see your point.

PR #177, opened 2019-05-12
Move around options on updating-users-passwords.md

aaronraimist:
and provide an example for using the admin API

Follow up to #175


spantaleev:
Nice! Thanks! :+1:

PR #178, opened 2019-05-14
Add a task to stop services (and remove containers)

eMPee584:
Complementary to the `start` tag, how about a `stop` tag? 😋


spantaleev:
Why not! Thank you! :+1:

Issue #174, opened 2019-05-10
Create a new "matrix-bridge" role to be reused with an existing installation


gnouts:
Hi all,

I already have my synapse installation (and don't want to change that part), but I'm interested in the bridge setup part...
So is it possible to use the bridges setup on my existing installation ? It's a classic debian install.

_I don't know Ansible, I'm a SaltStack guy :smile:_

spantaleev:
You can disable lots of playbook functionality, but disabling the `matrix-synapse` role is not possible (yet), because it's a crucial component that everything else builds around.

Additionally, all bridges are part of the `matrix-synapse` role and hardcode configuration values which assume a `matrix-synapse` systemd service.

Thinking about it, those bridges don't necessarily need to live in the `matrix-synapse` role and are not dependent on Synapse itself, but rather on on the [Application Service API](https://matrix.org/docs/spec/application_service/r0.1.0.html).

It might make sense to extract the brdiges into a new role (`matrix-bridge`) or have each bridge be its own separate role. I guess we'll do that at some point.

For now, I'm not sure how easy and useful it would be to disable everything else and just leave bridge installation only.

gnouts:
Thanks for the explanation. I'll keep an eye on this issue then :)

spantaleev:
I've started working on this.

I decided that instead of a single `matrix-bridge` role (which would tend to be large and increase with time), it'd be nicer if we have a separate role for each bridge.

I've already done the work that migrates the mautrix-telegram and mautrix-whatsapp bridges into their own roles. The work-in-progress code is available in the `separate-bridge-roles` branch. As it develops, it will get merged to `master` via #179.

If it's any of these 2 bridges that you'd like to use, feel free to give it a try even now, before it gets merged.

spantaleev:
I believe I've finished the work and merged the `separate-bridge-roles` branch to `master`.

Please open a new issue (or pull request) if something doesn't work as expected.

PR #179, opened 2019-05-14
Move bridges into separate roles

spantaleev:
The goal is to move each bridge into its own separate role.

Each bridge's role is meant to:

- depend only on the matrix-base role

- integrate nicely with the matrix-synapse role (if available)

- integrate nicely with the matrix-nginx-proxy role (if available and if
required). mautrix-telegram bridge benefits from integrating with
it.

- not break if matrix-synapse or matrix-nginx-proxy are not used at all

This has been provoked by #174 (Github Issue).


spantaleev:
For now, only the mautrix-telegram and mautrix-whatsapp bridges have been split out.
Work is ongoing.

spantaleev:
Bridges have been moved to their own role.

What is still missing from the `separate-bridge-roles` branch is the ability to **not** install Synapse.
We have a `matrix_synapse_enabled` variable that can be set to `false`, but it's not honored yet.
Once I do that, this will fix #174.

PR #180, opened 2019-05-14
Fix case of the mxisd ldap.connection.baseDNs option in matrix_mxisd_configuration_extension_yaml comment

eMPee584:
Else,
```
May 14 22:35:04 base docker[25166]: [main] ERROR App - You must specify at least one Base DN via the singular or plural config option                         
May 14 22:35:04 base docker[25166]: [main] ERROR App - Invalid or empty value for configuration item: ldap.connection.baseDNs
May 14 22:35:05 base systemd[1]: matrix-mxisd.service: Main process exited, code=exited, status=2/INVALIDARGUMENT                                             
```


spantaleev:
Good catch! :+1:

PR #181, opened 2019-05-14
Fix matrix_synapse_ext_password_provider_ldap_start_tls (it's boolean)

eMPee584:
With the leftover `|string` it was ending up as `start_tls: "False"` in `homeserver.yaml` which evaluates to `true`, causing unexpected
```
ldap_auth_provider - 420 - WARNING -  - Error during LDAP authentication: startTLS failed - protocolError
```


spantaleev:
Looks like a regression since #164.

Thanks for catching and fixing it! :+1:

PR #182, opened 2019-05-15
Provide a sample rest_auth_endpoint close to actual setup

eMPee584:
Well this does set it up so that the mxisd endpoint is reachable from the synapse container.


spantaleev:
Merged! :+1: 

Still, I hope this doesn't make people think that the REST auth password provider can only be used with mxisd.

I believe there might be lots of people who use it standalone and rather point it to their own server and not to mxisd. Using it doesn't require mxisd at all.

PR #185, opened 2019-05-20
Add a variable to control the stop tasks

Cadair:


Cadair:
If you ran the after playbook with no tags it starts all the services and then stops them again.

spantaleev:
Thank you for fixing this omission! :+1:

PR #188, opened 2019-05-23
Facebook API token not required

JasonLocklin:
validate_config.yml was copied from the telegram bridge code, but doesn't
apply to the Facebook bridge, as the Facebook bridge doesn't require a Facebook API/hash. 

I [confirmed](https://matrix.to/#/!PukFFdIcHgtaaHZflT:devture.com/$15586366102992gCVEF:maunium.net?via=devture.com&via=matrix.org&via=librem.one) with the bridge author.


spantaleev:
Oh, it looks like I completely missed that discussion.
Happy to hear we don't need those variables!

Thank you! :+1:

PR #190, opened 2019-05-27
Switch riot-web image (bubuntux/riot-web -> vectorim/riot-web)

spantaleev:
Switching to the official image (vectorim/riot-web) should ensure:
- there's less breakage, as it's maintained by the same team as riot-web
- there's fewer actors we need to trust
- we can upgrade riot-web faster, as newer versions should be released
on Docker hub at the same time riot-web releases are made

-----

Looks like this works, but I'd rather we not regress in terms of Docker image sizes.
See: https://github.com/vector-im/riot-web/pull/9848

We can wait a while and switch later, if/when the above gets merged.

### DIFF
--- a/roles/matrix-riot-web/defaults/main.yml
+++ b/roles/matrix-riot-web/defaults/main.yml
@@ -1,6 +1,6 @@
 matrix_riot_web_enabled: true
 
-matrix_riot_web_docker_image: "bubuntux/riot-web:v1.5.5"
+matrix_riot_web_docker_image: "vectorim/riot-web:v1.5.5"
 matrix_riot_web_docker_image_force_pull: "{{ matrix_riot_web_docker_image.endswith(':latest') }}"
 
 matrix_riot_web_data_path: "{{ matrix_base_data_path }}/riot-web"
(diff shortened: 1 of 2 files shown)


turt2live:
Just a heads-up that we're currently in "alpha" status for the image. We've not had a successful automated build, and haven't fully decided on keeping the image around. We do want to get to a point where a few releases go by without us having to interfere in Docker Hub before considering it stable.

spantaleev:
Thanks for the heads-up! :+1: 

I'll keep this open for now, so we can be reminded of it.

I'd love to switch to automatically-built (well, or even manually-built as part of the release checklist) images, done by the official riot-web team.

If you decide not to maintain a Docker image though, it's okay. In such a case, we can stick with what we have now.

aaronraimist:
@turt2live is there any update on the status of the riot-web docker image?

turt2live:
Our image seems largely stable and self-maintaining. We're changing our build system soonish, but that shouldn't affect the Docker image in any way.

I'd vote go for it and yell at us if it breaks.

spantaleev:
Awesome! Happy to merge this and keep closer to riot-web releases! :+1:

PR #194, opened 2019-06-02
Fixed error message to direct users to the right debug command

kingoftheconnors:
When an error happens in starting the server, it tells users to ` Try running systemctl status {{ item }} and systemctl -fu {{ item }}`, even though systemctl -fu doesn't exist, and it was meant to be journalctl.

It's a small fix, but finding out it was meant to be journalctl saved me a lot of grief, and it may help others too.


spantaleev:
Thanks for finding this! :+1:

PR #198, opened 2019-06-10
Add config option to be able to access database outside of container

aaronraimist:
Can probably do this with extra arguments but probably nice to just have this built in


spantaleev:
It's nice to have this role consistent with all the others and provide such an option!
Thank you! :+1: 

(by the way, I've made a minor change to the comment block)

aaronraimist:
> (by the way, I've made a minor change to the comment block)

Thanks

PR #199, opened 2019-06-11
Added a basic guide on migrating to another server

wabuMike:
A few days ago, I had to migrate matrix from one server to another.
This markdown file reflects the way I achieved migration. It's not a copy-paste step-by-step instruction, but rather a guide which I want to share. 
There might be bridges which don't migrate that well. However, I haven't experienced any issues with a basic and simple setup.


spantaleev:
Thank you! :+1:
I've adjusted your steps (in 8a7b3d5), to make it a bit easier and faster.

This also makes me think that it'd be helpful to add a migration guide for people running Synapse in another way (manually, etc.). This is going to be much more complicated though due to lots of variations.

PR #201, opened 2019-06-12
Allow default room version to be configured

aaronraimist:


spantaleev:
I've fixed up a minor double-quoting issue.

Thank you! :+1:

PR #205, opened 2019-06-21
WIP: Added Slack role

kingoftheconnors:
This is a role for using Matrix-Appservice-Slack.

There is no documentation yet, but to use it you need to define `slack_control_room_id` and `matrix_appservice_slack_enabled` in the inventory/host_vars/.../vars.yml file.

The `--tags=setup-all` and `start` work fine, but if you check the `journalctl -fu matrix-appservice-slack` you'll PROBABLY get a `Unhandled rejection Error: connect ECONNREFUSED` error. The matrix port connects to the homeserver itself, similar to the discord appservice and is right now set to 8008.

After I settle a few other projects, I hope to cycle back to this, but feel free to give it a whirl if you want!

### DIFF
--- a/roles/matrix-bridge-appservice-slack/defaults/main.yml
+++ b/roles/matrix-bridge-appservice-slack/defaults/main.yml
@@ -21,6 +21,8 @@ matrix_appservice_slack_user_prefix: 'slack_'
 matrix_appservice_slack_matrix_port: 9004
 matrix_appservice_slack_slack_port: 9003
 
+# Controls whether the appservice-slack container exposes its HTTP port (tcp/9003 in the container).
+#
 # Takes an "<ip>:<port>" or "<port>" value (e.g. "127.0.0.1:9999"), or empty string to not expose.
 matrix_appservice_slack_container_http_host_bind_port: ''
 
--- a/roles/matrix-bridge-appservice-slack/defaults/main.yml
+++ b/roles/matrix-bridge-appservice-slack/defaults/main.yml
@@ -1,0 +1,104 @@
+# matrix-appservice-slack is a Matrix <-> Slack bridge
+# See: https://github.com/matrix-org/matrix-appservice-slack
+
+matrix_appservice_slack_enabled: true
+
+matrix_appservice_slack_docker_image: "cadair/matrix-appservice-slack:latest"
+
+matrix_appservice_slack_base_path: "{{ matrix_base_data_path }}/appservice-slack"
+matrix_appservice_slack_config_path: "{{ matrix_appservice_slack_base_path }}/config"
+matrix_appservice_slack_data_path: "{{ matrix_appservice_slack_base_path }}/data"
+
+matrix_appservice_slack_public_endpoint: /appservice-slack
+matrix_appservice_slack_inbound_uri_prefix: "{{ matrix_homeserver_url }}{{ matrix_appservice_slack_public_endpoint }}"
+
+# Once you make a control room in Matrix, you can get its ID by typing any message and checking its source
+matrix_appservice_slack_control_room_id: ''
+matrix_appservice_slack_bot_name: 'slackbot'
+matrix_appservice_slack_user_prefix: 'slack_'
+
+# Controls the SLACK_PORT and MATRIX_PORT of the installation
+matrix_appservice_slack_matrix_port: 9004
+matrix_appservice_slack_slack_port: 9003
+
+# Takes an "<ip>:<port>" or "<port>" value (e.g. "127.0.0.1:9999"), or empty string to not expose.
+matrix_appservice_slack_container_http_host_bind_port: ''
+
+matrix_appservice_slack_homeserver_media_url: "matrix.{{ matrix_domain }}"
+matrix_appservice_slack_homeserver_url: "http://matrix-synapse:8008"
+matrix_appservice_slack_homeserver_domain: "{{ matrix_domain }}"
+matrix_appservice_slack_appservice_url: 'http://matrix-appservice-slack'
+
+# A list of extra arguments to pass to the container
+matrix_appservice_slack_container_extra_arguments: []
+
+# List of systemd services that matrix-appservice-slack.service depends on.
+matrix_appservice_slack_systemd_required_services_list: ['docker.service']
+
+# List of systemd services that matrix-appservice-slack.service wants
+matrix_appservice_slack_systemd_wanted_services_list: []
+
+matrix_appservice_slack_appservice_token: ''
+matrix_appservice_slack_homeserver_token: ''
+matrix_appservice_slack_id_token: ''
+
+matrix_appservice_slack_configuration_yaml: |
+    slack_hook_port: {{ matrix_appservice_slack_slack_port }}
+    inbound_uri_prefix: "{{ matrix_appservice_slack_inbound_uri_prefix }}"
+    bot_username: "{{ matrix_appservice_slack_bot_name }}"
+    username_prefix: {{ matrix_appservice_slack_user_prefix }}
+
+    homeserver:
+        media_url: "{{ matrix_appservice_slack_homeserver_media_url }}"
+        url: "{{ matrix_appservice_slack_homeserver_url }}"
+        server_name: "{{ matrix_domain }}"
+
+    dbdir: "/data"
+
+    matrix_admin_room: "{{ matrix_appservice_slack_control_room_id }}"
+
+
+matrix_appservice_slack_configuration_extension_yaml: |
+    #slack_hook_port: 9898
+    #inbound_uri_prefix: "https://my.server.here:9898/"
+    #bot_username: "slackbot"
+    #username_prefix: "slack_"
+    # Optional
+    #slack_master_token: "abc-123-def"
+    # Optional
+    #matrix_admin_room: "!aBcDeF:matrix.org"
+    #homeserver:
+    #    url: http://localhost:8008
+    #    server_name: my.server
+    # Optional
+    #tls:
+    #    key_file: /path/to/tls.key
+    #    crt_file: /path/to/tls.crt
+    #logging:
+    #    console: "info"
+    #    files:
+    #    - "./debug.log": "info"
+    #- "./error.log": "error"
+
+matrix_appservice_slack_configuration_extension: "{{ matrix_appservice_slack_configuration_extension_yaml|from_yaml if matrix_appservice_slack_configuration_extension_yaml|from_yaml else {} }}"
+
+matrix_appservice_slack_configuration: "{{ matrix_appservice_slack_configuration_yaml|from_yaml|combine(matrix_appservice_slack_configuration_extension, recursive=True) }}"
+
+matrix_appservice_slack_registration_yaml: |
+  id: "{{ matrix_appservice_slack_id_token }}"
+  as_token: "{{ matrix_appservice_slack_appservice_token }}"
+  hs_token: "{{ matrix_appservice_slack_homeserver_token }}"
+  namespaces:
+    users:
+    - exclusive: true
+      regex: '@{{ matrix_appservice_slack_user_prefix }}.*'
+    aliases:
+    - exclusive: false
+      regex: '#{{ matrix_appservice_slack_user_prefix }}.*'
+    rooms: []
+  url: "{{matrix_appservice_slack_appservice_url}}:{{ matrix_appservice_slack_matrix_port }}"
+  sender_localpart: slackbot
+  rate_limited: true
+  protocols: null
+
+matrix_appservice_slack_registration: "{{ matrix_appservice_slack_registration_yaml|from_yaml }}"
--- a/roles/matrix-bridge-appservice-slack/tasks/init.yml
+++ b/roles/matrix-bridge-appservice-slack/tasks/init.yml
@@ -1,0 +1,79 @@
+# If the matrix-synapse role is not used, `matrix_synapse_role_executed` won't exist.
+# We don't want to fail in such cases.
+- name: Fail if matrix-synapse role already executed
+  fail:
+    msg: >-
+      The matrix-bridge-appservice-slack role needs to execute before the matrix-synapse role.
+  when: "matrix_synapse_role_executed|default(False)"
+
+- set_fact:
+    matrix_systemd_services_list: "{{ matrix_systemd_services_list + ['matrix-appservice-slack'] }}"
+  when: matrix_appservice_slack_enabled|bool
+
+# If the matrix-synapse role is not used, these variables may not exist.
+- set_fact:
+    matrix_synapse_container_extra_arguments: >
+      {{ matrix_synapse_container_extra_arguments|default([]) }}
+      +
+      {{ ["--mount type=bind,src={{ matrix_appservice_slack_config_path }}/slack-registration.yaml,dst=/matrix-appservice-slack-registration.yaml,ro"] }}
+
+    matrix_synapse_app_service_config_files: >
+      {{ matrix_synapse_app_service_config_files|default([]) }}
+      +
+      {{ ["/matrix-appservice-slack-registration.yaml"] }}
+  when: matrix_appservice_slack_enabled|bool
+
+# If the matrix-synapse role is not used, `matrix_synapse_role_executed` won't exist.
+# We don't want to fail in such cases.
+- name: Fail if matrix-synapse role already executed
+  fail:
+    msg: >-
+      The matrix-bridge-appservice-slack role needs to execute before the matrix-synapse role.
+  when: "matrix_synapse_role_executed|default(False)"
+
+- block:
+  - name: Fail if matrix-nginx-proxy role already executed
+    fail:
+      msg: >-
+        Trying to append Slack Appservice's reverse-proxying configuration to matrix-nginx-proxy,
+        but it's pointless since the matrix-nginx-proxy role had already executed.
+        To fix this, please change the order of roles in your plabook,
+        so that the matrix-nginx-proxy role would run after the matrix-bridge-appservice-slack role.
+    when: matrix_nginx_proxy_role_executed|default(False)|bool
+
+  - name: Generate Matrix Appservice Slack proxying configuration for matrix-nginx-proxy
+    set_fact:
+      matrix_appservice_slack_matrix_nginx_proxy_configuration: |
+        location {{ matrix_appservice_slack_public_endpoint }} {
+        {% if matrix_nginx_proxy_enabled|default(False) %}
+        	{# Use the embedded DNS resolver in Docker containers to discover the service #}
+        	resolver 127.0.0.11 valid=5s;
+        	set $backend "{{ matrix_appservice_slack_appservice_url }}:{{ matrix_appservice_slack_slack_port }}";
+        	proxy_pass $backend;
+        {% else %}
+        	{# Generic configuration for use outside of our container setup #}
+        	proxy_pass http://127.0.0.1:{{ matrix_appservice_slack_slack_port }};
+        {% endif %}
+        }
+
+  - name: Register Slack Appservice proxying configuration with matrix-nginx-proxy
+    set_fact:
+      matrix_nginx_proxy_proxy_matrix_additional_server_configuration_blocks: |
+        {{
+          matrix_nginx_proxy_proxy_matrix_additional_server_configuration_blocks|default([])
+          +
+          [matrix_appservice_slack_matrix_nginx_proxy_configuration]
+        }}
+  tags:
+   - always
+  when: matrix_appservice_slack_enabled|bool
+
+- name: Warn about reverse-proxying if matrix-nginx-proxy not used
+  debug:
+    msg: >-
+      NOTE: You've enabled the Matrix Slack bridge but are not using the matrix-nginx-proxy
+      reverse proxy.
+      Please make sure that you're proxying the `{{ something }}`
+      URL endpoint to the matrix-appservice-slack container.
+      You can expose the container's port using the `matrix_appservice_slack_container_http_host_bind_port` variable.
+  when: "matrix_appservice_slack_enabled|bool and matrix_nginx_proxy_enabled is not defined"
--- a/roles/matrix-bridge-appservice-slack/tasks/setup_install.yml
+++ b/roles/matrix-bridge-appservice-slack/tasks/setup_install.yml
@@ -1,0 +1,46 @@
+---
+
+- name: Ensure Appservice Slack image is pulled
+  docker_image:
+    name: "{{ matrix_appservice_slack_docker_image }}"
+    source: "{{ 'pull' if ansible_version.major > 2 or ansible_version.minor > 7 else omit }}"
+
+- name: Ensure AppService Slack paths exist
+  file:
+    path: "{{ item }}"
+    state: directory
+    mode: 0750
+    owner: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_username }}"
+  with_items:
+    - "{{ matrix_appservice_slack_base_path }}"
+    - "{{ matrix_appservice_slack_config_path }}"
+    - "{{ matrix_appservice_slack_data_path }}"
+
+- name: Ensure Matrix Appservice Slack config installed
+  copy:
+    content: "{{ matrix_appservice_slack_configuration|to_nice_yaml }}"
+    dest: "{{ matrix_appservice_slack_config_path }}/config.yaml"
+    mode: 0644
+    owner: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_username }}"
+
+- name: Ensure appservice-slack registration.yaml installed
+  copy:
+    content: "{{ matrix_appservice_slack_registration|to_nice_yaml }}"
+    dest: "{{ matrix_appservice_slack_config_path }}/slack-registration.yaml"
+    mode: 0644
+    owner: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_username }}"
+
+- name: Ensure matrix-appservice-slack.service installed
+  template:
+    src: "{{ role_path }}/templates/systemd/matrix-appservice-slack.service.j2"
+    dest: "/etc/systemd/system/matrix-appservice-slack.service"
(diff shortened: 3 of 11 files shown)


kingoftheconnors:
Info about using this bridge further can be found at https://github.com/matrix-org/matrix-appservice-slack

It'll detail how to get the "control-room-id" (which is easy in Riot because you can just write any message in the control-room-channel and then use _check-source_) and how to invite the bot if you get that far.

roles/matrix-bridge-appservice-slack/defaults/main.yml
```diff
@@ -0,0 +1,92 @@
...
+matrix_appservice_slack_systemd_wanted_services_list: []
+
+matrix_appservice_slack_appservice_token: ''
+matrix_appservice_slack_homeserver_token: ''
+matrix_appservice_slack_id_token: ''
+
+matrix_appservice_slack_configuration_yaml: |
+    slack_hook_port: {{ matrix_appservice_slack_slack_port }}
+    #{{matrix_homeserver_url}} had been replaced with {{matrix_domain}} hmmmm....
+    inbound_uri_prefix: "{{ matrix_homeserver_url }}:{{ matrix_appservice_slack_slack_port }}"
```
spantaleev: I'd add another variable for this:

```yaml
matrix_appservice_slack_inbound_uri_prefix: "{{ matrix_homeserver_url }}/_matrix/slack"
```

The `/_matrix/slack` thing would need to be proxied by matrix-nginx-proxy, I guess.
See the Telegram bridge's `init.yml` to see how it injects additional nginx configuration.

This would be better than exposing the whole bridge on a public URL like `https://matrix.DOMAIN:9003`.
I don't think you can even expose it like this. Even though you can expose the `9003` port, you can't make it an HTTPS listener, because the bridge doesn't have access to SSL certificates.

.. so it's best to not expose the `9003` port at all and to let matrix-nginx-proxy handle SSL and proxying.

roles/matrix-bridge-appservice-slack/defaults/main.yml
```diff
@@ -0,0 +1,92 @@
...
+  hs_token: "{{ matrix_appservice_slack_homeserver_token }}"
+  namespaces:
+    users:
+    - exclusive: true
+      regex: '@slack_.*'
+    aliases:
+    - exclusive: false
+      regex: '#slack_.*'
+    rooms: []
+  url: "https://localhost:{{matrix_appservice_slack_matrix_port}}"
```
spantaleev: What you have (`https://localhost:{{matrix_appservice_slack_matrix_port}}`) can't work, because:
- `localhost`, as interpreted by the Synapse container means "our own (Synapse) container".. which is wrong.
- even if you make it reach the Slack container, `9003` is not an HTTPS port, so using `https://` is not okay

-----------

I would introduce a new variable for this:

```
matrix_appsevice_slack_appservice_url: http://matrix-appservice-slack:9003
```

This makes it possible to override it (if necessary).
Also, the new default value works better with containers (inside the container network).

roles/matrix-bridge-appservice-slack/tasks/setup_install.yml
```diff
@@ -0,0 +1,66 @@
+---
+
+# If the matrix-synapse role is not used, `matrix_synapse_role_executed` won't exist.
```
spantaleev: This kind of "Fail if matrix-synapse role already executed" check had better be moved to `init.yml`.
It's what we've done for the other bridges and it works better if it's there.

roles/matrix-bridge-appservice-slack/tasks/setup_install.yml
```diff
@@ -0,0 +1,66 @@
...
+    dest: "/etc/systemd/system/matrix-appservice-slack.service"
+    mode: 0644
+  register: matrix_appservice_slack_systemd_service_result
+
+- name: Ensure systemd reloaded after matrix-appservice-slack.service installation
+  service:
+    daemon_reload: yes
+  when: "matrix_appservice_slack_systemd_service_result.changed"
+
+# If the matrix-synapse role is not used, these variables may not exist.
```
spantaleev: This kind of configuration registrations works better if moved to `init.yml`.
It's what we've done for the other bridges and it works better if it's there.

Explanation for why it's better is here: 00383a73ac1b3eb791e8aec6e

roles/matrix-bridge-appservice-slack/defaults/main.yml
```diff
@@ -0,0 +1,92 @@
+# matrix-appservice-slack is a Matrix <-> Slack bridge
+# See: https://github.com/matrix-org/matrix-appservice-slack
+
+matrix_appservice_slack_enabled: true
+
+matrix_appservice_slack_docker_image: "cadair/matrix-appservice-slack:latest"
+
+matrix_appservice_slack_base_path: "{{ matrix_base_data_path }}/appservice-slack"
```
spantaleev: In other bridges, we're separating bridge configuration and data paths.

I'd introduce a few new variables:

```yaml
matrix_appservice_slack_config_path: "{{ matrix_appservice_slack_base_path }}/config"
matrix_appservice_slack_data_path: "{{ matrix_appservice_slack_base_path }}/data"
```

.. and then update everything to use those:
- `setup_install.yml` - needs to create all the paths
- `init.yml` registration needs to refer to `matrix_appservice_slack_config_path/registration.yaml`
- etc.

roles/matrix-bridge-appservice-slack/defaults/main.yml
```diff
@@ -0,0 +1,92 @@
...
+matrix_appservice_slack_configuration: "{{ matrix_appservice_slack_configuration_yaml|from_yaml|combine(matrix_appservice_slack_configuration_extension, recursive=True) }}"
+
+matrix_appservice_slack_registration_yaml: |
+  id: "{{ matrix_appservice_slack_id_token }}"
+  as_token: "{{ matrix_appservice_slack_appservice_token }}"
+  hs_token: "{{ matrix_appservice_slack_homeserver_token }}"
+  namespaces:
+    users:
+    - exclusive: true
+      regex: '@slack_.*'
```
spantaleev: I do wonder if the regex should be so generic (everything prefixed with `@slack_`), or if it should be more strict (mentioning the Matrix domain, etc.)

roles/matrix-bridge-appservice-slack/tasks/setup_install.yml
```diff
@@ -0,0 +1,66 @@
...
+
+- name: Ensure Appservice Slack paths exist
+  file:
+    path: "{{ matrix_appservice_slack_base_path }}"
+    state: directory
+    mode: 0750
+    owner: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_username }}"
+
+- name: Check if an old database file already exists
```
spantaleev: This check is unnecessary. We haven't had an older deployment of this, so there's no need for migrating data.

roles/matrix-bridge-appservice-slack/tasks/setup_install.yml
```diff
@@ -0,0 +1,66 @@
...
+    content: "{{ matrix_appservice_slack_configuration|to_nice_yaml }}"
+    dest: "{{ matrix_appservice_slack_base_path }}/config.yaml"
+    mode: 0644
+    owner: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_username }}"
+
+- name: Ensure appservice-slack registration.yaml installed
+  copy:
+    content: "{{ matrix_appservice_slack_registration|to_nice_yaml }}"
+    dest: "{{ matrix_appservice_slack_base_path }}/slack-registration.yaml"
```
spantaleev: This should use the new `matrix_appservice_slack_config_path` variable that I've mentioned above.

And instead of calling the file `slack-registration.yaml`, it can just call it `registration.yaml`.
The file is nested multiple levels into a Slack bridge directory, so it's obviously a Slack bridge registration file, without having to explicitly name it so.

roles/matrix-bridge-appservice-slack/templates/systemd/matrix-appservice-slack.service.j2
```diff
@@ -0,0 +1,37 @@
...
+ExecStartPre=-/usr/bin/docker kill matrix-appservice-slack
+ExecStartPre=-/usr/bin/docker rm matrix-appservice-slack
+
+# Intentional delay, so that the homeserver (we likely depend on) can manage to start.
+ExecStartPre=/bin/sleep 5
+
+ExecStart=/usr/bin/docker run --rm --name matrix-appservice-slack \
+			--log-driver=none \
+			--network={{ matrix_docker_network }} \
+			-v {{ matrix_appservice_slack_base_path }}:/usr/src/app/userconfig:z \
```
spantaleev: This likely needs to be adjusted, so that it mentions the 2 new variables: `matrix_appservice_slack_config_path` and `matrix_appservice_slack_data_path`.

Maybe it should be like this:

```
-v {{ matrix_appservice_slack_config_path }}:/usr/src/app/userconfig:z \
-v {{ matrix_appservice_slack_data_path }}:/data:z \
```

As for the data path, I don't see the bridge referring to any "data" or "database" directory or path, so I'm not sure how we'll use it at all. Maybe there's some hidden configuration setting or something.

In any case, it does sound better to at least make an attempt at separating `./config` from `./data`, even if we can't use it now and all files end up in `./config`.

roles/matrix-bridge-appservice-slack/defaults/main.yml
```diff
@@ -0,0 +1,92 @@
...
+
+# Once you make a control room in Matrix, you can get its ID by typing any message and checking its source
+slack_control_room_id: ''
+matrix_appservice_slack_bot_name: 'slackbot'
+
+# Controls the SLACK_PORT and MATRIX_PORT of the installation
+matrix_appservice_slack_matrix_port: 8008
+matrix_appservice_slack_slack_port: 9003
+
+matrix_appservice_slack_homeserver_address: 'http://matrix-synapse:{{ matrix_appservice_slack_matrix_port }}'
```
spantaleev: Renaming `matrix_appservice_slack_homeserver_address` to `matrix_appservice_slack_homeserver_url` is more consistent with how we use it in the configuration below.

Cadair:
@kingoftheconnors @spantaleev Is there anything in here that could be made easier by changes in the slack AS itself? I am currently making a bunch of changes to it, so I could fold in any feedback you have. 

One of the things I plan to add is making the path to the registration file configurable in the `config.json` file.

kingoftheconnors:
> 
> @kingoftheconnors @spantaleev Is there anything in here that could be made easier by changes in the slack AS itself? I am currently making a bunch of changes to it, so I could fold in any feedback you have.
> 
> One of the things I plan to add is making the path to the registration file configurable in the `config.json` file.

One thing that hurt organizationally was how the slack-registration file and databases had to be placed in a specific file together. If more freedom for where slack-registration and the database files could be found (such as saying "-d /data -f "/config" to signify looking for the databases in the /data file and the registration file in the /config file), that would help with organizing the files on the HS.

Also, the documentation could be better. The significance of 8008 and homeserverURL vs inboundURI is never explained.

spantaleev:
Off the top of my head, I can't figure out why we need to pass `-p MATRIX_PORT` to `app.js`, when we've already specified the homeserver URL in the configuration file.

Cadair:
@spantaleev it's the port the appservice listens on for pushes from synapse where as the HS url is the other way around?

@kingoftheconnors Yeah, I will look into separating the config and the data, probably have the data locations specified in the config json.

Cadair:
So it turns out there is a `dbdir:` option in `config.json` which allows you to specify where the data goes. Looks like I didn't document it anywhere though :man_facepalming:

Cadair:
I added config for the HS registration file in matrix-org/matrix-appservice-slack#143

spantaleev:
Looks like we should be able to add this to the configuration in `matrix_appservice_slack_configuration_yaml`:

```yaml
dbdir: /data
hs_registration_file: ./registration.yaml
```

This assumes that we've done the split into a `/data` directory and a config directory (mounted at `/usr/src/app/userconfig`).

I guess we also depend on the [cadair/matrix-appservice-slack:latest](https://hub.docker.com/r/cadair/matrix-appservice-slack) Docker image receiving an update. Looks like the last one is from 8 months ago.

Cadair:
I hope to push a new docker image for testing this week or next, and hopefully there will be a release of the slack AS itself in the next couple of weeks as well.

kingoftheconnors:
Awesome. Be sure to drop a comment here to let us know when the docker update gets online!

spantaleev:
According to matrix-org/matrix-appservice-slack#139, it looks like we may need to define `homeserver.media_url` as well.

Cadair:
Yes I think so.

Cadair:
I have pushed a new docker image with the tag "cadair", it folds in a lot of my open PRs at the moment. It should have the latest updates to specifying the registration path via the cli

roles/matrix-bridge-appservice-slack/defaults/main.yml
kingoftheconnors:
```
      matrix_mautrix_telegram_matrix_nginx_proxy_configuration: |
        location {{ matrix_mautrix_telegram_public_endpoint }} {
        {% if matrix_nginx_proxy_enabled|default(False) %}
        	resolver 127.0.0.11 valid=5s;
        	set $backend "matrix-mautrix-telegram:8080";
        	proxy_pass http://$backend;
        {% else %}
        	proxy_pass http://127.0.0.1:9006;
        {% endif %}
        }
```

I'm new to all of this, so bear with me
What should LOCATION `(matrix_mautrix_telegram_public_endpoint)` be for an appservice? Does matrix-appservice-slack need a specific location?

Cadair:
matrix-org/matrix-appservice-slack#143 is merged fyi.

roles/matrix-bridge-appservice-slack/defaults/main.yml
kingoftheconnors: This is something that the user should be able to customize, so I'll add a variable for it.

kingoftheconnors:
> 
> 
> According to [matrix-org/matrix-appservice-slack#139](https://github.com/matrix-org/matrix-appservice-slack/issues/139), it looks like we may need to define `homeserver.media_url` as well.

Would that be `matrix.{{matrix_domain}}` or just `{{matrix_domain}}`?

kingoftheconnors:
@Cadair is  `-p {{matrix_appservice_slack_matrix_port}}` still needed?

Cadair:
Yes, you have to specify to the AS what port the HS is attempting to connect to.

spantaleev:
> > According to [matrix-org/matrix-appservice-slack#139](https://github.com/matrix-org/matrix-appservice-slack/issues/139), it looks like we may need to define `homeserver.media_url` as well.
> 
> Would that be `matrix.{{matrix_domain}}` or just `{{matrix_domain}}`?

It should be `matrix.{{ matrix_domain }}`.

roles/matrix-bridge-appservice-slack/defaults/main.yml
spantaleev: I think you just need to match whatever you pass as the path in `inbound_uri_prefix`.

Example:

```yaml
matrix_appservice_slack_public_endpoint: /appservice-slack

matrix_appservice_slack_inbound_uri_prefix: "{{ matrix_homeserver_url }}{{ matrix_appservice_slack_public_endpoint }}"
```

`config.yaml`:

```yaml
inbound_uri_prefix: {{ matrix_appservice_slack_inbound_uri_prefix }}
```

And then for nginx you can use `location {{ matrix_appservice_slack_public_endpoint }}`

roles/matrix-bridge-appservice-slack/defaults/main.yml
spantaleev: For other bridges we do the obvious and safe thing (only allowing bridging for users on this homeserver). If people want to change it, we just ask them to override the whole registration configuration (e.g. `matrix_appservice_slack_registration_yaml`).

However, adding a variable sounds better and makes overriding easier!

roles/matrix-bridge-appservice-slack/defaults/main.yml
kingoftheconnors: So the endpoint **won't** be /_matrix/slack, but instead /appservice-slack? I assume it doesn't matter what the endpoint is as long as it matches the nginx proxy?

spantaleev: That's right. It doesn't matter what it is, as long as it's the same.

I first thought about `/_matrix/slack`, but I guess `/appservice-slack` is better. Putting it under `/_matrix` makes it sound like it's part of the Matrix Client-Server API or something of that sort. Well, that one is at `/_matrix/client`, so there is no confusion, but still..

Maybe in the future we'll support another Slack bridge (different than AppService Slack) and this one using something like `/slack` seems wrong. Better to use the full name (`appservice-slack`) anyway, regardless if it's under `/_matrix` or not.

kingoftheconnors: I tested out the current version today and I got an error from matrix-synapse.confd. It included the block:
```
[
a
n
s
i
b
l
e
u
n
d
e
f
i
n
e
d
]
```

Is this a fault of my nginx setup or something else?

spantaleev: The block was part of the configuration or you got it as an error message when you ran the playbook?

kingoftheconnors: It was part of the configuration, specifically `matrix-synapse.confd`

roles/matrix-bridge-appservice-slack/tasks/init.yml
```diff
@@ -0,0 +1,79 @@
...
+        {% endif %}
+        }
+
+  - name: Register Mautrix Telegram proxying configuration with matrix-nginx-proxy
+    set_fact:
+      matrix_nginx_proxy_proxy_matrix_additional_server_configuration_blocks: |
+        {{
+          matrix_nginx_proxy_proxy_matrix_additional_server_configuration_blocks|default([])
+          +
+          [matrix_mautrix_telegram_matrix_nginx_proxy_configuration]
```
spantaleev: This reference a variable used by the Telegram bridge. You'd wish to change it to `matrix_appservice_slack_matrix_nginx_proxy_configuration`.

This should fix the `undefined` problem you've mentioned.

roles/matrix-bridge-appservice-slack/defaults/main.yml
spantaleev: Looks like it's because `matrix_mautrix_telegram_matrix_nginx_proxy_configuration` needs to be changed to `matrix_appservice_slack_matrix_nginx_proxy_configuration`

kingoftheconnors:
So the bridge now can be reached, the proxied url (matrix.DOMAIN/appservice_slack) is reachable by the slack api, and I've been able to follow the directions by the appservice README, so a lot more progress has been made.

Right now there's a problem in linking channels, since the bridge doesn't seem to recognize when channels are linked and just says 
```Oops - SlackEventHandler failed: UnknownChannel { channel: 'CJVQ3ATT8' }```
to all messages, regardless of which channel it's in.

**Comment by spantaleev** [comment]
I wonder if that's a bridge problem or something that's wrong with our configuration.
@Cadair?

**Comment by Cadair** [comment]
I have also commented on the AS issue. Make sure that the bridge is getting events from synapse, specially the link commands.

**Comment by spantaleev** [comment]
Looks like this might be fully functional now, if networking is fixed?

**Comment by kingoftheconnors** [comment]
Networking has been fixed. There's a small to-do list left though:

- [ ] I'm having trouble with Slack->Matrix messages. I need to get to the bottom of that first
- [ ] I need to make a README
- [ ] I need to squash all the commits together to save space on the timeline

**Comment by kingoftheconnors** [comment]
Also, the slack role doesn't implement the `cadair` image. I know a few of cadair's PRs have been pulled in, so do I still need to do that?

**Comment by kingoftheconnors** [comment]
It works! Maybe try it out yourself, but if my bridge is anything to go off of, the slack role is functional!

... Although it still uses the `:latest` docker image rather than the `:cadair` one.

**Comment by Cadair** [comment]
Now 0.3 is out I should rebuild the latest image. Although I should find out about the official docker image as well.

**Comment by Cadair** [comment]
The latest tag of the docker image is now the master branch of the repo, which is the 0.3 release. So we should be good to go.

I have had a read over this and it looks great. I might try and deploy it on my personal server today.

**Comment by kingoftheconnors** [comment]
It seems media_URL may be wrong, since the profile pics aren't being bridged. I'll be taking a look at this later today.

**Comment by kingoftheconnors** [comment]
Alright, fixed! I just forgot to update my docker image. 😆

File: roles/matrix-bridge-appservice-slack/templates/systemd/matrix-appservice-slack.service.j2 [hunk 12]
```diff
@@ -0,0 +1,38 @@
...
+ExecStart=/usr/bin/docker run --rm --name matrix-appservice-slack \
+			--log-driver=none \
+			--network={{ matrix_docker_network }} \
+			-v {{ matrix_appservice_slack_config_path }}:/config:z \
+			-v {{ matrix_appservice_slack_data_path }}:/data:z \
+			{% for arg in matrix_appservice_slack_container_extra_arguments %}
+			{{ arg }} \
+			{% endfor %}
+			{{ matrix_appservice_slack_docker_image }} \
+            node app.js -c /config/config.yaml -f /config/slack-registration.yaml -p {{matrix_appservice_slack_matrix_port}}
```
spantaleev: Indentation is a little off?

roles/matrix-bridge-appservice-slack/templates/systemd/matrix-appservice-slack.service.j2
```diff
@@ -0,0 +1,38 @@
...
+
+[Service]
+Type=simple
+ExecStartPre=-/usr/bin/docker kill matrix-appservice-slack
+ExecStartPre=-/usr/bin/docker rm matrix-appservice-slack
+
+# Intentional delay, so that the homeserver (we likely depend on) can manage to start.
+ExecStartPre=/bin/sleep 5
+
+ExecStart=/usr/bin/docker run --rm --name matrix-appservice-slack \
```
spantaleev: Would be nice to make this `docker run` command more secure:
- using `--user={{ matrix_user_uid }}:{{ matrix_user_gid }}` or something similar -- running as a regular user, not `root`
- using `--cap-drop=ALL` to drop capabilities and additionally increase security

roles/matrix-bridge-appservice-slack/templates/systemd/matrix-appservice-slack.service.j2
```diff
@@ -0,0 +1,38 @@
...
+			--log-driver=none \
+			--network={{ matrix_docker_network }} \
+			-v {{ matrix_appservice_slack_config_path }}:/config:z \
+			-v {{ matrix_appservice_slack_data_path }}:/data:z \
```
spantaleev: Maybe we need an extra thing here, to handle the case where matrix-nginx-proxy is disabled.

Example:

```

			{% if matrix_appservice_slack_container_http_host_bind_port %}
			-p {{ matrix_appservice_slack_container_http_host_bind_port }}:{{ matrix_appservice_slack_slack_port }} \
			{% endif %}
```

.. which necessitates the following new variable in `defaults/main.yml`:

```yaml
# Controls whether the appservice-slack container exposes its HTTP port (tcp/9003 in the container).
#
# Takes an "<ip>:<port>" or "<port>" value (e.g. "127.0.0.1:9003"), or empty string to not expose.
matrix_appservice_slack_container_http_host_bind_port: ''
```

and the following definition in `group_vars/matrix_servers`:

```yaml
matrix_appservice_slack_container_http_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:{{ matrix_appservice_slack_slack_port }}' }}"
```

-------

I see that you already have a `matrix_appservice_slack_container_http_host_bind_port` variable in `defaults/main.yml`, but its comment block seems inaccurate.

spantaleev:
Seems like this is coming along nicely!

Thank you for all the hard work @kingoftheconnors and @Cadair!

spantaleev:
Anything else that needs to be done before this gets merged?

kingoftheconnors:
Not right now. There's currently work for an official docker image for the appservice, but that probably won't be up for a while. Once that's up, I'll probably open a new PR for that one. This role works and I'm using it for my own server as we speak!

Cadair:
There is already an official image for develop, just not stable yet.

On 18 August 2019 17:47:59 BST, kingoftheconnors <notifications@github.com> wrote:
>Not right now. There's currently work for an official docker image for
>the appservice, but that probably won't be up for a while. Once that's
>up, I'll probably open a new PR for that one. This role works and I'm
>using it for my own server as we speak!
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/205#issuecomment-522337279

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

kingoftheconnors:
Hm. Should we wait for that to stabilize or move over to that one now?

Cadair:
I will check with Halfy.

On 18 August 2019 17:54:00 BST, kingoftheconnors <notifications@github.com> wrote:
>Hm. Should we wait for that to stabilize or move over to that one now?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/205#issuecomment-522337685

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Cadair:
I suggest we stay with this one until we get a stable release of the official image. I think this is ready to merge.

spantaleev:
It's been a long ride, but I'm happy to finally merge this, so that everyone could start bridging to Slack easily!

Thank you for the great work and perseverance! :+1:

Announced in dd0f35562bd9b188.

PR #207, opened 2019-06-23
Whatsapp config playbook managed

tommes0815:
finally I was able to work on the playbook as promised a while ago. Since you converted the Telegram bridge already, I did it for the WhatsApp bridge. I basically copied your work. I tested it on my server and it works. Maybe we should ask more people to test?

Is anything else needed? Maybe it would help inexperienced admins to also extend the doc files for bridges and explain how to extend the default configurations with an example?

Best
`@thomas:linklayer.de`


spantaleev:
It looks great!

Thanks, Thomas! :+1:

spantaleev:
I've also just done some similar updates:
- making the Whatsapp bridge talk to Synapse using the in-container network - 3ff57ed74daf3ec
- using `|regex_escape` in `registration.yaml` - c876a7df1d8103cd3263

.. and getting inspiration from the stricter regex in the Whatsapp bridge's `registration.yaml` to make the other bridges the same - 37c8b96d06a080.

tommes0815:
happy to help (and to learn things on the way :D)

PR #213, opened 2019-06-29
Reduce journal logging

tommes0815:
Hey,
I used some local changes to reduce the logging to the systemd journal and thought I might upstream them. I basically added two options, one for setting an extra console logging level for synapse and one option to log the nginx access to a file instead of the console. With the synapse console level changed its still logs with the default level to the homeserver.log file (so you can have warning to the console and debug to the file). The nginx option is a bool and stops the access log to the journal completely.

Are those interesting for other playbook users?

### DIFF
--- a/roles/matrix-nginx-proxy/defaults/main.yml
+++ b/roles/matrix-nginx-proxy/defaults/main.yml
@@ -25,6 +25,9 @@ matrix_nginx_proxy_container_additional_volumes: []
 # A list of extra arguments to pass to the container
 matrix_nginx_proxy_container_extra_arguments: []
 
+# Controls whether nginx logs access to stdout of the container (and so to journald) or to the file /matrix/nginx-proxy/access.log
+matrix_nginx_access_log_to_file: false
+
 # Controls whether matrix-nginx-proxy should serve the base domain.
 #
 # This is useful for when you only have your Matrix server, but you need to serve
--- a/roles/matrix-nginx-proxy/tasks/setup_nginx_proxy.yml
+++ b/roles/matrix-nginx-proxy/tasks/setup_nginx_proxy.yml
@@ -195,7 +195,7 @@
     state: absent
   when: "not matrix_nginx_proxy_proxy_synapse_metrics_basic_auth_enabled|bool or not matrix_nginx_proxy_proxy_synapse_metrics|bool"
 
-- name: Ensure Matrix nginx-proxy configuration for base domain deleted
+- name: Ensure Matrix nginx-proxy access log deleted
   file:
     path: "{{ matrix_nginx_proxy_base_path }}/access.log"
     state: absent
--- a/roles/matrix-nginx-proxy/tasks/setup_nginx_proxy.yml
+++ b/roles/matrix-nginx-proxy/tasks/setup_nginx_proxy.yml
@@ -115,6 +115,14 @@
     daemon_reload: yes
   when: "matrix_nginx_proxy_enabled and matrix_nginx_proxy_systemd_service_result.changed"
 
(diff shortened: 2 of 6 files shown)


spantaleev:
I guess it's better to move in the other direction - try to get rid of as much file logging as possible and leave it to `journald`. This way, we don't litter the `/matrix` directory with useless logs everywhere, making it easier to backup the important parts, etc.

For bridges and other services, we've been getting rid of file logging where possible.
There may be some still remaining and it would be nice to get rid of them as well.

For Synapse, we still do log to a file, because it's quite chatty and has the possibility to cause journald to drop its messages on some distros (e.g. CentOS 7). It'd be nice to get rid of these log files as well and encourage people to tweak their rate limits in `/etc/systemd/journald.conf` (`RateLimitInterval=0` and `RateLimitBurst=0` work around it).

I do like the idea for a controllable `matrix_synapse_console_log_level` though. Is `INFO` the current default (when omitted)? Letting people increase/decrease logging to console sounds good.

I've seen a lot of issues about Synapse likely becoming less chatty (e.g. matrix-org/synapse#4751), but it doesn't seem like much is happening. So the situation is likely to stay the same for a long time still. We should probably just act bold and pull the plug on Synapse's logging to a file.

-------

If we go for optionally providing file logging, things will likely get complicated.
For example, for nginx, people would probably also want to specify `error_log` as well?

By the way, your need for nginx logging to a file can probably be solved by just mounting a volume to the original `/var/log/nginx/access.log` file (symlinked to `/dev/stdout` by default).
You can do this using:

```yaml
matrix_nginx_proxy_extra_args: ['--mount type=bind,src=/matrix/nginx-proxy/access.log,dst=/var/log/nginx/access.log']
```

You'd need to ensure the `/matrix/nginx-proxy/access.log` file is created and accessible first, though.

spantaleev:
I've moved the playbook in the opposite direction with ef5e4ad -- Synapse no longer logs to text files.

I think we should also fix up of any remaining services that log to text files (some bridges like the Whatsapp bridge seem to insist on doing it).

I think this direction makes this pull request obsolete, so I'm going to close it now.

tommes0815:
First, sorry for answering so late. Somehow I agree and disagree at the same time. I see your point of moving away of the file logging especially to the /matrix directory (maybe /var/log would have been better).

But on the other hand even through journald is powerful and comfortable it lacks the option to print the log without a certain unit (kind of an inverted -u). Thats why I find it not handy to use with very chatty services (like synapse, nginx access...). When you look at the complete log you just can not find anything anymore. I like to use it to get a general overview of the server with only major entries (like warnings). Thats why the option to log with two different levels to a file and the journal was useful.

Or maybe I just don't know some nice way of getting a general overview without showing chatty services, any recommendations?

I assume you won't change your mind. So I guess I will maintain the file based logging in my fork and rebase to you whenever there is an update.

spantaleev:
You can tail the whole log (without `-u`) and then invert-`grep` for `matrix-synapse`.
Example: `journalctl -f | grep -v matrix-synapse`

---------

As for keeping your own fork, if it's just for nginx's access log file, perhaps you don't need a fork if you use the `matrix_nginx_proxy_extra_args` method described in [my previous comment](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/213#issuecomment-507138856).

tommes0815:
Back when I started playing around with the logging I tested the invert `grep` and was somehow not satisfied. But I forgot why... Maybe I give the it an other try and see how satisfied I stay :) Maybe I will just set most of the logging to `warning` as I haven't used the synapse more verbose log anyway.
Thanks for the help

PR #218, opened 2019-07-17
introduce configuration to change riot branding / title

swarnat:
Not really big PR, but probably useful for somebody


spantaleev:
Great! Thank you! :+1:

PR #228, opened 2019-07-26
Fix apt message: docker doesn't support arch 'i386'

danbob:
This fixes the apt/apt-get warning
`N: Skipping acquire of configured file 'stable/binary-i386/Packages' as repository 'https://download.docker.com/linux/<distro> <version> InRelease' doesn't support architecture 'i386'`
when updating your system.


spantaleev:
Thanks! :+1: 

This does force us into the `x86_64` / `amd64` architecture, but.. even though Docker supports `armhf`, it's not like anything else in this playbook would work on another architecture anyway :)

PR #229, opened 2019-07-28
add hangouts bridge by copying facebook bridge and find-replacing

microchipster:
This is just a naive copy and find-replace of the facebook -> hangouts right now. I imagine there are some details that need to be worked out, but posting here for discussion.

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -125,6 +125,36 @@ matrix_mautrix_facebook_homeserver_token: "{{ matrix_synapse_macaroon_secret_key
 
 ######################################################################
 #
+# matrix-bridge-mautrix-hangouts
+#
+######################################################################
+
+# We don't enable bridges by default.
+matrix_mautrix_hangouts_enabled: false
+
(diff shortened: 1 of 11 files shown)


spantaleev:
I see that the proxying for `/login` has been fixed to go to `http://matrix-mautrix-hangouts:8080`.

Still, this PR modifies the `matrix-nginx-proxy` role, which it shouldn't, as [mentioned before](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/229#commitcomment-34484773).

microchipster:
FYI this config seems to work now.

spantaleev:
Looks like we just need some documentation page in `docs/` for this bridge.

Would you adapt the Facebook bridge's docs to Hangouts based on your experience?

Not that we can't merge it without it documentation, but it will be more difficult for people to get started.

spantaleev:
Anything else you'd like to do before we merge this?

I see that "admin mode" is still enabled for all homeserver users. We'd better change that before merging. It either needs to be made configurable, or we should at least hardcode safer values (no admin mode at all).

PR #232, opened 2019-08-04
Clarify sentence about dedicated user for access token

jooize:


spantaleev:
Sounds good! Thank you! :+1:

PR #236, opened 2019-08-08
Added possibility to enable guest access on synapse

oleg-fiksel:


roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -749,7 +749,7 @@ turn_shared_secret: {{ matrix_synapse_turn_shared_secret|string|to_json }}
 # connect to arbitrary endpoints without having first signed up for a
 # valid account (e.g. by passing a CAPTCHA).
 #
-turn_allow_guests: False
+turn_allow_guests: {{ matrix_synapse_turn_allow_guests }}
```
spantaleev: `|to_json` here would ensure we really pass a boolean.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -854,7 +854,7 @@ registration_shared_secret: {{ matrix_synapse_registration_shared_secret|string|
 # participate in rooms hosted on this server which have been made
 # accessible to anonymous users.
 #
-#allow_guest_access: false
+allow_guest_access: {{ matrix_synapse_allow_guest_access }}
```
spantaleev: `|to_json` here would ensure we really pass a boolean.

spantaleev:
Thank you! :+1:

PR #245, opened 2019-08-13
missing Map /.well-known/matrix/server for server discovery

dreamflasher:


examples/apache/matrix-synapse.conf
```diff
@@ -50,6 +51,15 @@
 		Header always set Content-Type "application/json"
 		Header always set Access-Control-Allow-Origin "*"
 	</Location>
+	# Map /.well-known/matrix/server for server discovery
+	Alias /.well-known/matrix/server /matrix/static-files/.well-known/matrix/server
+	<Files "/matrix/static-files/.well-known/matrix/server">
+		Require all granted
+	</Files>
+	<Location "/.well-known/matrix/server>
```
spantaleev: This probably needs an ending `"`?

Same goes for the existing `<Location "/.well-known/matrix/server>` line.

examples/apache/matrix-synapse.conf
```diff
@@ -50,6 +51,15 @@
...
+		Header always set Content-Type "application/json"
+		Header always set Access-Control-Allow-Origin "*"
```
spantaleev: `Header always set Access-Control-Allow-Origin "*"` is not necessary for the `server` file, as it's not meant to be used by browsers and CORS rules don't apply.

dreamflasher:
Both points right! Fixed.

examples/apache/matrix-synapse.conf
```diff
@@ -56,10 +56,6 @@
 	<Files "/matrix/static-files/.well-known/matrix/server">
 		Require all granted
 	</Files>
-	<Location "/.well-known/matrix/server>
```
spantaleev: Perhaps we still need the `Header always set Content-Type "application/json" header?

The [Server-Server spec](https://matrix.org/docs/spec/server_server/r0.1.3#get-well-known-matrix-server) says:

> The Content-Type for this response SHOULD be application/json, however servers parsing the response should assume that the body is JSON regardless of type.

.. so it doesn't seem to be strictly necessary to serve it with `Content-Type: application/json`, but it's better if it is.

spantaleev:
I've added the remaining thing -- `Content-Type` to `/.well-known/matrix/server` in d675cb3 and merged this.

Thank you! :+1:

PR #246, opened 2019-08-15
Update synapse (1.2.1 -> 1.3.0)

thedanbob:


spantaleev:
I haven't tested it, but it should work, so I'm merging as-is!
Thank you! :+1:

Ideally, we'd merge the new configuration options introduced by 1.3. I can do that tomorrow.

PR #247, opened 2019-08-16
Add a variable to disable server base setup (f.e. docker repo/package)

eMPee584:
We use debops to set up the server, and don't want to add the docker-ce repo but try the debian builtin instead. This commit is just an easy way to achieve that - how about the semantics? Any proposals how to do this very welcome.

### DIFF
--- a/roles/matrix-base/defaults/main.yml
+++ b/roles/matrix-base/defaults/main.yml
@@ -22,6 +22,9 @@ matrix_user_username: "matrix"
 matrix_user_uid: 991
 matrix_user_gid: 991
 
+# Wether to handle installation of docker and ntp.
+matrix_install_server_packages: true
+
 matrix_base_data_path: "/matrix"
 matrix_base_data_path_mode: "750"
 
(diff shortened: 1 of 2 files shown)


spantaleev:
I've heard of others potentially needing something like this too, so this would be a helpful addition! Thank you!

We normally use these `run_` variables to control which parts of the roles run, when the roles are used inside another playbook (people use our roles in their own playbooks).

In this case, you'd like to run the `matrix-base` role but rather skip some parts of it.
I suppose using another variable (default `matrix_install_server_packages: true` in `roles/matrix-base/defaults/main.yml`) would make more sense than using a `run_` variable.

eMPee584:
ok.. yeah the run variable was just a quick impromptu hack to invite your comment 😋
-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

eMPee584:
Ok I renamed the variable.. before merging I would suggest splitting off the package install into a `roles/matrix-base/tasks/install_server_packages.yml`.. — ?

spantaleev:
Do you mean renaming `setup_server_base.yml` to `install_server_packages.yml`?

.. because there's hardly anything in the current `setup_server_base.yml` that is not a package-install or post-install operation.

eMPee584:
well there's the installation part and the part ensuring these two services (docker and ntp) are running.. this patch's current state is not yet semantically convincing to me..

spantaleev:
Fixed by #312

PR #249, opened 2019-08-19
Describe setting up SMTP server with Sendgrid

Munfred:
I had some troubles with email because, as you correctly pointed out a few days ago, Google Cloud blocks port 25. For using an external SMTP server, I found that the Sendgrid free tier is pretty generous and hassle free, so I set it up on the friendserver I'm running. I'm documenting how to do it here because it took me a few tries to get all the configurations correct, and might be useful for other people using Google cloud.


spantaleev:
Looks helpful! Thank you! :+1:

PR #254, opened 2019-08-25
Add haproxy doc

Sudneo:
This PR adds brief documentation on how to use HAproxy in front of Matrix and proposes a sample configuration for HAproxy and for a dummy Nginx container. 

It might be a corner case, but this was the case I had to solve for myself and it might be useful for someone as well.

### DIFF
--- a/examples/haproxy/docker-compose.yml
+++ b/examples/haproxy/docker-compose.yml
@@ -1,8 +1,8 @@
 version: '3'
 services:
-  neginx:
+  nginx:
     image: local/nginx 
     ports:
       - 40888:80
     volumes:
-      - /matrix/static-files:/var/www/:rw
+      - /matrix/static-files:/var/www/:ro
--- a/examples/haproxy/docker-compose.yml
+++ b/examples/haproxy/docker-compose.yml
@@ -1,0 +1,8 @@
+version: '3'
+services:
+  neginx:
+    image: local/nginx 
+    ports:
+      - 40888:80
+    volumes:
+      - /matrix/static-files:/var/www/:rw
--- a/examples/haproxy/haproxy.cfg
+++ b/examples/haproxy/haproxy.cfg
@@ -39,6 +39,7 @@ frontend https-frontend
     # HAproxy wants the full chain and the private key in one file. For Letsencrypt manually generated certs (e.g., wildcard certs) you can use
     # cat /etc/letsencrypt/live/example.com/fullchain.pem /etc/letsencrypt/live/example.com/privkey.pem > /etc/haproxy/certs/star-example.com.pem
     bind *:443 ssl crt /etc/haproxy/certs/star-example.com.pem
+    #bind *:443 ssl crt /etc/haproxy/certs/riot.example.com.pem /etc/haproxy/certs/matrix.example.com.pem
     reqadd X-Forwarded-Proto:\ https
     option httplog
     option http-server-close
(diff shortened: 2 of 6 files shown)


examples/haproxy/docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+version: '3'
+services:
+  neginx:
```
spantaleev: Not that it matters, but, typo?

examples/haproxy/docker-compose.yml
```diff
@@ -0,0 +1,8 @@
...
+    image: local/nginx 
+    ports:
+      - 40888:80
+    volumes:
+      - /matrix/static-files:/var/www/:rw
```
spantaleev: Perhaps `ro` is enough?

examples/haproxy/nginx.conf
```diff
@@ -0,0 +1,18 @@
...
+}
+
+http {
+  server_tokens off;
+  server {
+    listen 80;
+    index index.html;
+    root /var/www;
+
+    }
```
spantaleev: Whitespace seems off around here.

examples/haproxy/haproxy.cfg
```diff
@@ -0,0 +1,96 @@
...
+	errorfile 500 /etc/haproxy/errors/500.http
+	errorfile 502 /etc/haproxy/errors/502.http
+	errorfile 503 /etc/haproxy/errors/503.http
+	errorfile 504 /etc/haproxy/errors/504.http
+
+frontend https-frontend
+    bind *:80
+    # HAproxy wants the full chain and the private key in one file. For Letsencrypt manually generated certs (e.g., wildcard certs) you can use
+    # cat /etc/letsencrypt/live/example.com/fullchain.pem /etc/letsencrypt/live/example.com/privkey.pem > /etc/haproxy/certs/star-example.com.pem
+    bind *:443 ssl crt /etc/haproxy/certs/star-example.com.pem
```
spantaleev: Maybe we should mention in `examples/haproxy/README.md` that the example requires a wildcard certificate, because it bundles all domains under a single HAProxy `frontend` with a single certificate?

Sudneo: Hey, thanks for taking the time to review all this.
A wildcard certificate is not needed. I personally used but I think that a matrix.example.com and riot.example.com certificate should be enough.

HAproxy supports multiple certificates in that line, so it would be enough to write something like

`bind *:443 ssl crt /etc/haproxy/certs/matrix.example.com.pem /etc/haproxy/certs/riot.example.com.pem `

Nevertheless, for letsencrypt generate certs, the "cat" trick to merge the fullchain and private key is still needed.

I will try to update the doc and clarify this point though.

examples/haproxy/nginx.conf
Sudneo: Fixed whitespaces/indentation.

examples/haproxy/docker-compose.yml
Sudneo: Yes, I also think it is enough. Good point.

examples/haproxy/docker-compose.yml
Sudneo: Yep. Didn't spot it even in my personal configuration. Thanks!

spantaleev:
Merged! Thank you! :+1:

PR #258, opened 2019-08-30
More Apache ProxyPass directives need 'nocanon'

julianfoad:
Invitations weren't working for me until I added 'nocanon' to these additional places. Until then, invitations failed with "Invalid signature for server ..." errors, as in https://github.com/matrix-org/synapse/issues/3294 .

I didn't check whether the user_directory/search proxy line also needs it, I just assumed it should have it too.

The other two proxy lines in this example also include a 'retry=0' parameter. That's a separate issue; I haven't touched it here.


spantaleev:
Good catch! Thank you! :+1:

PR #260, opened 2019-09-01
Move pipelining configuration to ansible.cfg

liquidat:
This is a rather small change that moves the pipelining option from `examples/hosts` to `ansible.cfg`.

# Reasoning

In my understanding the example section tries to be as simple and as easy to understand as possible. Most users however will not be able to understand what pipelining actually means and thus not change the option. Moving it away will hide it and makes the example host file 25% easier to read ;)

Also, while that option can be triggered via an environment variable like in the example hosts file, moving it to the config file enables users (and developers of this playbooks) to define it as a default value, and change it via `-e` on the fly if needed without worrying about the precedence of the environment variables.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,3 +1,5 @@
 [defaults]
 retry_files_enabled = False
+
+[connection]
 pipelining = True
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,2 +1,3 @@
 [defaults]
 retry_files_enabled = False
+pipelining = True
--- a/examples/hosts
+++ b/examples/hosts
@@ -11,4 +11,4 @@
 # consider adding an additional `ansible_connection=local` argument below.
 
 [matrix_servers]
-matrix.<your-domain> ansible_host=<your-server's external IP address> ansible_ssh_user=root ansible_ssh_pipelining=yes
+matrix.<your-domain> ansible_host=<your-server's external IP address> ansible_ssh_user=root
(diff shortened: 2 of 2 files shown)


spantaleev:
While I agree that it's nice to keep things simpler, I think it's relatively often that people need to change this setting.

Putting it in `ansible.cfg` and using `-e` to change it makes it much harder on people. They need to remember to always add `-e` to their `ansible-playbook` commands, now and forever. That's pretty jarring and likely not worth the benefit of removing it from the `hosts` file.

If we do remove it anyway, we still need to document how to use `-e` (adjusting the comments in the `hosts` file).

liquidat:
> While I agree that it's nice to keep things simpler, I think it's relatively often that people need to change this setting.
> 
> Putting it in `ansible.cfg` and using `-e` to change it makes it much harder on people. They need to remember to always add `-e` to their `ansible-playbook` commands, now and forever. That's pretty jarring and likely not worth the benefit of removing it from the `hosts` file.

My idea was that they can still alter the configuration the `host_vars` if they want to. There would not be a need to add varabiles on the command line.

spantaleev:
Looks like putting `ansible_ssh_pipelining` in a var file could work.

I've done some tests and it seems like overriding `ansible_ssh_pipelining` from the `hosts` file and from a ` host_vars` file both work.

I've updated the comments in `examples/hosts` to make them accurate and tell people how they need to do it.

As host line got shorter and simpler (no more explicit `ansible_ssh_pipelining=yes` there), but the comment got longer and more complicated. I'm unsure if we've accomplished too much in terms of simplification, but it's probably a little better. I'm merging it like this.

Thank you! :+1:

PR #265, opened 2019-09-17
Disable docker network tasks in check mode to allow running the playbook in check mode (--check --diff)

gusttt:
Allow to run the playbook in check mode:
https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html


spantaleev:
Thanks for adding support for this! :+1:

PR #267, opened 2019-09-20
Update riot-web (1.3.5 -> 1.3.6)

thedanbob:


aaronraimist:
(1.3.6 doesn't change anything for the web version, only Riot Desktop)

thedanbob:
Hmm you're right, I didn't look closely at the changelog.

spantaleev:
Good to know that there's nothing new!

Still, it doesn't hurt to be on the latest one.. At least to prevent others from thinking we should upgrade and taking an effort to create new pull requests ;)

Thanks! :+1:

PR #269, opened 2019-09-23
Update configuring-well-known.md

Archehandoro:
I had to add these changes to my existing /etc/haproxy/haproxy.cfg file to reverse proxy queries and setup service discovery.


spantaleev:
Awesome! Thank you! :+1:

PR #273, opened 2019-09-28
Bump riot web version 1.3.6 -> 1.4.0

Sporiff:

PR #275, opened 2019-10-03
WIP: Upgrade Synapse (1.3.1 -> 1.4.0)

aaronraimist:
#273 should happen at the same time

### DIFF
--- a/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
+++ b/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
@@ -105,6 +105,9 @@ federation_domain_whitelist: {{ matrix_synapse_federation_domain_whitelist|to_js
 # blacklist IP address CIDR ranges. If this option is not specified, or
 # specified with an empty list, no ip range blacklist will be enforced.
 #
+# As of Synapse v1.4.0 this option also affects any outbound requests to identity
+# servers provided by user input.
+#
 # (0.0.0.0 and :: are always blacklisted, whether or not they are explicitly
 # listed here, since they correspond to unroutable addresses.)
 #
@@ -131,8 +134,8 @@ federation_ip_range_blacklist:
 #
 #   type: the type of listener. Normally 'http', but other valid options are:
 #       'manhole' (see docs/manhole.md),
-#       'metrics' (see docs/metrics-howto.rst),
-#       'replication' (see docs/workers.rst).
+#       'metrics' (see docs/metrics-howto.md),
+#       'replication' (see docs/workers.md).
 #
 #   tls: set to true to enable TLS for this listener. Will use the TLS
 #       key/cert specified in tls_private_key_path / tls_certificate_path.
@@ -167,12 +170,12 @@ federation_ip_range_blacklist:
 #
 #   media: the media API (/_matrix/media).
 #
-#   metrics: the metrics interface. See docs/metrics-howto.rst.
+#   metrics: the metrics interface. See docs/metrics-howto.md.
 #
 #   openid: OpenID authentication.
 #
 #   replication: the HTTP replication API (/_synapse/replication). See
-#       docs/workers.rst.
+#       docs/workers.md.
 #
 #   static: static resources under synapse/static (/_matrix/static). (Mostly
 #       useful for 'fallback authentication'.)
@@ -311,6 +314,23 @@ listeners:
 #
 #allow_per_room_profiles: false
 
+# How long to keep redacted events in unredacted form in the database. After
+# this period redacted events get replaced with their redacted form in the DB.
+#
+# Defaults to `7d`. Set to `null` to disable.
+#
+#redaction_retention_period: 28d
+
+redaction_retention_period: {{ matrix_synapse_redaction_retention_period }}
+
+# How long to track users' last seen time and IPs in the database.
+#
+# Defaults to `28d`. Set to `null` to disable clearing out of old rows.
+#
+#user_ips_max_age: 14d
+
+user_ips_max_age: {{ matrix_synapse_user_ips_max_age }}
+
 
 ## TLS ##
 
@@ -442,7 +462,7 @@ acme:
     #
     # If unspecified, we will use CONFDIR/client.key.
     #
-    account_key_file: /data/acme_account.key
+    #account_key_file: /data/acme_account.key
 
 # List of allowed TLS fingerprints for this server to publish along
 # with the signing keys for this server. Other matrix servers that
@@ -493,7 +513,8 @@ event_cache_size: "{{ matrix_synapse_event_cache_size }}"
 
 ## Logging ##
 
-# A yaml python logging config file
+# A yaml python logging config file as described by
+# https://docs.python.org/3.7/library/logging.config.html#configuration-dictionary-schema
 #
 log_config: "/data/{{ matrix_server_fqn_matrix }}.log.config"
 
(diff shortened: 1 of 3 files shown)


roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -907,11 +934,45 @@ allow_guest_access: {{ matrix_synapse_allow_guest_access|to_json }}
...
+# background migration script, informing itself that the identity server all of its
+# 3PIDs have been bound to is likely one of the below.
+#
+# As of Synapse v1.4.0, all other functionality of this option has been deprecated, and
+# it is now solely used for the purposes of the background migration script, and can be
+# removed once it has run.
 {% if matrix_synapse_trusted_third_party_id_servers|length > 0 %}
 trusted_third_party_id_servers:
 {{ matrix_synapse_trusted_third_party_id_servers|to_nice_yaml }}
 {% endif %}
```
aaronraimist: What are you going to do with the identity server?

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -1046,73 +1126,117 @@ signing_key_path: "/data/{{ matrix_server_fqn_matrix }}.signing.key"
 #      "ed25519:auto": "abcdefghijklmnopqrstuvwxyzabcdefghijklmopqr"
 #  - server_name: "my_other_trusted_server.example.com"
 #
-# The default configuration is:
+trusted_key_servers: {{ matrix_synapse_trusted_key_servers|to_json }}
```
aaronraimist: This probably needs a note in the CHANGELOG. It isn't new that the default is matrix.org but it is newly configurable so people may not want to use matrix.org for privacy reasons.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
aaronraimist: Personally I don't need an identity server anymore. I would continue to run mxisd but I really don't have a purpose for it after Synapse 1.4.0 and it doesn't support the unbind API which leads to a broken experience.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -311,6 +314,23 @@ listeners:
...
+
+redaction_retention_period: {{ matrix_synapse_redaction_retention_period }}
+
+# How long to track users' last seen time and IPs in the database.
+#
+# Defaults to `28d`. Set to `null` to disable clearing out of old rows.
+#
+#user_ips_max_age: 14d
+
+user_ips_max_age: {{ matrix_synapse_user_ips_max_age }}
```
aaronraimist: These also need a CHANGELOG entry to say hey just so you know you are going to loose this stuff when you upgrade unless you set these to null.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
spantaleev: I'm getting the impression that we should still keep it for "background migration" purposes.
Or maybe that migration script had already been executed as part of a previous release..?

It doesn't hurt to keep `trusted_third_party_id_servers` around, at least initially.

----------

As for disabling our identity server, I think I won't need it as well.

However, if we disable our identity server right now, `matrix_synapse_trusted_third_party_id_servers` would default to `matrix_synapse_id_servers_public`, which is `['matrix.org', 'vector.im']`. The comment says this `trusted_third_party_id_servers` setting isn't used for anything so maybe that's okay.

----------

There are 2 other more **major** side-effects of disabling our identity server and having `matrix_synapse_trusted_third_party_id_servers` fall back to `['matrix.org', 'vector.im']` for us.


The first server in that `matrix_synapse_trusted_third_party_id_servers` list (in this case, `'matrix.org'`) ends up in `matrix_identity_server_url`.

This is used for riot-web's `default_is_url` setting (via the `matrix_riot_web_default_is_url` variable).
So, suddenly, riot-web starts pointing to matrix.org as a default server. That's probably not good.

Furthermore, `matrix_identity_server_url` is used in one other place -- the generation of the `/.well-known/matrix/client` file. All other clients (including Riot on mobile devices) which consult the `/.well-known/matrix/client` file will see that our server recommends talking to `matrix.org` as an identity server. That's probably not good as well.

----------------

So I think we should keep running an identity server.. Or set these defaults to something else (`null`?).

I'm not sure if clients like `null` values now or whether they decide to fall back to something else in such a case (`matrix.org` again?).

If mobile applications really do respect `null`, that means they disable their "Find contacts" feature. While I find that feature largely useless (and would never share my contacts with any app), we may wish to find a way to get it working (that is, by running an identity server by default?)

In any case, I'm not sure how useful such a feature is when pointed to our own self-hosted identity server. Delegation between identity servers only works between mxisd servers that are authoritative for their own domain. That's nice, but in the real world, most users are unfortunately using email addresses like `gmail.com`, etc. In such a centralized world, we may decide to kill the identity server thing (by default) for most users, and leave it as an option to enable for those who need it.

Still, lots of things to investigate before taking the right action..

spantaleev:
I've merged it like this.

Thank you, @aaronraimist! :+1: 

We should proceed thinking about whether to kill off the identity server by default or not. We can do this separately.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
aaronraimist: > So I think we should keep running an identity server.. Or set these defaults to something else (null?).

I got an Ansible error message if I set it to null but it is happy to accept `matrix_synapse_trusted_third_party_id_servers: []`

I agree that we should keep it around for a while while people upgrade. The background migration happens as soon as you upgrade to 1.4.0 is my understanding. For me the background migration process put the wrong info in since I initially set my server up a few years ago before this playbook or mxisd existed. When the background migration process ran it set the identity server too matrix.raim.ist which meant I couldn't unbind them and remove them from the actual server where they were (vector.im). It was easy enough for me to `UPDATE user_threepid_id_server SET id_server = 'vector.im' WHERE user_id = '@aaron:raim.ist';` in the database which allowed me to remove them from vector.im. Most people are probably not in that camp though so we probably don't need to mention it in the docs.


> This is used for riot-web's default_is_url setting (via the matrix_riot_web_default_is_url variable).
So, suddenly, riot-web starts pointing to matrix.org as a default server. That's probably not good.

If you set it to `matrix_synapse_trusted_third_party_id_servers: []` then the riot config looks like:
`"default_is_url": "",`
and `/.well-known/matrix/client` looks like
```
	"m.identity_server": {
		"base_url": ""
	}
```
You can probably just modify those to remove those sections when `matrix_synapse_trusted_third_party_id_servers` is set to `[]`. It does seem to work fine though. 
Screen Shot 2019-10-03 at 12 04 02 PM
Riot doesn't seem to ask for an identity server on login anymore.

spantaleev: It also did something weird with one of the email address bound to my account. I think it was validated by the same server, but I'm not sure when and how that happened..

Unbinding was not possible until I deleted that specific entry from `user_threepid_id_server`. I believe that not having any entry there makes Synapse think it's responsible for validating the 3pid and would just drop it without complaining. I'm fine with that, so I went that route.

Regular users would probably not experience this though..

--------

Thanks for confirming riot-web's behavior with:

```json
"m.identity_server": {
	"base_url": ""
}
```

If Riot/Android and Riot/iOS also behave well, we can try switching to that.
I'm curious what their "Contacts" tab would do in such a case though. I can probably test tomorrow or something.

aaronraimist: Riot iOS does not appear to like that on 0.9.5. When I try to login it says Invalid homeserver discovery response. Maybe it will behave better in the next version, the stuff about disconnecting from an intensity server hasn't been implemented yet in the mobile apps https://github.com/vector-im/riot-ios/issues/2746.

If I try to delete https://vector.im from the identity server URL field, Riot iOS just fills it back in. Setting it to localhost seems to work but I haven't verified that it isn't connecting to vector.im.

PR #277, opened 2019-10-03
Update riot-web (1.4.0 -> 1.4.1)

thedanbob:


spantaleev:
Thanks! :+1:

I've missed that one.. Looks like by the time Synapse got to v1.4.0, riot-web had already moved forward, hehe..

PR #278, opened 2019-10-04
Fix spelling ngnix -> nginx

ptman:


spantaleev:
Good catch! Thanks! :+1:

PR #281, opened 2019-10-14
Quote docker args in case inputs contain spaces

julianfoad:
Prompted by conversation between @elsatch:matrix.org and @aaronraimist about non-alphanumeric characters in password.


spantaleev:
Thanks guys! :+1:

julianfoad:
For the record:

The 'password=$2' lines don't need quotes, in Bash syntax, even though it's a bit surprising to look at.

I tested it in a modified copy of the script, modified so it prints the arguments of the final commands instead of actually running them.

PR #283, opened 2019-10-21
Add variable to enable verbose logging for mxisd

fnoah:
This is something that I have used to debug LDAP authentication with mxisd and found useful. Do you think we can add this variable to this project?


spantaleev:
That's nice, but.. it looks like this PR changes formatting in various unrelated places. Could you undo that and come up with a minimal patch?

fnoah:
My bad, I did not disable Prettier which messed with the formatting of all MD files i touched. Should be fixed now.

spantaleev:
Great! Thank you! :+1:

PR #286, opened 2019-10-23
Listen 8448 required for apache

dreamflasher:
Listen 8448 directive makes apache listen on that port, without it federation does not work.


spantaleev:
Thanks! :+1:

PR #293, opened 2019-11-02
Actually use matrix_synapse_storage_path

recklesscoder:
`matrix_synapse_storage_path` is already defined in `matrix-synapse/defaults/main.yml` (with a default of `"{{ matrix_synapse_base_path }}/storage"`), but was not being used for its presumed purpose in `matrix-synapse.service.j2`. As a result, if `matrix_synapse_storage_path` was overridden (in a `vars.yml`), the synapse service failed to start.


spantaleev:
Good point! Thank you! :+1:

PR #302, opened 2019-11-08
Mention that ARM is not supported

aaronraimist:
Feel free to reword but we have had a couple of issues created now asking about Raspberry Pi support.


spantaleev:
Looks good! Hopefully this prevents some future confusion.
Thank you! :+1:

Issue #304, opened 2019-11-09
Corrupt matrix-synapse.service


wabuMike:
I'm using the mautrix-whatsapp bridge. Pulling the current playbook changes from Github and running the playbook give me `--mount type=bind,src={{ matrix_mautrix_whatsapp_config_path }}/registration.yaml,dst=/matrix-mautrix-whatsapp-registration.yaml,ro \` in /etc/systemd/system/matrix-synapse.service.

PR #305, opened 2019-11-09
Replace constructs appending elements with variables to matrix_synapse_container_extra_arguments

wabuMike:
Replace constructs appending elements with variables to matrix_synapse_container_extra_arguments. Fixes issue https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/304

The only bridge I have activated is mautrix-whatsapp. Can anyone test these changes with other/more bridges activated?


spantaleev:
Thanks for reproducing and fixing the Ansible 2.9 issue that's been plaguing us!

It looks like it wasn't reproducible in all environments (even when the Ansible and Jinja versions were identical), so I'm not sure what the cause really was.

I can't seem to find any regressions with this fix, so I'm merging it as-is!

Thanks again! :+1:

PR #310, opened 2019-11-15
Don't install docker-ce if docker.io is installed

thedanbob:
I switched to docker.io on my matrix server since docker-ce doesn't support Ubuntu 19.10 yet. This PR skips installing docker-ce if docker.io is already present.

### DIFF
--- a/roles/matrix-base/tasks/setup_server_base.yml
+++ b/roles/matrix-base/tasks/setup_server_base.yml
@@ -38,6 +38,10 @@
     update_cache: yes
   when: ansible_os_family == 'Debian'
 
+- name: Gather package facts (Debian)
+  package_facts:
+  when: ansible_os_family == 'Debian'
+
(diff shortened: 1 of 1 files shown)


spantaleev:
I guess this won't hurt in most cases, so I'll merge it as it is.
Thank you! :+1: 

It'd be nice if we do it for RedHat-based distros as well.

What would probably be even better though is if we make Docker installation controllable via some variable like `matrix_docker_installation_enabled`.
Some of that is discussed [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/247#issuecomment-530277925).

We could even try to default that variable to something like `'docker.io' not in ansible_facts.packages` and allow people to explicitly set its value as well.

thedanbob:
I wasn't fully happy with it either but I figured it would work for now and can be refactored later. I'll put some thought into how it could be done better and maybe create another PR.

PR #311, opened 2019-11-16
Fix: Restrict a Debian-specific task to Debian OS.

hardye:
Since commit b9753635 the task 'Ensure docker-ce is installed (Debian)' fails with an error on CentOS although it should not even run on this OS.


spantaleev:
Good catch! :+1:

PR #312, opened 2019-11-18
Refactor setup server base

thedanbob:
I refactored setup_server_base.yml with two new variables: `matrix_docker_installation_enabled` and `matrix_docker_package_name`. This way the playbook can support alternate Docker packages (the only one I know about currently is docker.io on Debian, but maybe there are others or will be in the future) or manually maintaining Docker.


spantaleev:
Thank you very much for improving this! :+1: 

I assume you've tested it on Debian. I've just tested it on CentOS and it works well!

PR #314, opened 2019-11-20
Remove identity server section from .well-known/matrix/client if there is no identity server

aaronraimist:
Riot used to be fine with it being blank but now it complains. This creates an ugly looking comma when there is an identity server configured but I guess that's fine.


aaronraimist:
I think
```
	}{% if matrix_identity_server_url %},
	"m.identity_server": {
		"base_url": "{{ matrix_identity_server_url }}"
	}
	{% endif %}
```
probably looks worse

spantaleev:
It's not very pretty, but I guess it's okay.. Better to have it working than not have it at all!
Thank you! :+1:

PR #321, opened 2019-12-05
Allow Synapse manhole to be enabled

aaronraimist:
Can you double check that the way I have this set only exposes it locally? It is important that the manhole is not available to the outside world since it is quite powerful and the password is hard coded.

### DIFF
--- a/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
+++ b/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
@@ -229,11 +229,13 @@ listeners:
         compress: false
 {% endif %}
 
+{% if matrix_synapse_manhole_enabled %}
   # Turn on the twisted ssh manhole service on localhost on the given
   # port.
-  # - port: 9000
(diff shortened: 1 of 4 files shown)


roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -229,11 +229,13 @@ listeners:
...
 {% endif %}
 
+{% if matrix_synapse_manhole_enabled %}
   # Turn on the twisted ssh manhole service on localhost on the given
   # port.
-  # - port: 9000
-  #   bind_addresses: ['::1', '127.0.0.1']
-  #   type: manhole
+  - port: 9000
+    bind_addresses: ['0.0.0.0']
```
SuperSandro2000: This is globally.

spantaleev: It's the `-p` part for the `docker run` command that really matters. This here should be fine.

spantaleev:
Looks good! Thank you, @aaronraimist! :+1:

PR #322, opened 2019-12-05
Implement MSC1957: Integration manager discovery

aaronraimist:


roles/matrix-base/templates/static-files/well-known/matrix-client.j2
```diff
@@ -8,4 +8,14 @@
 		"base_url": "{{ matrix_identity_server_url }}"
 	}
 	{% endif %}
+	{% if matrix_dimension_integrations_rest_url and matrix_dimension_integrations_ui_url %},
```
spantaleev: Thank you for working on MSC1957!

---------

This does make the `matrix-base` role depend on `matrix-dimension`.

Maybe it's better if we introduce 2 new variables in `roles/matrix-base/defaults/main.yml`:

```yaml
matrix_integration_manager_rest_url: ~
matrix_integration_manager_ui_url: ~
```

and refer to those in `roles/matrix-base/templates/static-files/well-known/matrix-client.j2`.

And then the playbook could tie them like this in `group_vars/matrix_servers`:

```yaml
matrix_integration_manager_rest_url: "{{ matrix_dimension_integrations_rest_url if matrix_dimension_enabled else None }}"
matrix_integration_manager_ui_url: "{{ matrix_dimension_integrations_ui_url if matrix_dimension_enabled else None }}"
```

aaronraimist: Yep that makes sense

spantaleev:
Thanks, Aaron! :+1:

PR #323, opened 2019-12-08
Render vhost directives in https server block

yuri-moens:
Vhost directives were not being rendered in the https server block which made .well-known urls not work.


spantaleev:
Good catch! Thank you for fixing this federation-breaking thing for people who serve the base domain via the playbook! 👍🏻

Issue #296, opened 2019-11-06
Docker container custom labels


lpulley:
Hi all,

I'm currently using Traefik for my reverse proxy. Traefik grabs configuration data for containers from labels on running Docker containers, so I'm wondering if there is a way for me to assign labels to the Matrix containers from the `main.yml`. If not, is this something that could be easily added? I'm not familiar with how the containers are started in this playbook, but it shouldn't be hard to "inject" labels at runtime with `docker run -l` or something like that.

Doing things this way also requires the containers to be on the same Docker network; can I use `matrix_docker_network` to change the Docker network of the Matrix containers to match the network of my Traefik container?

Thanks!

spantaleev:
Hi!

There's a dedicated variable for each service which lets you inject additional `docker run` arguments.
For the Synapse service, it's called `matrix_synapse_container_extra_arguments`. You can use it like this:

```yaml
matrix_synapse_container_extra_arguments: ['--label key=value', '--label another=value']
```

You can use `matrix_docker_network` to change the Docker network to something else (defaults to `'matrix'` for most services).

lpulley:
Oh, sweet! Somehow I missed that. Thanks so much for your detailed reply! I really appreciate this project, thanks for helping so much!

spantaleev:
You're welcome! Happy to hear that!

lpulley:
@spantaleev Finally getting around to this. I can't get the Nginx proxy container to _not_ expose 80 and 443, so it won't start. (I need it to expose nothing, or unused ports, because I am using 80 and 443 for Traefik.)

Any idea on how I can use the built in Nginx container without any port mappings?

spantaleev:
@lpulley, as of right now, `matrix-nginx-proxy.service` always exposes ports 80 and 443 with no way to disable that. We can make it configurable though.

Until we do, I suggest that you modify `/etc/systemd/system/matrix-nginx-proxy.service` manually, to remove the port-mapping lines. Then do `systemctl daemon-reload` and restart it (`systemctl restart matrix-nginx-proxy`).

See if removing these port mappings and doing all your other tweaks leads you to a nice and working setup. If it does, we should definitely investigate how to remove the port mappings by tweaking playbook configuration.

lpulley:
Thanks!!
[...]

PR #325, opened 2019-12-10
Added Traefik configuration example to own webserver docs

lpulley:
@spantaleev helped me get this working with Traefik as a reverse proxy. I've added my working configuration to the docs with ample comments and explanation.

### DIFF
--- a/docs/configuring-playbook-own-webserver.md
+++ b/docs/configuring-playbook-own-webserver.md
@@ -115,5 +115,83 @@ All services would be served locally on `127.0.0.1:81` and `127.0.0.1:8449` (as 
 You can then set up another reverse-proxy server on ports 80/443/8448 for all of the expected domains and make traffic go to these local ports.
 The expected domains vary depending on the services you have enabled (`matrix.DOMAIN` for sure; `riot.DOMAIN` and `dimension.DOMAIN` are optional).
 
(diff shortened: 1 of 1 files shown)


spantaleev:
Thank you! :+1: 

-------

I've noticed that you're wiring federation directly to Synapse (`matrix-synapse:8048`).

You could probably make it go to nginx as well (`matrix-nginx-proxy:8448`) - that in-container `8448` port serves plain-HTTP traffic when `matrix_nginx_proxy_https_enabled` is `false`.

If so, everything would go to matrix-nginx-proxy and Synapse can remain as-is.

spantaleev:
Ah, I see [from your comment](https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/296#issuecomment-562819174) there may be some reason why you're making `matrix-nginx-proxy:8448` go unused and going for reverse-proxying directly to `matrix-synapse:8048`. Perhaps Traefik makes it more difficult to set up with multiple ports going to the same service.

lpulley:
Yes, the way the Traefik Docker provider works, I don't see a way to map two entrypoints to two different ports. Figured it would be fine to just patch through to Synapse since that's all Nginx does anyway.

spantaleev:
Got it! Sounds good then!

PR #326, opened 2019-12-11
Expose self service bridging in matrix_appservice_discord

dali99:
This allows for users to bridge already existing matrix rooms to discord


spantaleev:
Thank you! :+1:

PR #333, opened 2019-12-31
matrix-ma1sd support

eMPee584:
· Duplicates matrix-mxisd role for matrix-ma1sd fork (adapting strings and URLs)
· Replaces "mxisd" strings with ma1sd in docs, fixing URLs
· Renames variables (s/mxisd/ma1sd/) and adapts roles


spantaleev:
Besides this, we probably need to figure out how to hook it into `group_vars/matrix_servers`.

roles/matrix-ma1sd/vars/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+
+# Doing `|from_yaml` when the extension contains nothing yields an empty string ("").
+# We need to ensure it's a dictionary or `|combine` (when building `matrix_ma1sd_configuration`) will fail later.
+matrix_ma1sd_configuration_extension: "{{ matrix_ma1sd_configuration_extension_yaml|from_yaml if matrix_ma1sd_configuration_extension_yaml|from_yaml else {} }}"
```
SuperSandro2000: missing new line

eMPee584: uhmm.. thanks for the review.. this is the same as in the other roles though, isn't it?

eMPee584:
So finishing the pure string replaces and after a quick test, this seems to work. Now having a look at config file / var rename handling..

eMPee584:
.. I got stuck on ~~trying~~ researching how to regex-handle all the `matrix_mxisd_.*` vars instead of listing them individually in `roles/matrix-ma1sd/tasks/validate_config.yml` .. as in
```
- name: (Deprecation) Catch and report renamed ma1sd variables
  fail:
    msg: […]
  when: "item.old in vars"
  with_items:
    - {'old': 'matrix_mxisd_.*', 'new': 'matrix_ma1sd_.*'}
```
Is this possible?
Also moving the existant mxisd.db to /matrix/ma1sd/ma1sd.db was another migration task I didn't yet get to.. So if any ansible-wiz would like to move this further.. :rocket:

eMPee584:
Ok I rebased this and bumped the version.
The migration task mxisd db to ma1sd via ansible should be easy to do but I have still little clue about that.

group_vars/matrix_servers
```diff
@@ -448,41 +448,45 @@ matrix_mailer_enabled: true
...
 #
 ######################################################################
 
-# By default, this playbook installs the mxisd identity server on the same domain as Synapse (`matrix_server_fqn_matrix`).
+# By default, this playbook installs the ma1sd identity server on the same domain as Synapse (`matrix_server_fqn_matrix`).
 # If you wish to use the public identity servers (matrix.org, vector.im) instead of your own you may wish to disable this.
-matrix_mxisd_enabled: true
+matrix_ma1sd_enabled: true
 
+<<<<<<< HEAD
```
Shadowghost: Those merge comments should be removed 😄

Shadowghost:
In regards to the official doc [here](https://github.com/ma1uta/ma1sd/blob/master/docs/migration-from-mxisd.md) the migration is done by stopping mxisdb, moving files/mounting volumes on containers in the appropiate directories and starting ma1sd. Should be fairly simple to achieve with tasks running once if ansible recognizes that mxisdb was installed brefore.

roles/matrix-ma1sd/vars/main.yml
SuperSandro2000: git cares about this kinds of stuff and can generate stupid diffs if there is no new line at the end.

blaluc:
any progress with ma1sd? Since mxisd does not work anymore, we miss an import piece of the architecture :-(

roles/matrix-ma1sd/vars/main.yml
eMPee584: That may well be; however none of the other `roles/*/var/main.yml`s have a trailing new line.

group_vars/matrix_servers
eMPee584: oh good catch, how did that happen :sweat_smile:

roles/matrix-ma1sd/vars/main.yml
eMPee584: Oh just learned this is about an invisible `0A` byte at the end.. thanks for insisting on that xD

eMPee584:
Fairly simple.. yes. Learning this ansible witchcraft takes its time though..

eMPee584:
( Pull request ready for review though, I added the migration task.. )

spantaleev:
I've fixed up some things and have merged this!

Thanks a lot for your great work on this long-standing issue, @eMPee584! :+1:

cmuller:
Hello @eMPee584 

Thanks a lot for this migration! It went flawlessly for me except one tiny thing I have forgotten and it was a bit hard to debug because I was just watching the wrong traces and had no clue what was going on..

Le me explain:
 - login with riot or by API with our main URL (https://matrix.domain.com) was not working
 - I saw no errors in nginx or synapse or ma1sd (!)
 - ma1sd login was working (with local access to the container's IP address and port)
 - even synapse login was working! (with _matrix-internal API)

Explanation: I just forgot that we are using corporal and that it _intercepts_ communications and  forward login requests to the identity server with URL defined in `authCredential` fields of the `policy.json` file! So I had to change for all users:

```diff
-  "authCredential": "http://matrix-mxisd:8090/_matrix-internal/identity/v1/check_credentials",
+  "authCredential": "http://matrix-ma1sd:8090/_matrix-internal/identity/v1/check_credentials",
```

pushed the policy.. and then everything was working immediately :)

I figure that maybe it would be nice to add a note about this in the migration docs but I do not know where exactly (Changelog ? README ? some other docs). What do you think?
Anyway, thanks for this work. I feel much better to know that one of the services is not EOL..

PR #335, opened 2020-01-03
Notes when not using a ssh key for authentication

hannojg:
Fixed small typo, added an extra note on how to use the ansible docker image when not using a ssh key for authentication


spantaleev:
Thank you! :+1:

PR #336, opened 2020-01-08
Remove spurious 'mxisd' string from mautrix bridge templates comment

eMPee584:
These templates seemingly were copied originally from the mxisd one.. so remove the stray strings.


spantaleev:
Thanks!

PR #339, opened 2020-01-09
Add matrix-appservice-webhooks role.

tripleawwy:
Added matrix-appservice-webhooks to the playbook, adjusted group_vars and setup.yml.

### DIFF
--- a/docs/configuring-playbook-bridge-appservice-webhooks.md
+++ b/docs/configuring-playbook-bridge-appservice-webhooks.md
@@ -1,0 +1,58 @@
+# Setting up Appservice Webhooks (optional)
+
+The playbook can install and configure [matrix-appservice-webhooks](https://github.com/turt2live/matrix-appservice-webhooks) for you.
+
+Setup Instructions:
+
+loosely based on [this](https://github.com/turt2live/matrix-appservice-webhooks/blob/master/README.md)
+
+1. All you basically need is to adjust your `inventory/host_vars/matrix.<domain-name>/host-vars.yml`:
+    ```yaml
+    matrix_appservice_webhooks_enabled: true
+    matrix_appservice_webhooks_api_secret: '<your_secret>'
+    ```
+
+2. In case you want to change the verbosity of logging via `journalctl -fu matrix-appservice-webhooks.service`
+you can adjust this in `inventory/host_vars/matrix.<domain-name>/host-vars.yml` as well.
+<br>
+*Note*: default value is: `info` and availabe log levels are : `info, verbose`
+
+    ```yaml
+    matrix_appservice_webhooks_log_level: '<log_level>'
+    ```
+
+3. If you've already installed Matrix services using the playbook before, you'll need to re-run it (`--tags=setup-all,start`). If not, proceed with [configuring other playbook services](configuring-playbook.md) and then with [Installing](installing.md). Get back to this guide once ready.
+
+4. Invite the bridge bot user to your room:
+    - either with:
+      ```
+      /invite @_webhook:matrix.<domain.name>
+      ```
+        *Note*: Make sure you have administration permissions in your room
+    - or simply add the bridge bot to a private channel (personal channels imply you being an administrator)
+
+5. Send a message to the bridge bot in order to receive a private message including the webhook link.
+    ```
+    !webhook
+    ```
+
+6. The JSON body for posting messages will have to look like this:
+    ```
+    {
+        "text": "Hello world!",
+        "format": "plain",
+        "displayName": "My Cool Webhook",
+        "avatarUrl": "http://i.imgur.com/IDOBtEJ.png"
+    }
+    ```
+   - You can test this via curl like so:
+     ```
+     curl --header "Content-Type: application/json" \
+     --data '{
+     "text": "Hello world!",
+     "format": "plain",
+     "displayName": "My Cool Webhook",
+     "avatarUrl": "http://i.imgur.com/IDOBtEJ.png"
+     }' \
+     <the link you've gotten in 5.>
+     ```
--- a/docs/configuring-playbook-bridge-appservice-webhooks.md
+++ b/docs/configuring-playbook-bridge-appservice-webhooks.md
@@ -6,14 +6,14 @@ Setup Instructions:
 
 loosely based on [this](https://github.com/turt2live/matrix-appservice-webhooks/blob/master/README.md)
 
-1. All you basically need is to adjust your `inventory/host_vars/matrix.<domain-name>/host-vars.yml`:
+1. All you basically need is to adjust your `inventory/host_vars/matrix.<domain-name>/vars.yml`:
     ```yaml
     matrix_appservice_webhooks_enabled: true
     matrix_appservice_webhooks_api_secret: '<your_secret>'
     ```
 
 2. In case you want to change the verbosity of logging via `journalctl -fu matrix-appservice-webhooks.service`
-you can adjust this in `inventory/host_vars/matrix.<domain-name>/host-vars.yml` as well.
+you can adjust this in `inventory/host_vars/matrix.<domain-name>/vars.yml` as well.
 <br>
 *Note*: default value is: `info` and availabe log levels are : `info, verbose`
 
@@ -26,7 +26,7 @@ you can adjust this in `inventory/host_vars/matrix.<domain-name>/host-vars.yml` 
 4. Invite the bridge bot user to your room:
     - either with:
       ```
-      /invite @_webhook:matrix.<domain.name>
+      /invite @_webhook:<domain.name>
       ```
         *Note*: Make sure you have administration permissions in your room
     - or simply add the bridge bot to a private channel (personal channels imply you being an administrator)
--- a/roles/matrix-bridge-appservice-webhooks/defaults/main.yml
+++ b/roles/matrix-bridge-appservice-webhooks/defaults/main.yml
@@ -66,11 +66,6 @@ matrix_appservice_webhooks_configuration_yaml: |
     # The localpart to use for the bot. May require re-registering the application service.
     localpart: "_webhook"
 
-    # Appearance options for the Matrix bot
-    appearance:
-      displayName: "Webhook Bridge"
-      avatarUrl: "http://i.imgur.com/IDOBtEJ.png" # webhook icon
-
   # Provisioning API options
   provisioning:
     # Your secret for the API. Required for all provisioning API requests.
--- a/roles/matrix-bridge-appservice-webhooks/defaults/main.yml
+++ b/roles/matrix-bridge-appservice-webhooks/defaults/main.yml
@@ -1,0 +1,110 @@
+# matrix-appservice-webhooks is a Matrix <-> webhook bridge
+# See: https://github.com/turt2live/matrix-appservice-webhooks
+
+matrix_appservice_webhooks_enabled: true
+
+matrix_appservice_webhooks_docker_image: "turt2live/matrix-appservice-webhooks:latest"
+matrix_appservice_webhooks_docker_image_force_pull: "{{ matrix_appservice_webhooks_docker_image.endswith(':latest') }}"
+
+matrix_appservice_webhooks_base_path: "{{ matrix_base_data_path }}/appservice-webhooks"
+matrix_appservice_webhooks_config_path: "{{ matrix_appservice_webhooks_base_path }}/config"
+matrix_appservice_webhooks_data_path: "{{ matrix_appservice_webhooks_base_path }}/data"
+
+matrix_appservice_webhooks_public_endpoint: /appservice-webhooks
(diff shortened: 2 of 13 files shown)


tripleawwy:
Documentation will be added soon™.

roles/matrix-bridge-appservice-webhooks/tasks/init.yml
```diff
@@ -0,0 +1,79 @@
...
+  tags:
+   - always
+  when: matrix_appservice_webhooks_enabled|bool
+
+- name: Warn about reverse-proxying if matrix-nginx-proxy not used
+  debug:
+    msg: >-
+      NOTE: You've enabled the Matrix webhooks bridge but are not using the matrix-nginx-proxy
+      reverse proxy.
+      Please make sure that you're proxying the `{{ something }}`
```
spantaleev: `something` should probably be `matrix_appservice_webhooks_public_endpoint`?

roles/matrix-bridge-appservice-webhooks/defaults/main.yml
```diff
@@ -0,0 +1,110 @@
...
+
+    # Appearance options for the Matrix bot
+    appearance:
+      displayName: "Webhook Bridge"
+      avatarUrl: "http://i.imgur.com/IDOBtEJ.png" # webhook icon
+
+  # Provisioning API options
+  provisioning:
+    # Your secret for the API. Required for all provisioning API requests.
+    secret: 'warummussesdennsolangsein'
```
spantaleev: Probably need to make this configurable and a required variable?

roles/matrix-bridge-appservice-webhooks/defaults/main.yml
```diff
@@ -0,0 +1,110 @@
...
+  # Configuration related to the web portion of the bridge. Handles the inbound webhooks
+  web:
+    hookUrlBase:  "{{ matrix_appservice_webhooks_inbound_uri_prefix }}"
+
+  logging:
+    file: data/webhook.log
+    console: true
+    consoleLevel: info
+    fileLevel: verbose
+    writeFiles: true
```
spantaleev: We can probably consider flipping this to `false` to disable file-based logging.

Since we log to systemd/journald already and all other services avoid logging to files manually (where possible), it will be consistent with that.

I haven't tested whether this is the correct way to do it though.

spantaleev: Well, if we do that, we might want to bump `consoleLevel` to `verbose`.
I'm guessing this service doesn't spew too many log entries, so verbose could be a good default for it.
I'm fine to go with your recommendation though.

roles/matrix-bridge-appservice-webhooks/tasks/init.yml
```diff
@@ -0,0 +1,79 @@
...
+    when: matrix_nginx_proxy_role_executed|default(False)|bool
+
+  - name: Generate Matrix Appservice webhooks proxying configuration for matrix-nginx-proxy
+    set_fact:
+      matrix_appservice_webhooks_matrix_nginx_proxy_configuration: |
+        location {{ matrix_appservice_webhooks_public_endpoint }} {
+        {% if matrix_nginx_proxy_enabled|default(False) %}
+        {# Use the embedded DNS resolver in Docker containers to discover the service #}
+          resolver 127.0.0.11 valid=5s;
+          rewrite {{ matrix_appservice_webhooks_public_endpoint }}/(.*) /$1  break;
```
spantaleev: Perhaps removing the prefix could happen by fixing the `proxy_pass` directive to end with a slash.
We might then be able to get rid of this `rewrite` directive.

Example:

```diff
-proxy_pass {{ matrix_appservice_webhooks_appservice_url }}:{{ matrix_appservice_webhooks_matrix_port }};
+proxy_pass {{ matrix_appservice_webhooks_appservice_url }}:{{ matrix_appservice_webhooks_matrix_port }}/;
```

I haven't tested this though.

roles/matrix-bridge-appservice-webhooks/tasks/init.yml
```diff
@@ -0,0 +1,79 @@
...
+          proxy_pass {{ matrix_appservice_webhooks_appservice_url }}:{{ matrix_appservice_webhooks_matrix_port }};
+        {% else %}
+          {# Generic configuration for use outside of our container setup #}
+          proxy_pass http://127.0.0.1:{{ matrix_appservice_webhooks_matrix_port }};
```
spantaleev: No prefix-removal for this use-case (when matrix-nginx-proxy is disabled)?

We probably either need the `rewrite` rule applied here as well (or rather the previous rule being moved outside the `if`-statement.. Or better yet, getting rid of the `rewrite` rule altogether (see my other comment for a potential untested solution).

spantaleev:
Thank you for working on this!

I've given it a quick test (installation-wise) and have provided some comments and potential ideas for improvement. I haven't tested talking to the bot, etc., at least not yet.

roles/matrix-bridge-appservice-webhooks/tasks/init.yml
CybotTM: strange, we tried this and it failed, now we re-tried and it worked ... :-/
... seems to be some side effect whether using $backend variable proxy_pass target or not (which we already removed).

roles/matrix-bridge-appservice-webhooks/tasks/init.yml
tripleawwy: You're completely right about that.

roles/matrix-bridge-appservice-webhooks/defaults/main.yml
tripleawwy: Excuse me for accidentally commiting this portion. I somehow lost overview. Never the less, this is now adjusted.

roles/matrix-bridge-appservice-webhooks/defaults/main.yml
tripleawwy: Yet again, completely right about file based logging. I flipped ```writeFiles``` to ```false``` but I will stick with ```info``` for ```consoleLevel``` since the webhook bridge actually does spill out quite a lot of information.

tripleawwy: In order to change the verbosity level, I added ```matrix_appservice_webhooks_log_level``` variable.

spantaleev: Sounds good!

roles/matrix-bridge-appservice-webhooks/tasks/init.yml
tripleawwy: Since we were using the ````matrix-bridge-appservice-slack``` as basis for our configuration and failed to implement the slash in the end with the $backend variable, we decided to go with the rewrite thingie.

After seeing your comment, I decided to try out without using $backend variable and it magically worked.

roles/matrix-bridge-appservice-webhooks/tasks/init.yml
tripleawwy: Got rid of the rewrite rule.

roles/matrix-bridge-appservice-webhooks/tasks/init.yml
spantaleev: Happy to hear it works without it! Much simpler! :)

spantaleev:
Thanks a lot for fixing things up! Would you be adding documentation before merging this or would you rather do it later?

tripleawwy:
I'd rather create a documention before merging, since I tend to being pretty lazy ;p

spantaleev:
Sounds good! Looking forward to it then!

docs/configuring-playbook-bridge-appservice-webhooks.md
```diff
@@ -0,0 +1,58 @@
...
+loosely based on [this](https://github.com/turt2live/matrix-appservice-webhooks/blob/master/README.md)
+
+1. All you basically need is to adjust your `inventory/host_vars/matrix.<domain-name>/host-vars.yml`:
+    ```yaml
+    matrix_appservice_webhooks_enabled: true
+    matrix_appservice_webhooks_api_secret: '<your_secret>'
+    ```
+
+2. In case you want to change the verbosity of logging via `journalctl -fu matrix-appservice-webhooks.service`
+you can adjust this in `inventory/host_vars/matrix.<domain-name>/host-vars.yml` as well.
```
spantaleev: The file name is just `vars.yml`, not `host-vars.yml`.

docs/configuring-playbook-bridge-appservice-webhooks.md
```diff
@@ -0,0 +1,58 @@
...
+    ```yaml
+    matrix_appservice_webhooks_log_level: '<log_level>'
+    ```
+
+3. If you've already installed Matrix services using the playbook before, you'll need to re-run it (`--tags=setup-all,start`). If not, proceed with [configuring other playbook services](configuring-playbook.md) and then with [Installing](installing.md). Get back to this guide once ready.
+
+4. Invite the bridge bot user to your room:
+    - either with:
+      ```
+      /invite @_webhook:matrix.<domain.name>
```
spantaleev: This should probably just be `@_webhook:<domain.name>` (not a `matrix.` subdomain)

roles/matrix-bridge-appservice-webhooks/defaults/main.yml
```diff
@@ -0,0 +1,114 @@
...
+matrix_appservice_webhooks_enabled: true
+
+matrix_appservice_webhooks_docker_image: "turt2live/matrix-appservice-webhooks:latest"
+matrix_appservice_webhooks_docker_image_force_pull: "{{ matrix_appservice_webhooks_docker_image.endswith(':latest') }}"
+
+matrix_appservice_webhooks_base_path: "{{ matrix_base_data_path }}/appservice-webhooks"
+matrix_appservice_webhooks_config_path: "{{ matrix_appservice_webhooks_base_path }}/config"
+matrix_appservice_webhooks_data_path: "{{ matrix_appservice_webhooks_base_path }}/data"
+
+matrix_appservice_webhooks_public_endpoint: /appservice-webhooks
```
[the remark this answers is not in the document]
spantaleev: This is also passed to the bridge itself (as part of `matrix_appservice_webhooks_inbound_uri_prefix`), so it sounds like something that needs to be configured on any other reverse-proxy (not just matrix-nginx-proxy).

roles/matrix-bridge-appservice-webhooks/defaults/main.yml
```diff
@@ -0,0 +1,114 @@
...
+
+  # Configuration specific to the bridge. All fields (unless otherwise marked) are required.
+  webhookBot:
+    # The localpart to use for the bot. May require re-registering the application service.
+    localpart: "_webhook"
+
+    # Appearance options for the Matrix bot
+    appearance:
+      displayName: "Webhook Bridge"
+      avatarUrl: "http://i.imgur.com/IDOBtEJ.png" # webhook icon
```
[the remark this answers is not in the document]
tripleawwy: This apparently seems to be unused since the app's maintainer disabled the `updateBotProfile()` function. I have reached out to him and am awaiting an answer.
image

docs/configuring-playbook-bridge-appservice-webhooks.md
tripleawwy: Adjusted.

docs/configuring-playbook-bridge-appservice-webhooks.md
tripleawwy: I did adjust this as well and of course you're right.

roles/matrix-bridge-appservice-webhooks/defaults/main.yml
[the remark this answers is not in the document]
tripleawwy: I added a note that the app itself is listening on its own domain and not on /appservice-webhook directory.

roles/matrix-bridge-appservice-webhooks/defaults/main.yml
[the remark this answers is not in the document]
tripleawwy: image


Good thing: he's working on a rewrite!

roles/matrix-bridge-appservice-webhooks/defaults/main.yml
```diff
@@ -0,0 +1,114 @@
...
+
+  logging:
+    file: data/webhook.log
+    console: true
+    consoleLevel: {{ matrix_appservice_webhooks_log_level }}
+    fileLevel: verbose
+    writeFiles: false
+    rotate:
+      size: 52428800 # bytes, default is 50mb
+      count: 5
```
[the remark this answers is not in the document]
tripleawwy: I didn't think of just leaving it blank. Good idea and adjusted.

spantaleev:
Unless you have something else to add, I think this looks good for merging.

tripleawwy:
I think we're good to go.

spantaleev:
Merged! Thanks a lot for contributing this often-requested feature! :+1:

PR #351, opened 2020-01-20
Matrix change user admin status

prasket:
Added new script which allowed adding and removing of Administrator access. This replaces the `matrix-make-user-admin` script. 

I have also updated the documentation to reflect this and made some minor changes around the larger change in docs to help clarify things.


roles/matrix-postgres/templates/usr-local-bin/matrix-change-user-admin-status.j2
```diff
@@ -0,0 +1,19 @@
...
+
+docker run \
+	-it \
+	--rm \
+	--user=991:991 \
+	--cap-drop=ALL \
+	--env-file=/matrix/postgres/env-postgres-psql \
+	--network matrix \
+	postgres:12.1-alpine \
+	psql -h matrix-postgres -c "UPDATE users set admin=$2 WHERE name like '@$1:seattlematrix.org'"
```
spantaleev: Need to use `{{ matrix_domain }}` here ;)

roles/matrix-postgres/tasks/setup_postgres.yml
```diff
@@ -73,10 +73,10 @@
...
 
-- name: Ensure matrix-make-user-admin script created
+- name: Ensure matrix-change-user-admin-status script created
   template:
-    src: "{{ role_path }}/templates/usr-local-bin/matrix-make-user-admin.j2"
-    dest: "/usr/local/bin/matrix-make-user-admin"
+    src: "{{ role_path }}/templates/usr-local-bin/matrix-change-user-admin-status.j2"
+    dest: "/usr/local/bin/matrix-change-user-admin-status"
     mode: 0750
   when: matrix_postgres_enabled|bool
```
spantaleev: It's probably good to also add a cleanup task for the old file as well. Something like:

```yaml
- name: (Migration) Ensure old matrix-make-user-admin script deleted
  file:
    path: /usr/local/bin/matrix-make-user-admin
    state: absent
  when: matrix_postgres_enabled|bool
```

roles/matrix-postgres/templates/usr-local-bin/matrix-change-user-admin-status.j2
prasket: Doh! Ha I was testing it, I will get that fixed up soon and resubmit.

PR #354, opened 2020-01-22
uri.follow_redirects is now a string field

thedanbob:
Documented here: https://docs.ansible.com/ansible/2.9/modules/uri_module.html


spantaleev:
Looks like support for such string values (``none` and `safe` used by us) goes back to at least Ansible v2.5 (https://docs.ansible.com/ansible/2.5/modules/uri_module.html), so this should not cause backward-compatibility issues.

Thank you! :+1:

PR #357, opened 2020-01-24
Fix invalid password salts for slack webhooks

zekesonxx:
Under the previous salts, this resulted in the incredibly cryptic error of:

```
original message: An unhandled exception occurred while templating '{{ matrix_synapse_macaroon_secret_key | password_hash('sha512', 'webhooks-id-token') | to_uuid }}'.
Error was a <class 'ValueError'>, original message: invalid characters in sha512_crypt salt
```

Ansible password salts must be <= 16 characters and have a variety of restrictions on the contents, namely no `-` characters.


spantaleev:
Thank you! :+1:

PR #358, opened 2020-01-24
Install python3-docker if ansible is running python 3

thedanbob:


spantaleev:
I have no servers against which to test this, but I'm hoping it works well!

Thank you! :+1:

dhose:
I had an issue when the if case evaluates to false,  see #359 for a fix.

PR #359, opened 2020-01-24
Fix for when if case evaluates to false

dhose:
I got this error when the if case evalutes to false, so I added an else statement to fix the error.

```
FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: the inline if-expression on line 1 evaluated to false and no else section was defined.
```


funar:
I wish i had seen this before I submitted mine.,. haha

spantaleev:
Thanks folks! :+1: 

Thinking about it, it's fairly obvious and quite silly that this would not panic when it evaluates to `true`..

PR #360, opened 2020-01-24
Added condition if ansible is not running on python3

funar:
"'3' if ansible_python.version.major == 3" fatally exits on systems with Python versions less than 3.  Adding the "else" statement corrects the fatal error.


funar:
You can disregard this.  I did not see the previously submitted PR.

PR #361, opened 2020-01-25
Switch to upgraded & official matrix-appservice-irc Docker image

spantaleev:
77b919aec690a does most of the work to make this happen.

Still, it'd be nice to test this some more, especially by people who have been running the old bridge and have an existing database.


spantaleev:
Got feedback that the above works well, so I'll merge it in.

PR #362, opened 2020-01-25
Switch from kamax-matrix/matrix-synapse-rest-password-provider to ma1uta/matrix-synapse-rest-password-provider

spantaleev:
Synapse v1.9.0 changed some things which made the REST Auth Password
Provider break.

The ma1uta/matrix-synapse-rest-password-provider implements some
workarounds for now and will likely deliver a proper fix in the future.

Not much has changed between the 2 projects, so this should be a
painless transition.

### DIFF
--- a/roles/matrix-synapse/defaults/main.yml
+++ b/roles/matrix-synapse/defaults/main.yml
@@ -263,9 +263,9 @@ matrix_synapse_email_client_base_url: "https://{{ matrix_server_fqn_riot }}"
 
 
 # Enable this to activate the REST auth password provider module.
-# See: https://github.com/kamax-io/matrix-synapse-rest-auth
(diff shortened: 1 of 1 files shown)


eMPee584:
URL needs to be .. https://github.com/ma1uta/matrix-synapse-rest-password-provider/raw/ed377fb70513c2e51b42055eb364195af1ccaf33/rest_auth_provider.py
else synapse does not start with
```
   […]
   File "/usr/local/lib/python3.7/site-packages/rest_auth_provider.py", line 7
     <!DOCTYPE html>
     ^
 SyntaxError: invalid syntax
```
😁

spantaleev:
Ouch, that was a silly mistake. Thanks for reporting it! Should be fixed now!

Issue #365, opened 2020-02-04
Error when installing slack bridge


joao-p-marques:
When installing the slack bridge, reaching 
```
TASK [matrix-bridge-appservice-slack : Ensure appservice-slack registration.yaml installed]
``` 

I get the error:
```
crypt.crypt does not support 'sha512_crypt' algorithm
```

I've followed the guide on setting up the `vars.yml` file

spantaleev:
#357 seems somewhat related, but I don't see how it could have broken things.

From what I remember, Ansible's `password_hash` function was using one of two things (either Python's crypt module, or something else - can't remember what). It could be that whether this bug happens is dependent on which method is used and which Python version runs on the server.

Could you run this from the playbook's directory?

`ansible -i inventory/hosts -m setup 'matrix_servers' | grep -i ansible_python_version`

joao-p-marques:
Yes, I've looked into #357, but it seems to address only the webhooks bridge... I am using the Event Bridging...

The command outputed: `"ansible_python_version": "2.7.15+"`

spantaleev:
This is solved now, right?

joao-p-marques:
Yes, it was fixed by #368

PR #368, opened 2020-02-06
fix invalid password salt on slack bridge

joao-p-marques:
Change usage of `-` on slack bridge configuration files (password hash) to `.`, following the same principle as in #357 for slack webhooks, fixing #365.


spantaleev:
The error message seems different compared to #357, but I hope this helps.
Merged! Thank you! :+1:

PR #370, opened 2020-02-08
Fix matrix-appservice-irc database specification

verb:
This fixes an incorrect indentation in the database specification for appservice-irc which caused matrix-appservice-irc to refuse to start with the remarkably unhelpful error message:

```
ERROR:CLI Failed to run bridge.
```

This also updates doc links to the new upstream repo because the tedomum repo contains out-of-date documentation.


spantaleev:
Good catch! Thank you! :+1:

PR #374, opened 2020-02-10
matrix-nginx-proxy: adding additional configuration blocks

dhoffend:
Adding the ability to add additional configuration block not only the synapse vhost, but also to the nginx-http block, riot, dimension and domain vhost.

In my use case I needed to add additional arguments to import the real-ip (because my deployment is running behind a traefik reverse proxy).

Example usage:
```
# import real-ip from traefik reverse proxy
matrix_nginx_proxy_proxy_http_additional_server_configuration_blocks:
  - set_real_ip_from 172.16.0.0/12;
  - real_ip_header X-Real-Ip;
  - real_ip_recursive on;
```


spantaleev:
Looks good and useful! Thank you! :+1:

PR #381, opened 2020-02-18
add synapse config enable_group_creation

orangeman:


spantaleev:
I've fixed it up a bit to add `|to_json`, just in case.

Thank you! :+1:

Issue #382, opened 2020-02-19
matrix-appservice-webhooks incorrectly exposes matrix_port


sathiscode:
Configuring matrix_appservice_webhooks_container_http_host_bind_port should expose matrix_appservice_webhooks_webhooks_port, but instead it exposes matrix_appservice_webhooks_matrix_port

sathiscode:
#383 - PR for this issue

spantaleev:
Thanks for fixing it in #383!

PR #383, opened 2020-02-19
Fixed appservice webhooks container port mapping

sathiscode:
appservice webhooks container was mapped to matrix port instead of webhooks port.


sathiscode:
Issue for this PR - #382

spantaleev:
Good catch! Thank you! :+1:

PR #384, opened 2020-02-20
Configuring webhooks

sathiscode:
Added steps to configure appservice webhooks in the dimension manager.


spantaleev:
Thank you! :+1: 

I have modified wording a bit and linked to the docs page we have on Dimension, so that people can easily jump to it and set it up, if they haven't done so already.

PR #386, opened 2020-02-23
Raspberry pi

hooger:
I added some changes so it can be run on a raspberry pi. I also added some documentation.


docs/raspberry-pi.md
```diff
@@ -0,0 +1,17 @@
+# Raspberry Pi
+
+The playbook support for Raspberry Pi is now in beta(ish). The problem is that, docker containers are not architecture independent, and most of them are not build for arm. Some roles have been updated, so they built the necessary image on the host. It needs more space, as some build tools need to be present (like Java, for mxisd).
+
+To use these modification there is a variable that needs to be switched to enable this functionality. Add this to your vars.yaml file:
+```
+matrix_raspberry_pi = true
```
spantaleev: This is not true anymore.

In its place, we should probably introduce a new `matrix_container_images_self_build` variable in `roles/matrix-base/defaults/main.yml` (defaulting to `false`).

In `group_vars/matrix_servers`, we can wire it to all other `*_self_build` variables. Example:

```yaml
matrix_riot_web_self_build: "{{ matrix_container_images_self_build }}"
```

roles/matrix-riot-web/tasks/setup_riot_web.yml
```diff
@@ -19,7 +22,25 @@
...
+
+- name: Ensure Riot Web repository is present on self-build
+  git:
+    repo: https://github.com/vector-im/riot-web.git
+    dest: "{{ matrix_riot_web_docker_src_files_path }}"
+    version: "v{{ matrix_riot_web_docker_image.split(':')[1] }}"
+    force: "yes"
+  when: "matrix_riot_web_enabled|bool and matrix_riot_web_self_build"
+
+- name: Ensure Riot Web Docker image is build
```
spantaleev: This currently seems to fail:

> /usr/bin/git checkout --force vv1.5.12", "msg": "Failed to checkout vv1.5.12", "rc": 1, "stderr": "error: pathspec 'vv1.5.12' did not match any file(s) known to git.

roles/matrix-mxisd/tasks/setup_mxisd.yml
```diff
@@ -22,7 +23,29 @@
...
+- name: Ensure mxisd repository is present on self-build
+  git:
+    repo: https://github.com/kamax-matrix/mxisd.git
+    dest: "{{ matrix_mxisd_docker_src_files_path }}"
+    version: "v{{ matrix_mxisd_docker_image.split(':')[1] }}"
+    force: "yes"
+  when: "matrix_mxisd_enabled|bool and matrix_mxisd_self_build"
+
+- name: Ensure mxisd Docker image is build
+  shell: "./gradlew dockerBuild"
```
spantaleev: I wonder if there's a way to avoid running gradle and just invoke `docker build ..` directly. Well, via the `docker_image` Ansible module, like we do for the others.

But I guess you've already toyed with that and it's not as easy..

hooger: Unfortunately I haven't found any way. :/

hooger:
I pushed the changes you've asked for

roles/matrix-mxisd/tasks/setup_mxisd.yml
spantaleev: Indeed.. It looks like building the image relies on copying the jar file into it.

The actual building of the jar file happens outside and requires pulling in dependencies, all done using `gradle`.

spantaleev:
I've fixed up some minor things and have merged this!

Thanks a lot for doing all this work on getting this ready! :bowing_man: 
So many people can now experiment with the playbook on a Pi!

Hopefully, this will get polished by the community even more in the future - adding self-building support to more and more roles.. Or better yet, getting upstream projects to publish multi-arch images.

PR #392, opened 2020-03-02
Switch Webhooks Port, which is exposed for external nginx

redoo-networks:
Hy,

When using Webhooks together with external nginx, the default ports are wrong in my opinion. This Merge Request modify Webhooks Appservice so it is working.

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/roles/matrix-bridge-appservice-webhooks/templates/systemd/matrix-appservice-webhooks.service.j2

When you look at this file, you see node.js command, which is start with port "{{ matrix_appservice_webhooks_matrix_port }}" (Default: 6789)
But the Port, which is exposed by default is "{{matrix_appservice_webhooks_webhooks_port}}" (Default: 6788)
But on Port 6788 there nothing is listening in docker container.

So I switch the exposed variable to "matrix_appservice_webhooks_matrix_port", because I don't wanted to modify the ports for "non external nginx".
Normally shouldn't breaks something, because to make it working it was necessary to set both variables to same port.

I cannot imagine, this problem is related to our situation, because two ports, which must be same, are different and default values makes no sense.
But because there must be a reason to split both ports, maybe within the image, something was changed.

Regards,
Stefan

### DIFF
--- a/roles/matrix-bridge-appservice-webhooks/templates/systemd/matrix-appservice-webhooks.service.j2
+++ b/roles/matrix-bridge-appservice-webhooks/templates/systemd/matrix-appservice-webhooks.service.j2
@@ -23,7 +23,7 @@ ExecStart=/usr/bin/docker run --rm --name matrix-appservice-webhooks \
 			--cap-drop=ALL \
 			--network={{ matrix_docker_network }} \
 			{% if matrix_appservice_webhooks_container_http_host_bind_port %}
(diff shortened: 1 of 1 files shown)


spantaleev:
This was recently touched by #383.

Grepping for `matrix_appservice_webhooks_webhooks_port`, it doesn't look like it's used for anything meaningful, so I'm not sure why we have it.

I've merged your changes and have also completely removed `matrix_appservice_webhooks_webhooks_port` in bbe7bff.

Hopefully there's nothing I've missed.

Thank you! :+1:

PR #394, opened 2020-03-03
Customization welcome

swarnat:
Because we are starting using Matrix/Riot productive, I checked Riot-Web README for new configurations and found some I'm searching for some time time, because both logos on welcome page are handled different and wallpaper cannot changed.

Now customization of logo during login and wallpaper is also able to configure.

Pay attention: 
The way I choose will change default behavior of UI, because in past the second "Login Logo" cannot configured. 
Now it is set to same logo as on welcome startpage. So this logo maybe changed from riot logo to branding.
When default of "matrix_riot_web_welcome_authlogo" is set to "false", the old behavior will used.

About "matrix_riot_web_welcome_headline", this isn't always my preferred method to insert "_t(..)", too, but I think the translation cannot modified by config. So this can also overwritten.

### DIFF
--- a/roles/matrix-riot-web/defaults/main.yml
+++ b/roles/matrix-riot-web/defaults/main.yml
@@ -43,13 +43,13 @@ matrix_riot_web_welcome_text: "_t('Decentralised, encrypted chat &amp; collabora
 
 # Links, shown in footer of welcome page:
 # [{"text": "Link text", "url": "https://link.target"}, {"text": "Other link"}]
-matrix_riot_web_welcome_footerlinks: false
+matrix_riot_web_branding_authFooterLinks: ~
 
 # URL to image, shown during Login
-matrix_riot_web_welcome_authlogo: "{{ matrix_riot_web_welcome_logo }}"
+matrix_riot_web_branding_authHeaderLogoUrl: "{{ matrix_riot_web_welcome_logo }}"
 
 # URL to Wallpaper, shown in background of welcome page
-matrix_riot_web_welcome_background: false
+matrix_riot_web_branding_welcomeBackgroundUrl: ~
 
 # By default, there's no Riot homepage (when logged in). If you wish to have one,
 # point this to a `home.html` template file on your local filesystem.
--- a/roles/matrix-riot-web/defaults/main.yml
+++ b/roles/matrix-riot-web/defaults/main.yml
@@ -32,7 +32,20 @@ matrix_riot_web_welcome_user_id: "@riot-bot:matrix.org"
 # Branding of riot web
 matrix_riot_web_brand: "Riot"
 
+# Links, shown in footer of welcome page:
+# [{"text": "Link text", "url": "https://link.target"}, {"text": "Other link"}]
+matrix_riot_web_welcome_footerlinks: false
+
+# URL to image, shown during Login
+matrix_riot_web_welcome_authlogo: false
+
+# URL to Wallpaper, shown in background of welcome page
+matrix_riot_web_welcome_background: false
+
+# URL to Logo on welcome page
 matrix_riot_web_welcome_logo: "welcome/images/logo.svg"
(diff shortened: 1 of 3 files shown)


roles/matrix-riot-web/defaults/main.yml
```diff
@@ -32,6 +32,25 @@ matrix_riot_web_welcome_user_id: "@riot-bot:matrix.org"
...
+
+# URL of link on welcome image
+matrix_riot_web_welcome_logo_link: "https://riot.im"
+
+matrix_riot_web_welcome_headline: "_t('Welcome to Riot.im')"
+matrix_riot_web_welcome_text: "_t('Decentralised, encrypted chat &amp; collaboration powered by [matrix]')"
+
+# Links, shown in footer of welcome page:
+# [{"text": "Link text", "url": "https://link.target"}, {"text": "Other link"}]
+matrix_riot_web_welcome_footerlinks: false
```
spantaleev: It's more consistent with the rest of the playbook to name this `matrix_riot_web_branding_authFooterLinks`. Similarly for the other variables which end up in the `config.json.j2` template.

If some variable serves multiple purposes or is used elsewhere (`welcome.html.j2`), it can probably be named differently.

Is `false` a good default value for this, indicating "do nothing special"? What if we send an empty list (`[]`)?
I assume you've tested this, by it doesn't hurt asking just to be sure :)

roles/matrix-riot-web/defaults/main.yml
```diff
@@ -32,6 +32,25 @@ matrix_riot_web_welcome_user_id: "@riot-bot:matrix.org"
...
+
+# URL to image, shown during Login
+matrix_riot_web_welcome_authlogo: "{{ matrix_riot_web_welcome_logo }}"
+
+# URL to Wallpaper, shown in background of welcome page
+matrix_riot_web_welcome_background: false
```
spantaleev: Likewise, I'd like to confirm that `false` is a good value for what would normally be a string.

What if we use an empty string or null (`~`)?

redoo-networks:
Just adjusted the variable names and now I understand the naming structure. 

**About false or `~`?**
Both it correct. Condition about footerLinks in Riot-web, for example of footer links, is [here](https://github.com/vector-im/riot-web/blob/421cd4127942817b9bde8d710e22351bebefe7c0/src/components/views/auth/VectorAuthFooter.js#L30)
In case, where "optimized" developer get rid of a value, to check with, I tend to set default values to "false", like "Is it set?" "false".
But everything which is empty, will fulfill it.
Probably `~` is a more Ansible like way and look cleaner. That's why I also changed it.

All are tested and, except wallpaper, in used in our system on redoo.chat. Wallpaper now is set to null in our system and works, too.

spantaleev:
Thank you for your good work and diligence! :+1:

teutat3s:
After using `matrix_riot_web_welcome_headline` I get this on the new welcome / login site:
```
missing translation: en-us|Welcome to riot.my.domain
```
could you please try if this is reproduceable? As you can see, I just changed the domain, nothing else.

spantaleev:
Well, the `_t()` function invokes the translation system, which has a translation for the verbatim string `'Welcome to Riot.im'`, but not for `'Welcome to riot.my.domain'`.

You can either:

- refrain from calling into the translation system and just hardcode text in whichever language you choose: `matrix_riot_web_welcome_headline: "Hello! Welcome to riot.my.domain"`. The downside is that the text will appear as you see it, regardless of the language selected in Riot.

- try to call into the translation system and perform a string replace on the final result. Example: `matrix_riot_web_welcome_headline: "_t('Welcome to Riot.im').replace('Riot.im', 'riot.my.domain')"`. However, I'm not sure how the `welcome.html` template is processed by Riot and whether you can call `.replace()` like this there.. Likely not.

PR #410, opened 2020-03-20
how to activate dimensions

ayounggun:
I am new to synapse. Thanks so much for these playbooks!

I wasn't sure how you actual activate dimensions after setting these variables. Should you re run 

ansible-playbook -i inventory/hosts setup.yml --tags=setup-all
ansible-playbook -i inventory/hosts setup.yml --tags=start

? If so perhaps you could tell the readers to do this somewhere in the help file.

PS I don't know how to use gitbub so sorry if what I'm doing is incorrect.

Thanks =]


spantaleev:
Thank you for improving the docs! :+1:

PR #415, opened 2020-03-23
Add Jitsi support

### LINKED ISSUE #402: Self host Jitsi
https://microblog.shivering-isles.com/@sheogorath/103833245934684102

spantaleev:
Fixes #402.

Current downsides:
- requires a `develop` version of riot-web, until a stable version is released
- requires pointing Jitsi JVB to an external STUN server (we use Google's STUN servers by default). Pointing it to our own Coturn leads to incorrect external IP address discovery and a non-working service. **No longer that big of a problem** (see below).
- ~~changing most `matrix_jitsi_*` variables will likely not result in changes taking effect, because of the way the Jitsi containers work -- they use our environment variables only the first time, to generate some configuration files, and they don't obey environment variables anymore, unless the configuration is deleted. We may wish to add some `--setup-jitsi-clean` Ansible tag or something, which will stop `matrix-jitsi-*` services, deletes `/matrix/jitsi/*/config` directories and then rebuilds the configuration.~~


spantaleev:
My research shows that there's 2 STUN server definitions, used for different things.

The STUN servers passed to JVB only seem to be used for its own sake -- so that it can discover its public IP address. Using Google's STUN servers or some other public ones doesn't seem like too big of a problem for this. Yes, it does leak some data, but it's at the server level.

-----------------

Then there are the STUN servers used in the web UI. Those are hardcoded to the same values (Google STUN servers) in the `jitsi/web` Docker image's default [config.js](https://github.com/jitsi/docker-jitsi-meet/blob/master/web/rootfs/defaults/config.js) file. This file is modified slightly based on environment variables the first time the `jitsi/web` image is started. Relevant code is here: https://github.com/jitsi/docker-jitsi-meet/blob/76acc6502734d9bca5fa78b67102cfbc7a4ad29c/web/rootfs/etc/cont-init.d/10-config#L64

It doesn't seem like we can control what ends up in `stunServers` for the web UI via environment variables, so clients are forced to use Google's STUN servers.

We either need to request such a new environment variable or make a pull request to [jitsi/docker-jitsi-meet](https://github.com/jitsi/docker-jitsi-meet). We'd then need to wait for a new release of `jitsi/web`, which may take some time.

Alternatively, we could look into moving away from these environment variables and managing our own configuration files via Ansible. This will give us much more control. At the same time, it will also fix our "configuration does not get regenerated" problem (described above).

Looking at the "startup code" for `jitsi/web`, at least, it seems like we can generate `config.js` ourselves and let it run to generate all other config files on its own. So we can do it file by file, and not go all in from the start. Some services (like jitsi/web) have multiple such configuration files and it would be annoying to recreate all that via Ansible from the start. Other services (like jitsi/jvb) are much simpler and we can easily move away from environment variables for them.

spantaleev:
In d605b21 I've done some changes to how we generate the configuration for the various components.

We now try to generate it ourselves and not rely on Jitsi containers generating it the first time.

Because of this, playbook variable changes have a higher chance to take effect. For most things it should be alright. It's not a guarantee, however, because we still let Jitsi containers generate some of the remaining configuration files.

I've also fixed the STUN server problem. From now on, JVB uses Google STUN servers by default, to discover its own external IP address. This is not ideal, but it's okay. We may wish to provide a few more default STUN servers, so that it won't be breaking in regions which don't have access to Google services.

The web UI and what clients use for STUN is pointed to our own Coturn by default from now on (if Coturn is enabled), so individual users do not communicate with Google STUN servers.

Issue #64, opened 2019-01-07
Don't supervise the docker binary with systemd


jcgruenhage:
The docker binary is just a rest client that is talking to the docker daemon, which means that you aren't supervising the services but just the docker binary. I'm not sure what the reason for this is, but it means you can't use this playbook on alpinelinux, voidlinux, gentoo and possibly more. 

I have multiple suggestions on how to solve this:
 - Switch to a different container stack, that doesn't rely on a background daemon to run containers (podman.io for example), which interacts more nicely with supervision suites, and add support for more supervisors
 - Don't supervise the docker binary but let the docker daemon manage the containers completely (which it will anyway, it's just a matter of not wrapping it in systemd)

I strongly prefer option 1 for my usecases, but since podman isn't available to most users, that probably won't be possible.

spantaleev:
That's right.. Unfortunately, we are not really supervising the actual containers, but just the Docker client.

Still, having systemd "supervise" is useful because:

- all output generated by each container propagates to its corresponding systemd serivce. The output is then logged in a standard way (`journald`). We intentionally suppress Docker logging (`--log-driver=none`) to prevent double-logging (once by Docker and once by systemd) or going out of disk space in `/var/lib/docker` due to excessive logging.

- the Docker client will die when the container process dies.. Which will report to the systemd service and an automated restart will happen, as configured

Docker is an implementation detail (admittedly, one we can't easily get away from - see the next section below).
We'd rather Docker is not managing containers, restarts, logging or anything of that sort by itself. The fact that we can't really supervise the actual process is not ideal, but changing to "just let Docker do it all" (option 2 from your description) sounds even worse.

--------

About option 1, I've thought about going for another container stack, but it seems like it won't be easy (or possible).

I've thought about [rkt](https://coreos.com/rkt/) before, but it lacked some features. Things may be better now - I haven't checked.

Looking at [podman](https://podman.io/), it doesn't seem like it's available for most distros, like you said.

Even if it is, we need to check if it supports everything that we use (or if we can squeeze into whatever it supports). While running multiple containers and having them talk to one another should be possible with podman too, we also rely on some implementation details in Docker (embedded DNS resolver being at `127.0.0.11` -- we use this in `matrix-nginx-proxy` vhosts).

[...]

PR #418, opened 2020-03-24
Made directory variables for system files

mooomooo:
Wrap system paths in vars:

matrix_systemd_path: "/etc/systemd/system"
matrix_cron_path: "/etc/cron.d"
matrix_local_bin_path: "/usr/local/bin"

which can then be overridden in inventory/host_vars if you want to use some other container management system.  This helps with #64 

N.B. These paths must already exist on your system.  Tasks can certainly be added to the playbook to ensure they exist, but I wanted to make the minimum change first.


spantaleev:
Thank you! :+1: 
Glad to hear it helps you with the composify thing!

PR #420, opened 2020-03-26
clarify the config path

PlanetSmasher:
added path to the correct vars.yaml in matrix-synapse to help better configuration for beginners


aaronraimist:
No that is not where you edit the config. The only config file you edit is `inventory/host_vars/matrix.<your-domain>/vars.yml`. Don't change things inside the `roles` folder.

spantaleev:
Clarifying is always a good idea. Aaron is right though, you're supposed to use `inventory/host_vars/matrix.<your-domain>/vars.yml`.

I've adjusted your patch and merged it. Thanks! :+1:

PR #423, opened 2020-03-28
Flip around Jitsi port format to match other ports

aaronraimist:
This matches the way the ports are written in prerequisites.md and is the format that UFW likes.


spantaleev:
Makes sense! Thank you! :+1:

PR #426, opened 2020-03-29
Fix bad network and bad path to postgres env in docker run

bnounours:


spantaleev:
Better than hardcoding! Thanks! :+1:

PR #432, opened 2020-04-03
fix jitsi guests mode according to docker-jitsi-meet repo

teutat3s:
According to [docker-jitsi-meet](https://github.com/jitsi/docker-jitsi-meet/blob/fabfb2a676175894510f78ec8de5d10ae8c10e8c/web/rootfs/etc/cont-init.d/10-config#L89-L99) repo


spantaleev:
Thank you! 👍🏻

PR #434, opened 2020-04-04
add missing jitsi auth URL conditional

teutat3s:
While double checking on the `internal` authentication in [docker-jitsi-meet](https://github.com/jitsi/docker-jitsi-meet) I noticed a missing conditional, that this PR adds.

For reference, see [sip-communicator.properties](https://github.com/jitsi/docker-jitsi-meet/blob/7c0c795fa7d29d04248ec6201ef7b879f4dc537f/jicofo/rootfs/defaults/sip-communicator.properties#L13-L15) in that repo.


spantaleev:
Thank you! :+1:

PR #435, opened 2020-04-04
jitsi-web: add more defaults

alindt:


spantaleev:
Nice! Thank you! :+1:

PR #436, opened 2020-04-05
add optional Jitsi internal auth+guests docs

teutat3s:
This PR adds documentation for how to setup Jitsi with authentication and guests mode.

@spantaleev I'd like to add reviewers, but it seems disabled for this repo.


spantaleev:
I've tested this and added a bit more information. It looks like it works well!

Thank you for chasing down these things and making authenticated Jitsi work! :+1:

PR #437, opened 2020-04-05
add ntpd defaults

alindt:


spantaleev:
Thanks for cleaning things up and make it more configurable! :+1:

PR #438, opened 2020-04-05
Update Caddy example in configuring-well-known to Caddy version 2

SeanPONeil:


spantaleev:
I'm not familiar with Caddy or the differences between v1 and v2, but I'll take your word for it ;)
Thank you for improving the docs! :+1:

PR #440, opened 2020-04-06
Facebook self build

hooger:
Slight modification to force self-build images when the git master branch has been modified, as currently the docker image xy:latest is always present.


spantaleev:
Thanks! :+1:

PR #444, opened 2020-04-08
Jitsi security update

teutat3s:
For background information click [here](https://github.com/jitsi/docker-jitsi-meet#security-note), [here](https://github.com/jitsi/docker-jitsi-meet/blob/master/CHANGELOG.md) and [here](https://github.com/jitsi/docker-jitsi-meet/commit/768b6c4a50d75b143ca5006e63182f1eb924076c).

This PR adds a bash script inspired by https://github.com/jitsi/docker-jitsi-meet/blob/master/gen-passwords.sh to override the default Jitsi passwords with stronger ones.

It also updates the Jitsi docker containers to the newest version, which will stop working with the default passwords, see their [changelog](https://github.com/jitsi/docker-jitsi-meet/blob/master/CHANGELOG.md).

We should probably add a hint in the changelog (docs, too?) for everyone who's already running Jitsi.
They should only need to remove their prosody config directory, as all other configs get generated by ansible AFAIK. The following should be enough to get them ready for the update:

```bash
# on the matrix host
rm -r /matrix/jitsi/prosody

# on the local laptop / computer
cd inventory/scripts
bash generate-jitsi-passwords.sh
ansible-playbook -i inventory/hosts setup.yml --tags=setup-jitsi,start

```

I'm curious about what @spantaleev thinks about adding a bash script to this clean ansible repo - maybe there's a better and more ansible-like way to do this - I'd be happy to learn.


inventory/scripts/generate-jitsi-passwords.sh
```diff
@@ -0,0 +1,50 @@
...
+JIBRI_RECORDER_PASSWORD=`generatePassword`
+JIBRI_XMPP_PASSWORD=`generatePassword`
+
+echo "" >> ../host_vars/${domain}/vars.yml
+echo "Jitsi passwords generated by inventory/scripts/gen-passwords.sh" >> ../host_vars/${domain}/vars.yml
+echo "matrix_jitsi_jicofo_component_secret: $JICOFO_COMPONENT_SECRET" >> ../host_vars/${domain}/vars.yml
+echo "matrix_jitsi_jicofo_auth_password: $JICOFO_AUTH_PASSWORD" >> ../host_vars/${domain}/vars.yml
+echo "matrix_jitsi_jvb_auth_password: $JVB_AUTH_PASSWORD" >> ../host_vars/${domain}/vars.yml
+echo "matrix_jitsi_jibri_recorder_password: $JIBRI_RECORDER_PASSWORD" >> ../host_vars/${domain}/vars.yml
+echo "matrix_jitsi_jibri_xmpp_password: $JIBRI_XMPP_PASSWORD" >> ../host_vars/${domain}/vars.yml
```
SuperSandro2000, line 50: Please take a look 
```suggestion
{
	echo ""
	echo "Jitsi passwords generated by inventory/scripts/gen-passwords.sh"
	echo "matrix_jitsi_jicofo_component_secret: $JICOFO_COMPONENT_SECRET" 
	echo "matrix_jitsi_jicofo_auth_password: $JICOFO_AUTH_PASSWORD"
	echo "matrix_jitsi_jvb_auth_password: $JVB_AUTH_PASSWORD"
	echo "matrix_jitsi_jibri_recorder_password: $JIBRI_RECORDER_PASSWORD" 
	echo "matrix_jitsi_jibri_xmpp_password: $JIBRI_XMPP_PASSWORD" 
} >> ../host_vars/${domain}/vars.yml
```

See https://github.com/koalaman/shellcheck/wiki/SC2129

inventory/scripts/generate-jitsi-passwords.sh
```diff
@@ -0,0 +1,50 @@
...
+}
+
+cd ../host_vars
+get_domain_dir
+
+JICOFO_COMPONENT_SECRET=`generatePassword`
+JICOFO_AUTH_PASSWORD=`generatePassword`
+JVB_AUTH_PASSWORD=`generatePassword`
+JIBRI_RECORDER_PASSWORD=`generatePassword`
+JIBRI_XMPP_PASSWORD=`generatePassword`
```
SuperSandro2000, line 42:
```suggestion
JICOFO_COMPONENT_SECRET=$(generatePassword)
JICOFO_AUTH_PASSWORD=$(generatePassword)
JVB_AUTH_PASSWORD=$(generatePassword)
JIBRI_RECORDER_PASSWORD=$(generatePassword)
JIBRI_XMPP_PASSWORD=$(generatePassword)
```

docs/configuring-playbook-jitsi.md
```diff
@@ -25,6 +25,17 @@ Add this to your `inventory/host_vars/matrix.DOMAIN/vars.yml` configuration:
...
+## Securing your Jitsi instance with strong passwords
+
+Please use the bash script provided in this repo to generate strong passwords for your Jitsi instance.
+Execute the following commands in your terminal from the root of this repo:
+```bash
+cd inventory/scripts
+bash generate-jitsi-passwords.sh
+```
+
+The script will add the corresponding ansible variables and passwords generated with `openssl rand -hex 16` to the bottom of your `inventory/host_vars/matrix.DOMAIN/vars.yml` configuration.
```
spantaleev, line 37: It would probably be better if the script was just generating passwords and telling you what to put in your `vars.yml` file.

It's less complex and would work fine for people who have more than 1 host in their Matrix inventory.

I guess that if you have more than 1 host, you're likely a more advanced user and can figure out how to generate the passwords manually by yourself, but.. It might be better to be less magical in any case.

spantaleev: I've pushed a change (see f798605836e46cc73), which would catch people not defining these settings.

It would also catch people who have previously installed Jitsi with the default secrets (because the defaults are now undefined, so they effectively have no secrets either).

I think what's left to do now is: simplify `generate-jitsi-passwords.sh`, so that it would merely do generation and tell people what to copy/paste and where.

We may also look into simplifying the documentation page a bit.

What are your thoughts on this?

spantaleev:
Thank you for getting us started on this!

I've added some reivew comments above and have also done some work (f798605836e46cc7) on catching mistakes (variables not being defined) or old installations with default passwords, so that we can point these people to the Jitsi setup docs page again.

As for having a bash script, it's probably not ideal, but I don't think it's too bad. Especially if we stick to simple things which work on both Linux and Mac.  I've simplified the script and docs page (1b27e0f812c652e) and have then renamed it (dcce276b869), so that it would be more manageable if we ever get to have many scripts inside `inventory/scripts`.

I think it's good to merge this as-is, but would by happy to wait a while for some testing and feedback from you guys!

docs/configuring-playbook-jitsi.md
spantaleev: I've gone ahead and done this. Looking forward to your feedback though.

inventory/scripts/generate-jitsi-passwords.sh
spantaleev: Fair point!

I've changed the script to not write to a file though, so this is no longer applicable.

spantaleev:
Looks like the new Jitsi JVB container required an `JVB_AUTH_PASSWORD` environment variable to be defined. Without it, it would refuse to start.

I've addressed this in 14786071afaf55ac.

teutat3s:
Looks great to me, thank you for the feedback and improvements!

spantaleev:
Merged! Thank you! :+1:

PR #446, opened 2020-04-09
Jitsi related firewall and port changes

skoal:


spantaleev:
I was wondering whether `5349` is `stuns` or `turns`.. Well, jitsi/web likely doesn't understand `turns`, so..

Thank you! :+1:

skoal:
please revert the turns change,its not valid,it shoud be stun anyway,sry,tested it here:
https://webrtc.github.io/samples/src/content/peerconnection/trickle-ice/

PR #448, opened 2020-04-09
Add global var to control skype image self build

rbelem:
Gergely Horváth:
A single line is missing from group_vars/matrix_servers: matrix_mx_puppet_skype_container_image_self_build: "{{ matrix_container_images_self_build }}",
so if the global variable is set, everything is being self-built


spantaleev:
Thanks! :+1: 

If this works well, we should adjust `docs/self-building.md` to mention that the Skype bridge can also be self-built.

PR #452, opened 2020-04-11
Fix to avoid putting an incorrect empty jitsi.preferredDomain when no…

jgoerzen:
… local jitsi


spantaleev:
Thanks for noticing and fixing this! :+1:

PR #455, opened 2020-04-13
matrix-appservice-slack: improved documentation

daudo:
better described how to get channel id, added a section how to unlink a channel and a troubleshooting section

### DIFF
--- a/docs/configuring-playbook-bridge-appservice-slack.md
+++ b/docs/configuring-playbook-bridge-appservice-slack.md
@@ -63,7 +63,7 @@ Note that the bot's domain is your server's domain **without the `matrix.` prefi
 
     * Invite the bot user to both the Slack and Matrix channels you would like to bridge using `/invite @slackbot` for slack and `/invite @slackbot:MY.DOMAIN` for matrix.
 
-    * Determine the "channel ID" that Slack uses to identify the channel, which can be found in the url https://XXX.slack.com/messages/<channel id>/.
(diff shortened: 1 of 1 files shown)


spantaleev:
Wasn't the original meaning the same? Open the channel in the browser and check the URL?

I guess the original didn't explain very much, but what we have after this patch is essentially the same thing explained in 2 different ways and these 2 ways being offered as alternatives?

Or am I misunderstanding something?

Maybe the explanation should be like this:

> * Determine the "channel ID" that Slack uses to identify the channel. You will see it when you open a given Slack channel in a browser. The URL reads something like this: `https://app.slack.com/client/XXX/<the channel id>/details/`.

daudo:
Even after some clicking around, I wasn't able to figure out where the original URL came from, but I thought I'd let the original URL there, just in case I am missing something.

So yes, your explanation makes perfect sense. I'll update accordingly.

As I walk through the steps laid out in the documentation, I've noted some additional things (most notably troubleshooting stuff) that I'll add to my PR, if you wish.

spantaleev:
If these other things are unrelated to the Slack bridge, we can probably do them in another PR :)

Thanks for keeping track of things and improving the docs!

daudo:
I've updated the PR to only have one way to retrieve the channel id plus added a section about unlinking and troubleshooting

spantaleev:
Much more detailed! Thank you! :+1:

PR #456, opened 2020-04-14
Synapse workers

eMPee584:
· first version: grokking this was a bit of 🤯 .. and there are probably still faults in the wiring
· :warning: ~~not tested impact on UX yet~~
· update 2020/04/21: live testing on our 7k+ uni matrix instance :sweat_smile: :sweat_drops: 
· update 2020/06/30: **works very well** to spread load, now with 10k+ users
· update 2020/09/09: currently working on expanding the patch queue to cover recent changes to synapse workers.. not much left but it might need another week.
· update 2020/09/29: about to merge with Max's contributions in #642 PR ✊♻️
· update 2020/10/14: ok I've merged and mostly reconciled [Max's](https://github.com/eMPee584/matrix-docker-ansible-deploy/tree/feature/add-worker-support) & my branch (this PR) & with MDAD* mainline, now working to get the frigging PIDs from the worker processes and fixing the removal of leftovers
· update 2020/10/28: nearly finished, only missing support for setups not using inbuilt nginx.. **REVIEWS, PLEASE**
· update 2021/01/08: yes, still nearly finished :grin: 

* (hey @spantaleev this project needs an official abbrevation/nick name lol)


eMPee584:
Turns out, even with a `when` condition on an `import_tasks` action evaluating to false, loops within these tasks will be executed. How **!clever** a design choice is that. 🥳

eMPee584:
@spantaleev this template file has a **lot** of duplicate statements.. I had started to factor some of that out into a `{%  macro render_proxy_directives() %}`, but oh the indentation... this would have been nice https://github.com/pallets/jinja/pull/919
We tried it on our test instance though and it seems to work fine. Still waiting for a colleague to provide us with his load testing script..

spantaleev:
One of the things that grabs my attention is the use of `matrix_synapse_*` variables in the `matrix-nginx-proxy` role (e.g. `matrix_synapse_use_presence`, `matrix_nginx_proxy_synapse_workers_list`).

We try to keep roles as separate as possible and to instead pass data along via `group_vars/matrix_servers`.

The easiest way to fix that would probably be by defining a new default like this in the `matrix-nginx-proxy` role:

```yaml
# List of Synapse workers.
# Example: [{"worker": "federation_reader", "port": 5000}]
matrix_nginx_proxy_synapse_workers_list: []
```

.. and then connecting it to `matrix_synapse_workers_enabled_list` from `group_vars/matrix_servers`:

```yaml
matrix_nginx_proxy_synapse_workers_list: "{{ matrix_synapse_workers_enabled_list }}"
```

We probably don't need to pass `matrix_synapse_use_presence` to `matrix-nginx-proxy`. It doesn't seem to hurt having a rule capturing `location ~ ^/_matrix/client/(api/v1|r0|unstable)/presence/[^/]+/status` and forwarding it to Synapse in any case (even if presence is disabled).

-------

I'm guessing when workers are enabled, we don't support `matrix_nginx_proxy_enabled: false`?
If the proxy is disabled, we attempt to generate a configuration that can be used with a local nginx server (one not running in a container). In such a case, we tell other services to expose their ports on `127.0.0.1` and we reverse-proxy to those ports, and not to `container-name:port`.

I guess making workers work with this would be complicated.

I see that you've modified `roles/matrix-synapse/templates/synapse/systemd/matrix-synapse.service.j2` to expose all worker ports (on `0.0.0.0` even). Is it necessary to do this? We may wish to expose them on `127.0.0.1` in some rare cases (like `matrix_nginx_proxy_enabled: false`), but I don't see why we'd do it on `0.0.0.0` all of the time.

--------

I'm guessing that when workers are enabled, [matrix-corporal](https://github.com/devture/matrix-corporal) likely breaks, as it's no longer capturing all traffic. Rather, some traffic gets captured in these new `location` blocks and goes to Synapse workers instead.

Using matrix-corporal + Synapse workers is some advanced use-case that will be difficult to solve. You should ignore it. I'm just mentioning it here for completeness.

lovelaced:
I don't have anything to add here, but just wanted to add this would be **massively** useful to my organization, and workers are a REALLY important feature. Would love to be able to use the officially supported ansible playbook for this rather than throwing something together.

eMPee584:
(ok thanks for the feedback.. this will need couple more days to get right.. )

ptman:
@eMPee584 any update? smooth sailing?

eMPee584:
Ah, sorry, I was busy with school the recent weeks and the convoluted nginx template also made me avoid finally tackling this. I'll try and get it done in the coming week..
Yes, it does work very well.

spantaleev:
Ah, good to here you're still on it! Looking forward to merging this!

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2
```diff
@@ -255,6 +376,19 @@ server {
 		ssl_ciphers "EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH";
 	{% endif %}
 
+	{% set federation_reader_worker = matrix_synapse_workers_enabled_list|selectattr('worker', 'equalto', 'federation_reader')|first %}
+	{% if federation_reader_worker %}
+	{# c.f. https://github.com/matrix-org/synapse/blame/master/docs/workers.md#L160 #}
+	location ~ ^(/_matrix/federation/v1/event/|/_matrix/federation/v1/state/|/_matrix/federation/v1/state_ids/|/_matrix/federation/v1/backfill/|/_matrix/federation/v1/get_missing_events/|/_matrix/federation/v1/publicRooms|/_matrix/federation/v1/query/|/_matrix/federation/v1/make_join/|/_matrix/federation/v1/make_leave/|/_matrix/federation/v1/send_join/|/_matrix/federation/v2/send_join/|/_matrix/federation/v1/send_leave/|/_matrix/federation/v2/send_leave/|/_matrix/federation/v1/invite/|/_matrix/federation/v2/invite/|/_matrix/federation/v1/query_auth/|/_matrix/federation/v1/event_auth/|/_matrix/federation/v1/exchange_third_party_invite/|/_matrix/federation/v1/user/devices/|/_matrix/federation/v1/send/|/_matrix/federation/v1/get_groups_publicised$|/_matrix/key/v2/query|/_matrix/federation/v1/groups/) {
```
lovelaced, line 382: /_matrix/federation/v1/groups/ should only be allowed GET requests; this configuration will cause federated community invites to fail.

eMPee584: Thanks for the feedback.. Is this theoretical or have you experienced it on an actual system?
I think there probably is a different mistake in the config to cause this.. we are also having some minor troubles with our setup.
I've just created an `awk` script to parse the endpoints directly from the upstream synapse workers documentation.. but it remains kinda nightmarish to work with.. :thinking:

lovelaced: Yes, I experienced it myself when trying to set up a fresh homeserver so I just removed the groups endpoint so the main synapse thread is handling it instead for now.

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2
```diff
@@ -101,6 +101,114 @@
...
+		proxy_pass http://synchrotron$request_uri;
+		proxy_set_header Host $host;
+		proxy_set_header X-Forwarded-For $remote_addr;
+	}
+	{% endif %}
+
+	{% set client_reader_worker = matrix_synapse_workers_enabled_list|selectattr('worker', 'equalto', 'client_reader')|first %}
+	{% if client_reader_worker %}
+	{# c.f. https://github.com/matrix-org/synapse/blame/master/docs/workers.md#L252 #}
+	location ^/_matrix/client/(versions$|(api/v1|r0|unstable)/(publicRooms$|rooms/.*/joined_me|rooms/.*/context/.|rooms/.*/members$|rooms/.*/messages$|rooms/.*/state$|login$|account/3pid$|keys/query$|keys/changes$|voip/turnServer$|joined_groups$|publicised_groups$|publicised_groups/|pushrules/.*$|groups/.*$|register$|auth/.*/fallback/web$)) {
```
lovelaced, line 131: Two of the endpoints here can only handle GET requests, and there are also two more not included here.

```
^/_matrix/client/(api/v1|r0|unstable)/pushrules/.*$
^/_matrix/client/(api/v1|r0|unstable)/groups/.*$
^/_matrix/client/(api/v1|r0|unstable)/user/[^/]*/account_data/
^/_matrix/client/(api/v1|r0|unstable)/user/[^/]*/rooms/[^/]*/account_data/
```

julianfoad: The "location ^/..." here, for a regex match needs to be "location ~ ^..." doesn't it?

roles/matrix-synapse/templates/synapse/systemd/matrix-synapse-worker@.service.j2
```diff
@@ -0,0 +1,29 @@
...
+
+# systemd ftw 🤦‍♂️
+# https://github.com/systemd/systemd/issues/14895#issuecomment-594123923
+ExecStart=/bin/sh -c "WORKER=%i; WORKER=$${WORKER%%:*}; \
+			exec /usr/bin/docker exec \
+			--user={{ matrix_user_uid }}:{{ matrix_user_gid }} \
+			matrix-synapse \
+			python -m synapse.app.$${WORKER} -c /data/homeserver.yaml -c /data/worker.%i.yaml"
+
+ExecStop=/usr/bin/docker exec matrix-synapse pkill -f %i
```
nuved, line 22: I can't stop a worker service with systemd because matrix-synapse container hasn't pkill command. 

`/usr/bin/docker exec matrix-synapse pkill -f 18082
OCI runtime exec failed: exec failed: container_linux.go:349: starting container process caused "exec: \"pkill\": executable file not found in $PATH": unknown
`

spantaleev: This is not related to this Pull Request. It's a general issue affecting the whole playbook, because of Synapse's recent (in v1.18.0) switch from an Alpine to a Debian base. See: https://github.com/matrix-org/synapse/pull/7839

spantaleev: Oh, I'm sorry.. I did not look into that much. We only use this `pkill` thing here. It is indeed an issue that only affects the workers setup.

maxklenk:
> Redis support was added in v1.13.0 with it becoming the recommended method in v1.18.0.
https://github.com/matrix-org/synapse/blob/master/docs/workers.md

The workers in the current synapse version now communicate via Redis, so we should setup Redis as well to add worker support.

roles/matrix-redis/defaults/main.yml
```diff
@@ -0,0 +1,10 @@
+matrix_redis_enabled: true
```
spantaleev, line 1: Enabling Redis is a good default for the `matrix-redis` role.

Can we adjust `group_vars/matrix_servers`, to add a new `matrix-redis` section, which would disable Redis by default. Or at least hook it up like this:

```yaml
matrix_redis_enabled: "{{ matrix_synapse_workers_enabled }}"
```

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -2426,16 +2461,17 @@ opentracing:
...
   # Uncomment the below to enable Redis support.
   #
-  #enabled: true
+  enabled: true
 
   # Optional host and port to use to connect to redis. Defaults to
   # localhost and 6379
   #
-  #host: localhost
+  host: matrix-redis
```
spantaleev, line 2469: It probably makes sense to introduce a `matrix_synapse_redis_host` variable in the `matrix-synapse` role, with a default value of `""`.

And maybe add a check in `roles/matrix-synapse/tasks/validate_config.yml` to ensure it's not empty if `matrix_synapse_workers_enabled`.

And then to have something like this in `group_vars/matrix_servers`:

```yaml
matrix_synapse_redis_host: "{{ 'matrix-redis' if matrix_redis_enabled else '' }}"
```

This way, people could disable the integrated Redis (`matrix_redis_enabled: false`) and point `matrix_synapse_redis_host` to some other (potentially external host) if they wish to do so.

Well, I guess we may need configuration options for the Redis port and potentially password as well.

This is not very crucial and it's something which can be changed later on as well, so it doesn't hurt to skip it if you're busy with other things.

eMPee584:
... sOo, back to business.. sorry had this nearly working, then deadlines for a written objection and our 2020.2 https://schulstick.org/ release came in the way 🌊😅💦
It's much appreciated @maxklenk had driven this forward, his way of setting the worker list in YAML is awesome..
I had created a perl script extracting and optimizing the regex from the workers doc using Regexp::Assemble and Regexp::Optimizer, but while the optimization obviously shrinks the regex, it also obscures the individual endpoints and that is stupid. So I changed my original awk script to reproduce the format Max committed and that means it should be super easy to keep in sync. Also, started writing a little github action to notify of upstream documentation changes..
Speaking of which, because of a tram strike in Dresden our English teacher missed class today and I had the opportunity to go through all changes of the involved repositories (max's fork, master, synapse dox & related) and compare our implementations.. It's good to hear he already has it in production use, that means it's nearly good to go.
I think the detailed redis configuration has too many knobs, but that's assuming these internal workers are the only relevant redis use case in the matix context ever 🤔
Anyways, will try and drive this further tomorrow. 🖖

eMPee584:
· update 2020/10/14: ok I've merged and mostly reconciled [Max's](https://github.com/eMPee584/matrix-docker-ansible-deploy/tree/feature/add-worker-support) & my branch (this PR) & with MDAD* mainline, now working to get the frigging PIDs from the worker processes and fixing the removal of leftovers


* (hey @spantaleev this project needs an official abbrevation/nick name lol)

eMPee584:
[…] also: the recent hiccups with rogue synapse processes is fixed in 1.21..

```
Bugfixes
--------
Synapse 1.21.1 (2020-10-13)
Synapse 1.21.0 (2020-10-12)
Synapse 1.21.0rc3 (2020-10-08)
==============================
- Fix duplication of events on high traffic servers, caused by PostgreSQL `could not serialize access due to concurrent update` errors. ([\#8456](https://github.com/matrix-org/synapse/issues/8456))

```

eMPee584:
> We probably don't need to pass `matrix_synapse_use_presence` to `matrix-nginx-proxy`. It doesn't seem to hurt having a rule capturing `location ~ ^/_matrix/client/(api/v1|r0|unstable)/presence/[^/]+/status` and forwarding it to Synapse in any case (even if presence is disabled).

Well IIUC, the point is to disable presence status processing but handle the client traffic for it.. so the location should not be redirected to the worker when presence is enabled.

> If `use_presence` is False in the homeserver config, it can also handle REST
> endpoints matching the following regular expressions:
> 
>     ^/_matrix/client/(api/v1|r0|unstable)/presence/[^/]+/status
> 
> This "stub" presence handler will pass through `GET` request but make the
> `PUT` effectively a no-op.


> I see that you've modified `roles/matrix-synapse/templates/synapse/systemd/matrix-synapse.service.j2` to expose all worker ports (on `0.0.0.0` even). Is it necessary to do this?

Yes you're right, that should most likely be changed...
on our server, we also don't use the inbuilt nginx container but one running on the host.. so will have to have a look at that use case regardless 😅💦

eMPee584:
Oh I forgot to say that I rebased this PR on Max's changes in #642 .. only thing missing is adding the scripted generation of the endpoint locations and linking nginx and synapse role variable spaces via `group_vars` .. hope to get that done today.

localguru:
Question: Why is e.g. for federation_sender "  - { worker: federation_sender, port: 18301 }" in matrix-synapse/defaults/main.yml with port 18301 and this port is not used anywhere in synapse config? Same with pusher and appservice. Don't understand that.

localguru:
What about postgres "tuning" when using workers?

eMPee584:
At last, with the latest commit, this PR could basically be consider as finished, only missing support for setups not using inbuilt nginx.. so **TEST, PLEASE**
The previous fluke of setting ports for workers which don't even serve any traffic as well as exposing worker ports publicly on the host both have been mended. And the script to fetch the list of endpoints directly from the synapse docs has been included, comments on whether to include the resulting `roles/matrix-synapse/vars/workers.yml` file into the repository appreciated @spantaleev ✌️
oh and sorry for my time estimates during the journey being somewhat off #slowcoding 😆

PC-Admin:
@eMPee584 No need to apologize, this is very complicated and you're helping a lot of people out. :)

TEST TIME! First i merged the PR like so:
```
$ git clone https://github.com/spantaleev/matrix-docker-ansible-deploy.git
$ cd ./matrix-docker-ansible-deploy/
$ git fetch origin pull/456/head:eMPtesting
$ git checkout eMPtesting
```

I then disabled the previous worker setup from Max by re-running the previous script i was using.

I edited my vars.yml like so:
```
matrix_synapse_workers_enabled: true
matrix_synapse_workers_enabled_list:
  - { type: generic_worker, port: 18111 }
  - { type: generic_worker, port: 18112 }
  - { type: federation_sender, port: 0 }
  - { type: media_repository, port: 18221 }
```

I run your PR and it spits out this error:
```
TASK [matrix-synapse : Download synapse workers doc] ***************************
fatal: [matrix.perthchat.org]: FAILED! => {"changed": false, "msg": "Destination /home/chatoasis/Documents/workers-eMPee584/matrix-docker-ansible-deploy/roles/matrix-synapse/files does not exist"}
```

Changed "matrix_synapse_workers_enabled: false" then ran it again:
```
TASK [matrix-synapse : Ensure worker configs are cleaned] **********************
 [WARNING]: Unable to find '/matrix/synapse/config' in expected paths (use
-vvvvv to see paths)


TASK [matrix-synapse : Remove workers from synapse.wants list] *****************
fatal: [matrix.perthchat.org]: FAILED! => {"msg": "'matrix_synapse_workers_avail_list' is undefined"}
```

I hope this helps a bit, i ended up jumping back to the upstream after. Didn't break anything. :sweat_smile:

localguru:
Great work. Thank you!

Isn't it a bit risky to trust that the formatting of [worker.md](https://raw.githubusercontent.com/matrix-org/synapse/master/docs/workers.md) will not change and that the parsing will never fail?

eMPee584:
> ```
> TASK [matrix-synapse : Download synapse workers doc] ***************************
> fatal: [matrix.perthchat.org]: FAILED! => {"changed": false, "msg": "Destination /home/chatoasis/Documents/workers-eMPee584/matrix-docker-ansible-deploy/roles/matrix-synapse/files does not exist"}
> ```
Wow that certainly is a strange error. I tried to reproduce it in a fresh repository checkout but do not get any errors (with just `--tags=setup-synapse,setup-nginx-proxy` though, does that work for you?)
you' sure there isn't something fishy with the filesystem location? What happens if you check out in `/tmp` ?

> ```
> TASK [matrix-synapse : Ensure worker configs are cleaned] **********************
>  [WARNING]: Unable to find '/matrix/synapse/config' in expected paths (use
> -vvvvv to see paths)
`-v^5` certainly sounds interesting :rofl: 

Several of theses cases of undefined variables when disabling/"uninstalling" workers have just been fixed in a commit though.. try to find me some more :cowboy_hat_face: 

> Isn't it a bit risky to trust that the formatting of worker.md will not change and that the parsing will never fail?
@localguru yes totally. An issue could be filed with the synapse repo requesting a machine-readable format.. for, now, this works. And synapse will probably be replaced by dendrite within a year.. so :shrug:

PC-Admin:
> > ```
> > TASK [matrix-synapse : Download synapse workers doc] ***************************
> > fatal: [matrix.perthchat.org]: FAILED! => {"changed": false, "msg": "Destination /home/chatoasis/Documents/workers-eMPee584/matrix-docker-ansible-deploy/roles/matrix-synapse/files does not exist"}
> > ```
> 
> Wow that certainly is a strange error. I tried to reproduce it in a fresh repository checkout but do not get any errors (with just `--tags=setup-synapse,setup-nginx-proxy` though, does that work for you?)
> you' sure there isn't something fishy with the filesystem location? What happens if you check out in `/tmp` ?
> 

Tried the same code in /tmp, with setup-all,start and setup-synapse,setup-nginx-proxy, same error for both:
```
TASK [matrix-synapse : Download synapse workers doc] ***************************
fatal: [matrix.perthchat.org]: FAILED! => {"changed": false, "msg": "Destination /tmp/matrix-docker-ansible-deploy/roles/matrix-synapse/files does not exist"}
```

It works in any location with the upstream code. Tried your latest commit too, exact same error. :S

eMPee584:
It's not conflicting, just haven't merged latest upstream commits bumping synapse version..
and `git ls-files roles/matrix-synapse/files` sure enough does give
```
roles/matrix-synapse/files/usr-local-bin/matrix-synapse-worker-write-pid
roles/matrix-synapse/files/workers-doc-to-yaml.awk
```
yeah?
Please rerun `--tags=setup-synapse` with `-vvvvv` and post the full log file.. and please upload your checked out MDAD repo as an archive (without your inventory files of course) or send via mail..

PC-Admin:
> and `git ls-files roles/matrix-synapse/files` sure enough does give
> 
> ```
> roles/matrix-synapse/files/usr-local-bin/matrix-synapse-worker-write-pid
> roles/matrix-synapse/files/workers-doc-to-yaml.awk
> ```
> 
> yeah?
> Please rerun `--tags=setup-synapse` with `-vvvvv` and post the full log file.. and please upload your checked out MDAD repo as an archive (without your inventory files of course) or send via mail..

Those files definitely exist, it's a strange error indeed. :S Sending you the logs and my copy of the MDAD repo now, good luck friend.

localguru:
With SAML and multiple generic.workes you run into this problem:

https://github.com/matrix-org/synapse/issues/7530

Solution is not to catchg these paths by generic.worker regex:
```
^/_matrix/client/(r0|unstable)/auth/.*/fallback/web$ 

^/_matrix/client/(api/v1|r0|unstable)/login/sso/redirect$
^/_matrix/saml2/authn_response$
```

eMPee584:
k, also got reported that the media endpoints can be requested on the federation port, will add that.. what a monster of maintenance burden; let's hope dendrite obsoletes all of this within a year :rofl:

localguru:
> k, also got reported that the media endpoints can be requested on the federation port, will add that.. what a monster of maintenance burden; let's hope dendrite obsoletes all of this within a year rofl

Yes, if homeserver A sends an image to homeserver B, homerserver B is asking for the image on the federation port 8448 to get the media on URL `/_matrix/media/`. If `worker_app: synapse.app.media_repository` is enabled the regex `/_matrix/media/`for the media worker must point to the internal media worker port in nginx vhost for federation port 8448 too. Otherwise homeserver B gets a 404 (not found) on the federation port 8448. The mean thing is that on homeserver B, the user/client gets the message "Error decrypting image" and not something like "not found".

localguru:
@eMPee584 Hi, I think there is another issue in the worker setup with uploading bigger files. In a setup without workers `client_max_body_size` in `roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2` matches on the path `location ~* ^(\/_matrix|\/_synapse\/client)` for client an `/` for federation. Both include `/_matrix/media`:

client vhost:
```
server {
	listen {{ 8443 if matrix_nginx_proxy_enabled else 443 }} ssl http2;
	[...]
	location ~* ^({{ matrix_nginx_proxy_proxy_matrix_client_api_forwarded_location_prefix_regexes|join('|') }}) {

		client_body_buffer_size 25M;
		client_max_body_size {{ matrix_nginx_proxy_proxy_matrix_client_api_client_max_body_size_mb }}M;
		proxy_max_temp_file_size 0;
```
federation vhost:
```
server {
	{% if matrix_nginx_proxy_https_enabled %} listen 8448 ssl http2;
	[...]
	location / {

		client_body_buffer_size 25M;
		client_max_body_size {{ matrix_nginx_proxy_proxy_matrix_federation_api_client_max_body_size_mb }}M;
		proxy_max_temp_file_size 0;
```

The problem is now, that `client_body_buffer_size`, `client_max_body_size` and `proxy_max_temp_file_size` don't access/catch the **previous** locations for the workers!

I can't estimate what effect have missing `client_body_buffer_size`, `client_max_body_size` e.g. on the `generic.worker`, but with the `media_repository` worker, it causes that `client_max_body_size` for the path `/_matrix/media/` is not set and an upload of up to 50 MB is not possible.

In the case of `media_repository` worker at least the `/_matrix/media/` location should look like this:

```
			location ~ ^/_matrix/media/ {
				proxy_pass http://media_repository_upstream$request_uri;
				proxy_set_header Host $host;

				# this must match *max_upload_size* in homeserver.yaml
				client_max_body_size matrix_nginx_proxy_proxy_matrix_client_api_client_max_body_size_mb }}M;

			}
```

To make it even more complicated: the federation vhost has a has a `client_max_body_size` value three times greater. The question now is whether this larger value should be set here too, which would mean that another `/_matrix/media` location would have to be generated for the federation vhost:

```
location ~ ^/_matrix/media/ {
	proxy_pass http://media_repository_upstream$request_uri;
        proxy_set_header Host $host;

	# this must match *max_upload_size* in homeserver.yaml
	client_max_body_size matrix_nginx_proxy_proxy_matrix_federation_api_client_max_body_size_mb }}M;
}
```

eMPee584:
@PC-Admin from the log you sent me, there was a space in the name of a parent folder. I added the necessary quoting to the shell command fixing that.
Mh, most of these issues seem to be addressed. .. @spantaleev , any other change requests? Should the resulting `roles/matrix-synapse/vars/workers.yml` also be put into the git?
Support for outside nginx instances and the nginx template macro consolidation can go in after.. ?

ptman:
@eMPee584 there seems to be a conflict as well

eMPee584:
Key thanks; the minor conflict has been dealt with.

maxklenk (changes requested):
Please fix the mentioned bugs and add the generated files to `.gitignore`:
```
/roles/matrix-synapse/files/workers.upstream-documentation.md
/roles/matrix-synapse/vars/workers.yml
```

roles/matrix-synapse/files/workers-doc-to-yaml.awk
```diff
@@ -0,0 +1,146 @@
...
+# URL=https://github.com/matrix-org/synapse/raw/master/docs/workers.md
+# curl -L ${URL} | awk -f workers-doc-to-yaml.awk -
+
+function worker_stanza_append(string) {
+    worker_stanza = worker_stanza string
+}
+
+function line_is_endpoint_url(line) {
+    # probably API endpoint if it starts with white-space and ^ or /
+    return (line ~ /^ +[\^/].*\//)
```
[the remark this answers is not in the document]
eMPee584: Hm, your reasoning is valid, but the result with these proposed adjustments produces the same output. Escaping of the slash in the square brackets is optional (as the parser knows it is not meant as regex end but part of the character list.. awk's man page:
`[abc...]   A  character  list:  matches any of the characters abc....  You may include a range of characters by separating them with a dash.  To include a literal dash in the list, put it first or last.`)
The SSO endpoints are already being commented out with a line `# FIXME: ADDITIONAL CONDITIONS REQUIRED: to be enabled manually` and followed by the note about issue 7530 which you added in https://github.com/matrix-org/synapse/pull/8701 ..

roles/matrix-synapse/tasks/workers/setup_install.yml
```diff
@@ -0,0 +1,88 @@
...
+
+- name: Download synapse workers doc
+  get_url:
+    url: https://github.com/matrix-org/synapse/raw/master/docs/workers.md
+    dest: "{{ role_path }}/files/workers.upstream-documentation.md"
+
+- name: Download synapse workers doc and convert into YAML
+  shell:
+    cmd: "awk -f '{{ role_path }}/files/workers-doc-to-yaml.awk' -- '{{ role_path }}/files/workers.upstream-documentation.md' > '{{ role_path }}/vars/workers.yml'"
+    creates: "{{ role_path }}/vars/workers.yml"
```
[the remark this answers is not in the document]
eMPee584: Oh wow, thanks. Ansible has its pecularities..

roles/matrix-synapse/tasks/workers/setup_install.yml
```diff
@@ -0,0 +1,88 @@
...
+    enabled: true
+  with_items: "{{ matrix_synapse_workers_enabled_list }}"
+
+- name: Ensure previous worker configs are cleaned
+  file:
+    path: "{{ item }}"
+    state: absent
+  with_fileglob:
+    - "{{ matrix_synapse_config_dir_path }}/worker.*.yaml"
+
```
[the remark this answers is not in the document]
eMPee584: Change adopted; thanks!

roles/matrix-synapse/files/workers-doc-to-yaml.awk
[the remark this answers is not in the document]
maxklenk: The script fails on my mac if the backslash is not escaped:
```
awk: nonterminated character class ^ +[\^
 source line number 14 source file workers-doc-to-yaml.awk
 context is
            return (line ~ /^ >>>  +[\^/ <<< ].*\//)
```
I am using `awk version 20200816`. I think it is save to escape it, even if it is not necessary on your machine.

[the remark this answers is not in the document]
maxklenk: I tested the SSO routes again. The section below `# OpenID Connect requests.` is not automatically disabled when I run your script. Only with my filter these routes are removed. Are they disabled when you parse it?

roles/matrix-synapse/templates/synapse/worker.yaml.j2
```diff
@@ -0,0 +1,29 @@
+#jinja2: lstrip_blocks: "True"
+worker_app: synapse.app.{{ item.worker }}
+
+worker_replication_host: 127.0.0.1
+worker_replication_port: {{ matrix_synapse_replication_tcp_port }}
+worker_replication_http_port: {{ matrix_synapse_replication_http_port }}
+
+{% if item.worker not in [ 'appservice', 'federation_sender', 'pusher' ] %}
+worker_listeners:
```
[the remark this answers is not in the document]
eMPee584: Ok added exposing of metrics ports, you're invited to test..

roles/matrix-synapse/files/workers-doc-to-yaml.awk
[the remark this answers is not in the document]
eMPee584: Yes they are.. probably another regex incompatibility between our `awk` versions in line 123: `line ~ /\<[Ii]f\>|\<[Ff]or\>/` ?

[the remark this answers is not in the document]
eMPee584: Ok awk compatibility seems to be fixed now 🤦‍♂️

roles/matrix-synapse/files/usr-local-bin/matrix-synapse-worker-write-pid
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+# Find a synapse worker's PID and write it to a file so systemd can manage it as a service
```
jdreichmann, line 2: Was [docs/systemd-with-workers](https://github.com/matrix-org/synapse/tree/develop/docs/systemd-with-workers) considered? It might be possible to use systemd's targets to archieve the functionality without needing to "manually" run scripts to record PIDs

roles/matrix-synapse/tasks/workers/setup_install.yml
```diff
@@ -0,0 +1,100 @@
...
+- name: Download synapse workers doc and convert into YAML
+  local_action:
+    module: shell
+    cmd: "awk -f '{{ role_path }}/files/workers-doc-to-yaml.awk' -- '{{ role_path }}/files/workers.upstream-documentation.md' > '{{ role_path }}/vars/workers.yml'"
+    creates: "{{ role_path }}/vars/workers.yml"
+  vars:
+    ansible_become: no
+
+- name: Load list of available worker apps and endpoints
+  include_vars: "{{ role_path }}/vars/workers.yml"
```
jdreichmann, line 20: I'd suggest `delegate`ing the previous three tasks to localhost (and `set_fact`ing the vars after the extraction+conversion), or even going so far as to hard-code the endpoints in the playbook and maintain them.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -235,6 +235,39 @@ listeners:
...
+{% endif %}
+{% if matrix_synapse_workers_enabled_list|selectattr('type', 'equalto', 'pusher')|list  %}
+start_pushers: false
+{% endif %}
+{% if matrix_synapse_workers_enabled_list|selectattr('type', 'equalto', 'user_dir')|list  %}
+update_user_directory: false
+{% endif %}
+
+# rather let systemd handle the forking
+daemonize: false
```
jdreichmann, line 269: SystemD only calls docker anyway, so i dont think there is much benefit to setting this, as it will only affect the docker container, and if the service file is set to `Type=oneshot`, that might cause problems

roles/matrix-synapse/templates/synapse/systemd/matrix-synapse.service.j2
```diff
@@ -43,6 +43,18 @@ ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-synapse \
 			{% if matrix_synapse_manhole_enabled and matrix_synapse_container_manhole_api_host_bind_port %}
 			-p {{ matrix_synapse_container_manhole_api_host_bind_port }}:9000 \
 			{% endif %}
+			{% for worker in matrix_synapse_workers_enabled_list %}
+			{% if matrix_synapse_workers_enabled and not matrix_nginx_proxy_enabled|default(False) %}
+			{# Expose worker ports (by default 18xxx range) on host if not using internal nginx proxy #}
```
jdreichmann, line 48: I am not sure what exactly this does, why would worker ports be mapped in the main synapse container (back to the host)? Shouldn't the mapping be (workers|synapse)->host (and therefore the worker mapping be seperate from synapse port mappings)?

roles/matrix-synapse/files/workers-doc-to-yaml.awk
```diff
@@ -0,0 +1,146 @@
...
+            print "# ]"
+        }
+    }
+}
+
+END {
+    print "\nmatrix_synapse_workers_avail_list:"
+    print workers | "sort"
+}
+
```
jdreichmann, line 145: I'm unsure if parsing the docs to yaml is really a great way, i would agree that mainting the paths (and their worker mapping) is not the greatest job in the world, but this seems.. rather hacky.

roles/matrix-synapse/files/usr-local-bin/matrix-synapse-worker-write-pid
eMPee584: Yes it was, but that is about setting up synapse directly on the host. This repository puts all components in docker containers, which are not actually designed for multi-entry programs.. So launching the worker processes after the main homeserver via docker-exec requires this bit of script trickery to extract host-perspective PIDs from the synapse container, as the resulting docker-processes itself are host-side control processes.

Defining a matrix target principally is a nice idea, but functionally will not award us any improvement I'd presume. If you find any, please file a top-level issue.

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2
```diff
@@ -159,6 +226,50 @@
...
+{% set user_dir_workers = matrix_nginx_proxy_synapse_workers_list|selectattr('type', 'equalto', 'user_dir')|list %}
+{% set frontend_proxy_workers = matrix_nginx_proxy_synapse_workers_list|selectattr('type', 'equalto', 'frontend_proxy')|list %}
+{% if matrix_nginx_proxy_synapse_workers_enabled %}
+	# Round Robin "upstream" pools for workers
+
+	{% if generic_workers %}
+	upstream generic_worker_upstream {
+		# ensures that requests from the same client will always be passed
+		# to the same server (except when this server is unavailable)
+		ip_hash;
```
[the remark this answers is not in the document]
eMPee584: No. Which `key` for the http://nginx.org/en/docs/http/ngx_http_upstream_module.html#hash directive would you propose for that?
 The average user might have two devices online with different IPs. Balancing by IP would get the traffic of both be handled by different workers, balancing by authorization would make them go to the same.
It might marginally improve the load because of less cache faults.. Than again, no idea how well the new `redis` IPC will catches that.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
eMPee584: Please test this specifically instead of posting vague assumptions.
Does it cause problems?

roles/matrix-synapse/templates/synapse/systemd/matrix-synapse.service.j2
eMPee584: Because the worker processes run in the synapse container alongside the main homeserver process. When not using the builtin nginx container instance, they have to be accessible by a process outside the matrix docker network, thus exposing the ports on the host.

If you think it would improve things if worker ran in its own container, please test that.

roles/matrix-synapse/files/workers-doc-to-yaml.awk
eMPee584: Obviously it is not the greatest way, and it is rather hacky. But the People demand worker support, this does work now and can be fixed if anything changes enough to break it.

Be encouraged to open an issue with synapse for a more machine-readable endpoint mapping.

roles/matrix-synapse/tasks/workers/setup_install.yml
eMPee584: Ok admittedly, I'm an ansible n0ob so it would be nice if you could explain to me what difference it would make (for the first option).
For the latter, that would certainly be feasible. What advantage would it have though? :thinking:

roles/matrix-synapse/files/workers-doc-to-yaml.awk
jdreichmann: With this setup, the playbook will fail to extract the endpoint mapping if either github is down, the format of the endpoints changed in the docs or the synapse docs moved. There is no good way to be notified of this, except noticing it no longer working. Maintaing a manual endpoint mapping list is imo cleaner, as it only needs to be reviewed (and adjusted, if necessary) when updating synapse.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
jdreichmann: The statement that systemd would handle the forking is false, as the systemd service calls `docker run`, which then runs the synapse process. Adding `daemonize: false` could imply that systemd is actually in control of the forking of the synapse process, which it cannot be. It will affect the docker containers in their `docker attach $container` behaviour.

roles/matrix-synapse/tasks/workers/setup_install.yml
jdreichmann: The advantage is that the tasks get run on the host used to deploy the server, i.e. the computer you are sitting infront of. Delegating a task like "download something from github" to localhost allows the download to succeed, even if the actual server might have a firewall which does not allow arbitrary egress traffic (i.e. blocking github (or other sites, in a corporate setting))

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2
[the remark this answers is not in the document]
jdreichmann: `hash $http_authorization` should balance connections by the value of the Authorization header. For a single-user multi-device setups, there is little benefit, but if you have e.g. 10 users with each 2-3 sessions, each user w/ all their sessions would be bound to one backend, enabling fast `/sync`s and not stalling other users too much if someone does an `/initialsync` - this applies only to the `/[initial]sync` endpoints (and 2-3 others), the rest of the CS-API-endpoints can be balanced by source IP aswell.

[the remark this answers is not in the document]
spantaleev: How would you make all of the user's sessions land on the same worker? Doesn't each session gets its own access token and thus has its own unique `Authorization` header.

The best that should be possible with this is to have each session land the same worker, but not have all these sessions (for the same user) do that.

Unless there's something smart which could parse the macaroon out of the `Authorization` header, read the user id out of it and then balance based on that value.

I see that `workers.md` says the same:

> For /sync and /initialSync requests it will be more efficient if all requests from a particular user are routed to a single instance. Extracting a user ID from the access token or Authorization header is currently left as an exercise for the reader.

Well, how does one solve that exercise (in nginx), I don't know.

spantaleev:
I've been thinking of how we can solve enabling Synapse workers and using matrix-corporal at the same time.

Right now, it's not possible. This is okay. We don't need to redo things to merge this PR. We should probably just indicate it somewhere and make the playbook error out if it detects that both workers and corporal are enabled at the same time.

Below is my attempt to explain the problem and a potential solution:

`matrix.DOMAIN` traffic usually goes to the nginx vhost configured in `matrix-synapse.conf`.

`proxy_pass`ing is usually pointed to Synapse (`http://matrix-synapse:8008`) or to matrix-corporal. This "workers PR" changes `matrix-synapse.conf` a lot, making it do URL-matching and forwarding requests to various upstreams, not just to `http://matrix-synapse:8008`.

If corporal is enabled, `matrix.DOMAIN` is usually pointed to matrix-corporal. matrix-corporal is checking incoming requests and proxying them to Synapse (`http://matrix-synapse:8008`). Very simple.

With workers, matrix-corporal proxying to a single place (`http://matrix-synapse:8008`) means we always hit the same worker. Our simple solution no longer works. Also, there's so much configuration in `matrix-synapse.conf` right now, that requests won't even hit corporal, in favor of some other upstream ;)

---------------

Maybe a good solution would be to have 1 more level of `proxy_pass` indirection in nginx. I'm thinking of 2 vhosts:

- `matrix.DOMAIN` - as an entrypoint
- `matrix-synapse-proxy` - internal vhost that looks very much like `matrix-synapse.conf.j2` right now - this is where all the complicated URL-matching and dispatching to upstreams (workers) happens

In the regular case (no corporal), `matrix.DOMAIN` requests will just `proxy_pass` to `matrix-synapse-proxy`, which `proxy_pass`es to Synapse (being aware of the various Synapse workers, etc.).

If matrix-corporal is enabled, `matrix.DOMAIN` requests will `proxy_pass` to matrix-corporal. Then, whenever matrix-corporal needs to talk to Synapse, it will talk to the `matrix-synapse-proxy` internal vhost, instead of talking to `http://matrix-synapse:8008` directly. This way, matrix-corporal doesn't need to be aware of workers and which one handles what and it will reuse all the things in the `matrix-synapse-proxy` configuration.

------------------

Problem with the above solution may be that it's an additional nginx vhost.. And that probably makes it somewhat ugly (and slower?).

But it's probably a great idea anyway, because Dendrite/Conduit are coming, aren't they..? This `matrix-synapse-proxy` (our `matrix-synapse.conf.j2`) complicated configuration (URL matching, lots of upstreams, ..) had better become its own vhost. Then the actual `matrix.DOMAIN` entrypoint will be a simple one which merely points us to the appropriate upstream (`matrix-synapse-proxy`, `matrix-dendrite-proxy`, `matrix-conduit-proxy`, `matrix-corporal`, etc.)

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2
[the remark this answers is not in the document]
bjeanes: https://github.com/bungle/awesome-resty#authentication-and-authorization has a macroon lib, so in principle it might be possible using OpenResty (nginx distribution with lots of lua modules)

jolly-jump:
> I see that you've modified `roles/matrix-synapse/templates/synapse/systemd/matrix-synapse.service.j2` to expose all worker ports (on `0.0.0.0` even). Is it necessary to do this? We may wish to expose them on `127.0.0.1` in some rare cases (like `matrix_nginx_proxy_enabled: false`), but I don't see why we'd do it on `0.0.0.0` all of the time.

I have the same question as @spantaleev : Do the 0.0.0.0:19xxx ports need to be exposed to the public?
If yes, we should mention that the prerequisits have to be modified and the firewall has to be opened on these ports, if there is a firewall.

Otherwise: I will live test tomorrow at 8:00 sharp, if the workers (using the default configuration) really do work and distribute some traffic and computer power...

spantaleev:
Can you merge `master` into this (or rebase it)?

Looks like merging `master` into this branch would go smoothly. I have no working worker setup yet though, so I can't be sure I can test it correctly.

I'd like to review this and finally make progress on merging this though.

-----------------

I also agree with the comments above that pulling a Synapse documentation file and running awk against is.. powerful and smart, but likely fragile.

If we are to do this, we should at least pull the file for the branch/tag we're trying to run as a Docker image (and not necessarily Synapse `master`). It's probably better if we run this script manually while maintaining the playbook though, and not do it every time from every installation.

It'd be nice if we can get notifications of changes made to it.

eMPee584:
Ok, so I synced this branch with master, committed the current worker endpoints YAML instead of regenerating it on each deployment (still needs f.e. a github action to monitor changes of the synapse doc file).
And up to now I hadn't recognized `docker-run -p` could be used to expose ports on localhost only.. so finally got around to set this limit (#RTFMFTW).
maybe you can test if that works as well as before @jolly-jump ?

eMPee584:
Also, @anoadragon453 has recently created https://github.com/matrix-org/synapse/pull/9162 which is a very similar workers setup especially for testing with [matrix-org/complement](https://github.com/matrix-org/complement), the "Matrix compliance test suite".. if anyone feels the urge to dig into that and identify further improvement potential to this PR, feel very much encouraged :wink:

jolly-jump:
it works so far with the test server.
In production I'll have to wait till tomorrow or even later... Dont have lots of time for a maintenance (postgres update necessary)...

roles/matrix-synapse/defaults/main.yml
```diff
@@ -286,6 +298,35 @@ matrix_synapse_metrics_port: 9100
...
+# - no endpoints / doesn't need port mapping if port ends on 0
+# - single-instance-only if 2nd last digit of port number is 0
+matrix_synapse_workers_enabled_list:
+  - { type: generic_worker, port: 18111, metrics_port: 19111 }
+  - { type: generic_worker, port: 18112, metrics_port: 19112 }
+  - { type: generic_worker, port: 18113, metrics_port: 19113 }
+  - { type: generic_worker, port: 18114, metrics_port: 19114 }
+  - { type: generic_worker, port: 18115, metrics_port: 19115 }
+  - { type: generic_worker, port: 18116, metrics_port: 19116 }
+  - { type: pusher, port: 00, metrics_port: 19200 }
```
spantaleev, line 314: From what I could find in the code, we don't make a distinction between `00` and `0` (but I may have missed it).

If that's really the case, we should probably just annotate these `00` entries with some comment and turn them into a `0`.

If we do need to make some distinction, it's probably better to add more fields (`multiple_instances_supported: false`) and use those.

What do you think?

jdreichmann:
> Also, @anoadragon453 has recently created [matrix-org/synapse#9162](https://github.com/matrix-org/synapse/pull/9162) which is a very similar workers setup especially for testing with [matrix-org/complement](https://github.com/matrix-org/complement), the "Matrix compliance test suite".. if anyone feels the urge to dig into that and identify further improvement potential to this PR, feel very much encouraged wink

I'd like to add that running multiple workers in the same container is _not_ advised:

> Note that this is not currently intended to be used in production. *If you'd like to use Synapse workers with Docker, instead make use of the official image, with one worker per container.* The purpose of this dockerfile is currently to allow testing Synapse in worker mode with the Complement test suite.

This PR does *not* do that, which is why i'd currently advise against it in this form.

roles/matrix-synapse/vars/workers.yml
```diff
@@ -0,0 +1,308 @@
+---
+
+matrix_synapse_workers_generic_worker_endpoints:
```
jdreichmann, line 3: I'd argue that splitting these endpoints directly into `matrix_synapse_workers_generic_worker_endpoints.[sync|federation|...]`, and combining groups where needed would be more maintenance friendly.
It makes it easy to route only federation/only sync/... requests

roles/matrix-synapse/files/workers-doc-to-yaml.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+# Fetch the synapse worker documentation and extract endpoint URLs
+# matrix-org/synapse master branch points to current stable release
+
+URL=https://github.com/matrix-org/synapse/raw/master/docs/workers.md
+curl -L ${URL} | awk -f workers-doc-to-yaml.awk > ../vars/workers.yml
```
jdreichmann, line 6: Is this still needed after `vars/workers.yml` was added?

spantaleev:
I do wonder though.. we run workers by spawning more processes into the same container using `docker exec`.

One could consider this as "running all workers in the same container", albeit it in a different way - one that spawns them later on from outside.

--------------------

An alternative would be to replace that with `docker run` -- spawning a new container on the `matrix` network for each of these workers. I'm not sure what we win/lose by doing that, whether it's worth it and how difficult it is to reorganize the current playbook code to adopt it.

I suspect that it may make things simpler:
- no longer having the need to keep track pids, so simpler systemd service files
- no need for the `matrix-synapse-worker-write-pid` script
- no need for `jq` (currently used by the `matrix-synapse-worker-write-pid` script).
 
By the way, `jq` installation also currently fails on CentOS, unless one has manually installed `epel-release` first (`yum install -y epel-release`). We should either fix that somehow anyway, or hope for the dependency to go away.

The `docker run` idea has probably been considered though, and disregarded in favor of `docker exec` for some reason?

roles/matrix-base/templates/usr-local-bin/matrix-remove-all.j2
```diff
@@ -15,11 +15,26 @@ if [ "$sure" != "Yes, I really want to remove everything!" ]; then
...
+	# and stop these instantiated services as well.
+	for s in $(find {{ matrix_systemd_path }}/ -xtype f -name "matrix-*" -printf "%f\n" | grep -v '@.service' | uniq); do
 		systemctl stop $s
+	done
+
+	# Get rid of regular service files, as well as symlinks like
+	# '/etc/systemd/system/matrix-synapse.service.wants/matrix-synapse-worker@generic_worker:18111.service'
+	# and even
+	# '/etc/systemd/system/multi-user.target.wants/matrix-synapse.service'.
+	for s in $(find {{ matrix_systemd_path }}/ -xtype f -name "matrix-*" -printf "%p\n"); do
```
jdreichmann, line 32: Maybe do `systemctl disable $service` first before cleaning up stuff with `rm`, ideally that should remove most symlinks already and is cleaner. If disabling services is not cleaning up everything, i'd get out the `rm`...

jdreichmann (changes requested):

jdreichmann:
> One could consider this as "running all workers in the same container", albeit it in a different way - one that spawns them later on from outside.

It doesn't matter if they get spawned at the sime time or later - running them all in the same container seems very not-intended, and i'd rather use the official image with one worker per container, as is recommended by anyone actually running synapse at scale

roles/matrix-base/templates/usr-local-bin/matrix-remove-all.j2
spantaleev: I've switched to `systemctl disable --now $s` in c05d3d09bd22, but it still seems like we need to `rm` symlinks, as explained in that commit.

roles/matrix-synapse/vars/workers.yml
spantaleev: You're suggesting that we pre-split these early on, as part of the awk script?

I guess that would be better compared to the runtime splitting I did in 5ca68210cdd. But then again, one needs to rework the awk script somewhat, which.. I haven't dipped my toes into yet.

roles/matrix-synapse/files/workers-doc-to-yaml.sh
spantaleev: I guess this shell script is just some self-executing documentation telling developers how to trigger the regeneration of `vars/workers.yml`. This is also documented in `workers-doc-to-yaml.awk`.

spantaleev:
I think we should attempt to switch to running a container per-worker using `docker run` and stop doing `docker exec`.

I'm working on some other large changes now (to get matrix-corporal running and to make us more Dendrite-friendly), as described in [this old comment of mine](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/456#issuecomment-749401133).

I could give that `docker run` thing a shot after that, unless someone else could.

roles/matrix-base/templates/usr-local-bin/matrix-remove-all.j2
jdreichmann: `disable`ing atleast tells systemd to expect that this unit is gone, and should remove it from `multi-user.target.wants`. If specified in-spec, it should (afaik) also be possible to be auto-removed from `matrix-synapse.service.wants`.

roles/matrix-synapse/files/workers-doc-to-yaml.sh
jdreichmann: Yes, but then it should maybe be moved to `contrib` or similar, if it is not used in the playbook, as leaving it at this place implies it is a used component which needs maintenance.

roles/matrix-base/templates/usr-local-bin/matrix-remove-all.j2
spantaleev: True.

We are specifying them in `Wants=..` for `matrix-synapse.service`. Not sure why systemd doesn't auto-remove them from the directory.

roles/matrix-synapse/vars/workers.yml
jdreichmann: I would suggest splitting into arrays of endpoints (`sync: [], federation: [], media: [], ...`), if it was generated with the awk script, it should also be added there, but having them split in the source allows other contributors to cleanly use f.ex. the set of all federation endpoints.

jdreichmann:
Running all workers in their own container also makes a clear 1:1 mapping of systemd-service <> container, and gives the ability to define a healthcheck _per worker_ on the docker layer (see `HEALTHCHECK`), giving easy troubleshooting when doing `docker ps` and f.ex. a single worker is unhealthy, which is afaik not easily doable when running all workers in the same container

spantaleev:
70796703d36efc just landed. Workers now run in dedicated containers. We also don't have a dependency on `jq` anymore. Yay for simplicity!

I've changed a lot of core things, so people running this should be careful. I may have introduced some bad regressions.

spantaleev:
Regarding d3ecc6f017c80,

I found some bridges weren't working (they were failing when trying to access the media repository), because bridges were pointed to `http://matrix-synapse:8008`. This directly goes to the main Synapse process, which doesn't support such things (we do `enable_media_repo: false` in `homeserver.yaml`).

The solution was to point them to `http://matrix-nginx-proxy:12080` instead, which then forwards to Synapse or to a worker.

I've define a new `matrix-base` variable to control this `matrix_homeserver_container_url` and `group_vars/matrix_servers` tries to point it correctly:

- if `matrix-nginx-proxy` is enabled, it goes there (`http://matrix-nginx-proxy:12080`), which takes care of things
- if not, make it go straight to Synapse (`http://matrix-synapse:8008`) regardless of whether workers are enabled. However, this will fail for some requests if workers are enabled, because the main process does not handle everything.

I'm not sure what a good solution would be. We can stop disabling certain endpoints (`enable_media_repo: false`) for the main process. Is it really that bad if some request squeezes through?

I guess this has gotten quite complicated (in this and other areas), because we support so much flexibility (with proxy, without proxy; with Synapse terminating SSL by itself; or not; with workers; without workers; with matrix-corporal in front; without it). I find it hard to imagine all possible scenarios and to ensure they won't be broken. And this will only get more complicated once we get Dendrite support and will need to be even more dynamic in terms of where traffic goes, etc.

I guess something will have to break..

-----------------------

Another thing I found is that we currently only support worker communication via Synapse's replication protocol.

The Synapse `workers.md` docs recommend that we don't do that and use Redis instead.

We do setup Redis, and I've tried to make it use that instead of Synapse's replication protocol (`matrix_synapse_replication_listener_enabled: "{{ matrix_synapse_workers_enabled and not matrix_redis_enabled }}"`, or you can just hardcode it to `false`), but for some reason communication fails.

Synapse logs weren't very helpful in telling me why yet..

roles/matrix-synapse/templates/synapse/systemd/matrix-synapse-worker@.service.j2
spantaleev: This has been redone. Should be fixed.

roles/matrix-base/templates/usr-local-bin/matrix-remove-all.j2
spantaleev: This has been redone. Now that we don't use systemd instantiated services and don't have `[Install] WantedBy=...` clauses in workers, things are simpler and we're back to something similar to what we had before:

```sh
	for s in $(find {{ matrix_systemd_path }}/ -type f -name "matrix-*" -printf "%f\n"); do
		systemctl disable --now $s
		rm -f {{ matrix_systemd_path }}/$s
	done
```

spantaleev:
To be notified of changes to the `workers.md` file in Synapse's repository, we can get inspired by [this HN comment](https://news.ycombinator.com/item?id=25903821), and subscribe to these RSS (well, Atom) feeds:

- (master) https://github.com/matrix-org/synapse/commits/master/docs/workers.md.atom
- (develop) https://github.com/matrix-org/synapse/commits/develop/docs/workers.md.atom

spantaleev:
It might make sense to add a "preset" variable that people can control to easily toggle the workers.

Just like `matrix_nginx_proxy_ssl_preset` (added in https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/755): https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/1aec5f9735160f68a58ff0e8e874bab50a1f16e9/roles/matrix-nginx-proxy/defaults/main.yml#L218-L229

You'd then be able to set `matrix_synapse_workers_preset` to some value like `basic-federation` to enable just the federation sender, or to `small`, `large` etc.

-------------

Maybe it's good to come up with another way of defining these workers. I'd rather say `matrix_synapse_workers_generic_worker_count: 5`, than have to specify them manually (along with their ports, etc).

We would then be able to do validation more easily as well (e.g. "You've requested 3 appservice workers, but this type of worker can only have a single instance at most").

Presets could then be a simpler definition, like:

- `santas-little-outbound-federation-helper`, leading to:
  - `matrix_synapse_workers_federation_sender_count = 1`
  - all other type of worker having a `_count` of `0`
- `one-of-each`, leading to:
  - `matrix_synapse_workers_generic_worker_count = 1`
  - `matrix_synapse_workers_federation_sender_count = 1`
  - `matrix_synapse_workers_media_repository_count = 1`
  - etc.
- a few other types that it's reasonable to have

Besides using a preset, people can also set these `_count` variables manually.

------

It may make sense to define a `_port_range_start` and `_metrics_range_start` variable for each type of worker, which is what we'd need to use to dynamically construct the `matrix_synapse_workers_enabled_list`.

-------------

Comments?

jolly-jump:
Hi,
something is broken:
```
TASK [matrix-synapse : Ensure unnecessary worker systemd services are stopped and disabled] **********************************************
failed: [matrix.my-domain.org] (item={'path': '/etc/systemd/system/matrix-synapse-worker@.service', 'mode': '0644', 'isdir': False, 'ischr': False, 'isblk': False, 'isreg': True, 'isfifo': False, 'islnk': False, 'issock': False, 'uid': 0, 'gid': 0, 'size': 1350, 'inode': 1835211, 'dev': 2049, 'nlink': 1, 'atime': 1612048538.3485785, 'mtime': 1611011784.3944073, 'ctime': 1611011784.862405, 'gr_name': 'root', 'pw_name': 'root', 'wusr': True, 'rusr': True, 'xusr': False, 'wgrp': False, 'rgrp': True, 'xgrp': False, 'woth': False, 'roth': True, 'xoth': False, 'isuid': False, 'isgid': False}) => {"ansible_loop_var": "item", "changed": false, "item": {"atime": 1612048538.3485785, "ctime": 1611011784.862405, "dev": 2049, "gid": 0, "gr_name": "root", "inode": 1835211, "isblk": false, "ischr": false, "isdir": false, "isfifo": false, "isgid": false, "islnk": false, "isreg": true, "issock": false, "isuid": false, "mode": "0644", "mtime": 1611011784.3944073, "nlink": 1, "path": "/etc/systemd/system/matrix-synapse-worker@.service", "pw_name": "root", "rgrp": true, "roth": true, "rusr": true, "size": 1350, "uid": 0, "wgrp": false, "woth": false, "wusr": true, "xgrp": false, "xoth": false, "xusr": false}, "msg": "Service is in unknown state", "status": {}}
```

spantaleev:
85a05f3 added support for dynamically generating `matrix_synapse_workers_enabled_list` based on other variables.
I believe that having this makes it much easier for people to configure workers.

I think having some presets might be a useful feature that can be added easily now. Most people would then be able to just switch a single other variable (`matrix_synapse_workers_preset`) which controls all these many other `matrix_synapse_workers_*_workers_count` variables.

It'd be interesting to brainstorm what some good presets would be. I guess it could be like this:
- `minimal` - a very minimal (but still useful) one is to just enable a federation worker and leave all other workers disabled
- `one-of-each` - the next preset could be "one of each" worker type
- `larger` - after that, there could be one is like "one of each", but has multiple generic workers (3 or 5, for example)

Note that we currently don't support running multiple `federation_sender` workers (see 61e427d), so we can't increase those right now.

People that need more workers than these 3 presets, can go for a hybrid approach: start with a preset (say "one of each") and then increase the number of a certain worker that they need more of (e.g. `matrix_synapse_workers_generic_workers_count: 15`).

Another question is, which should be the default preset? I think `one-of-each` is most helpful and reasonable. People that enable workers are obviously on a somewhat bigger server, so going for the middle option is good. People who want to save resources can go the extra mile and go for `minimal` or for additional tweaking of `matrix_synapse_workers_*_workers_count` variables. People who want a bigger worker setup can also use `one-of-each` and tweak the variables that seem worth tweaking.

-----------

Do you have other ideas for useful presets? Or comments about these ones? Does having a preset option sound useful at all or should we just go for "one of each" by default and ask people to tweak the worker counts?

ptman:
It would be great to get (voluntary) feedback on what actual worker counts end up being popular. One could start with one, or no, presets and add more as patterns are identified.

spantaleev:
I've made a breaking change to the format of `matrix_synapse_workers_enabled_list` in d6c4d41.

People that define that manually will need to add an `instanceId` property to each worker instance.

----------

In any case, it's better to refrain from defining `matrix_synapse_workers_enabled_list`, if possible in favor of `matrix_synapse_workers_*_workers_count` variables.

spantaleev:
I think we have a cause for celebration! This is now complete and merged to `master`! :tada: 

Huge thanks to everyone who has worked on making this happen!

----------------

There are minor things that could be improved.

For example, `pusher` and `federation_sender` workers can only run as a single instance (at most) for us. This is solvable, but requires more work and testing. It's probably not too important as well, given that TU Dresden scaled to 10k+ even without it.

Another thing that may be improved is metrics. Since #838, we do support metrics collection. When workers are enabled, graphs become somewhat sparse. I'm guessing we'd need to collect data from the various workers when workers are enabled. Maybe we need some different graphs (splitting them per worker or something?).

ptman:
https://github.com/matrix-org/synapse/issues/7773 https://github.com/matrix-org/synapse/blob/develop/docs/metrics-howto.md#monitoring-workers

jolly-jump:
Congratulations to those who worked on this! This saved my butt. We have 1k users only, but it seems this was one necessary step.

I have two other questions: 
- nginx worker connections
- postgres worker connections

They are not handled by this configuration of workers, I suppose. But, if I tune postgres to use "max_parallel_workers" as proposed here https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/532 and if I use more nginx worker connections by tampering with `/matrix/nginx-proxy/nginx.conf` is it wise to count all the workers together to avoid too much overprovisioning of CPU cores (I know, that is not a real thing).

On the other hand: would it be difficult to let the nginx-config (at least) be managed by this workers-configuration or is this all about synapse only and therfore it should be a separate configuration?

spantaleev:
We can adjust both of these (nginx workers; Postgres max connections) automatically when Synapse workers are enabled via `group_vars/matrix_servers` settings.

We can define variables for these things in their respective roles and then override them from `group_vars/matrix_servers` to some better values (for higher scale). We can even do it per preset (especially if we add larger presets that will require changes). Still, people may wish to adjust the values manually if they're running many additional workers.

eMPee584:
> > It's probably not too important as well, given that TU Dresden scaled to 10k+ even without it.
> 
> Just to get an idea, how many resources/workers are needed at that scale?

From what I remember, very few.. a couple (3-6) synchrotron workers and the resultant parallelization from using just a single worker for the others already had a sufficient impact compared to a monolithic synapse process.
I'll ask the current admin team for the worker config in use now..

PR #459, opened 2020-04-15
Upstream set mx-puppet-skype default port to 8438

rbelem:
Default port was set to 6000 and upstream developer updated to 8438 to avoid port clash.


spantaleev:
Since these ports are never exposed within the container, there's no possibility of clash, is there?

In any case, doesn't hurt to have it be different. Thanks! :+1:

spantaleev:
Ahh.. To answer my own question - upstream was observing clashes and they're changing the port for their own reasons.
We're just following what they do, even though it's not strictly necessary for us. I get it now :)

PR #460, opened 2020-04-16
Fix hardcoded postgres template values

cnvandijk:
There was a few hardcoded values in the postgres script template `matrix-change-user-admin-status`, which are fixed by this PR to correctly used the templated variables. Additionally, some inconsistent whitespace usage I noticed in these scripts is fixed by this PR.


spantaleev:
Thank you! :+1:

cnvandijk:
And thanks to you for this great project, it has saved us a lot of time.

spantaleev:
Happy to hear this! Thank you for the kind words!

PR #461, opened 2020-04-17
Expose allowed local 3pids as a configurable option

lampholder:
Trivial change to expose `allowed_local_3pids` (which allows you to restrict registration to users who provide a 3pid matching the specified pattern or patterns) as a configurable option.


spantaleev:
Thank you! :+1:

PR #462, opened 2020-04-18
Add support for running dimension in an unfederated environment

cnvandijk:
This PR adds support for running dimension in an unfederated environment, following the upstream [documentation](https://github.com/turt2live/matrix-dimension/blob/master/docs/unfederated.md).

I'm not sure about the DNS requirements in an unfederated environment, so left that section of the [docs](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-dimension.md#prerequisites) alone.

Happy to hear any advice on a better/cleaner way to make the switch in the config.

### DIFF
--- a/roles/matrix-dimension/defaults/main.yml
+++ b/roles/matrix-dimension/defaults/main.yml
@@ -58,7 +58,7 @@ matrix_dimension_configuration_yaml: |
     # The URL that Dimension should use when trying to communicate with federated APIs on
     # the homeserver. If not supplied or left empty Dimension will try to resolve the address
     # through the normal federation process.
-    federationUrl: "http://matrix-synapse:8048"
+    federationUrl: "http://matrix-synapse:{{ 8048 if matrix_synapse_federation_enabled|bool else 8008 }}"
 
     # The URL that Dimension will redirect media requests to for downloading media such as
     # stickers. If not supplied or left empty Dimension will use the clientServerUrl.
--- a/roles/matrix-dimension/defaults/main.yml
+++ b/roles/matrix-dimension/defaults/main.yml
@@ -32,6 +32,8 @@ matrix_dimension_integrations_rest_url: "https://{{ matrix_server_fqn_dimension 
 matrix_dimension_integrations_widgets_urls: ["https://{{ matrix_server_fqn_dimension }}/widgets"]
 matrix_dimension_integrations_jitsi_widget_url: "https://{{ matrix_server_fqn_dimension }}/widgets/jitsi"
 
+matrix_dimension_homeserver_federationUrl: "http://matrix-synapse:8048"
+
 # Default Dimension configuration template which covers the generic use case.
 # You can customize it by controlling the various variables inside it.
 #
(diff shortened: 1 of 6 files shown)


roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -226,7 +226,11 @@ listeners:
     x_forwarded: true
 
     resources:
-      - names: [client]
+      - names:
```
spantaleev, line 229: For this one, we should probably introduce a new variable in `roles/matrix-synapse/defaults/main.yml` like this:

```yaml
matrix_synapse_http_listener_resource_names: ['client']
```

We can use its value here like this:
```
    resources:
      - names: {{ matrix_synapse_http_listener_resource_names|to_json }}
        compress: false
```

In `group_vars/matrix_servers`, we can do the override depending on Dimension configuration. Example:

```yaml
matrix_synapse_http_listener_resource_names: |
  {{
    ['client']
    +
    (['openid'] if matrix_dimension_enabled and not matrix_synapse_federation_enabled else [])
  }}
```

tomkel:
Is there any extra logic needed if federation is enabled but no servers are whitelisted?

spantaleev:
Not sure about this. We can ignore this special case for now, I guess.

cnvandijk:
> Is there any extra logic needed if federation is enabled but no servers are whitelisted?

As far as I can infer from [this bug report](https://github.com/turt2live/matrix-dimension/issues/247), the dimension domain needs to be whitelisted. Since this PR only deals with the federation url and openid listener, the two seem unrelated. Furthermore, this PR does not prevent operation in federated environments with whitelisted domains, the user just needs to whitelist the dimension domain.

It could be an idea to verify what I wrote above and add it to the documentation, but this is outside the scope of this PR. Since we run an unfederated system, I can't help with testing.

PS: when no servers are whitelisted synapse accepts connections from all domains, but I assume you mean when there *are* whitelisted servers.

roles/matrix-dimension/defaults/main.yml
```diff
@@ -58,7 +58,7 @@ matrix_dimension_configuration_yaml: |
     # The URL that Dimension should use when trying to communicate with federated APIs on
     # the homeserver. If not supplied or left empty Dimension will try to resolve the address
     # through the normal federation process.
-    federationUrl: "http://matrix-synapse:8048"
+    federationUrl: "http://matrix-synapse:{{ 8048 if matrix_synapse_federation_enabled|bool else 8008 }}"
```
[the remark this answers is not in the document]
cnvandijk: Resolved in 3f4bc9b.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
cnvandijk: Resolved in 3f4bc9b.

cnvandijk:
Thanks for the feedback @spantaleev, I agree your suggested method is cleaner. I have pushed 3f4bc9b which should resolve your comments. Please let me know if you would like 3ddb8cd and 3f4bc9b squashed together before merging.

spantaleev:
Looks good! Thank you for your diligence! :+1:

PR #463, opened 2020-04-19
Architecture

hooger:
Add better support for non amd64 architectures, hopefully replacing the self-building nomenclature which (in my opinion) is a little user hostile. :)


spantaleev:
Thank you! :+1:

PR #470, opened 2020-04-26
Add information to make jitsi work in Local Network

bnounours:
I struggled several days to make it work.
Jitsi was not working in LAN but was ok from WAN to WAN.
I found that JVB send the local Docker IP to client in LAN (172.X.X.X) which is not accessible easily.

The doc of jitsi in docker suggest to add **DOCKER_HOST_ADDRESS**  in environment variable to make it work in LAN.
https://github.com/jitsi/docker-jitsi-meet#running-behind-nat-or-on-a-lan-environment

### DIFF
--- a/docs/configuring-playbook-jitsi.md
+++ b/docs/configuring-playbook-jitsi.md
@@ -27,6 +27,12 @@ matrix_jitsi_enabled: true
 # We only need this temporarily - until Jitsi integration in riot-web is finalized.
 # Remove this line in the future, to switch back to a stable riot-web version.
 matrix_riot_web_docker_image: "vectorim/riot-web:develop"
+
+# Put this parameter to make jitsi work on local network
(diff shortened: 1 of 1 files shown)


spantaleev:
Hi! Thanks for figuring this out, and in a clean way!

Would you rebase it against the current docs page? Looks like we've made some changes that yours doesn't have and there are merge conflicts.

--------------

Also, since most playbook users are probably unlikely to need LAN access to their Jitsi server, so we should probably not make this setting so prominent at the top of the Jitsi docs page.

I think it would be better to add yet another "(Optional) Making your Jitsi server work on a LAN", which explains the problem, mentions that it's just for people who access Jitsi from the LAN, and then tells them how to do it:

```yaml
matrix_jitsi_jvb_container_extra_arguments:
  - '--env "DOCKER_HOST_ADDRESS=<Local IP adress of the host>"'
```

Would you adjust your PR like that?

bnounours:
ok

PR #472, opened 2020-04-27
add missing keys 'permalinkPrefix' and 'default_theme' for configuring riot web

GuillauG:
### Description

Add two useful keys for configuring riot-web : `permalinkPrefix` and `default_theme`

### Why

Those two keys were missing and are useful to further configure Riot-Web, both being specifications of the [riot config file](https://github.com/vector-im/riot-web/blob/develop/docs/config.md)

- `permalinkPrefix` allows to redirect users to non-federated instance of Riot at `riot.DOMAIN` or `matrix.to.DOMAIN` instead of the default `matrix.to`
- `default_theme` allows to edit default_theme for Riot users

### DIFF
--- a/roles/matrix-riot-web/defaults/main.yml
+++ b/roles/matrix-riot-web/defaults/main.yml
@@ -29,6 +29,7 @@ matrix_riot_web_integrations_ui_url: "https://scalar.vector.im/"
 matrix_riot_web_integrations_rest_url: "https://scalar.vector.im/api"
 matrix_riot_web_integrations_widgets_urls: ["https://scalar.vector.im/api"]
 matrix_riot_web_integrations_jitsi_widget_url: "https://scalar.vector.im/api/widgets/jitsi.html"
+matrix_riot_web_permalinkPrefix: "https://matrix.to"
 # Riot public room directory server(s)
 matrix_riot_web_roomdir_servers: ['matrix.org']
 matrix_riot_web_welcome_user_id: "@riot-bot:matrix.org"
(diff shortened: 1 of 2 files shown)


spantaleev:
Looks like there's no need for a `matrix_riot_web_change_default_theme` variable. We can just check like this: `{% if matrix_riot_web_default_theme != '' %}`.

Ideally, if we can provide a sane default value (`'light'`?) and do it without an `{% if .. %}` block, that would be even better.

----

Also, the place you've added these new variables is a little unsuitable - it breaks up the comment we had above the `matrix_riot_web_settingDefaults_custom_themes` variable.

------

As for `matrix_riot_web_permalinkPrefix`, would it work well if people set its value like this:

```yaml
matrix_riot_web_permalinkPrefix: "https://{{ matrix_server_fqn_riot }}"
```

Well, I'm not sure we should make this the default. I guess sticking to `matrix_riot_web_permalinkPrefix: "https://matrix.to"` is better.

spantaleev:
.. and thank you for your contribution, by the way!

Also, I'd like to note that one can extend the configuration without having to introduce new variables. Take a look at the `matrix_riot_web_configuration_extension_json` variable.

Still, for some frequently-used things like this (the default theme thing at least!), it's good to provide variables explicitly.

GuillauG:
Thank your for the good points !
I've reworked the pull request.

I think it's better to keep `https://matrix.to` as the default.
Let me know if you see anything else.

And feel free to finalize if you like !

spantaleev:
Looks good! Thank you! :+1:

PR #475, opened 2020-04-28
Update Jitsi docker images to stable 4416

teutat3s:
As far as I can see, there should be no further changes needed to our special ansible setup.

https://github.com/jitsi/docker-jitsi-meet/compare/stable-4384-1...stable-4416


spantaleev:
Thanks! :+1:

PR #478, opened 2020-04-29
Add support for Jitsi LDAP authentication

cnvandijk:
This PR adds support for the [saslauthd method](https://github.com/jitsi/jitsi-meet/wiki/LDAP-Authentication#ldap-authentication-for-jitsi-meet-via-cyrussaslauthd) of authenticating Jitsi users via LDAP (by following the [docker-jitsi-meet](https://github.com/jitsi/docker-jitsi-meet#authentication-using-ldap) docs).

I was considering trying to harmonise the Jitsi LDAP configuration added in this PR with the other LDAP auth methods we already support ([synapse ldap3](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-ldap-auth.md#setting-up-the-ldap-authentication-password-provider-module-optional-advanced) and [ma1sd ldap](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-ma1sd.md#authentication)) under the assumption that most users would be using the same LDAP server for all services, however I feel this will get quite messy. Suggestions welcome on how to do it in a clean way.

Note that there are three authentication methods available, [internal, ldap and jwt](https://github.com/jitsi/docker-jitsi-meet#authentication). I left JWT out because I am not able to test this method.

### DIFF
--- a/docs/configuring-playbook-jitsi.md
+++ b/docs/configuring-playbook-jitsi.md
@@ -4,7 +4,7 @@ The playbook can install the [Jitsi](https://jitsi.org/) video-conferencing plat
 
 Jitsi installation is **not enabled by default**, because it's not a core component of Matrix services.
 
-The setup done by the playbook is very similar to [docker-jitsi-meet](https://github.com/jitsi/docker-jitsi-meet).
+The setup done by the playbook is very similar to [docker-jitsi-meet](https://github.com/jitsi/docker-jitsi-meet). You can refer to the documentation there for many of the options here.
 
 
 ## Prerequisites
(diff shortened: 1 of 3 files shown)


spantaleev:
It looks great! Thank you! :+1: 

Hooking it to ma1sd LDAP or the Synapse LDAP configuration is probably difficult to do in a generic way. I guess we should leave it manual, as it is now.

cnvandijk:
Yes I agree that manual is fine for now, it is the most flexible. There will always be small differences in the config, for example I needed a different LDAP filter for sasl and ma1sd.

Cheers for the documentation improvements, on both pages.

PR #479, opened 2020-04-29
Document how to point dimension jitsi widgets at your own jitsi instance

cnvandijk:
This can be removed once turt2live/matrix-dimension#345 is resolved.


spantaleev:
Thank you! :+1: 

I've improved the Jitsi docs page a bit in 0029b37b9690cf26, mentioning the same Dimension problem.

PR #480, opened 2020-05-01
Add caddy 1 well-known example

andresterba:
The provided example for federation seems to be for Caddy 2.
As I am stuck with Caddy 1 for now, this is the proxy setting for Caddy 1.


spantaleev:
Thank you! :+1:

PR #481, opened 2020-05-01
fix skype bridge being rebuilt when upstream changed

hooger:


SuperSandro2000:
Just curious: Why did you change that?

spantaleev:
You use the result if the `git pull` in other places, instead of hardcoding to `yes`, don't you?

spantaleev:
Ah, I see.. The `when` statement handles that:

```yaml
when: "matrix_mx_puppet_skype_enabled|bool and matrix_mx_puppet_skype_container_image_self_build and matrix_mx_puppet_skype_git_pull_results.changed"
```

Thank you! :+1:

PR #484, opened 2020-05-04
Clarify domain of Dimension admin users in docs by using matrix_domain variable

cnvandijk:
A minor improvement in the Dimension docs, as discussed in ﻿#matrix-docker-ansible-deploy:devture.com.


spantaleev:
Thank you! :+1:

PR #485, opened 2020-05-04
Fix group name gid

cnvandijk:
This PR addresses a few related issues i noticed regarding the ownership and permissions of files and directories managed by the playbook.

It should be highlighted that user intervention during migration is likely needed to fix the group ownership of `dimension.db`. I put the suggested commands in the commit message of 741064a, but it is basically just `chown -R`. How do we want to handle `CHANGELOG.md`? As far as I can tell there are no tags, just a date and a description of the action required, correct?

The migration worked for me, but it may warrant some more testing before merging into master.

The last commit 7585bcc is optional, I find it cleaner to use a separate variable for user and group but am open to discussion. I don't need it for my use case, it just felt like the right thing to do given the other changes in this PR.

On the topic of permissions, there is a mix used for both files and directories throughout the playbook. Even at the top level there is some variation:
```
drwxr-x--- matrix matrix coturn
drwxrwx--- matrix matrix dimension
drwxr-x--- matrix matrix jitsi
drwxr-x--- matrix matrix ma1sd
drwxr-x--- matrix matrix mailer
drwxr-x--- matrix matrix nginx-proxy
drwx------ matrix matrix postgres
drwxr-x--- matrix matrix riot-web
drwxrwx--- matrix matrix ssl
drwxr-xr-x matrix matrix static-files
drwxr-x--- matrix matrix synapse
```
Not sure if I am brave enough to tackle this though. At least in the subfolders I am sure there is a reason for some of the modes. Much of it could however be harmonised via a variable, like with `matrix_base_data_path_mode`.

### DIFF
--- a/roles/matrix-base/defaults/main.yml
+++ b/roles/matrix-base/defaults/main.yml
@@ -22,6 +22,7 @@ matrix_server_fqn_dimension: "dimension.{{ matrix_domain }}"
 matrix_server_fqn_jitsi: "jitsi.{{ matrix_domain }}"
 
 matrix_user_username: "matrix"
+matrix_user_groupname: "matrix"
 matrix_user_uid: 991
 matrix_user_gid: 991
 
--- a/roles/matrix-base/tasks/setup_matrix_base.yml
+++ b/roles/matrix-base/tasks/setup_matrix_base.yml
@@ -2,7 +2,7 @@
 
 - name: Ensure Matrix group is created
   group:
-    name: "{{ matrix_user_username }}"
+    name: "{{ matrix_user_groupname }}"
     gid: "{{ matrix_user_gid }}"
     state: present
 
@@ -11,7 +11,7 @@
     name: "{{ matrix_user_username }}"
     uid: "{{ matrix_user_uid }}"
     state: present
-    group: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_groupname }}"
 
 - name: Ensure Matrix base path exists
   file:
@@ -19,7 +19,7 @@
     state: directory
     mode: "{{ matrix_base_data_path_mode }}"
     owner: "{{ matrix_user_username }}"
-    group: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_groupname }}"
   with_items:
     - "{{ matrix_base_data_path }}"
 
--- a/roles/matrix-base/tasks/setup_well_known.yml
+++ b/roles/matrix-base/tasks/setup_well_known.yml
@@ -8,7 +8,7 @@
     state: directory
     mode: 0755
     owner: "{{ matrix_user_username }}"
-    group: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_groupname }}"
   with_items:
     - "{{ matrix_static_files_base_path }}/.well-known/matrix"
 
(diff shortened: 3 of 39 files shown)


cnvandijk:
Lol ok I clearly did not test this thoroughly enough -- the Dimension container is failing to start due to the `node` user inside the container having a uid:gid of 1000:1000. I will read up a bit on alternative ways to do this, as I can't have containers running on our server with 1000:1000.

As a side note, these changes did pass `--tags=self-check`. Some extra tests may thus need to be added...

spantaleev:
`--tags=self-check` mostly checks for Synapse, DNS records, `.well-known`, etc.

It doesn't do an exhaustive check of all services.

spantaleev:
Why couldn't you use `--user 1000:1000` for Dimension?

Although I do have a user with this `uid`, such a user doesn't necessarily need to exist on the host system, I believe.

spantaleev:
I've merged #486 and then this one (had to resolve some conflicts).

Thank you for introducing a `matrix_user_groupname` variable! :+1: 

---------------

Now I'm thinking about Dimension though..

We should probably let it run as its own user, as it used to. Or we should find a way to run the Dimension Docker image with the `matrix` user. Right now it's a mix of both and it probably won't work, even if ownership of its database is fixed up, as per 741064a178b8bd67.

cnvandijk:
Apologies for the slow response, yesterday was a public holiday here. I think the merge may have been a bit premature, see my comment about 741064a:
> the Dimension container is failing to start

As for the Dimension user, as best I understand I think we have three options:
1. Rebuild the Dimension image with a different uid
2. Use a tool like [uidfix](https://github.com/boxboat/fixuid) (not recommended in production)
3. Stick with 1000:1000

I am currently looking into a fourth option of remapping using subordinate user and group IDs, see [here](https://docs.docker.com/engine/security/userns-remap/#about-remapping-and-subordinate-user-and-group-ids). I will report back once I understand it a bit more.

As a side note, I am a bit confused with the history of the repo at the moment, how did you do the merge into master in 36c61b5?

cnvandijk:
> Why couldn't you use `--user 1000:1000` for Dimension?
> 
> Although I do have a user with this `uid`, such a user doesn't necessarily need to exist on the host system, I believe.

For us it is a bit the other way around, we have an existing LDAP group 1000 with users that should not be able to access the files. Of course we have many layers that should prevent them getting this far, but the more the better.

cnvandijk:
> `--tags=self-check` mostly checks for Synapse, DNS records, `.well-known`, etc.
> 
> It doesn't do an exhaustive check of all services.

It is probably a good idea to at least make sure that all services are running. Do you think this belongs in `--tags=self-check`, or even in `--tags=start`? Let me know, then I will open an appropriate issue to track this (not sure if I will have time to personally work on it though).

spantaleev:
I've since reverted the Dimension stuff to how it used to work before (see ae1b1be3f4b6a683 and dd527d596850).

We should do another PR to move away from using `1000:1000`, I guess. Maybe it's best if we fixup the original image and sent the contribution upstream.


-------------

> For us it is a bit the other way around, we have an existing LDAP group 1000 with users that should not be able to access the files. Of course we have many layers that should prevent them getting this far, but the more the better.

Yep.. These files are in `/matrix/dimension`, which should not be accessible via a `1000:1000` user. I think you should be safe enough. Of course, not using `1000:1000` would be ideal.


-----------------

> As a side note, I am a bit confused with the history of the repo at the moment, how did you do the merge into master in 36c61b5?

I merged it on the command-line. Your commits are listed as one of the parents.

---------------

> It is probably a good idea to at least make sure that all services are running. Do you think this belongs in --tags=self-check, or even in --tags=start? Let me know, then I will open an appropriate issue to track this (not sure if I will have time to personally work on it though).

`--tags=start` currently checks if services start successfully. I guess that's good enough.

The same service checks could be added to `--tags=self-check` as well, I guess.

These checks are a little ugly at the moment, because of some systemd workarounds on newer distros :/

PR #487, opened 2020-05-05
Added a documentation enhancement regarding external TURN server

christianlupus:
Just by chance stumbled upon a small piece of missing information during reading.

### DIFF
--- a/docs/configuring-playbook-turn.md
+++ b/docs/configuring-playbook-turn.md
@@ -31,3 +31,12 @@ matrix_synapse_turn_uris:
 - turn:HOSTNAME_OR_IP?transport=udp
 - turn:HOSTNAME_OR_IP?transport=tcp
 ```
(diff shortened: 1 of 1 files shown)


spantaleev:
This happens automatically, thanks to `group_vars/matrix_servers`: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/7c4bc0207058b36315559c61188876359c7b294d/group_vars/matrix_servers#L492-L500

I guess we don't need documentation adjustments.

christianlupus:
This is up to you to decide. I thought this way you could avoid connections to external servers when you want to keep it in a restricted environment.

spantaleev:
Ohh.. You addition is to the "Using your own TURN server" section. I overlooked that.
It certainly makes sense to add it. Thank you! :+1:

PR #492, opened 2020-05-09
Add note about ask-become-pass

therealklanni:
This flag is necessary for Ansible to ask for the sudo password when using the non-root option as documented in the `hosts` file. Otherwise, Ansible errors out with `missing sudo password`.


spantaleev:
Thanks! :+1:

PR #495, opened 2020-05-10
Document how to administer synapse database via tunneled web-frontend

eMPee584:
.. because less pain :sweat_smile:


spantaleev:
Thanks! :+1:

PR #498, opened 2020-05-13
Add matrix_skip_setup variable to not run setup tasks

eMPee584:
On Debian, even though the `matrix_docker_installation_enabled` allows to skip activation of the docker.com repository and installation of its packages, `roles/matrix-base/tasks/server_base/setup_debian.yml` triggers other actions that might not be wanted. For example, the various
```
  apt:
    update_cache: yes
```
will trigger an `apt update` action to refresh all apt repositories, which will only succeed when no apt interface is running. Furthermore, `apt-transport-https` has been included in `apt` since a while:

>    This is a dummy transitional package - https support has been moved into 
>    the apt package in 1.5. It can be safely removed.

https://pkgs.org/search/?q=apt shows that `stretch` was the last distro to need that package, but maybe someone is still running this on such a system, so m'eh.
Then the last package unconditionally installed is `fuse`, which in turn causes `fuse3` from unstable to be uninstalled, as well as its dependencies. Might create another patch for these issues..
For now, define a variable to allow completely skipping server-base setup on all platforms.

### DIFF
--- a/roles/matrix-base/tasks/main.yml
+++ b/roles/matrix-base/tasks/main.yml
@@ -13,7 +13,9 @@
     - setup-all
 
 - import_tasks: "{{ role_path }}/tasks/server_base/setup.yml"
-  when: run_setup|bool
+  when:
+    - run_setup|bool
+    - not matrix_skip_setup_server_base|bool
   tags:
     - setup-all
 


spantaleev:
I wonder if `matrix_skip_setup` should skip all other things as well..?

Or, if it's just mean to control package installation, it might make sense to  call it `matrix_skip_setup_package_installation` or something. Perhaps it's better to call it `matrix_setup_package_installation_enabled` in that case (defaulting to `true`).

-----------------

As for `fuse`, we should probably install that on demand (in the `matrix-synapse` role), only when Goofys is enabled.

eMPee584:
uhm yeah semantics :sweat_smile: 
what about `matrix_skip_setup_server_base` ?

spantaleev:
And that would skip this whole thing, I imagine?

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5b8fd44aa7a0659e67ce8bdb7bfb3448e9656d51/roles/matrix-base/tasks/main.yml#L15-L18

eMPee584:
NEW: more simpler xD

spantaleev:
Looks good.. Except, we need to define `matrix_skip_setup_server_base: true` in `roles/matrix-base/defaults/main.yml` ;)

eMPee584:
Ah now i go through the worker patches i remember this little fun fact:
```
# a negative when condition will not actually prevent
# ansible from executing loops in imported tasks!                                                         
```

`ansibe-insanity --why 🤦‍♂️`

PR #499, opened 2020-05-13
Prevent 404s when Riot tries to retrieve domain-specific config.json

eMPee584:
Riot unconditionally asks for a config.${document.domain}.json, c.f.
https://github.com/vector-im/riot-web/blame/develop/src/vector/getconfig.ts#L24
So with some ease, we provide that.


eMPee584:
uuuhm .. i think `${document.domain}` actually equals `{{matrix_server_fqn_riot}}` :facepalm:

aaronraimist:
It's a harmless error but sure. https://github.com/vector-im/riot-web/commit/e669c681e29e4b757cf7c593ad6878c985011827

> Fetch both config.json-s at the same time, first one fails 99% of the time

spantaleev:
Looks good! Thanks! :+1:

PR #507, opened 2020-05-16
Typo STMP -> SMTP

chagai95:


spantaleev:
Good catch! Thanks! :+1:

PR #508, opened 2020-05-16
Jitsi 4548-1 update

teutat3s:
This PR updates the Jitsi containers to the latest version which is 4548-1 atm.
Also a new volume for plugins is added to the prosody container.

Possible breaking upstream changes:
- Prosody auth method changed default from `internal_plain` to `internal_hashed`. I guess people using auth need to reconfigure all passwords. Will report back after more testing. See https://github.com/jitsi/docker-jitsi-meet/commit/d44230e2b121e72542a60b6dfbce9b07e84ed706#diff-af6bb4ab6f170892599e22f681468018

Open question:
- How closely do we want to follow the Jitsi docker upstream repo? E.g. [this commit](https://github.com/jitsi/docker-jitsi-meet/commit/0177765f2dfb7c54afa33158d5e25e8f3bcefe4d#diff-4e5e90c6228fd48698d074241c2ba760) adds compatibility for SElinux `enforcing` mode, do we want that, too? Or do we wait with the implementation until the need arises?

### DIFF
--- a/roles/matrix-jitsi/defaults/main.yml
+++ b/roles/matrix-jitsi/defaults/main.yml
@@ -50,7 +50,7 @@ matrix_jitsi_jibri_recorder_user: recorder
 matrix_jitsi_jibri_recorder_password: ''
 
 
-matrix_jitsi_web_docker_image: "jitsi/web:4416"
+matrix_jitsi_web_docker_image: "jitsi/web:4548-1"
 matrix_jitsi_web_docker_image_force_pull: "{{ matrix_jitsi_web_docker_image.endswith(':latest') }}"
 
 matrix_jitsi_web_base_path: "{{ matrix_base_data_path }}/jitsi/web"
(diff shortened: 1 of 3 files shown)


spantaleev:
We probably don't need to do anything about the SELinux change, unless we want to support SELinux as well. I have no way to test it though.

---------

About authentication, looks like `AUTH_TYPE` still defaults to `internal` for `env.example`: https://github.com/jitsi/docker-jitsi-meet/blob/3472ab0b19f13e675b9e438f8882bd09c0979d8e/env.example#L105

So we likely don't need to change our own `matrix_jitsi_auth_type` variable.

It'd be nice to figure out if people need to reconfigure all passwords because of the internal change though.

------------

About `matrix_jitsi_prosody_plugins_path`, we should probably make the playbook explicitly create the directory here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/63c4538400c436b0175e1c0b1550c991584e01e5/roles/matrix-jitsi/tasks/setup_jitsi_prosody.yml#L7-L17

Right now (when using `-v` to mount it), Docker would auto-create it, but we'd better not rely on that. It's most likely created with inappropriate permissions that way.

teutat3s:
I just did a quick test again, the internal change should only affect those who delete their prosody config file `prosody/config/prosody.cfg.lua` without deleting their `prosody/config/data` directory. So whenever issues arise, people should delete the whole prosody config folder anyways, to start with a fresh config.

spantaleev:
Awesome! Thank you for figuring it all out! :+1:

Issue #505, opened 2020-05-15
3PIDs not shown in user settings / reverse proxy issue?


rootuser:
Hello,

I am trying to find the reason for an issue I discovered yesterday:
3PIDs are not shown on a users settings page (I have only tried that thoroughly with email addresses though).

At first I posted this error in the riot web issue tracker, so all the details can be found there, if necessary [https://github.com/vector-im/riot-web/issues/13033](url)

In short it's like this: Emails are not shown in user settings in riot web and riotX (Android). In the old riot.im app emails are displayed correctly.
Also, I can successfully request the 3PID list via client-server API.
So I guess riot web and riot x use the identity service API, while riot.im uses client api.

Now I noticed in my browser (with developer tools active) that there seems to be an issue with the identity server; when opening the setings page I get:

> Access to XMLHttpRequest at 'https://matrix.ohoco.de/_matrix/identity/v2/account/register' from origin 'https://riot.ohoco.de' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

The identity server responds with a
> HTTP/1.1 405 Method Not Allowed

The question is now: why is that header not send? In the issue linked above, t3chguy suggests it might be a reverse proxy issue.
I have gone through your installation instructions again, but could not find any error in my configuration.

Do you perhaps have an idea what I could look for?
Thanks for any hint!

Best, Tom

ns130291:
same as #503

Issue #503, opened 2020-05-14
Email addresses/phone numbers do not show up in the Settings tab


schopenhauer:
The issue has already been reported in the following [bug report](https://github.com/vector-im/riot-web/issues/13033#) logged in the `riot-web` repo.

Potential solution would be to add the ACAO directive to the reverse proxy - to be tested.

```
add_header Access-Control-Allow-Origin *;
```

ns130291:
Adding the header in the reverse proxy (in my case haproxy) solves the CORS error in the browser but doesn't help with the `405 Method Not Allowed` error. 
I guess the header isn't set when this error occurs as all other responses from the identity module have `Access-Control-Allow-Origin: *` set.

t2d:
running with `-t self-check` doesn't report any problems but this issue persists

0x46616c6b:
Maybe that is related to #511. If `ma1sd` not configured with enabled `v2` api prefix Riot reports a problem with CORS (405 Method not allowed).

ns130291:
with the new change I'm no longer getting the 405 error, but now it says `401 Unauthorized`

rootuser:
I can confirm that the actual release fixes the CORS header problem.
Alas I have issues with ma1sd instead; 

The identity server is requested for 3PIDs, but fails with the followin error (from the Javascript console:)

> Unable to reach identity server at https://matrix.myserver.de to check for 3PIDs bindings in Settings
> Error: Unsupported identity server: bad response

Do you think that would be fixed in the new ma1sd release with v2 enabled?

spantaleev:
You can try the ma1sd RC by putting this in your `vars.yml` file:

```yaml
matrix_ma1sd_docker_image: "ma1uta/ma1sd:2.4.0-rc1"
```

Let us know if it helps! If not, you can report the issue at the [ma1sd](https://github.com/ma1uta/ma1sd) repository.

rootuser:
Thanks for your reply!

I did try it but, alas, to no avail. I will report the bug.

spantaleev:
Bug seems to be reported here: https://github.com/ma1uta/ma1sd/issues/48

jmastr:
@spantaleev ma1sd v2.4.0 is now released, but the bug is still present. I can confirm this error:
```
Unable to reach identity server at https://matrix.trust2core.rocks to check for 3PIDs bindings in Settings
```

[...]

PR #511, opened 2020-05-19
enable by default v2 api prefix for ma1sd

0x46616c6b:
I stumpled upon this while installing and configuring Matrix. The client (Riot) was not able to use the configured Identity Server. Riot returned

```
Access to XMLHttpRequest at 'https://matrix.example.com/_matrix/identity/v2/account/register' from origin 'vector://vector' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

First it looks like a CORS problem but with a deeper look in ma1sd I noticed that `v2` is not enabled by default at the moment. They will change that in a upcoming release. So I think it is a useful change to enable that by default or have it configurable anyways.

### DIFF
--- a/roles/matrix-ma1sd/defaults/main.yml
+++ b/roles/matrix-ma1sd/defaults/main.yml
@@ -76,6 +76,10 @@ matrix_ma1sd_self_check_validate_certificates: true
 # According to: https://github.com/ma1uta/ma1sd/blob/master/docs/troubleshooting.md#increase-verbosity
 matrix_ma1sd_verbose_logging: false
 
+# Setting up support for API prefixes
+matrix_ma1sd_v1_enabled: true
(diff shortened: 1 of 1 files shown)


spantaleev:
It'd be nice to stick to upstream defaults, but...

.. given that not having API v2 enabled breaks other important components we install by default (riot-web), we may as well do something about it and force-enable v2.

Thank you for fixing #503 and #505! :+1:

teutat3s:
It seems version `2.4.0` is going to enable API v2 by default
There's only a RC for now.
https://github.com/ma1uta/ma1sd/releases/tag/2.4.0-rc1

spantaleev:
Nice! It will happen soon enough then :)

PR #513, opened 2020-05-21
Fix/mx puppet bridge

joao-p-marques:
Small changes to fix the MX Puppet Bridge setup.

1. OAuth was not enabled by default
2. The wrong variable was being used in `/roles/matrix-bridge-mx-puppet-slack/tasks/init.yml`. It should be `matrix_mx_puppet_slack_enabled`, and not `matrix_appservice_slack_enabled`


spantaleev:
Thank you! :+1:

PR #514, opened 2020-05-22
added matrix-sms-bridge

benkuly:
Currently not tested. Wanted to know, if it is a good idea to create one systemd-service for appservice and database and also if I made any mistakes, which could destroy my matrix-setup.


spantaleev:
Thank you for creating this Matrix <-> SMS bridge and adding support to the playbook!

I took a bit of a look through your PR. My findings are below.

---------

I also think it'd be better if we kept the database separate.

The same role could create an extra systemd `.service` for the database and make the main service depend on the database service (using `After=` and `Requires=`).

-----

I also think this part may be a bit worrisome:

```yaml
matrix_sms_bridge_database_password: "{{ matrix_synapse_macaroon_secret_key | password_hash('sha512', 'sms.db.password') | to_uuid }}"
```

Looking at the [neo4j Docker image documentation on authentication](https://neo4j.com/docs/operations-manual/current/docker/introduction/#docker-auth), it seems like `NEO4J_AUTH` is only used initially (when creating the database). Any subsequent changes to it won't affect an existing database.

If `matrix_synapse_macaroon_secret_key` were to change in the future, we'll end up with a mismatch between the initially-assigned database password and the one we try to use later on, leading to an inaccessible database.

So I think people assigning the password manually and explicitly might be better. I see that `validate_config.yml` already guards against `matrix_sms_bridge_database_password` being empty.

---------

Looking at the current systemd `.service`, it looks like you're using potentially `matrix_sms_bridge_container_http_host_bind_port` for both the database and the bridge. You probably need a separate variable for the database.

I guess there's no need to expose the port of the NEO4J database though. So removing `matrix_sms_bridge_container_http_host_bind_port` usage from its `docker run` command is probably better.

---------

The `ExecStart` lines in the systemd service are likely missing some `\` slashes on some lines.

benkuly:
Thank you very much!
I added docs, seperated the services for database and fixed your suggestions.
I will now try if it works.

benkuly:
Currently this cotains some issues while running, but I will fix it tomorrow.

spantaleev:
It looks good! Thanks for these updates! :+1: 

Anything else you'd like to do (testing that it works?) or is it ready for merging?

benkuly:
Currently everything is working except, that the user has not enough rights to access the modem...

spantaleev:
Perhaps `--cap-drop=ALL` is too much. Maybe if we get rid of that it would work? Or do we need to not use `--user` as well?

Also, instead of not doing `--cap-drop=ALL`, it'd be better to use `--cap-add=...` and add whatever is necessary.. But that could be the next step - once you get it working :)

benkuly:
Or maybe the `--user` just needs to be in a group, that has access to the device? In ubuntu it's `dialout`.

spantaleev:
If you create a user in the Docker image and it has some static uid/gid, we can use those in the `--user` argument, instead of `matrix_user_uid` and `matrix_user_gid`.

The downside is that it would be creating files in `/data` using that foreign user, which would likely cause other issues down the road. Hmm..

benkuly:
And adding the matrix uid to the group is not possible? I will try to find a way the next days, when I have sparetime.

spantaleev:
When you use `--user`, the Docker image process starts as that user/group. You can't manipulate group information with a regular user.

Some images start as root and drop to another user later on. They may even receive the user id / group id as environment variables and dynamically switch to it (creating the user/group, if it doesn't exist).

We try to avoid it though. We'd rather not trust images to start as `root` and drop privileges. We'd rather start them with a non-privileged user.

-----------

What if you use `--user={{ matrix_user_uid }}:THE_dialout_GROUP_ID`? Maybe then the process would be creating files owned by `matrix_user_uid`, but it'd also be part of the `dialout` group and be able to access the modem?

benkuly:
I'm aware of all this. I meant that we could add the Unix user that are used for all docker container to the dialup group. Or is this no good approach?

spantaleev:
This could be a manual step for enabling this bridge (something like `usermod -a -G dialout matrix`).

It likely depends on the distro as well. Can you verify that doing that on the host distro is enough?

benkuly:
adding with usermod didn't work, because it does not give the container the group. I added it manually (see commit). But I think this only works on debian based distributions...

benkuly:
Its working :)

spantaleev:
From what I understand, the group name passed to `--group-add` must exist inside the Docker image. So I guess your Docker image contains `dialout` and it's fine.

However, if the host is a completely different distro (than the container) and it's another group that owns the modem device, maybe it won't match and it won't work..?

benkuly:
Yeah I think thats possible... I could add this fact to the docs, so that users of other distros know to create a group dialout on the host manually and change the group of the device to dialout...

spantaleev:
Oh, so you can manage the device's group manually? Perhaps they can just `chown` it to the `matrix` group then and it would just work..?

benkuly:
You are right. Thats also possible.

spantaleev:
Anything else we should do here? Or some doc updates remaining to be done (the `chown` thing)?

benkuly:
Yep, the docs. But currently I'm on vacation...

spantaleev:
No rush! :)

benkuly:
I added the docs :)

benkuly:
Sorry, I need to do more changes, because when the device get reseted or replugged, it does not work anymore.

spantaleev:
Ah.. Does the device disappear and then reappear?

Perhaps mounting all of `/dev` would be the solution. We may need to do it like this: `-v /dev:dev:slave`.

Check the Docker documentation on [bind mounts and bind propagation](https://docs.docker.com/storage/bind-mounts/#configure-bind-propagation).

benkuly:
Yeah my problem is, that the LTE-Stick needs a SoftReset every hour and then the connection to it does not work anymore. I'm going to try it with `slave`

benkuly:
Works for now. I think you can merge it. If I find more optimazations I will create a new PR.

spantaleev:
To make it consistent with all other bridge roles, can we also rename this role from `matrix-sms-bridge` to `matrix-bridge-sms`, so that it has a `matrix-bridge-` prefix?

benkuly:
Does that include all vars and the systemd services?

spantaleev:
We don't prefix vars with `matrix_bridge_` in any of the other roles, so I guess it's good as it is.

spantaleev:
Thanks a lot for contributing SMS bridging and for your patience troubleshooting everything! :+1:

benkuly:
And thanks for your help :) I learned many about ansible!

PR #516, opened 2020-05-23
Added matrix backup service

hungrymonkey:
Hi Spantaleev,

You have not elaborated on how you envision rsync backup should work. Before I attempt to test for more configurations for s3 backup, here is my progress.

Issue #467

Contains S3 backup

setup tag: setup-matrix-backup

Introduced these configurable variables
matrix_backup_enabled
matrix_backup_bucket
matrix_backup_bucket_endpoint
matrix_backup_bucket_awscli_docker_image_latest
matrix_backup_bucket_key_id
matrix_backup_bucket_key_secret
matrix_backup_rsync_target
matrix_backup_cron_day

Disabled on default
Default cron job will run weekly

Added compatible s3 services
Added simple backup acl
Added awscli utility container
Created environment file
Created one shot systemd service file

Tested matrix-backup/env-backup deployment
Tested matrix-backup cleanup
Tested IAM Role s3 backup access

Not tested
Credential Access


spantaleev:
Seems somewhat strange to put this in a systemd service, doesn't it? Or do you do that so we can get nice logging data in case something fails?

The name of the service is also pretty generic (`matrix-backup.service`), while it actually only deals with Postgres dumping and uploading to S3.

How about we rename it to `matrix-backup-postgres-s3.service`, along with everything relate to it.

And then someone could create a different one (like `matrix-backup-postgres-rsync`), which dumps the database and rsyncs it somewhere.. Or `matrix-backup-postgres-dump-over-ssh` or whatever.

It would be nice to have lots of separate backup methods (not just for Postgres, but for the other data as well), and users being able to choose the one that works best for them.

Of course, we can start with yours (targeting S3), but there may be others in the `matrix-backup` role in the future.

So I guess each one should have its own name, special variables that enable/disable it, subdirectories (for its tasks, templates), etc.

What do you think?

hungrymonkey:
@spantaleev , You never answered my previous thread about adding rsync. We can change the utilities or use a different feature set. Linux has too many backup options. I implemented s3 buckets because the conversation become easier when anybody has something to criticize. We can always add utilities to the same directory or create different files.  I would not mind creating rsync to one the same patch but I kinda need an idea on how you wanted it implemented.


I created systemd one shot services because I wanted to email failures one day. Systemd have nice logging services.

Can we agree on the directory structure?

```
matrix-docker-ansible-deploy/roles/matrix-backup/templates/matrix-backup-postgres-rsync.service
matrix-docker-ansible-deploy/roles/matrix-backup/templates/matrix-backup-postgres-s3.service
matrix-docker-ansible-deploy/roles/matrix-backup/tasks/setup-s3.yml
matrix-docker-ansible-deploy/roles/matrix-backup/tasks/setup-rsync.yml
```

Features I want to implement on the first patch

* Rsync/s3 backup
* Documentation
* Time stamped snapshots (My current implementation cannot take advance of COW)

Future Features
* Restore procedures compatible with https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/maintenance-postgres.md#upgrading-postgresql
* Different utilizes
* Encryption.
* Emailing failures.

hooger:
Why aren't you put the actual backup code into a shell script and run it by systemd? Logging could also work through the SyslogIdentifier and StandardOutput options of systemd.

I think implementing everything in a systemd service file makes it more difficult to extend the role.

hungrymonkey:
@hooger. Backup is somewhat opinionated in general. I wrote a quick, dirty, work backup script because I personally needed something to implemented quickly but I understand upstream have much larger demands in order to maintain it.  I decided to write a PR to start a conversation about it. To be honest, I am not sure how I should implement it without the drag of maintenance. I know the maintainer expressed interest with a script and cron job.

hungrymonkey:
Awesome, I will close my pr.
https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/868

hungrymonkey:
I will probably rewrite with that pr in mind

Issue #519, opened 2020-05-25
no proper video conference, whenever more than two participants are involved ...


andreas-dc:
Hello,

Once again I need to point out the excellence of the project first. I really like it.

I followed the configuration/installation script to the letter. My Jitsi section of vars.yml is:

matrix_jitsi_enabled: true
matrix_jitsi_enable_auth: true
matrix_jitsi_enable_guests: true
matrix_jitsi_jicofo_component_secret: xxx
matrix_jitsi_jicofo_auth_password: xxx
matrix_jitsi_jvb_auth_password: xxx
matrix_jitsi_jibri_recorder_password: xxx
matrix_jitsi_jibri_xmpp_password: xxx

All necessary ports are open. I do arrive at the Jitsi server directly (skipping Matrix) from regular home networks (without port restrictions). Even when I turn off the firewall of the docker host completely, the Jitsi server fails to serve a conference to more than 2 participants. There is no obvious crash, just all video screens (of all conference members) turning black. Whenever I go back to p2p, the session revitalises properly.

Do you have any idea why? Is there any chance to study logs? Since the services are not started normally (s6 is used), I can't find jicofo.log, prosody.log or jvb.log in their usual locations. Probably I simply don't recognise the path of the files being put. Maybe someone can point me into proper directions? Maybe there is some rerouting to any kind of docker logs happening. I'm new to ansible and docker.

Much thx for reading,
Andreas

andreas-dc:
It seems like setting "--log-driver=json-file" within the systemd service description on the docker host does the trick in logging. But so far the messages were unfortunately of no help.

spantaleev:
Logs go straight to journald. Use `journalctl -fu matrix-jitsi-SOMETHING`.

You don't need to do the `--log-driver=json-file` hack :)

If you're trying to use the Jitsi server from the local network of your Matrix server, you may need to take a look at #473.

Also, Jitsi reportedly has some problems on Firefox, so you may wish to try another browser (like Chromium/Chrome)

andreas-dc:
Much thx for your very quick response, spantaleev! "Journalctl" is enough for me, I immediately undid the hack. ;)

Did you mean, if I was trying to access the Jitsi server locally? Clients and server in the same network? This is not the case. The docker host (incl. Matrix and Jitsi containers) is located in a remote data centre, being accessible via public ip.

I've been using Opera, whenever it comes to video conferencing, for a while.

[...]

PR #521, opened 2020-05-27
Clarify documentation on port forwarding of Jitsi/TRP behind NAT

cnvandijk:
Depending on your firewall/NAT setup, incoming RTP packets may have as destination address the external IP of your firewall and not your Matrix/Jitsi server. This is important if your port forwarding rules are filtered by destination IP.

This caught me for a while, and may be the issue behind #519, so I figured it would be good to add it to the docs.


spantaleev:
`../roles/matrix-jitsi/defaults/main.yml#L147` will probably get outdated quickly.

Maybe it's better to link to the file and mention `matrix_jitsi_jvb_stun_servers`.

I'm not sure how one should correct the problem though? Can you elaborate some more on what you did to make it work for you?

spantaleev:
Thanks! :+1:

cnvandijk:
Our firewall filters not only based on port number but destination address. All other packets intended for this playbook come in with the IP defined in the DNS, but the RTP packets come in with the firewall's external IP. Until we come up with a better solution, we just added an extra rule for these packets.

For the record it is also possible to skip STUN completely and just hardcode the IP with something like ([docs](https://github.com/jitsi/ice4j/blob/master/doc/configuration.md#mapping-harvesters)):
```
org.ice4j.ice.harvest.NAT_HARVESTER_LOCAL_ADDRESS=<container-ip>
org.ice4j.ice.harvest.NAT_HARVESTER_PUBLIC_ADDRESS=<dns-ip>
```
but it requires putting the container on a fixed IP in the docker network.

andreas-dc:
Adding "DOCKER_HOST_ADDRESS=xxx" to ../matrix-docker-ansible-deploy/roles/matrix-jitsi/templates/jvb/env.j2 is the only thing, you need to let the jvb start-up script define

org.ice4j.ice.harvest.NAT_HARVESTER_LOCAL_ADDRESS=<container-ip>
org.ice4j.ice.harvest.NAT_HARVESTER_PUBLIC_ADDRESS=<docker host ip>

... by itself (making it obsolete to use aws, tcp harvesting or stun).

PR #522, opened 2020-05-27
fix for importing sqllite database

Tupsi:
the current version fails the import, because the volume for the media is missing. It still fails if you have the optional shared secret password provider is enabled, so that might need another mount. Commenting out the password provider in the hoimeserver.yaml during the run works as well.


spantaleev:
I guess mounting any other directory (besides `matrix_synapse_config_dir_path`) would also work.. I wonder if something like `/dev/null` works?

I'll merge it as it is for now. Thanks for finding and fixing it! :+1:

PR #524, opened 2020-05-27
Remove hardcoded paths to commands on the host machine

cnvandijk:
The `ExecStartPre` command shown below does not work on Ubuntu 18.04 LTS, as `chown` is located in `/bin` and not `/usr/bin` ([file link](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/f8157a94ef70c2eb0bb317dbccd84b78301be493/roles/matrix-dimension/templates/systemd/matrix-dimension.service.j2#L13)):
```
ExecStartPre=-/usr/bin/chown {{ matrix_dimension_user_uid }}:{{ matrix_dimension_user_gid }} {{ matrix_dimension_base_path }}/dimension.db
```
We need the full path here since systemd added path lookup to `ExecStart*` in v239, but Ubuntu 18.04 LTS is still on v237.

This PR replaces the hardcoded path with `/usr/bin/env chown`, as the location will vary between distros.

I may have gotten carried away, but I replaced all usage of hardcoded paths on the host machine with ansible variables pointing to `/usr/bin/env` by default (including `docker`). This may have been a bit excessive as in many cases (e.g. the ansible `shell` module) path lookup is anyway performed, but I felt it was better to use the same `matrix_host_command_*` variable. It would be great to get some feedback on what people think here.

### DIFF
--- a/roles/matrix-base/defaults/main.yml
+++ b/roles/matrix-base/defaults/main.yml
@@ -36,6 +36,7 @@ matrix_host_command_docker: "/usr/bin/env docker"
 matrix_host_command_sleep: "/usr/bin/env sleep"
 matrix_host_command_chown: "/usr/bin/env chown"
 matrix_host_command_fusermount: "/usr/bin/env fusermount"
+matrix_host_command_openssl: "/usr/bin/env openssl"
 
 matrix_ntpd_package: "ntp"
 matrix_ntpd_service: "{{ 'ntpd' if ansible_os_family == 'RedHat' or ansible_distribution == 'Archlinux' else 'ntp' }}"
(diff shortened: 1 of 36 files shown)


spantaleev:
Thank you for diligently fixing all such instances and making things more portable! :+1:

cnvandijk:
For the record, an alternative approach could be to use the Ansible `find` module to search for the executables on the host, and use the result in the unit file templates. We would need to get the environment variable `PATH` for where to search.

The advantage is removing the runtime overhead of searching with `env`, and more robustness because the paths are fixed. Downside is that we recreate `env` in Ansible.

Thoughts?

spantaleev:
Good point!

I was also wondering about the same thing yesterday and was looking at Ansible's [lookup plugins](https://docs.ansible.com/ansible/latest/plugins/lookup.html) for such a utility, but couldn't find anything.

I'm guessing the runtime performance of the playbook when using [find](https://docs.ansible.com/ansible/latest/modules/find_module.html) would be much worse though. It would also be more complicated to organize.

I guess the runtime performance of `/usr/bin/env` isn't so bad and it keeps things simple and easily overridable.

cnvandijk:
By runtime I meant when running the unit files, not when running Ansible.

Instead of the `find` module, something like the output of the `which` command (or a more posix-compliant alternative) could be used.

I think the robustness and security aspects are more important than the performance aspects, BTW.

PR #531, opened 2020-06-06
Add Docker Network for matrix-mautrix-telegram-db ...

dasTholo:
.. for Telegram Bridge…with Postgress

Postgres setup like:
```
matrix_mautrix_telegram_configuration_extension_yaml: |
  appservice:
    database: "postgres://XXX:XXX@matrix-postgres:5432/mxtg"
```

 will fail without the Dockernetwork


spantaleev:
Thank you! 👍🏻

PR #538, opened 2020-06-11
Fix missing conf for double puppeting in mx-puppet-slack

izissise:
Simple PR that copy the conf for double puppeting from mx-puppet-skype to mx-puppet-slack


spantaleev:
Looks good! Thanks! :+1:

Issue #469, opened 2020-04-24
Include mx-puppet-instagram and mx-puppet-discord


Hexalyse:
I just wanted to know if there is any plan to include the https://github.com/Sorunome/mx-puppet-instagram bridge to the playbook ?

And same question for the https://github.com/matrix-discord/mx-puppet-discord instead of the current Discord bridge (this one supports bridging guilds transparently, and also private and group DMs).

If not, I've never used Ansible before using this playbook, so is there a LOT of work to include such a bridge, or just a few variables to configure so that it can be automatically deployed and configured ? In which case, I might try to add mx-puppet-instagram.

spantaleev:
@rbelem said he'd be working on mx-puppet-discord support soon.

For mx-puppet-instagram, I haven't heard of anyone mentioning it. Please go ahead with adding support, if you're interested and would like to jump into Ansible. I guess you can copy some of the other mx-puppet bridge roles and start from there. If you've got any questions or troubles with development, we can provide help in our Matrix room (`#matrix-docker-ansible-deploy:devture.com`).

jdreichmann:
For mx-puppet-instagram, see #551 
I could already test it successfully locally

jdreichmann:
With #540 and #551 being merged and announced, i'd vote to close this issue (ping @spantaleev).

spantaleev:
Thanks for the ping! It's nice to see support for both bridges getting merged within the same week :)

PR #540, opened 2020-06-11
Add mx-puppet-discord

izissise:
The one last issue is that there is no official docker image, the one specified is 3 months old, and gives no guarantees on what's in it 

I opened an issue on mx-puppet-discord
https://github.com/matrix-discord/mx-puppet-discord/issues/80

#469 
https://github.com/matrix-discord/mx-puppet-discord

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,8 @@ This playbook sets up your server using the following Docker images:
 
 - [sorunome/mx-puppet-twitter](https://hub.docker.com/r/sorunome/mx-puppet-twitter) - the [mx-puppet-twitter](https://github.com/Sorunome/mx-puppet-twitter) bridge to [Twitter](https://twitter.com) (optional)
 
+- [pneumaticat/mx-puppet-discord](https://hub.docker.com/r/pneumaticat/mx-puppet-discord) - the [mx-puppet-discord](https://github.com/matrix-discord/mx-puppet-discord) bridge to [Disocrd](https:/discordapp.com) (optional)
+
 - [turt2live/matrix-dimension](https://hub.docker.com/r/turt2live/matrix-dimension) - the [Dimension](https://dimension.t2bot.io/) integrations manager (optional)
 
 - [jitsi/web](https://hub.docker.com/r/jitsi/web) - the [Jitsi](https://jitsi.org/) web UI (optional)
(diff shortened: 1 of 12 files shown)


Cadair:
A while ago I hacked together something that builds the docker container locally as part of the playbook: https://github.com/Cadair/matrix-docker-ansible-deploy/blob/dangersalad/roles/matrix-bridge-mx-puppet-discord/tasks/setup_install.yml#L11L34

Not sure if that's a better solution or not.

spantaleev:
We do such "self-building" for various roles. We only currently do it for other architectures (non-amd64) or when explicitly requested.

Learn more here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/self-building.md

If we can, it'd be better if we use a ready-made image, instead of building as part of running the playbook. Would be nice if we can get a more up-to-date image, perhaps by poking upstream :)

spantaleev:
What does logging in as a user vs a bot refer to?

If it works well enough, I guess we could merge this.

izissise:
Sadly, the problem of the source image remains, the current value is the most downloaded mx-puppet-discord image on docker hub.
there is an open issue here https://github.com/matrix-discord/mx-puppet-discord/issues/80

One solution could be enabling image building by default for this bridge

jdreichmann:
@spantaleev Logging in a user is against discords ToS, but allows to bridge DMs and manage friends via matrix.

jdreichmann:
I should add that the version of the docker image the playbook currently uses is outdated enough that the bridge fails to log in.
I succeeded in actually using the bridge by creating a newer image at https://github.com/matrix-discord/mx-puppet-discord/issues/80#issuecomment-650720309, the old version just fails with stacktraces when logging in (probably because discord updated parts of their API)

jdreichmann:
https://github.com/matrix-discord/mx-puppet-discord/issues/80 seems to be resolved, we could point the playbook to `sorunome/mx-puppet-discord:latest` and retest with that

izissise:
Awesome, I'll make the change :)

izissise:
Should be good :)

jdreichmann:
Just tested with the image, everything seems to work

izissise:
I removed the merge commit and rebased from master branch

spantaleev:
Thanks a lot to @izissise for contributing this bridge and for @jdreichmann for testing and pushing for an official image! :+1:

PR #541, opened 2020-06-11
Update matrix-sms-bridge

benkuly:


roles/matrix-bridge-sms/defaults/main.yml
```diff
@@ -3,8 +3,8 @@
 
 matrix_sms_bridge_enabled: true
 
-matrix_sms_bridge_docker_image: "folivonet/matrix-sms-bridge:latest"
-matrix_sms_bridge_docker_image_force_pull: "{{ matrix_sms_bridge_docker_image.endswith(':latest') }}"
+matrix_sms_bridge_docker_image: "folivonet/matrix-sms-bridge:0.2.1.RELEASE"
+matrix_sms_bridge_docker_image_force_pull: "{{ matrix_sms_bridge_docker_image.endswith(':0.2.1.RELEASE') }}"
```
spantaleev, line 7: Why change `matrix_sms_bridge_docker_image_force_pull`? It's probably fine the way it was `:latest` being force-pulled as it often changes where it points to. Other tagged-images don't need force-pulling, as they're static.

benkuly: Thanks, dind't know that and changed it in d49ee51.

spantaleev:
Thank you! :+1:

PR #544, opened 2020-06-16
nginx Variables worker_

schneif2:
Just a little change using variables for worker_processes and worker_connections. 

We had some performance issues with nginx using around 1000 clients. Specially when all tried to sync, download images and collect chat history.

Made it changeable in configuration for each setup with default of current static values.

### DIFF
--- a/roles/matrix-nginx-proxy/defaults/main.yml
+++ b/roles/matrix-nginx-proxy/defaults/main.yml
@@ -6,8 +6,8 @@ matrix_nginx_proxy_enabled: true
 matrix_nginx_proxy_docker_image: "nginx:1.19.0-alpine"
 matrix_nginx_proxy_docker_image_force_pull: "{{ matrix_nginx_proxy_docker_image.endswith(':latest') }}"
 
-matrix_nginx_proxy_proxy_synapse_worker_processes: 1
-matrix_nginx_proxy_proxy_synapse_worker_connections: 1024
+matrix_nginx_proxy_worker_processes: 1
+matrix_nginx_synapse_worker_connections: 1024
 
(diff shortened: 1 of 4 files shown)


roles/matrix-nginx-proxy/defaults/main.yml
```diff
@@ -6,6 +6,9 @@ matrix_nginx_proxy_enabled: true
 matrix_nginx_proxy_docker_image: "nginx:1.19.0-alpine"
 matrix_nginx_proxy_docker_image_force_pull: "{{ matrix_nginx_proxy_docker_image.endswith(':latest') }}"
 
+matrix_nginx_proxy_proxy_synapse_worker_processes: 1
```
spantaleev, line 9: Looks great! And thank you for this contribution! :+1: 

Why put `_synapse_` in the variable name when it affects all matrix-nginx-proxy vhosts?

Perhaps we should rename these variables to `matrix_nginx_proxy_worker_processes` and `matrix_nginx_synapse_worker_connections`.

schneif2: Yes you are right. I will rename the variables later and provide contribution again.
Thank you and all other contributors for this great project.

spantaleev:
`matrix_nginx_synapse_worker_connections` still seems to remain in use? Would you be changing that as well or would you like me to help out with these renames?

spantaleev:
I'm closing this, since it's been open for a long time.. and it currently has merge conflicts (because we no longer have a `matrix-riot-web` role).

If you'd like to get this merged, please fix it up and send a new Pull Request!

Thanks for your help!

PR #547, opened 2020-06-21
Small fixes in DNS documentation

AlwaysLivid:


aaronraimist:
That might be how you set it up but that isn't what the documentation tells you to do

AlwaysLivid:
> That might be how you set it up but that isn't what the documentation tells you to do or what the default setup is.

I think that there isn't a clear enough differentiation between `<your-domain>` of the `Target` row and the `<your-domain>` further down below.

spantaleev:
`<your-domain>` refers to your base domain. The default `vars.yml` file (`examples/host-vars.yml`) also tells you to use a base (bare) domain in `matrix_domain`.

If you've used `matrix.DOMAIN` there, then things will change like you propose. However, most people don't wish to do that, as it makes MXIDs uglier (`@user:matrix.DOMAIN`, instead of `@user:DOMAIN`).

----------

What you're proposing can also be accomplished while retaining pretty usernames, but it requires:
- using a bare domain in `matrix_domain`
- tweaking other variables - `matrix_server_fqn_riot`, `matrix_server_fqn_dimension`, `matrix_server_fqn_jitsi` (example: `matrix_server_fqn_riot: "riot.matrix.{{ matrix_domain }}"`)

PR #549, opened 2020-06-24
Added config hint for SMS verification

ctwelve:
After some excellent help by @aaronraimist, and his suggestion that we document this particular use-case, I thought I'd give it a go. It's trivially easy to make SMS happen but finding the resultant options to twiddle wasn't intuitive.

Since I suspect many people would very much want to enable this option, I here suggest a small addition to the documentation to point the way.


spantaleev:
Thank you for documenting this! :+1:

PR #550, opened 2020-06-26
Improve the example Traefik config.

julianfoad:
Use playbook variables instead of textual place-holders for the domain names.


spantaleev:
Makes better sense to those familiar with the Ansible code. I hope it's not confusing for others. Let's see how it goes!

Thank you! :+1:

PR #551, opened 2020-06-27
Add mx-puppet-instagram

jdreichmann:
Adds https://github.com/Sorunome/mx-puppet-instagram.

Still needs documentation and testing

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -394,6 +394,37 @@ matrix_mx_puppet_slack_login_shared_secret: "{{ matrix_synapse_ext_password_prov
 
 ######################################################################
 #
+# matrix-bridge-mx-puppet-instagram
+#
+######################################################################
+
+# We don't enable bridges by default.
+matrix_mx_puppet_instagram_enabled: false
+
+matrix_mx_puppet_instagram_container_image_self_build: "{{ matrix_architecture != 'amd64'}}"
+
+matrix_mx_puppet_instagram_systemd_required_services_list: |
+  {{
+    ['docker.service']
+    +
+    (['matrix-synapse.service'] if matrix_synapse_enabled else [])
+  }}
+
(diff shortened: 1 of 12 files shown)


group_vars/matrix_servers
```diff
@@ -392,6 +392,37 @@ matrix_mx_puppet_slack_login_shared_secret: "{{ matrix_synapse_ext_password_prov
...
+matrix_mx_puppet_instagram_container_image_self_build: "{{ matrix_architecture != 'amd64'}}"
+
+matrix_mx_puppet_instagram_systemd_required_services_list: |
+  {{
+    ['docker.service']
+    +
+    (['matrix-synapse.service'] if matrix_synapse_enabled else [])
+  }}
+
+matrix_mx_puppet_instagram_appservice_token: "{{ matrix_synapse_macaroon_secret_key | password_hash('sha512', 'mxslk.as.tok') | to_uuid }}"
```
spantaleev, line 413: The `mxslk.as.tok` key here and `mxslk.hs.tok` in `matrix_mx_puppet_instagram_homeserver_token` conflict with the keys used for the Slack bridge. If both are enabled, Synapse will complain that tokens are not unique.

Let's use a different key for this bridge. It should be a relatively short string to be compatible with various versions of Ansible (e.g. `insta.as.tok` and `insta.hs.tok`).

jdreichmann: Adressed in 2073cd6

jdreichmann:
Added some documentation and thanks @spantaleev for the hint with the colliding access-tokens, i overlooked that

spantaleev:
Looks good! Think we're ready to merge this or would you like to test it some more first?

jdreichmann:
I tested it on my HS and then marked it ready for review so other can test if they want, but from my side i consider it ready for merge.

spantaleev:
That's good enough! Thanks a lot for increasing the number of bridges we support! :+1:

PR #554, opened 2020-06-28
Update Jitsi to stable-4627-1

teutat3s:
- Update jitsi containers to latest release [stable 4627-1](https://github.com/jitsi/docker-jitsi-meet/releases)
- Notable changes: https://github.com/jitsi/docker-jitsi-meet/compare/stable-4548-1...stable-4627-1

The update should not contain any breaking changes.


spantaleev:
Looks good! Thanks! :+1:

SuperSandro2000:
This version number should probably be put in a variable cause it is the same for every image.

PR #555, opened 2020-06-29
Add mx-puppet-twitter

tulir:


spantaleev:
Looks good! Thanks a lot for increasing the number of bridges we support! +1

PR #557, opened 2020-07-01
Add mx-puppet-steam

izissise:
Successfully tested and working

https://github.com/icewind1991/mx-puppet-steam

It's basically the mx-puppet-discord commit with discord replaced to steam and the right docker image set.


spantaleev:
Awesome! Thank you! :+1:

PR #568, opened 2020-07-04
Run all API self checks in check_mode

xshadow:
This PR allows running the matrix playbooks also in check_mode without failure due to undefined variables. 

It will enforce the execution of the uri module in check_mode, so all necessary variables are gathered.


spantaleev:
Thank you! :+1:

PR #569, opened 2020-07-05
Fixed Steam link

AlwaysLivid:

PR #571, opened 2020-07-05
Fixed repository link for mx-puppet-steam

AlwaysLivid:
- https://github.com/icewind1991/mx-puppet-steam is the link that's referenced by the documentation.
- The previous link, https://github.com/matrix-steam/mx-puppet-steam, is invalid/inaccessible to the public.


AlwaysLivid:
I made that change because I noticed that it effectively broke the playbook as soon as I tried to use it. However, it was pointed out to me in the Matrix chat that the documentation is also inaccurate.

spantaleev:
Thanks! :+1: 

This was indeed breaking the playbook when self-building was used for this role (for architectures other than `amd64`).

The docs page seems correct now (since your #569 fix and my b6b5435746edd9167 fix).

PR #574, opened 2020-07-10
Remove unused validate_config.yml, since it causes ansible warnings

xshadow:
Remove unused validate_config.yml from matrix-base.yml. It only causes ansible warnings.


spantaleev:
Thanks! :+1:

PR #578, opened 2020-07-13
Fix default SSL path for federation api in docs

### LINKED ISSUE #577: javax.net.ssl.SSLPeerUnverifiedException: Certificate for <matrix.<your-domain>> doesn't match any of the subject alternative names: [<your-domain>]

jmastr:
One could also remove the two variables from the docs completely,
because they are set by the playbook automatically.

Error: javax.net.ssl.SSLPeerUnverifiedException: Certificate for
<matrix.<your-domain>> doesn't match any of the subject alternative
names: [<your-domain>]

Fixes #577 (Github Issue).


spantaleev:
Thank you! :+1: 

Indeed, these match the defaults and they're not necessary. I think the intention was to tell people how to use a custom certificate (which might be named differently or be in another location). It's probably better to keep it for reference.

PR #580, opened 2020-07-13
importing-postgres.md: Add 'Troubleshooting' w/workaround for ownership

bertiebaggio:
**Note**: This PR addresses documenting the workaround only, it may be worth fixing the underlying issue.

If a Postgres dump contains ALTER TABLE ... OWNER_TO <username>
statements which set the owner to a username different from
'synapse' the post Postgres import task will fail complaining
about lack of role.

Changing the matrix_postgres_connection_username group var has no
effect. However, the ALTER TABLE statements (and accompanying comments)
can be rewritten to change the username to 'synapse', which permits the
import task to succeed.

From a sample of 1, having the owner set in this was causes no
discernable side effects on the homeserver.


spantaleev:
Wow, wonderfully written! Thanks a lot for this diligence! :+1:

bertiebaggio:
Thank you :-) It was my pleasure to contribute to such a cool and useful project!

PR #588, opened 2020-07-17
Fix issues building ma1sd Docker image

mattcen:
The tag format used in the `ma1sd` repo have change. Versions no longer
start with 'v', and when building for non-amd64, we also need to strip
off the '-$arch' bit from the Docker image name.

Further, when building the .jar file, `ma1sd` currently names the .jar
based on the project's directory, which we call 'docker-src'. This means
other parts of the `ma1sd` build can't find the .jar file. Remedy this
by ensuring that the dir is called `docker-src/ma1sd`.
(I've proposed a patch to `ma1sd` on PR https://github.com/ma1uta/ma1sd/pull/55 to fix this: https://github.com/ma1uta/ma1sd/pull/55/commits/7b992c1060c58e2a46c814ce76a8fcc9435d89c2)

### DIFF
--- a/roles/matrix-ma1sd/defaults/main.yml
+++ b/roles/matrix-ma1sd/defaults/main.yml
@@ -11,7 +11,7 @@ matrix_ma1sd_docker_image: "ma1uta/ma1sd:2.4.0-{{ matrix_ma1sd_architecture }}"
 matrix_ma1sd_docker_image_force_pull: "{{ matrix_ma1sd_docker_image.endswith(':latest') }}"
 
 matrix_ma1sd_base_path: "{{ matrix_base_data_path }}/ma1sd"
-matrix_ma1sd_docker_src_files_path: "{{ matrix_ma1sd_base_path }}/docker-src"
+matrix_ma1sd_docker_src_files_path: "{{ matrix_ma1sd_base_path }}/docker-src/ma1sd"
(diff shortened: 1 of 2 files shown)


mattcen:
Ah, there's another issue here. When the Docker image is built with Gradle, it's tagged as `ma1uta/ma1sd:2.4.0`, but the playbook looks for an image tagged `ma1uta/ma1sd:2.4.0-amd32`. I'm not sure of the best way to fix this without breaking the situation where the image is pulled from Docker Hub, rather than built, and then *will* have the architecture as part of the tag.

spantaleev:
Can we add another task when self-building is done, so that we'd retag `ma1uta/ma1sd:VERSION` to `{{ matrix_ma1sd_docker_image }}`?

mattcen:
Yeah, that'd do the trick. I'm not sure off the top of my head how to ensure all dependencies are met so the task runs at the right times etc. (my Ansible-fu is a little weak), but I'll see if I can work it out.

mattcen:
Looks like my latest commit correctly fixes the misnamed image.

spantaleev:
I've done a few fixes and merged this! Thanks a lot for keeping an eye on it and patching things up! :+1:

mattcen:
Thanks for the fixes and merge! I wasn't sure about the indentation on that `when` block; my editor kept trying to fix it but it was unclear whether the indents would change the behaviour!
I'll go ahead and test the Element changes now :)

PR #594, opened 2020-07-19
Update configuring-dns.md

ronanduddy:
Adding a comment with regard to DNS propagation.

This addition may be a useful explanation for those who cannot reach their installation right away and therefore lessening any frustrations that may be felt during the installation.


spantaleev:
Looks good! Thank you! :+1:

ronanduddy:
Great :+1:

PR #597, opened 2020-07-22
Nginx proxy headers

TwoTwenty:
These headers add some security and I believe they are used by app.element.io and chat.mozilla.com.

I have tested them roughly and they do not seem to break our internal matrix/element/jitsi/dimension services.


spantaleev:
Looks good! Thank you! :+1:

rebruCH:
Wouldn't it be a better idea for the dimension to allow the URL of the home server in relation to "x-frame"?

Because I have the problem that in the Desktop Client element the default dimensions are still loaded and on the web it aborts because of the violation of the x-frame option.

```
Refused to display 'https://dimension.<matrixdomain>/element?scalar_token=<token>&room_id=<roomid>&room_name=<roomname>&theme=light' in a frame because it set 'X-Frame-Options' to 'sameorigin'.
```

PR #601, opened 2020-07-23
Fix capability names for synapse-admin

thedanbob:


spantaleev:
Thanks! :+1:

Seems to have worked with the `CAP_` prefix for me, but maybe certain Docker versions are not so permissive.

thedanbob:
Oh, I bet you're right. I forgot I had switched to docker.io because docker-ce wasn't available for the distro I was running at the time (Ubuntu). Since then I switched to Debian, but being Debian the docker.io package is a whole major version out of date. I've switched back to docker-ce now, but at least these caps now work for anyone else in the same situation.

spantaleev:
I see, sounds reasonable! We'd better go for the more standard and generally supported way of doing things!

PR #606, opened 2020-07-25
Update configuring-playbook-ma1sd.md

rebruCH:
Add an example to open all domains for registration in relation to black- and whitelists.


spantaleev:
Thank you! :+1:

PR #607, opened 2020-07-26
Added config sample for synapse admin with traefik 2

matthiasjo:
I used the config sample at https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-own-webserver.md and extended it for the new synapse admin container, so that it would be available at https://matrix.DOMAIN/synapse-admin.
Please review and edit this PR.

### DIFF
--- a/docs/configuring-playbook-synapse-admin.md
+++ b/docs/configuring-playbook-synapse-admin.md
@@ -57,6 +57,6 @@ matrix_synapse_admin_container_extra_arguments:
     # (The 'default' certificate resolver must be defined in Traefik config)
     - '--label "traefik.http.routers.matrix-synapse-admin.tls.certResolver=default"'
 
-    # The Synapse Admin container uses port 8766 internally
-    - '--label "traefik.http.services.matrix-synapse-admin.loadbalancer.server.port=8766"'
+    # The Synapse Admin container uses port 80 by default
+    - '--label "traefik.http.services.matrix-synapse-admin.loadbalancer.server.port=80"'
 ```
--- a/docs/configuring-playbook-synapse-admin.md
+++ b/docs/configuring-playbook-synapse-admin.md
@@ -6,6 +6,7 @@ It's a web UI tool you can use to **administrate users and rooms on your Matrix 
 
 See the project's [documentation](https://github.com/Awesome-Technologies/synapse-admin) to learn what it does and why it might be useful to you.
 
+
 ## Adjusting the playbook configuration
 
 Add the following configuration to your `inventory/host_vars/matrix.DOMAIN/vars.yml` file:
@@ -14,6 +15,7 @@ Add the following configuration to your `inventory/host_vars/matrix.DOMAIN/vars.
 matrix_synapse_admin_enabled: true
 ```
 
+
 ## Installing
 
 After configuring the playbook, run the [installation](installing.md) command again:
@@ -21,6 +23,7 @@ After configuring the playbook, run the [installation](installing.md) command ag
 ```
 ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start
 ```
+
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


docs/configuring-playbook-synapse-admin.md
```diff
@@ -6,7 +6,6 @@ It's a web UI tool you can use to **administrate users and rooms on your Matrix
 
 See the project's [documentation](https://github.com/Awesome-Technologies/synapse-admin) to learn what it does and why it might be useful to you.
 
-
```
spantaleev, line 9: Can you restore this and the other removed lines?

docs/configuring-playbook-synapse-admin.md
```diff
@@ -24,11 +22,38 @@ After configuring the playbook, run the [installation](installing.md) command ag
...
+# Don't bind any HTTP or federation port to the host
+# (Traefik will proxy directly into the containers)
+matrix_synapse_admin_container_http_host_bind_port: ""
+
+matrix_synapse_admin_container_extra_arguments:
+    # May be unnecessary depending on Traefik config, but can't hurt
+    - '--label "traefik.enable=true"'
+
+    # The Synapse Admin container will only receive traffic from this subdomain and path
+    - '--label "traefik.http.routers.matrix-synapse-admin.rule=(Host(`{{ matrix_server_fqn_matrix }}`) && Path(`{matrix_synapse_admin_public_endpoint}`))"'
```
spantaleev, line 49: Maybe `{matrix_synapse_admin_public_endpoint}` needs to be `{{ matrix_synapse_admin_public_endpoint }}`?

docs/configuring-playbook-synapse-admin.md
```diff
@@ -24,11 +22,38 @@ After configuring the playbook, run the [installation](installing.md) command ag
...
+
+    # (Define your entrypoint)
+    - '--label "traefik.http.routers.matrix-synapse-admin.entrypoints=web-secure"'
+
+    # (The 'default' certificate resolver must be defined in Traefik config)
+    - '--label "traefik.http.routers.matrix-synapse-admin.tls.certResolver=default"'
+
+    # The Synapse Admin container uses port 8766 internally
+    - '--label "traefik.http.services.matrix-synapse-admin.loadbalancer.server.port=8766"'
```
spantaleev, line 58: I may be confusing things, but.. shouldn't this be port `80`?

The Synapse Admin container binds to port `80` internally. We expose it on port `8766` outside when `matrix_nginx_proxy_enabled: false` (in that case we set `matrix_synapse_admin_container_http_host_bind_port: '127.0.0.1:8766'` via a statement in `group_vars/matrix_servers`).

Your example configuration explicitly disables such exposure via `matrix_synapse_admin_container_http_host_bind_port: ""`, so I don't see how port `8766` will come into play at all.

matthiasjo: I used this [comment](https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/296#issuecomment-562819174) as a template. When traefik is used `matrix_nginx_proxy_enabled: false`. Thats why synapse admin is on port 8766 internally and traefik pointed to this port.

docs/configuring-playbook-synapse-admin.md
matthiasjo: you're correct. changed it :)

docs/configuring-playbook-synapse-admin.md
matthiasjo: sorry. the lines are restored

docs/configuring-playbook-synapse-admin.md
spantaleev: It says "Traefik maps to internal port of Synapse/nginx container"

In the synapse-admin case, the internal port of synapse-admin is 80, not 8766.

I don't think pointing Traefik to 8766 will work. Have you tested this setup?

matthiasjo: I've tested it and it worked. Thats strange bc it shouldn't have. You are corrrect. `docker ps` shows sysnapse-admin running on port 80. I guess that bc I disabled the exposure with `matrix_synapse_admin_container_http_host_bind_port: ""` I'll change the traefik port.

spantaleev:
Thank you! :+1:

PR #611, opened 2020-07-29
Make ansible check mode runs silent, for non idempotent tasks

xshadow:
Make ansible check mode runs silent, for non idempotent tasks:

    do not enforce service starts and stops in check_mode

### DIFF
--- a/roles/matrix-common-after/tasks/start.yml
+++ b/roles/matrix-common-after/tasks/start.yml
@@ -1,21 +1,23 @@
 ---
 
-- name: Ensure systemd reloaded
+- name: Ensure systemd is reloaded
   service:
     daemon_reload: yes
 
-- name: Ensure Matrix services stopped
+- name: Ensure Matrix services are stopped
   service:
     name: "{{ item }}"
     state: stopped
   with_items: "{{ matrix_systemd_services_list }}"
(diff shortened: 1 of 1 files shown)


spantaleev:
The `ansible_check_mode` change sounds reasonable.

Suppressing reporting changes when things really change doesn't sound like a good idea though. If a new (`:latest`) image gets force-pulled, that is indeed a change that Ansible did. It doesn't make sense to hide it from the user. The same probably applies to creating the SSL directory. If its permissions change due to certbot running (I think they do sometimes), we should address that instead.

xshadow:
1) Regarding the :latest image I agree (my system seemed to be out of date)

2) I kindly disagree. Imagine you run ansible regulary against your system to check if something has changed it should report on changes on which you have to interfere. Those SSL files in the given directories will change and will never have these open permissions. But maybe my setup is too specific. To set those permissions recurse will almost always show a change.

xshadow:
Then I will only introduce the non-discusworthy changes :)

spantaleev:
You're right.. SSL directory chowning probably needs to happen in some other way..

I'll merge the straightforward things for now! Thank you for fixing check mode! :+1:

PR #614, opened 2020-07-29
Enable ma1sd hashing by default

vractal:


spantaleev:
Looks good! Thank you! :+1: 

Probably fixes #503 and #505.

PR #615, opened 2020-07-29
Add a secondary option on how to get access token

vaivars:
Potentially the CURLing could be a better approach, since then you don't leave spare sessions hanging.

### DIFF
--- a/docs/configuring-playbook-dimension.md
+++ b/docs/configuring-playbook-dimension.md
@@ -37,7 +37,8 @@ We recommend that you create a dedicated Matrix user for Dimension (`dimension` 
 Follow our [Registering users](registering-users.md) guide to learn how to register **a regular (non-admin) user**.
 
 You are required to specify an access token (belonging to this new user) for Dimension to work.
-To get an access token for the Dimension user, follow these steps:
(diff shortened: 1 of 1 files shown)


docs/configuring-playbook-dimension.md
```diff
@@ -46,6 +47,17 @@ To get an access token for the Dimension user, follow these steps:
...
 4. Close the private browsing session. **Do not log out**. Logging out will invalidate the token, making it not work.
 
+*With CURL* 
+
+```
+curl -X POST --header 'Content-Type: application/json' -d '{
+    "identifier": { "type": "m.id.user", "user": "YourDimensionUsername" },
+    "password": "YourDimensionPassword",
+    "type": "m.login.password"
+}' 'http://localhost:8008/_matrix/client/r0/login'
```
spantaleev, line 57: Most people run with `matrix_nginx_proxy_enabled: true` (the default). In that case, Synapse is not exposed to `localhost:8008`. We'd better change this to `https://matrix.DOMAIN`

vaivars: Good point, made the change.

spantaleev:
Looks good! Thank you! :+1:

PR #620, opened 2020-08-04
Allow self-build images for Synapse-Admin

NachvollCiba:


spantaleev:
Thank you! :+1: 

I've updated the docs to reflect this!

PR #623, opened 2020-08-05
Python requires docker package and other small improvements

merklaw:
Hello all.

Having never used Ansible, the guide was a great help in getting a Matrix server up and running. I installed it locally to experiment and only ran into a few issues. I want to share my experience for anyone else who might run into these issues.

- The playbook requires the docker package.

- ma1sd is the 2nd most memory hungry container, so if you don't need the features you can disable it.

- Being installed locally, i used a self-signed SSL certificate. You have to add a security exception to Matrix by visiting `matrix.<your-domain>` (as well as Element of course).

I have a few questions, which I hope someone can answer:

- I am new to Ansible, so is it safe to disable ma1sd? Are there other configuration options that are depending on it?
- I had to run the ansible commands with sudo. I was thinking about adding either "sudo" in front of the command or adding "as superuser" above it (in installing.md). What do you think?

Please review and comment. You are free to merge if there are no problems.

### DIFF
--- a/docs/configuring-playbook-client-element.md
+++ b/docs/configuring-playbook-client-element.md
@@ -6,7 +6,7 @@ If that's okay, you can skip this document.
 
 ## Disabling Element
 
-If you'd like for the playbook to not install Element (or to uninstall the previously installed Element), you can disable it in your configuration file (`inventory/host_vars/matrix.<your-domain>/vars.yml`):
+If you'd like for the playbook to not install Element (or to uninstall it if it was previously installed), you can disable it in your configuration file (`inventory/host_vars/matrix.<your-domain>/vars.yml`):
 
 ```yaml
 matrix_client_element_enabled: false
--- a/docs/configuring-playbook-client-element.md
+++ b/docs/configuring-playbook-client-element.md
@@ -6,7 +6,7 @@ If that's okay, you can skip this document.
 
 ## Disabling Element
 
-If you'd like for the playbook to not install (or to uninstall the previously installed Element), you can disable it in your configuration file (`inventory/host_vars/matrix.<your-domain>/vars.yml`):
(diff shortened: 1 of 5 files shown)


spantaleev:
Well, the playbook installs `python-docker` by default, so installing it manually shouldn't be necessary.

Indeed, some distros don't have Python or there's a misunderstanding between which Python version Ansible would use, which could cause issues. If Python 3 is used, we try to install `python3-docker`.

See here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/fffe9da5a9a750eb9fffeca320f4225d835df722/roles/matrix-base/tasks/server_base/setup_debian.yml#L40

This may be what was causing your issue. You probably need to play with `ansible_python_interpreter`. See here: https://docs.ansible.com/ansible/latest/reference_appendices/python_3_support.html

---------

The other documentation adjustments look good!

merklaw:
Thank you for the feedback. I realize something now. I had set `matrix_docker_installation_enabled: false` because I installed docker myself, but that disables the installation of `docker-python` (I am running CentOS). What do you think about moving that line you cited up to the `name: Ensure APT packages are installed` section (in the 4 different server_base files) instead? That would make certain the package is installed regardless.

merklaw:
Wait, nevermind. That might fail because if `matrix_docker_installation_enabled: false` is set, it doesn't add the Docker repo anyways.

I added a note to `roles/matrix-base/defaults/main.yml`.

spantaleev:
Ah, it all makes sense now then!

Looks good! Merging as is! Thank you! :+1:

tomkel:
I remember there being some concern about everything not working 100% correctly with no ID server configured. I think it was on the matrix-synapse issues ?

PR #627, opened 2020-08-08
Update README.md -> 'many times' -> 'more than once'

MatthewCroughan:
Stating "many times" makes me think there is a finite amount of times I can run it before it *will* start causing trouble. But this is not true. So just state it can me ran more than once without causing trouble :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ This is similar to the [EMnify/matrix-synapse-auto-deploy](https://github.com/EM
 
 - this one installs and hooks together **a lot more Matrix-related services** for you (see above)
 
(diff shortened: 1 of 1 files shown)


spantaleev:
Thinking about it **too much**, "can be executed more than once" makes me think of simultaneous executions being supported (which is likely not true.. and doesn't even make sense).

I guess I should refrain from thinking so much about it though.. Your proposed change is probably better in most cases. Thank you! :+1:

MatthewCroughan:
Yes, English is a terrible language. If people fork this and execute it more than once, I do not know what they expect to achieve, so we can't support that case :D

PR #628, opened 2020-08-09
Update CORS for ma1sd

ctwelve:
Even with the v2 updates listed in #503 and partially addressed in #614, this is still needed to enable identity services to function with Element Desktop/Web. Testing on multiple clients with a clean config has confirmed this, at least for my installation.

This could be a case of a stubborn client, however making the change seems to fix the user experience. Here's the error in latest release of Element Desktop on macOS:

image


spantaleev:
Makes sense! Thank you! :+1:

merklaw:
Is it safe to put a * (wildcard) in Access-Control-Allow-Origin? I am asking because I don't know the answer and it sounds risky.

ctwelve:
In this case, yes. The reason has to do with how the relevant web service is called by Element. It isn’t a risk so long as the server software isn’t compromised, and to be honest if it is, this wouldn’t add any security anyway.

PR #629, opened 2020-08-09
Undo ill-advised change

ctwelve:
In #628 I proposed a CORS change that turns out not to be the root of the issue. Caffeine-addled diagnosis leads to sloppy thinking, and this change should be reverted. In fact, if left it will cause problems for new installations.

### DIFF
--- a/roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2
+++ b/roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2
@@ -64,7 +64,6 @@
 
 		proxy_set_header Host $host;
 		proxy_set_header X-Forwarded-For $remote_addr;
-		add_header Access-Control-Allow-Origin *;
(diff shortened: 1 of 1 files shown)


spantaleev:
I'm fine with reverting this as well. I'm curious though, why would it be causing problems?

ctwelve:
Well, it turns out that something goofy about my total setup was masking the underlying problem, so I misdiagnosed the problem. It's probably a deeper issue with ma1sd, and I'm now getting M_UNAUTHORIZED in my setup.

ctwelve:
So what ended up happening is two different values got sent for CORS, but only sometimes, and any browser-based client (so, Element Desktop too) freaks out when that happens.

spantaleev:
I see! Thanks for the explanation! I've merged the revert! :+1:

ctwelve:
Thanks! I *think* the issue is regarding ma1sd's registration `policies` block so I'm pursuing that right now. We'll see!

PR #640, opened 2020-09-07
Change anonymized to general

asalminen:
Change anonymized to more proper term as server vice this is not anonymized. Server name is the first parameter that is collected. And if server happens to be for individual use these statistics would be at personal level without any anonymising.


spantaleev:
Sounds reasonable! Thanks! :+1:

PR #647, opened 2020-09-12
Update configuring-playbook-jitsi.md

banym:
modified example for jitsi ldap authentication to show all available variables.


spantaleev:
Thank you! :+1:

PR #649, opened 2020-09-14
Use the same naming convention as the other mx-puppet suite.

scottcrossen:
For consistency


spantaleev:
I think "bridge" sounds better, but.. if this is consistent with the others, let's go for consistency ;)

Thank you! :+1:

PR #650, opened 2020-09-15
New docker image appservice-slack

dwiegreffe:
The used Docker image is no longer maintained and is outdated so it does not work anymore. 

Before upgrading existing installations, the data in the appservice-slack folder on the host server must be deleted, as the files are not compatible with the new version.

### DIFF
--- a/roles/matrix-bridge-appservice-slack/defaults/main.yml
+++ b/roles/matrix-bridge-appservice-slack/defaults/main.yml
@@ -3,7 +3,7 @@
 
 matrix_appservice_slack_enabled: true
 
-matrix_appservice_slack_docker_image: "cadair/matrix-appservice-slack:cadair"
+matrix_appservice_slack_docker_image: "matrixdotorg/matrix-appservice-slack"
 matrix_appservice_slack_docker_image_force_pull: "{{ matrix_appservice_slack_docker_image.endswith(':latest') }}"
 
 matrix_appservice_slack_base_path: "{{ matrix_base_data_path }}/appservice-slack"
--- a/roles/matrix-bridge-appservice-slack/defaults/main.yml
+++ b/roles/matrix-bridge-appservice-slack/defaults/main.yml
@@ -3,7 +3,7 @@
 
 matrix_appservice_slack_enabled: true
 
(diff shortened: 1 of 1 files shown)


spantaleev:
Can we use a tagged Docker image instead? Example: `matrixdotorg/matrix-appservice-slack:release-1.5.0`.

I'm guessing we'll also need to somehow handle existing installations. Perhaps an entry in the CHANGELOG would be easiest and hopefully good enough.


--------

We've tried switching to the official image before, but data migration was (still) broken, so we were stuck on this image (see this issue: https://github.com/matrix-org/matrix-appservice-slack/issues/347).

Destroying the data directory and starting fresh with the official image doesn't sound ideal, but maybe a better migration path won't be available. That issue has been open for too long without any work happening there.

-----------

Is `cadair/matrix-appservice-slack:cadair` really so broken nowadays that it doesn't work at all for anyone? @Cadair?

dwiegreffe:
Of course it is possible to tag the image, but then you have to check regularly if there are updates. Slack is currently massively rebuilding its API and scopes, so old versions can become non-functional quite quickly.

The cadair image is currently broken since it uses deprecated methods which are denied by the Slack API.

spantaleev:
The way it is now (no `:latest` suffix, even though it's implied and `:latest` will be used), the playbook will **never** update the image. You'll be forever stuck on the first image that you pull. This is not good.

If you specify `:latest` explicitly, the playbook will try to pull a new image each time you run it. This is also not good, because we don't know what `latest` is. Is that built from `master` directly? Perhaps it's sometimes unpolished/broken. We try to have the playbook lead to no surprises (unless a CHANGELOG entry warns you about a surprise). So, where possible, we try to use tagged releases which we know to work.

Using a release is best. We can update it to another release when a new one becomes available. It'd be great if someone keeps track of these and updates us on time, so we can stay ahead of Slack breakage. I can subscribe to the RSS feed for that bridge and do updates myself, but.. I can't test them.. because I don't use this bridge. This is the reason some things become outdated. I'd rather not update certain images if I can't test that they work. I'd be happy for someone else to help with that.

dwiegreffe:
Then you should rather use a tagged image.

spantaleev:
Thanks! :+1:

PR #652, opened 2020-09-15
Corrects the Caddyfile

0hlov3:
I did a failure, so the identity server wasn't reachable. But now it should work.


spantaleev:
Thanks for fixing it up! :+1:

PR #655, opened 2020-09-24
Allow configuration of Element's bug_report_endpoint_url and showLabsSettings

aaronraimist:
showLabsSettings is the new enableLabs I guess. enableLabs doesn't seem to do anything anymore. It had been deprecated for a while. I've left it enabled by default since it was previously but maybe it shouldn't be.

This PR also removes @riot-bot:matrix.org as the default welcome_user_id since it doesn't exist anymore.


spantaleev:
Thank you! :+1:

PR #657, opened 2020-09-26
Client well known compatibility

cnvandijk:
Support both the `im.vector.riot` and `io.element` variants in `.well-known/matrix/client`.

According to the docs, `e2ee` [is already under](https://github.com/vector-im/element-web/blob/develop/docs/e2ee.md#disabling-encryption-by-default) `io.element` however `jitsi` [is still under](https://github.com/vector-im/element-web/blob/develop/docs/jitsi.md#configuring-element-to-use-your-self-hosted-jitsi-server) `im.vector.riot`.

For now let's just maintain backward and forward compatibility for both settings since the client version is out of the control of this playbook. If somebody has a better proposal I am happy to hear it.

Also, I fixed a few remaining `*_riot_*` variables. Note that this is again a breaking change (similar to 2020-07-17), due to the variable name change.

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -18,8 +18,6 @@
 
 matrix_identity_server_url: "{{ ('https://' + matrix_server_fqn_matrix) if matrix_ma1sd_enabled else None }}"
 
-matrix_riot_jitsi_preferredDomain: "{{ matrix_server_fqn_jitsi if matrix_jitsi_enabled else '' }}"
-
 ######################################################################
 #
 # /matrix-base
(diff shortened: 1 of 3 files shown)


group_vars/matrix_servers
```diff
@@ -18,8 +18,6 @@
 
 matrix_identity_server_url: "{{ ('https://' + matrix_server_fqn_matrix) if matrix_ma1sd_enabled else None }}"
 
-matrix_riot_jitsi_preferredDomain: "{{ matrix_server_fqn_jitsi if matrix_jitsi_enabled else '' }}"
```
spantaleev, line 21: Why not rename this to `matrix_client_element_jitsi_preferredDomain` instead?

cnvandijk: Because `matrix_client_element_jitsi_preferredDomain` already existed in `group_vars/matrix_servers`. It was in twice, once used by `roles/matrix-base/templates/static-files/well-known/matrix-client.j2`, the other by `roles/matrix-client-element/templates/config.json.j2`. I assumed this was a bug.

spantaleev: You're right! Seems like it is a bug!

spantaleev:
Thank you for keeping us up-to-date! :+1:

PR #660, opened 2020-09-29
Element web : update welcome page template

### LINKED ISSUE #659: Missing translation in welcome screen
After setting ```matrix_client_element_disable_guests : true``` , the welcome screen show some missing translations.

Screenshot 2020-09-29 at 12 12 59

This seems related to the way the translation is parsed in Element, now using the brand name.

- In the ansible playbook 

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/7eb8192a5121369002e6241e3733eb2b3d25dbfa/roles/matrix-client-element/templates/welcome.html.j2#L182

- In Element's  ```EN_en-xxxx.json``` : ```"Chat with %(brand)s Bot": "Chat with %(brand)s Bot"```

NB: I have a custom brand name on the server

Also Room directory has a cap at directory but appears without in i18n file ```"Room directory": "Room directory"``` https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/7eb8192a5121369002e6241e3733eb2b3d25dbfa/roles/matrix-client-element/templates/welcome.html.j2#L188

The "Need Help?" is not present in i18n files so maybe it has been removed.

From what I have read online, I thought first it might be a cache issue but the issue is persistent on different browsers and devices.

Seems like the updated template is here 

https://github.com/vector-im/element-web/blob/b720c8a138e4094fdc7cd3faeb5045c5ef5bc827/res/welcome.html

I could make a PR with updated translation, just I have never tested a playbook locally...

clemsos:
fix #659 missing translations when guest users are allowed on Element

### DIFF
--- a/roles/matrix-client-element/templates/welcome.html.j2
+++ b/roles/matrix-client-element/templates/welcome.html.j2
@@ -97,17 +97,8 @@ h1::after {
     color: #2e2f32 !important;
 }
 
-.mx_ButtonHeadline {
-    margin-bottom: 14px;
-}
-
 .mx_ButtonLabel {
     margin-left: 20px;
-}
-
-.mx_ButtonWrapperText {
-    font-size: 13px;
-    margin-bottom: 10px;
 }
 
 .mx_Header_title {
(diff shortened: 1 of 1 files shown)


clemsos:
this is based on element-web v1.7.8

spantaleev:
Thank you! :+1:

PR #663, opened 2020-10-02
Use up-to-date PostgreSQL container version for backups.

hardye:
This follows up on the recent upgrade of the PostgreSQL container to v13.0.


spantaleev:
Thanks! :+1:

PR #666, opened 2020-10-06
Adding '.python-version' to .gitignore

dhose:


spantaleev:
Doesn't hurt having it. Thanks! :+1:

Issue #536, opened 2020-06-10
Mention how to get back some disk space from docker


aaronraimist:
We might want to mention somewhere that you may want to run `docker system prune -a` once in a while. I don't know that much about Docker so I always forget to do it. I just ran it and got back 6GB from old docker images.

PR #668, opened 2020-10-10
Add docker prune

FanchTheSystem:
Related to https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/536

Tested on my own instance.

I have put it in **matrix-common-after**, but maybe it should be in **matrix-base**, or in a new role ... I don't know...

### DIFF
--- a/docs/maintenance-and-troubleshooting.md
+++ b/docs/maintenance-and-troubleshooting.md
@@ -33,6 +33,12 @@ matrix_synapse_root_log_level: "INFO"
 
 Re-run the playbook after making these configuration changes.
 
+## Remove unused Docker data
+
+You can free some disk space from Docker, see [docker system prune](https://docs.docker.com/engine/reference/commandline/system_prune/) for more information.
+```bash
+ansible-playbook -i inventory/hosts setup.yml --tags=run-docker-prune
+```
 
 ## Postgres
 
--- a/roles/matrix-common-after/tasks/main.yml
+++ b/roles/matrix-common-after/tasks/main.yml
@@ -11,3 +11,7 @@
 - import_tasks: "{{ role_path }}/tasks/dump_runtime_results.yml"
   tags:
     - always
+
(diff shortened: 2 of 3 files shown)


spantaleev:
Thank you! :+1:

thedanbob:
One thing to note, this command will delete the `certbot` image since it's not used by a container except when it's run periodically. But this probably isn't a problem since `docker run` should automatically pull the image if it doesn't exist.

Unfortunately, there doesn't seem to be a way currently to exclude certain images from being pruned except by custom label, not tags.

spantaleev:
True.. I guess we can't solve it by labeling images and excluding them from purging, because we'd still like to purge older versions of these images (so we'll need to unlabel them somehow later on?).

I guess it's no big deal that `certbot` gets pruned, as it will get pulled again later on. The output from pulling it may make the cronjob send an email, but that's not too horrible. Pruning happens only once in a while anyway.

-----------

An alternative solution that is probably good is to have all roles inject the images they require into some global list (similar to how they inject themselves into `matrix_systemd_services_list` from `init.yml`). Then, after pruning, we can iterate that list and try to pull each image. Most images would already be there, since pruning wouldn't have deleted them while active. `certbot` would get pulled.

thedanbob:
I managed to come up with a bash script that deletes only the outdated images. It's a little ridiculous and not something I'd stick into the playbook, but I'm also kind of proud of it so I'll leave it here in case it's useful to someone.

```bash
docker rmi $(docker images --format "{{.Repository}}" | sort | uniq --repeated | xargs -I% sh -c 'docker images % --format "{{.ID}}" | tail -1' | xargs)
```

With comments:

```bash
docker rmi $( \ # Delete the following IDs
  docker images --format "{{.Repository}}" | \ # Output just the image names
  sort | uniq --repeated | \ # Find duplicate names
  xargs -I% sh -c 'docker images % --format "{{.ID}}" | tail -1' | \ # For each image, output only the last (oldest) ID
  xargs \ # Gather all the IDs on one line for `docker rmi`
)
```

PR #673, opened 2020-10-13
Added role for dynamic dns

scottcrossen:
Tell me what you think! (I'm sure there are some yml improvements)

I've been testing a homeserver I stood up on a free google cloud f1-micro (1 shared core, 614 MB) a few months ago. I've learned a few things that could be added to this playbook and give matrix access to more casual developers.


roles/matrix-dynamic-dns/tasks/validate_config.yml
```diff
@@ -0,0 +1,19 @@
...
+    - "matrix_dynamic_dns_username"
+    - "matrix_dynamic_dns_password"
+    - "matrix_domain"
+    - "matrix_dynamic_dns_provider"
+    - "matrix_dynamic_dns_mode"
+
+- name: Fail if dynamic dns mode is incorrect
+  fail:
+    msg: >-
+      matrix_dynamic_dns_mode needs to be set to 'daemon' or 'startup'
```
spantaleev, line 18: This says `'daemon'` or `'startup'`, but the check below is for `'daemon'` and `'dhcp'`

roles/matrix-dynamic-dns/tasks/validate_config.yml
```diff
@@ -0,0 +1,19 @@
...
+
+- name: Fail if required settings not defined
+  fail:
+    msg: >-
+      You need to define a required configuration setting (`{{ item }}`).
+  when: "vars[item] == ''"
+  with_items:
+    - "matrix_dynamic_dns_username"
+    - "matrix_dynamic_dns_password"
+    - "matrix_domain"
```
spantaleev, line 11: Perhaps a new variable (`matrix_dynamic_dns_domain`), with an empty default value would be better.

We can then define that in terms of `matrix_domain` in `group_vars/matrix_servers`:

```yaml
matrix_dynamic_dns_domain: "{{ matrix_domain }}"
```

setup.yml
```diff
@@ -5,6 +5,7 @@
 
   roles:
     - matrix-base
+    - dynamic-dns
```
spantaleev, line 8: This should probably be `matrix-dynamic-dns`?

roles/matrix-dynamic-dns/tasks/install.yml
```diff
@@ -0,0 +1,23 @@
...
+    dest: "/etc/ddclient.conf"
+    mode: 0600
+  register: matrix_dynamic_dns_ddclient_domain_config
+
+- name: Ensure ddclient client config directory exists
+  file:
+    path: "/etc/default"
+    state: directory
+    mode: 0700
+    owner: "{{ user_username }}"
```
spantaleev, line 15: What is `user_username` and `user_groupname`?

I wonder if creating/changing `/etc/default`, which already exists on a lot of (all) distros is a good idea?

On CentOS 7, `/etc/default` is owned by `root:root` with `755` permissions.

docs/configuring-playbook-budget-builds.md
```diff
@@ -0,0 +1,21 @@
+# Tips for deploying Matrix on a Budget
+
+## Dynamic DNS
+
+Most cloud providers / ISPs will charge you extra for a static IP address. If you're
+not hosting a highly reliable homeserver you can workaround this via dynamic DNS. To
+set this up, you'll need to get the username/password from your DNS provider. For
+google domains, this process is described [here](https://support.google.com/domains/answer/6147083).
+After you've gotten the proper credentials you can add the following config to your `host-vars`:
```
spantaleev, line 9: Saying `inventory/host_vars/matrix.DOMAIN/vars.yml` instead of `host-vars` is probably more helpful to users.

roles/matrix-dynamic-dns/tasks/platform/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+
+- include_tasks: "{{ role_path }}/tasks/platform/centos.yml"
```
spantaleev, line 3: I suppose there's no good way to run `dhclient` in a Docker container, like we do with the rest of the services?

This would make it more platform-independent and simplify things.

spantaleev (changes requested):
Hey! Thank you for your contribution!

I've submitted some feedback.

roles/matrix-dynamic-dns/tasks/platform/main.yml
scottcrossen: AFAIK we actually could use a docker container, it just wouldn't have access to the DHCP operation mode. One of the nice things about ddclient is that it's configurable to run the dynamic dns routine whenever dhcp claims a new IP and I believe this can't be done in a container. It's a little more overhead to run the daemon but not horrible.

I'll make a new PR (and close this one for posterity) with the [container approach](https://hub.docker.com/r/linuxserver/ddclient). I think we should do this anyway because this will interfere least with other services on the host, provide modularity, and be platform agnostic. We can always add the DHCP "mode" back as another option which opts-in to a host-level ddclient.... but I think most people will be happy with the docker solution.

scottcrossen: Oh BTW I just started a new job and so opensource contributions are not my highest priority right now. This may take me a few weeks to do. I hope to also make PRs against you that enable forward proxying of bridges like mautrix-facebook, enable swap, add a LinkedIn bridge (still in development), and also workaround a docker issue encountered on smaller machines. Just be patient as I get these out!

spantaleev: Awesome! Looking forward to these contributions, whenever you get the chance!

Good luck at your new job!

PR #681, opened 2020-10-19
Dynamic DNS

scottcrossen:
Fixes issues in https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/673

Tested on Debian with default values


roles/matrix-dynamic-dns/defaults/main.yml
```diff
@@ -0,0 +1,40 @@
+# Whether dynamic dns is enabled
+matrix_dynamic_dns_enabled: false
+
+# The dynamic dns daemon interval
+matrix_dynamic_dns_daemon_interval: '300'
+
+# The docker container to use when in mode
+matrix_dynamic_dns_docker_image: 'linuxserver/ddclient'
```
spantaleev, line 8: Maybe this should have an explicit `:latest` tag here? This would also allow `matrix_dynamic_dns_docker_image_force_pull` to work as expected.

We also prefer sticking to fixed versions which we know to work, where available. So we may go for something like `v3.9.1-ls44` instead.

Looks like all of these tags supported multiple architectures. This would work nicely for us, without further complications (no need to enable self-building for non-`amd64` or something).

roles/matrix-dynamic-dns/defaults/main.yml
```diff
@@ -0,0 +1,40 @@
+# Whether dynamic dns is enabled
+matrix_dynamic_dns_enabled: false
```
spantaleev, line 2: We usually set `_enabled` to `true` in each role, so that people using our roles outside of the playbook have them enabled without doing anything.

For the playbook itself, we enable/disable roles by overriding such `_enabled` variables in `group_vars/matrix_servers`.

docs/configuring-playbook.md
```diff
@@ -33,6 +33,7 @@ When you're done with all the configuration you'd like to do, continue with [Ins
 
 - [Setting up the Jitsi video-conferencing platform](configuring-playbook-jitsi.md) (optional)
 
+- [Setting up budget builds or resource-constrained builds](configuring-playbook-budget-builds.md) (optional)
```
spantaleev, line 36: Maybe renaming this to `configuring-playbook-dynamic-dns.md` is better for the Dynamic DNS feature (which may be usable for others as well)?

Or, we could have 2 docs pages linking to each other.

docs/configuring-playbook-budget-builds.md
```diff
@@ -0,0 +1,21 @@
...
+## Dynamic DNS
+
+Most cloud providers / ISPs will charge you extra for a static IP address. If you're
+not hosting a highly reliable homeserver you can workaround this via dynamic DNS. To
+set this up, you'll need to get the username/password from your DNS provider. For
+google domains, this process is described [here](https://support.google.com/domains/answer/6147083).
+After you've gotten the proper credentials you can add the following config to your inventory/host_vars/matrix.DOMAIN/vars.yml:
+
+```
+matrix_dynamic_dns_username: XXXXXXXXXXXXXXXX
```
spantaleev, line 12: Adding `matrix_dynamic_dns_enabled: true` is probably beneficial here, since this role is not enabled by default.

spantaleev (changes requested):
Awesome! Thanks for making this run in a container in a portable manner!
Looks like it will even run on architectures other than `amd64` without any modification.

I've posted some last minor additional things we could improve. I guess we could merge it after that :blush:

roles/matrix-dynamic-dns/defaults/main.yml
scottcrossen: Great tips! Done.

roles/matrix-dynamic-dns/defaults/main.yml
scottcrossen: I've updated this line. I think the life we want is to only enable this when there is work to be done. i.e. that there exists nonzero domains to set dynamically. Check out this update!

docs/configuring-playbook.md
scottcrossen: Good call! I've rebranded this article

docs/configuring-playbook-budget-builds.md
scottcrossen: See comment on first line of defaults/main.yml

scottcrossen:
I've redone the config data model a bit so that multiple domains can use different providers. Ex:

```
matrix_dynamic_dns_domain_configurations: |
  {{
    [{'provider': 'domains.google.com', 'protocol': 'dyndn2', 'username': 'XXXXXXXXXXXXXXXX', 'password': 'XXXXXXXXXXXXXXXX', 'domain': matrix_domain}, {'provider': 'another.provider.com', 'protocol': 'dyndn2', 'username': 'XXXXXXXXXXXXXXXX', 'password': 'XXXXXXXXXXXXXXXX', 'domain': element_domain}]
  }}
```

I've also changed the nginx playbook a bit so that dynamic dns runs before certificate retrieval (needed when using domain challenges ofc)

scottcrossen:
Retested on debian, although I reused the dynamic-dns playbook recently on a pi too.

roles/matrix-dynamic-dns/defaults/main.yml
```diff
@@ -0,0 +1,32 @@
+# Whether dynamic dns is enabled
+matrix_dynamic_dns_enabled: "{{ matrix_dynamic_dns_domain_configurations is defined }}"
```
spantaleev, line 2: This is smart, but all our optional roles require an explicit `_enabled: true` to enable them, so I feel like this is a little strange.

Also, given that `matrix_dynamic_dns_domain_configurations` is set to `[]` in `group_vars/matrix_servers`, I think this would currently make `matrix_dynamic_dns_domain_configurations is defined` yield `true` and enable this role for everyone by default.

I think it's best that we set `matrix_dynamic_dns_enabled` to `true` in the role itself (so that anyone who includes it into their own other playbook will have it enabled by default), and then `matrix_dynamic_dns_enabled: false` in `group_vars/matrix_servers` (so that our "distribution" would have it be an optional component, disabled by default).

scottcrossen: Done

spantaleev (approved):

spantaleev:
I've fixed a few things and merged it!

Thanks a lot for your hard work on bringing Dynamic DNS support to the playbook! :+1:

PR #684, opened 2020-10-24
Fix docker mount instruction in documentation

deSagaz:
The SSL certificates should be accessible to the nginx proxy container, not the synapse container.


spantaleev:
Good catch! Thanks! :+1:

PR #698, opened 2020-10-28
Upgrade Element (1.7.10 -> 1.7.11)

jgbresson:


spantaleev:
Thanks for this! :+1: 

Looks like Element v1.17.12 is out, and I've just upgraded to that in c1c6eaefff7eb, making this PR obsolete.

PR #702, opened 2020-11-01
keycloak SSO

makayabou:
add nginx conf for /_synapse/oidc/callback, used for keycloak SSO.   

It had been broken by commit 63a49bb2dc7780e023b2801a7230cda529b2b3c1


spantaleev:
Thanks for catching this and taking the time to make a Pull Request! :+1: 

There was a cleaner way to do it (which also doesn't mention Keyclock, which is just one of the OpenID Connect providers).

I've implemented it in 350c39d7457f325

PR #713, opened 2020-11-09
ssl retrieval with nginx running on host - fixes #712

### LINKED ISSUE #712: can't use builtin certbot if my nginx is running on host
I use my nginx proxy on host to serve apps.
So, as port 80 is already used, I can't have playbook retrieving ssl certs for me with lets encrypt.

makayabou:
It may be a dirty fix, and it only works with nginx,

but I just tried to level up and be helpful.

Thanks for the project


spantaleev:
Well, that would let you retrieve the certificates.. and then still leave you with something which won't work.
None of these domains would serve what they need to serve, unless you inject even more configuration and do even more changes (as described in our [using your own webserver docs page](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-own-webserver.md)).

I think we're better off not doing such hacky solutions.

makayabou:
i meant level up my own capabilities ;-)  As you saw, last time i wasn't even able of using the templates . Thanks again for the project and help on matrix chan

PR #717, opened 2020-11-12
Non root user and Matrix group

borgmanJeremy:
This might be obvious to more experienced users but it took me a while to realize non-root users needed to be a member of the matrix group.

### DIFF
--- a/docs/installing.md
+++ b/docs/installing.md
@@ -12,6 +12,11 @@ ansible-playbook -i inventory/hosts setup.yml --tags=setup-all
 
 **Note**: if you **do** use SSH keys for authentication, **and** use a non-root user to *become* root (sudo), you may need to add `-K` (`--ask-become-pass`) to the above (and all other) Ansible commands.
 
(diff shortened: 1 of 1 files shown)


spantaleev:
Are you not using `--ask-become-pass` or something to elevate your user to `root` for the installation process?

The `matrix` group does not exist before installing, so this may be a bit of a catch-22.

borgmanJeremy:
I was using -K.

I ran into an issue during the "start" phase even with -K because the user I was becoming was not a member of the matrix group and could not read the files in /matrix

spantaleev:
I don't know how this could work. Your intention is to `become` someone who is not `root`.

Given that you're trying to become a non-superuser, I don't see how the playbook will be able to do everything it needs (that is, creating systemd service files in `/etc/systemd/system/`, etc.)

This seems like a strange setup and the additional note you're wishing to add to the documentation will be confusing to most users (since there's no `matrix` group initially.. it's a group that gets created by Ansible later on).

PR #718, opened 2020-11-13
updated matrix-sms-bridge

benkuly:


spantaleev:
Thanks for the detailed instructions! :+1:

PR #723, opened 2020-11-17
update mautrix-telegram to 0.9.0

jdreichmann:
https://github.com/tulir/mautrix-telegram/releases/tag/v0.9.0


spantaleev:
I wonder if it has some new configuration options (and/or new defaults) that we'd like to pull.

Good to hear that it works as-is though! Thanks! :+1:

jdreichmann:
E2BE encryption could be turned on, but it is off by default upstream aswell

PR #725, opened 2020-11-22
Fix issues with mautrix dbs on resource-constrained systems

scottcrossen:
The issue:

Whenever I restart my server or docker daemon, Mautrix-based bridges with alembic databases refuse to start because of this error:

```
docker: Error response from daemon: Conflict. The container name "/matrix-mautrix-something-db" is already in use by container "blahblahblah".
You have to remove (or rename) that container to be able to reuse that name.
```

It's because the `--rm` step times-out and the old container can't be removed. For some reason the container can't be removed by the daemon.

This fix will first try to kill the database container, then if that times out it forcefully removes the container and the network interface.


spantaleev:
Does that network disconnect even do anything when you've removed the container and then force-removed it? It should be completely gone, shouldn't it.

Thinking about it now, we can probably run the migrations container without networking (`--network=none`), at least for now, until we've migrated these bridges to use Postgres.

scottcrossen:
I found that the a `rm --force` does retain artifacts in the network bridge even though it removes the container. You won't be able to start another container because the network identifier is still in use on that bridge. I'm not really sure what these "*-db" containers are used for. Are you saying that these containers (ending in "*-db") are "migrations containers"? Are these the ones you want on `--network=none`? Right now 2/3 of them run on the default bridge.

scottcrossen:
According to this. It looks like something might fail if the non-matrix docker bridge is used:

https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/331c77a6510b8b15f9d8e603c7f43e5f74ce74cd

spantaleev:
You're right. Some people may be overriding the configuration and pointing it to `matrix-postgres`, in which case it would need to be on the `matrix` network.

scottcrossen:
> You're right. Some people may be overriding the configuration and pointing it to `matrix-postgres`, in which case it would need to be on the `matrix` network.

Is this just for telegram or for all mautrix bridges?

spantaleev:
Probably for all bridges (not just mautrix ones). Well, not for now, because none of them use a DB that's on the network. But they should in the future. Especially after #740 and some related work lands.

spantaleev:
I'd rather we don't merge this as it appears to be a hacky workaround for some weird container issue, which is not experienced commonly and may have been been fixed in newer Docker releases (20.10 appeared recently).

If you're still experiencing it, let's see how we can reproduce it and come up with a similar fix.

Thanks anyway! :+1:

scottcrossen:
Hey cool with me! Sorry about the delay (holidays)

PR #730, opened 2020-11-25
updates matrix-sms-bridge (changed SMS provider)

benkuly:


spantaleev:
Looks like things become much simpler with this. Thanks for keeping us updated! :+1:

benkuly:
Yep, we have far less problems without gammu and it's much easier to setup :)

PR #732, opened 2020-11-26
Note about the max length of the postgres password

CodingAnonymous:
That could save someone a lot of trouble...


spantaleev:
Thanks! :+1: 

I've also updated `roles/matrix-postgres/tasks/validate_config.yml` to enforce it.

PR #740, opened 2020-11-29
postgres: create databases for all services

jdreichmann:
If a service is enabled, a database for it is created in postgres with a uniqque password. The service can then use this database for data storage instead of relying on sqlite.

- [x] Point all bridges which support it to the postgres container (if `matrix_postgres` is enabled)
- [x] Roles should default to `sqlite`, so should the playbook as a whole if `postgres` is running externally
- [x] Roles should auto-detect existing SQLite DBs and warn/error/auto-migrate before they change the config to the postgres-default
- [x] Testing
- [x] Migration can be done using `--tags=import-generic-sqlite-db` (from the `matrix-postgres` role) to import sqlite DBs (we still have this tag, even though we now do automatic migration for everything and don't advertise it at all)

Implementation:

- [x] appservice_discord (done and tested; automatic migration supported)
- [x] appservice_irc  (done and ~tested; automatic migration supported)
- [x] appservice_slack (done and ~tested; automatic migration supported)
- [x] mautrix_facebook (done and tested; automatic migration supported)
- [x] mautrix_whatsapp (done, but **untested**; automatic migration supported)
- [x] mautrix_telegram (done, but **untested**; automatic migration supported)
- [x] mautrix_hangouts (done, but **untested**; automatic migration supported)
- [x] mx_puppet_skype (done and ~tested; automatic migration supported, mx-puppet-* added by @jdreichmann )
- [x] mx_puppet_slack (done and ~tested; automatic migration supported))
- [x] mx_puppet_discord (done and ~tested; automatic migration supported)
- [x] mx_puppet_steam (done and ~tested; automatic migration supported)
- [x] mx_puppet_instagram (done and ~tested; automatic migration supported)
- [x] mx_puppet_twitter (done and ~tested; automatic migration supported)
- [x] dimension (done and ~tested; automatic migration supported)
- [x] ma1sd (done and tested; automatic migration supported)
- [x] matrix-reminder-bot (done and tested; automatic migration supported)
- [x] matrix-registration (done and tested; ~~defaulting to Postgres blocked by ZerataX/matrix-registration#44~~; automatic migration supported)

Feel free to pick a bridge and implement the migration :)


roles/matrix-postgres/tasks/setup_postgres.yml
```diff
@@ -162,3 +162,33 @@
     - matrix-change-user-admin-status
     - matrix-postgres-update-user-password-hash
   when: "not matrix_postgres_enabled|bool"
+
+# Create additional databases
+- name: Retrieve IP of postgres container
```
pushytoxin, line 167: Why not use the database server's host name? The Docker DNS server should resolve it for you - {{ matrix_synapse_database_host }}

jdreichmann: Can you reach a docker container by it's hostname on the host? Last time i checked, a similar feature was only available for docker-on-mac - and ansible (i.e.: this module) runs on the host. The IP however should be routable from the host, regardless of docker version / OS distribution

jdreichmann:
Bridges compatible with postgres DB store:

- [ ] matrix-appservice-discord
- [ ] matrix-appservice-slack
(incomplete)

spantaleev:
Thank you for the great start on this! :+1: 

It's been a good inspiration and I've spent a lot of time progressing tonight based on it.

I believe to have made some significant progress.
I've pushed my commits to your branch and I've tried to keep them atomic and understandable.

As an overview of all the changes:

- I'm trying to move away from bridge roles reaching into `matrix_postgres_additional_databases`. It'd be nicer if we can keep roles separate as much as possible with `group_vars/matrix_servers` wiring them together. See how I've reversed things around for the Appservice Discord bridge. This is to be done for the other bridges as well.

- I'm trying to let people still use SQLite for their bridges (and even having that as the default in some cases), instead of forcing them into Postgres. See how it's done for the Appservice Discord bridge.

- when people use the roles separately from the playbook,  the goal is to have the bridge roles default to whatever they can do in such a configuration (which is: use SQLite)

- when people use an external Postgres server, we also aim for defaulting to SQLite, because we're not ready to automatically create users/databases in something which is not `matrix-postgres`. Ideally, we'd be able to, but for now we make a lot of assumptions and hardcode stuff. I don't have such an external-Postgres setup on hand right now, nor the motivation to deal with these edge cases myself. Still, we don't break things for these people - we default to whatever worked (SQLite). They can also prepare their own Postgres users/databases and repoint bridges there, so whoever is inclined enough can do it manually. Or they can submit a PR, of course ;)

- I've added `|to_uuid` to the logic for password generation to avoid escaping issues (and even potential "password too long" issues). I've hit some escaping issues when importing data with [pgloader](https://pgloader.io/), but we may have the same problem in other places. Using UUIDs for passwords is probably a safe enough default

- I've migrated the tasks away from using Postgres Ansible modules (`postgresql_user` and `postgresql_db`), because they run on the host and can't reach the `matrix-postgres` container. Also, they require additional libraries (`psycopg2`) to be installed on the server, which may be a pain to do, especially for all the bunch of distros that we support.

- aim for having an `*_engine: postgres` default for bridges (see the Appservice Discord bridge), but still have the bridge role tasks detect an old database and error out before overwriting the configuration and pointing the bridge to Postgres

- as part of this erroring out, we provide 2 resolution paths (letting people stay on SQLite; or telling them how to migrate to Postgres)

- to do the (SQLite -> Postgres) migration we have a new command `--tags=import-generic-sqlite-db` living in the `matrix-postgres` role. It's to be used to import any database (from any bridge, etc.), into Postgres. Whenever it completes, it renames the database file (`file.db` -> `file.db.backup`). Re-running the playbook after that should avoid the aforementioned "erroring out" and should set up a bridge to use Postgres

As mentioned in the summary above, most of these things are only implemented for the Appservice Discord bridge. I'm yet to spend time on doing the same for the other bridges. Well, if someone else wants to help with that, that'd be great. Now that the complicated part is done, it should be relatively easy to fix these things up.

Talking about the Appservice Discord bridge, I've already used the playbook's instructions to migrate my data from SQLite to Postgres. It seems like it worked, but I'm yet to see any Discord activity and confirm whether it all works.

I'll likely spend time migrating the Facebook bridge tomorrow. At least that's a bridge I can test easily.

jdreichmann:
Thanks for the continuation, and some input on how the final featureset would make sense. It would be nice to have an automatic Import with pg_loader per default, as that would avoid people running into errors just because their DB will be migrated. For people running the absolute defaults, i'd be in support of forcing them to postgres, but the roles should default to sqlite, i agree there.

roles/matrix-bridge-appservice-discord/defaults/main.yml
```diff
@@ -41,6 +41,23 @@ matrix_appservice_discord_bridge_homeserverUrl: "http://matrix-synapse:8008"
...
+# To use SQLite, stick to these defaults.
+#
+# To use Postgres:
+# - change the engine (`matrix_appservice_discord_database_engine: 'postgres'`)
+# - adjust your database credentials
+matrix_appservice_discord_database_engine: 'sqlite'
+matrix_appservice_discord_database_filename: "/data/{{ matrix_appservice_discord_database_filename_name }}"
+matrix_appservice_discord_database_filename_name: "discord.db"
+matrix_appservice_discord_database_connString: 'postgresql://{{ matrix_appservice_discord_database_connString_username }}:{{ matrix_appservice_discord_database_connString_password }}@{{ matrix_appservice_discord_database_connString_hostname }}:{{ matrix_appservice_discord_database_connString_port }}/{{ matrix_appservice_discord_database_connString_db_name }}'
+matrix_appservice_discord_database_connString_username: 'matrix_appservice_discord'
```
jdreichmann, line 55: I'd argue to drop the `_connString_` part of all `$role_database_*` variables, as it's not the username/password/... _of_ the connString, the connString is a compound of the set. Also, i'm lazy and dont want to type that much and shorter variable names are better, and `matrix_appservice_discord_database_username` is still descriptive.

roles/matrix-bridge-appservice-discord/tasks/setup_install.yml
```diff
@@ -1,5 +1,28 @@
...
 
+- block:
+  - set_fact:
+      matrix_appservice_discord_sqlite_db_path: "{{ matrix_appservice_discord_data_path }}/{{ matrix_appservice_discord_database_filename_name }}"
+
+  - name: Check if an SQLite database already exists
+    stat:
+      path: "{{ matrix_appservice_discord_sqlite_db_path }}"
+    register: matrix_appservice_discord_sqlite_db_path_stat_result
+
```
jdreichmann, line 11: Maybe we could do sth here with `import_tasks` for an automigration

roles/matrix-bridge-appservice-discord/templates/config.yaml.j2
```diff
@@ -58,8 +58,11 @@ database:
   # If you are migrating, see https://github.com/Half-Shot/matrix-appservice-discord/blob/master/docs/howto.md#migrate-to-postgres-from-sqlite
   # WARNING: You will almost certainly be fine with sqlite unless your bridge
   # is in heavy demand and you suffer from IO slowness.
-  filename: "/data/discord.db"
-  # connString: "postgresql://user:password@localhost/database_name"
+  {% if matrix_appservice_discord_database_engine == 'sqlite' %}
+  filename: {{ matrix_appservice_discord_database_filename|to_json }}
+  {% else %}
+  connString: {{ matrix_appservice_discord_database_connString|to_json }}
```
jdreichmann, line 64: `|to_json`? That should be a string already, and if it weren't, there's `|string`..

roles/matrix-postgres/tasks/main.yml
```diff
@@ -26,6 +26,12 @@
   tags:
     - import-synapse-sqlite-db
 
+# Perhaps we need a new variable here, instead of `run_postgres_import_sqlite_db`.
+- import_tasks: "{{ role_path }}/tasks/import_generic_sqlite_db.yml"
```
jdreichmann, line 30: with a `vars: { ... }` block, we should be able to dynamically execute this in every role for auto-migration

roles/matrix-postgres/tasks/import_generic_sqlite_db.yml
```diff
@@ -0,0 +1,84 @@
...
+  fail:
+    msg: "The `sqlite_database_path` variable needs to be provided to this playbook, via --extra-vars"
+  when: "sqlite_database_path is not defined or sqlite_database_path.startswith('<')"
+
+- name: Fail if playbook called incorrectly
+  fail:
+    msg: >-
+      The `postgres_db_connection_string` variable needs to be provided to this playbook, via `--extra-vars`.
+      Example: `--extra-vars="postgres_db_connection_string=postgresql://username:password@localhost:<port>/database_name`"
+  when: "postgres_db_connection_string is not defined or not postgres_db_connection_string.startswith('postgresql://')"
```
jdreichmann, line 20: We could provide that to the user, we should know the `connString` of `matrix-postgres` already, and avoid people potentially having their db-password in their bash_history

roles/matrix-postgres/tasks/import_generic_sqlite_db.yml
```diff
@@ -0,0 +1,84 @@
...
+
+- name: Ensure matrix-postgres is started
+  service:
+    name: matrix-postgres
+    state: started
+    daemon_reload: yes
+  register: matrix_postgres_service_start_result
+
+- name: Wait a bit, so that Postgres can start
+  wait_for:
```
jdreichmann, line 51: There is a feature somewhere which can do stuff like `wait_for` a specific port to be available:
```
- wait_for:
    port: 5432
    delay: 5
    timeout: "{{ postgres_start_wait_time }}"
```
(though this requires getting the IP of `matrix-postgres, but that's not hard)

roles/matrix-postgres/templates/init-additional-db-user-and-role.sql.j2
```diff
@@ -0,0 +1,19 @@
+-- `CREATE USER` does not support `IF NOT EXISTS`, so we use this workaround to prevent an error and raise a notice instead.
+-- Seen here: https://stackoverflow.com/a/49858797
+DO $$
+BEGIN
+  CREATE USER {{ additional_db.username }};
+  EXCEPTION WHEN DUPLICATE_OBJECT THEN
+  RAISE NOTICE 'not creating role {{ additional_db.username }}, since it already exists';
```
jdreichmann, line 7: 'not creating user `{{ additional_db.username }}` ...'? - the `role` is (imo) a bit confusing, does postgres really call their users "role", or do the users _have_ roles?

roles/matrix-bridge-appservice-discord/defaults/main.yml
spantaleev: True!

I've gone for prefixing them with `_connString_` because:

- we already use a similar pattern in other places

Example:

```yaml
matrix_synapse_docker_image: "{{ matrix_synapse_docker_image_name_prefix }}matrixdotorg/synapse:{{ matrix_synapse_docker_image_tag }}"
matrix_synapse_docker_image_name_prefix: "{{ 'localhost/' if matrix_synapse_container_image_self_build else 'docker.io/' }}"
# The if statement below may look silly at times (leading to the same version being returned),
# but ARM-compatible container images are only released 1-7 hours after a release,
# so we may often be on different versions for different architectures when new Synapse releases come out.
matrix_synapse_docker_image_tag: "{{ 'v1.24.0' if matrix_architecture == 'amd64' else 'v1.24.0' }}"
matrix_synapse_docker_image_force_pull: "{{ matrix_synapse_docker_image.endswith(':latest') }}"
```

.. but I could see how having these variables being more standalone makes sense too (`matrix_synapse_docker_image_tag` -> `matrix_synapse_version`)

- for bridges we aim for sensible mapping between variable names and configuration values

For example `matrix_appservice_discord_database_connString` *likely* mapping to the `database.connString` configuration option of the Appservice Discord bridge.

I think for some role we even had the prefix be `(matrix_service_config_STUFF_HERE`) making it even more explicit.

`matrix_appservice_discord_database_username` somewhere near `matrix_appservice_discord_database_connString` may make people believe there's actually also a `database.username` field in the configuration.

Putting the username under the `matrix_appservice_discord_database_connString` variable prefix and having it used inside probably makes it clear enough that it's just some way to split the `connString` into sub-parts for easier definition/redefinition.

----------

You're right though.. For people who override these configuration values, putting something like `matrix_appservice_discord_database_connString_username` in their `vars.yml` file sounds weird. We should probably change to what you're proposing.

roles/matrix-bridge-appservice-discord/templates/config.yaml.j2
spantaleev: It should be a string already, but if it contains some special characters, it has the chance of breaking YAML parsing.

Using `|to_json` ensures that it would do proper double-quote wrapping and escape any other weird characters that may be inside, so that it would be valid YAML syntax.

It's probably unlikely and mostly affecting `matrix_appservice_discord_database_connString` and not `matrix_appservice_discord_database_filename`, but it doesn't hurt having it there.

I think our default connection strings yield rather safe values, but people might set a weird password or something sometimes.

We've recently had complains of ma1sd breaking, because someone's Postgres password contained a weird sequence (e.g. `:{space}`). That was fixed here: https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/7372480e95d9fa60c05270236cc6f4d6753c3e67

roles/matrix-bridge-appservice-discord/tasks/setup_install.yml
spantaleev: Auto-migration would be ideal. At first I started on that, but then I figured it might be somewhat dangerous.

For example, importing the Appservice Discord database **always** gives me at least a few errors ("... cannot have multiple primary keys in `emoji` table", etc). As of yet, I'm still not sure how perfectly that migration went.

So it makes me see it as something a bit dangerous and a somewhat big change, that users should do explicitly (similar to how we don't auto-upgrade between Postgres versions -- although that's even more dangerous).

Doing auto-migration also seemed more difficult to do, as it involves:

- (*already done*) detecting whether we need to migrate
- stopping the bridge service, so that we can work with the database exclusively
- doing the migration via `import_tasks` (should be easy). This likely involves doing the same `file.db` -> `file.db.backup` rename
- then doing all the configuration tasks -- we need to rebuild a bridge's `config.yml` after all, as there's no `file.db` database anymore and it's all in Postgres already. If the bridge were to be started at this moment, it would create a new empty database (likely not too catastrophic though).
- then starting the bridge again if it had been stopped in step 2, above. Is this to be done in the role itself or postponed until the very end of playbook execution (`matrix-common-after`)? If someone does just `--tags=setup-appservice-discord`, then we won't even make it to where we expect to in `matrix-common-after`. Well, that's rare, but ..

It should be possible to have auto-migration though. What we have now is a little annoying to use.

roles/matrix-postgres/tasks/import_generic_sqlite_db.yml
spantaleev: You're right! This is quite raw right now. We should provide a way to just pass a "service name" and have the connection string extracted out of `matrix_postgres_additional_databases`.

That is, if we're staying with such a manual-migration setup.

roles/matrix-bridge-appservice-discord/defaults/main.yml
jdreichmann: I've put the variables without the infix `_connString_` in the `group_vars/matrix_server` for now, as i'm for hiding implementation details (such as specific config.yaml layout) from the user where it would have little benefit.

roles/matrix-postgres/tasks/import_generic_sqlite_db.yml
spantaleev: We can extract it, and I was initially redoing your extraction logic to not require `jq`, etc., but I still don't think that we can reach that IP from the host itself.

Has that worked for you somehow?

I'd think that a container binding to ports on some network (well, any network besides `--net=host`) will not be reachable from the host itself (or from any other container network), unless ports have been exposed (`-p 5432:5432`) or some other kind of bridging is configured.

-----------------

What we could do is use something like [waitforit](https://github.com/maxcnunes/waitforit) from within a container on the same network (`--network=matrix`) and wait for `matrix-postgres:5432`. We'd need to find a container image for it though. And it'd be nice it it worked on ARM architectures as well.

Having said that, our current pgloader container image (`docker.io/illagrenan/pgloader:3.6.2`) only works on `amd64`, so people on ARM don't have a good alternative :/ Still, that's something that only affects people on current installations, who want to move to Postgres.

But if we are to do this `waitforit` thing, it would execute for everyone who enables at least one service (bridge) that uses an additional database.

Maybe waiting for 10-15 seconds is not too terrible, given that it's easy to do and doesn't break ARM support.

roles/matrix-postgres/templates/init-additional-db-user-and-role.sql.j2
spantaleev: Postgres seems to call these roles, yes. And some roles have login support (`CREATE ROLE .. WITH LOGIN`) while others are just about data ownership, but don't provide login support (`CREATE ROLE ... WITH NOLOGIN`).

`CREATE USER` is a shortcut for saying `CREATE ROLE .. WITH LOGIN`.

I've fixed this notice message in f1e85f7112ffa9dd

roles/matrix-postgres/tasks/import_generic_sqlite_db.yml
jdreichmann: Something like this has worked for me in the past. we can `docker inspect` the container, register the output and use `|from_json` and lookup the IP, and on all docker daemon variants, the IP itself should always be routable from the host to the container.

If the pg_loader is `amd64` only right now, `arm64` people wont be able to migrate anyway

roles/matrix-bridge-appservice-discord/defaults/main.yml
spantaleev: Alright! Let's get rid of that `_conString_` infix thing. Or should I do it for Appservice Discord?

roles/matrix-postgres/tasks/import_generic_sqlite_db.yml
spantaleev: That's an easy way to get it without needing `jq`: `docker inspect matrix-postgres --format '{{.NetworkSettings.Networks.matrix.IPAddress}}'`

It yields something like `172.18.0.3` for me right now.

Connecting doesn't work for me though (testing `telnet 172.18.0.3 5432` from the host).

And I'd be surprised if it did, as that's the whole point of containers -- isolating these processes and their network away from the host (and any other containers), unless specifically requested to bridge.

This does work though: `docker run -it --rm --network=matrix docker.io/alpine:3.12 /bin/sh -c 'apk add --no-cache busybox-extras && telnet 172.18.0.3 5432'`

spantaleev: As for the ARM thing, yes, ARM users won't be able to migrate anyway as things are right now. But at least new playbook users can still do a full installation (which uses bridges with Postgres) without any breakage. If run `waitforit` in some `amd64`-only container, we'd be breaking `--tags=setup-all` for anyone who has enabled anything that uses additional Postgres databases, unless we find or build our own ARM image.

I feel it's better if we don't break the main installation path and rather just break some migration command for some subset of the users. For now, at least. Not that these are related. They're both fixable, separately ;) And we should

roles/matrix-bridge-appservice-discord/defaults/main.yml
spantaleev: I've redone these variable names in d91aa5a060ab0. Let me know what you think!

I've even gone for `matrix_appservice_discord_postgres_username`, instead of `matrix_appservice_discord_database_username` (`_postgres_` instead of `_database_`), to make it even more explicit what it's about.

roles/matrix-postgres/tasks/import_generic_sqlite_db.yml
spantaleev: Fixed in 5dba0c038b6c19a9b.

The import command is something like this now:

```
ansible-playbook -v -i inventory/hosts setup.yml \
--tags=import-generic-sqlite-db \
--extra-vars='sqlite_database_path=/matrix/appservice-discord/data/discord.db postgres_connection_string_variable_name=matrix_appservice_discord_database_connString'
```

roles/matrix-bridge-appservice-discord/defaults/main.yml
jdreichmann: i am not sure if `_postgres_` is better then `_database_` - the connString format supports a lot of databases, and they all have the require information like  `host`, `port`, `user`, ... - the generic `_database_` is probably more accurate, as with that, a user could run a mysql-DB aswell and use the variables and just changeout the protocol-prefix in the connString and change the `*_database_engine` to `mysql`

spantaleev:
The mautrix bridges are done now (Facebook, Whatsapp, Telegram, Hangouts). I've updated the list at the top of this issue.

Of these, **only `mautrix-facebook` that has been tested** and appears to be working.

-----------------------

## Testing bridges

If you'd like to help out test the other bridges, that'd be nice. For each one that you use, you can:

- try running the playbook (``ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start``) and see if it's erroring out telling you to explicitly switch to SQLite or to migrate your existing data to Postgres
- try explicitly opting into SQLite (e.g. `matrix_mautrix_whatsapp_database_engine: 'sqlite'` in your `vars.yml` file) and confirm that the bridge still works the same way it did before
- try getting rid of this `matrix_mautrix_whatsapp_database_engine: 'sqlite'` thing from `vars.yml`, re-run the playbook (`--tags=setup-all,start`), but this time follow the (SQLite -> Postgres) importing instructions from the error message. Make sure the bridge works and can send/receive messages, etc.

**If you hit a problem** with the migration, you can:
- stop all services (``ansible-playbook -i inventory/hosts setup.yml --tags=stop`)
- restore the old SQLite database file (e.g. `mv /matrix/mautrix-whatsapp/data/mautrix-whatsapp.db.backup /matrix/mautrix-whatsapp/data/mautrix-whatsapp.db`)
- explicitly opt into SQLite once again (e.g. `matrix_mautrix_whatsapp_database_engine: 'sqlite'` in your `vars.yml` file)
- re-run the playbook: `ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start`

roles/matrix-bridge-appservice-discord/defaults/main.yml
spantaleev: Haha.. Good point! Switching from `_postgres_` back to `_database_` does make some sense.

The protocol schema thing can be its own variable too (e.g. `matrix_appservice_discord_database_protocol_schema`).

jdreichmann: we _could_ extract the schema from the `*_database_engine`, but for now, i think it's fine w/ just `*_database_engine` and building the connString with a static scheme (postgres)

roles/matrix-bridge-appservice-discord/tasks/setup_install.yml
spantaleev: An alternative to both auto-migration and erroring-out like we do now is to: detect that there's leftover SQLite data and automatically switch to using SQLite when generating the configuration.

We could leave the `matrix_postgres_additional_databases` definitions in place and have these empty databases get created and left unused.

Whenever we detect that leftover data and auto-switch to SQLite, we could push a warning (notice) message to be displayed at the end of playbook execution. We do such a thing for major Postgres upgrades. We automatically keep people on the old version (instead of trying to auto-upgrade them), and we tell them with a message, letting them run the upgrade whenever they feel comfortable. Code is here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/6e1dfb62f0d371fdb82fc7a9e635dd32d0179e86/roles/matrix-postgres/tasks/setup_postgres.yml#L21-L31

What do you think? To me, it sounds reasonable and simple to implement.. It's also not dangerous.

roles/matrix-bridge-appservice-discord/defaults/main.yml
spantaleev: I've restored naming from `_postgres_` to `_database_` in 087dbe4ddc80ba. I guess this discussion can be resolved?

roles/matrix-bridge-appservice-discord/tasks/setup_install.yml
jdreichmann: That adds an extra layer of "let the user choose", and i think it becomes harder and harder to keep the change small. ideally, we would automigrate everything and have the roles themselves with sqlite as default, supporting delaying, external postgres, etc etc blows up the scope, and users would except that a merged feature was tested with every combination.
TL;DR: could do, but might bite us later, and we should encourage users to migrate to postgres by making the path to migrate easy (encourage users to actually do it and not perpetually delay it)

spantaleev:
Thinking about ma1sd now.. That's something we have enabled by default for everyone. Asking for a manual (SQLite -> Postgres) migration sounds like a huge annoyance for a very large number of people.. I've looked into doing automatic migrations, as previously discussed.

Automatic migrations are implemented for `matrix-reminder-bot`. For ARM users, we fail early and tell them they should explicitly switch to the `sqlite` engine. For now, at least. We may soon be able to modify the importing utility (`roles/matrix-postgres/tasks/util/migrate_db_to_postgres.yml`) to support ARM.

Looking into enabling automatic migrations for the other bridges and for ma1sd.

spantaleev:
I've added (SQLite + Postgres) support to the remaining services.

In dd797ba, I've also fixed `--tags=upgrade-postgres` and `--tags=import-postgres` to not choke with all these new databases that we're creating.

I think that what remains to be done before we can merge this, is:
- fix up the `mx-puppet-*` bridges, so that they support automatic migration like all the other services
- find someone to test Slack/IRC bridge importing. If we can't, we can always default what we can't test to `sqlite` for now
- find someone to test the other bridges (those that support automatic migration already)

It might also be nice to get pgloader running on ARM, but we can also leave this for another time.

spantaleev:
Lately, I've:

- added automatic (SQLite -> Postgres) migration support to all `mx-puppet-*` bridges too (using pgloader)
- added automatic (nedb -> Postgres) migration support to appservice-irc and appservice-slack (using scripts that are part of their Docker images)
- added self-building support for the pgloader component. This allows us to do this "migrate everything to Postgres" thing on ARM at the same time we release it for `amd64`

I've tested as much as I can, especially given that I don't use most of these bridges. It's somewhat unfortunate that we didn't get much outside testing for this, so I guess we'll have to test it upon "release". I'll put up a warning about it in the changelog.

chagai95:
I tested this again today and I messed up a few things so it was not a standard test - it worked for a moment fine, then I added the relaybot for whatsapp by changing the template and changed the template of telegram as well to allow all to use it, got kicked out everywhere and the bridges are now regenerating rooms but it might be my fault for using the wrong template at the beginning which made them use sqlite because I used my old templates that had sqlite databases in them. Also facing a problem still with the WhatsApp bridge and haven't tested the slack bridge yet...

PR #743, opened 2020-12-05
Drop the old workaround for an Ansible bug that has been fixed three years ago

pushytoxin:
This PR cleans up an old workaround.

No issues running like this for at least six months.

```
$ ansible --version
ansible 2.9.13
...
```

### DIFF
--- a/roles/matrix-base/tasks/setup_matrix_base.yml
+++ b/roles/matrix-base/tasks/setup_matrix_base.yml
@@ -19,31 +19,10 @@
     mode: '0660'
   when: "matrix_vars_yml_snapshotting_enabled|bool"
 
-# `docker_network` doesn't work as expected when the given network
-# is a substring of a network that already exists.
-#
-# See:
-# - https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/12
-# - https://github.com/ansible/ansible/issues/32926
-#
-# Due to that, we employ a workaround below.
-#
-# - name: Ensure Matrix network is created in Docker
-#   docker_network:
-#     name: "{{ matrix_docker_network }}"
-#     driver: bridge
-
-- name: Check existence of Matrix network in Docker
-  shell:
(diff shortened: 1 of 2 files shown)


pushytoxin:
What versions of Ansible does the playbook support (what versions of Ansible are people using with the playbook)?

aaronraimist:
Currently the playbook supports Ansible 2.7.0 or newer https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/ansible.md#supported-ansible-versions

pushytoxin:
The fix was backported and appears in Ansible tags v2.6.8 and v2.7.1

Delay merging this until the minimum supported ansible version gets bumped to v2.7.1 then.

spantaleev:
I guess distros that are on 2.7.x should have 2.7.1 (or newer) available to them, so we may bump the requirements.

I've tested these distros and the version of Ansible in their default repositories as of right now (2021-01-03):

- Ubuntu 16.04: Ansible 2.0.0.2 on Python 2.7.12
- Ubuntu 16.10: *could not check; non-LTS and too old, so repositories have been moved to archive or something*
- Ubuntu 18.04: Ansible 2.5.1 on Python 2.7.17
- Ubuntu 18.10: *could not check; non-LTS and too old, so repositories have been moved to archive or something*
- Ubuntu 20.04: Ansible 2.9.6 on Python 3.8.5
- Ubuntu 20.10: Ansible 2.9.9 on Python 3.8.6
- Debian 9: Ansible 2.2.1.0 on Python 2.7.13
- Debian 10: Ansible 2.7.7 on Python 3.7.3
- Centos 7: Ansible 2.9.16 on Python 2.7.5

Seems like most distros either have newer than 2.7.1 or are on a too-old version (in which case people either run Ansible in Docker or can install newer via `pip` or something).

I've bumped the minimum Ansible version to 2.7.1 in 6e652e10 and 2c09111a3, and have merged this!

Thank you! :+1:

PR #744, opened 2020-12-06
Remove the recording button from the Jitsi UI if recording is disabled.

hardye:
Recording in Jitsi was disabled by default in commit e0d7d5f0, but the button is still visible in the UI. This PR removes the button from the Jitsi Web toolbar if recording is disabled.


spantaleev:
Awesome! Thanks! :+1:

PR #745, opened 2020-12-06
fixes config.js

wabuMike:
comments in config.js must begin with two slashes


spantaleev:
Good catch! Thank you! :+1:

PR #757, opened 2020-12-20
add option to disable nginx access log

0x46616c6b:
Would be nice to have an option to disable the nginx access log for the proxy. This should be optional but possible.


spantaleev:
Looks good!

But it would make more sense to call this variable `matrix_nginx_proxy_access_log_enabled` and have it default to `true`. This makes it consistent with other variables in the playbook (`*_enabled`).

Indentation seems a little off in `roles/matrix-nginx-proxy/templates/nginx/nginx.conf.j2`.

0x46616c6b:
I updated the PR. Thanks for the quick response and the useful hints.

spantaleev:
Looks good! Thank you! :+1:

PR #758, opened 2020-12-21
added section to import database with a password

PC-Admin:
Noticed the import command doesn't work when a postgresql connection password is set. Throws this error:

`psql: error: could not connect to server: FATAL:  password authentication failed for user \"synapse\"`

Here is a small PR I've made that sets the connection password as an environmental variable in the postgres container, if it's actually defined. Cheers!

### DIFF
--- a/roles/matrix-postgres/tasks/import_postgres.yml
+++ b/roles/matrix-postgres/tasks/import_postgres.yml
@@ -72,11 +72,32 @@
       --mount type=bind,src={{ server_path_postgres_dump }},dst=/{{ server_path_postgres_dump|basename }},ro
       --entrypoint=/bin/sh
       {{ matrix_postgres_docker_image_latest }}
+      -c "export PGPASSWORD='{{ matrix_postgres_connection_password }}'; cat /{{ server_path_postgres_dump|basename }} |
+      {{ 'gunzip |' if server_path_postgres_dump.endswith('.gz') else '' }}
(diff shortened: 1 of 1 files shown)


spantaleev:
Hey! Thanks, but I don't see why this is necessary.

By default, Postgres has a password for us. `group_vars/matrix_servers` sets `matrix_postgres_connection_password`:

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/d0ee86e0a54f5ae93c872550d8b66f5b660d0b1d/group_vars/matrix_servers#L875

This password then ends up as `PGPASSWORD=...` in `{{ matrix_postgres_base_path }}/env-postgres-psql`: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/d0ee86e0a54f5ae93c872550d8b66f5b660d0b1d/roles/matrix-postgres/templates/env-postgres-psql.j2#L2-L4

We pass these environment variables to the import `docker run` command using `--env-file`:

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/d0ee86e0a54f5ae93c872550d8b66f5b660d0b1d/roles/matrix-postgres/tasks/import_postgres.yml#L71

So it should work. And it's been tested to work. Our Postgres database has always had a password, albeit a predictable one that we use for every installation :/

PC-Admin:
Makes sense! It's possible it failed for some other reason perhaps. I'll have another look into it during my next export/import, apologies! :)

PR #766, opened 2020-12-23
Fix concatenation of additional databases

lxp:
Otherwise the postgres upgrade fails with the following error:

```
Unexpected templating type error occurred on ({{
  [matrix_postgres_connection_username]
  +
  matrix_postgres_additional_databases|map(attribute='username')
}}
): can only concatenate list (not "generator") to list
```


spantaleev:
I haven't had an issue with, but you're probably on some other (older?) Ansible version. I'll take your word for it. Thank you! 👍🏻

lxp:
Just for completeness, I was using Ansible 2.9.16.

PR #769, opened 2020-12-27
Check for buggy version of Ansible that Ubuntu 20.04 provides

aaronraimist:
I have no idea if this works, I haven't tested it but it should theoretically check for the buggy version of Ansible that Ubuntu 20.04 comes with and provide a nicer warning message.


spantaleev:
Haven't tested it, but it looks good! Thank you! :+1:

PR #772, opened 2020-12-29
Example Files for Caddy2

0hlov3:
Added the tls directive again, but added a Readme-File to link to the Documentation and Describe a bit of the directives.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,5 +1,5 @@
 [defaults]
 retry_files_enabled = False
-ansible_python_interpreter=/usr/bin/python3
+
 [connection]
 pipelining = True
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,5 +1,5 @@
 [defaults]
 retry_files_enabled = False
-
+ansible_python_interpreter=/usr/bin/python3
 [connection]
 pipelining = True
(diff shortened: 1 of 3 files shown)


spantaleev:
Thanks!

I have a few questions

- why discard the log output for Jitsi?
- it looks like you've gone for a configuration that is commented by default (likely because Jitsi is an optional service). However, Dimension also is, yet we're not doing that. I wonder if we should comment the Dimension configuration just the same or uncomment the Jitsi one
- without the `tls` options, is a certificate not created by default? Or does it error out if you don't define a global configuration. The current vhost configuration makes me think that a certificate will not be created at all

0hlov3:
> Thanks!
> 
> I have a few questions
> 
> * why discard the log output for Jitsi?
Ah that's the default setting on my Server to discard the Logs, I'll edit it, so that logs will be written.
> * it looks like you've gone for a configuration that is commented by default (likely because Jitsi is an optional service). However, Dimension also is, yet we're not doing that. I wonder if we should comment the Dimension configuration just the same or uncomment the Jitsi one
Yes I think, I will comment the Dimension block also, cause its not a default.
> * without the `tls` options, is a certificate not created by default? Or does it error out if you don't define a global configuration. The current vhost configuration makes me think that a certificate will not be created at all
Without the tls option, no tls is set for the Site and not certificate will be created, you can add your own certificates to the tls block, but if you want to use lets encrypt, you have to add your ACME-E-Mail-Address.

I will edit the pull request and update it.

spantaleev:
Thanks! :+1:

PR #778, opened 2021-01-09
Call API list tokens enpoint

june07:
ansible-playbook -i inventory/hosts setup.yml --tags=list-matrix-registration-tokens


spantaleev:
Thanks! 👍 

I've documented it in 0f9be8321c

PR #786, opened 2021-01-12
Local rebuild for Telegram

drpaneas:
I have tested it today with my RPi4 and it works. Without this, you cannot use the telegram bridge unless you have a 64-bit intel/amd processor.

### DIFF
--- a/docs/configuring-playbook-bridge-mautrix-telegram.md
+++ b/docs/configuring-playbook-bridge-mautrix-telegram.md
@@ -12,18 +12,6 @@ matrix_mautrix_telegram_api_id: YOUR_TELEGRAM_APP_ID
 matrix_mautrix_telegram_api_hash: YOUR_TELEGRAM_API_HASH
 ```
 
-**Note**: For architectures different from `amd64` you need to trigger a local rebuild:
-
-```yaml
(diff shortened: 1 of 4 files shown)


docs/configuring-playbook-bridge-mautrix-telegram.md
```diff
@@ -12,6 +12,17 @@ matrix_mautrix_telegram_api_id: YOUR_TELEGRAM_APP_ID
 matrix_mautrix_telegram_api_hash: YOUR_TELEGRAM_API_HASH
 ```
 
+**Note**: For architectures different from `amd64` you need to trigger a local rebuild:
```
spantaleev, line 15: None of this is necessary.

For other bridges we do it automatically. See `group_vars/matrix_servers`.

We can do the same for this one by adding `matrix_mautrix_telegram_container_self_build: "{{ matrix_architecture != 'amd64' }}"` there.

spantaleev:
Thanks for adding self-building support to the Telegram bridge!

Please make self-building automatic on non-`amd64` as described in my other comment and let's merge this!

spantaleev:
I've fixed up a few minor things and merged this. Thank you! 👍

PR #793, opened 2021-01-13
[WIP] appservice-slack: use arm64/amd64 images; selfbuild for others

drpaneas:
Currently this bridge uses only the 64-bit image. However there is already an available image for arm64 provided from the same maintainer (see [here](https://mau.dev/tulir/mautrix-telegram/container_registry/1)).

My intention with this PR is to pull the correct image (for both amd64 and arm64) based on `matrix_architecture`. For all the others, trigger a rebuild.

Sadly, this currently fails with:

```
TASK [matrix-bridge-appservice-slack : Ensure Matrix Appservice Slack config installed] **********************************************************************************************************************************************************************
fatal: [matrix.xxxxx.net]: FAILED! => {"changed": false, "checksum": "07486873673decfc8c694af1cf076ac788010922", "msg": "Destination directory /matrix/appservice-slack/config does not exist"}
```

Any help or pointers to what might be wrong would be greatly appreciated :)

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -110,7 +110,7 @@ matrix_appservice_webhooks_systemd_required_services_list: |
 # We don't enable bridges by default.
 matrix_appservice_slack_enabled: false
 
-matrix_appservice_slack_container_self_build: "{{ matrix_architecture != 'amd64' }}"
+matrix_appservice_slack_container_self_build: "{{ matrix_architecture not in ['amd64', 'arm64'] }}"
 
 # Normally, matrix-nginx-proxy is enabled and nginx can reach matrix-appservice-slack over the container network.
 # If matrix-nginx-proxy is not enabled, or you otherwise have a need for it, you can expose
@@ -444,7 +444,7 @@ matrix_mx_puppet_skype_database_password: "{{ matrix_synapse_macaroon_secret_key
 # We don't enable bridges by default.
 matrix_mx_puppet_slack_enabled: false
 
-matrix_mx_puppet_slack_container_image_self_build: "{{ matrix_architecture not in ['amd64', 'arm64'] }}"
+matrix_mx_puppet_slack_container_image_self_build: "{{ matrix_architecture != 'amd64'}}"
 
 matrix_mx_puppet_slack_systemd_required_services_list: |
   {{
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -110,6 +110,8 @@ matrix_appservice_webhooks_systemd_required_services_list: |
 # We don't enable bridges by default.
 matrix_appservice_slack_enabled: false
 
+matrix_appservice_slack_container_self_build: "{{ matrix_architecture != 'amd64' }}"
+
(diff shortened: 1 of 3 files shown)


group_vars/matrix_servers
```diff
@@ -442,7 +444,7 @@ matrix_mx_puppet_skype_database_password: "{{ matrix_synapse_macaroon_secret_key
 # We don't enable bridges by default.
 matrix_mx_puppet_slack_enabled: false
 
-matrix_mx_puppet_slack_container_image_self_build: "{{ matrix_architecture != 'amd64'}}"
+matrix_mx_puppet_slack_container_image_self_build: "{{ matrix_architecture not in ['amd64', 'arm64'] }}"
```
drpaneas, line 447: I think I found my error. I've got confused with the two slack bridges.

spantaleev:
[matrixdotorg/matrix-appservice-slack](https://hub.docker.com/r/matrixdotorg/matrix-appservice-slack) doesn't have `arm64` builds, yes.

Neither does [sorunome/mx-puppet-slack](https://hub.docker.com/r/sorunome/mx-puppet-slack), so I guess this pull request serves no purpose..?

drpaneas:
True, for some reason the maintainer builds in arm64 https://mau.dev/tulir/mautrix-telegram/container_registry/1
but they don't publish it in dockerhub.

Anyway, I will make adjustments to the PR to make it self-build for non `amd64` build.

spantaleev:
Just talked to @tulir.

Tagged images are multi-arch (`arm64` and `amd64`). I've adjusted the playbook to not self-build for these architectures. For `arm32` we still fall back to self-building.

drpaneas:
I just fetched your latest changes in the repo, and it fails on my RPi4:

```
failed: [matrix.boobalar.net] (item=matrix-appservice-slack) => {"ansible_loop_var": "item", "changed": false, "item": "matrix-appservice-slack", "msg": "matrix-appservice-slack was not detected to be running. It's possible that there's a configuration problem or another service on your server interferes with it (uses the same ports, etc.). Try running `systemctl status matrix-appservice-slack` and `journalctl -fu matrix-appservice-slack` on the server to investigate."}
```

The error:

```
$ sudo /usr/bin/env docker run --rm --name matrix-appservice-slack --log-driver=none --user=997:1001 --cap-drop=ALL --network=matrix -v /matrix/appservice-slack/config:/config:z -v /matrix/appservice-slack/data:/data:z docker.io/matrixdotorg/matrix-appservice-slack:release-1.5.0 node app.js -p 9004 -c /config/config.yaml -f /config/slack-registration.yaml


docker: Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"node\": executable file not found in $PATH": unknown.
```

spantaleev:
That's normal. appservice-slack only provides an `amd64` image and we don't even support self-building for it.

So running it on a Pi results in the `amd64` image being pulled and that image can't work there.

We need to add self-building support, similar to how you've done it for mautrix-telegram in #786.

I see that you've started on it here. Maybe it should work..? Your code has some references to "Mautrix Telegram" though - you may wish to fix them.

PR #796, opened 2021-01-15
Enable message retention configuration policy.

brndnmtthws:
This allows people to enable message retention policies on their homeserver.

One note: there are a few extra features that are available, but I decided to err on the side of simplicity in the implementation. For example, `purge_jobs` takes an array with a bunch of possible parameters, but I just simplified it and stuck with the defaults. If preferred, I can change the settings so that `purge_jobs` uses an array for the config values instead of extracting the parameters.

### DIFF
--- a/docs/configuring-playbook-synapse.md
+++ b/docs/configuring-playbook-synapse.md
@@ -26,3 +26,13 @@ Certain Synapse administration tasks (managing users and rooms, etc.) can be per
 ## Synapse + OpenID Connect for Single-Sign-On
 
 If you'd like to use OpenID Connect authentication with Synapse, you'll need some additional reverse-proxy configuration (see [our nginx reverse-proxy doc page](configuring-playbook-nginx.md#synapse-openid-connect-for-single-sign-on)).
+
+## Message Retention Policy
+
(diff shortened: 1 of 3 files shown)


spantaleev:
Thank you! 👍 

While this is very nice, I wonder if we should keep introducing more and more dedicated variables for controlling `homeserver.yaml`. There are hundreds of options there (some more flexible and harder to control than others).

Generally, it's nice if we can keep things simpler (less dedicated options in the playbook) and rely on the universal mechanism we have for extending the configuration - `matrix_synapse_configuration_extension_yaml`. Using that, anyone can override and inject any configuration key, and we won't have  to keep growing the list of variables we define as part of the playbook.

On a related note, `retention.enabled` is a very unfortunate name for this configuration option (but that's the fault of the Synapse folks). It defaults to `false`, which seems to imply that nothing is retained at all. `retention_policy.enabled` may have been a better name.

brndnmtthws:
I'm a little unclear as to where your comment means you want me to update the PR or not. If there is something specific you want me to change, please let me know.

robbyoconnor (approved):
I love this!

spantaleev:
This has been superseded by 5bf09f5fdcc0fd7c7125432e659bb3129996bc25 and various fixup patches that followed it.

PR #797, opened 2021-01-17
Add Synapse OIDC configuration through variables

### LINKED ISSUE #784: OIDC configurable for Synapse
OIDC should be configurable. 

oidc_config variable in roles/matrix-synapse/templates/synapse/homeserver.yaml.j2

I need this for my employer, so i might put together a merge request.

tntclaus:
I did it for myself but noticed that there were other requests for same functionality (e.g. #784). 

There is synapse template and documentation in this pull request.

### DIFF
--- a/docs/configuring-playbook-synapse.md
+++ b/docs/configuring-playbook-synapse.md
@@ -27,8 +27,7 @@ matrix_synapse_password_config_enabled: true
 Add this to allow seamless forwarding to element web app and element android app. Without this setting matrix will ask the user if he trusts the app he tries to login.  
 
 ```yaml
-matrix_synapse_sso:
-  client_whitelist:
+matrix_synapse_sso_client_whitelist:
   - "https://element.{{ matrix_domain }}/"
   - element://element
 
--- a/docs/configuring-playbook-synapse.md
+++ b/docs/configuring-playbook-synapse.md
@@ -17,6 +17,35 @@ Alternatively, **if there is no pre-defined variable** for a Synapse setting you
 
 - or, if extending the configuration is still not powerful enough for your needs, you can **override the configuration completely** using `matrix_synapse_configuration` (or `matrix_synapse_configuration_yaml`). You can find information about this in [`roles/matrix-synapse/defaults/main.yml`](../roles/matrix-synapse/defaults/main.yml).
 
+## Synapse OpenID Connect
+
+If you plan to rely only on SSO and don't want your users to ever login with password credentials disable this:
+```yaml
+matrix_synapse_password_config_enabled: true
+```
+
+Refer to [official doc](https://github.com/matrix-org/synapse/blob/develop/docs/openid.md) for examples.
+
+
+This is the bare minimum config example:
+```yaml
+# Configure open id connect
+matrix_synapse_oidc_enabled: true
+matrix_synapse_oidc_issuer: "https://openid.example.com"
+matrix_synapse_oidc_client_id: "your-client-id"
+matrix_synapse_oidc_client_secret: "s0m3v3RyS3cr3tStr!ng"
+```
+
+Depending on oauth provider you may also like to change following settings:
+
+```yaml
+matrix_synapse_oidc_scopes: ["openid"]
+matrix_synapse_oidc_allow_existing_users: false
+matrix_synapse_oidc_client_auth_method: client_secret_basic
+matrix_synapse_oidc_subject_claim: "sub"
+matrix_synapse_oidc_localpart_template: "{% raw %}{{ user.preferred_username }}{% endraw %}"
+matrix_synapse_oidc_display_name_template: "{% raw %}{{ user.name }}{% endraw %}"
+```
 
 ## Synapse Admin
 
--- a/docs/configuring-playbook-synapse.md
+++ b/docs/configuring-playbook-synapse.md
@@ -24,7 +24,23 @@ If you plan to rely only on SSO and don't want your users to ever login with pas
 matrix_synapse_password_config_enabled: true
 ```
 
-Refer to [official doc](https://github.com/matrix-org/synapse/blob/develop/docs/openid.md) for examples.
+Add this to allow seamless forwarding to element web app and element android app. Without this setting matrix will ask the user if he trusts the app he tries to login.  
+
+```yaml
+matrix_synapse_sso:
+  client_whitelist:
+  - "https://element.{{ matrix_domain }}/"
+  - element://element
+
+```
+
+If you use `nginx proxy role` then you should add this setting as well to successfully pass redirects: 
+```yaml
+matrix_nginx_proxy_proxy_matrix_client_api_forwarded_location_synapse_oidc_api_enabled: true
+
+```
+
+Refer to [official Synapse doc](https://github.com/matrix-org/synapse/blob/develop/docs/openid.md) for oidc setup examples.
 
 
 This is the bare minimum config example:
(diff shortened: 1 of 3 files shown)


docs/configuring-playbook-synapse.md
```diff
@@ -24,7 +24,23 @@ If you plan to rely only on SSO and don't want your users to ever login with pas
 matrix_synapse_password_config_enabled: true
 ```
 
-Refer to [official doc](https://github.com/matrix-org/synapse/blob/develop/docs/openid.md) for examples.
+Add this to allow seamless forwarding to element web app and element android app. Without this setting matrix will ask the user if he trusts the app he tries to login.  
+
+```yaml
+matrix_synapse_sso:
```
spantaleev, line 30: I don't see `matrix_synapse_sso` being defined and used anywhere. Is something missing from your branch?

tntclaus: matrix_synapse_sso is used in `matrix-synapse/templates/synapse/homeserver.yaml`:
https://github.com/eraga/matrix-docker-ansible-deploy/blob/master/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2#L1855

and in `matrix-synapse/defaults.yml`: 
https://github.com/eraga/matrix-docker-ansible-deploy/blob/master/roles/matrix-synapse/defaults/main.yml#L466

The last one should be fixed though (it is misconfigured), gimme a minute, I'll push another commit.

tntclaus: done

spantaleev: Why isn't this converted to something more flat?

```yaml
matrix_synapse_sso:
  client_whitelist: []
```

Seems like there's no point to having a multi-level object and then having to reach into it?

tntclaus: To keep same structure as synapse has.

spantaleev: This makes it harder for us to introduce new defaults that the user can use as-is, forcing them to always redefine the whole `matrix_synapse_sso` object.

For example, we can define: `matrix_synapse_sso_client_whitelist` and put a default value of empty in the role, and a new default in `group_vars/matrix_servers` like this:

```yaml

matrix_synapse_sso_client_whitelist: |
  {{
    []
    +
    ([matrix_server_fqn_element] if matrix_client_element_enabled else [])
  }}
```

If Synapse adds more keys under `sso`, we would be able to support them using variables like `matrix_sso_*`, and the `matrix_synapse_sso` object won't get in the way.

tntclaus: understood. I'll correct it

tntclaus: ok, done :)

docs/configuring-playbook-synapse.md
```diff
@@ -17,6 +17,51 @@ Alternatively, **if there is no pre-defined variable** for a Synapse setting you
...
+If you plan to rely only on SSO and don't want your users to ever login with password credentials disable this:
+```yaml
+matrix_synapse_password_config_enabled: true
+```
+
+Add this to allow seamless forwarding to element web app and element android app. Without this setting matrix will ask the user if he trusts the app he tries to login.  
+
+```yaml
+matrix_synapse_sso_client_whitelist:
+  - "https://element.{{ matrix_domain }}/"
```
spantaleev, line 31: This can be `- "{{ matrix_server_fqn_element }}"`.

Should we consider having these as defaults even, instead of asking people to set them.

The `matrix-synapse` role could have an empty list (`matrix_synapse_sso_client_whitelist: []`), while `group_vars/matrix_servers` does something like this:

```yaml
# We always whitelist `element://element` (even if the web client is disabled).
# Perhaps this could be made configurable, if necessary.
matrix_synapse_sso_client_whitelist: |
  {{
    (["{{ matrix_server_fqn_element }}"] if matrix_client_element_enabled else [])
    +
    ['element://element']
  }}
```

I haven't tested this, by the way.

docs/configuring-playbook-synapse.md
```diff
@@ -17,6 +17,51 @@ Alternatively, **if there is no pre-defined variable** for a Synapse setting you
...
+  - element://element
+
+```
+
+If you use `nginx proxy role` then you should add this setting as well to successfully pass redirects: 
+```yaml
+matrix_nginx_proxy_proxy_matrix_client_api_forwarded_location_synapse_oidc_api_enabled: true
```
spantaleev, line 38: Can we also automatically toggle this via `group_vars/matrix_servers` based on some other variable?

```yaml
matrix_nginx_proxy_proxy_matrix_client_api_forwarded_location_synapse_oidc_api_enabled: "{{ matrix_synapse_oidc_enabled if matrix_synapse_enabled else false }}"
```

It makes things easier for people.

docs/configuring-playbook-synapse.md
```diff
@@ -17,6 +17,51 @@ Alternatively, **if there is no pre-defined variable** for a Synapse setting you
 
 - or, if extending the configuration is still not powerful enough for your needs, you can **override the configuration completely** using `matrix_synapse_configuration` (or `matrix_synapse_configuration_yaml`). You can find information about this in [`roles/matrix-synapse/defaults/main.yml`](../roles/matrix-synapse/defaults/main.yml).
 
+## Synapse OpenID Connect
+
+If you plan to rely only on SSO and don't want your users to ever login with password credentials disable this:
```
spantaleev, line 22: This part can be confusing, I think.

We say "disable this", but.. users don't have it explicitly enabled in their `vars.yml` file. Maybe they don't even know what it is.
It's not ready for copy-pasting, it needs to be adapted and you need to be careful to read the text above the code block.

Perhaps we should just put it somewhere else (down there with the other settings, below the bare-minimum config you've given). It could default to `true`, as it is here, and there could be a comment saying "you may wish to disable this".

Alternatively (even better, I think), it can be like this:

```yaml
# Some other settings here

# Authenticating against Synapse's localdb or other password providers is also possible by default.
# The former is controlled by `matrix_synapse_password_config_localdb_enabled`.
# Uncomment the line below to disable all authentication methods other than SSO.
# matrix_synapse_password_config_enabled: false
```

roles/matrix-synapse/defaults/main.yml
```diff
@@ -426,3 +426,47 @@ matrix_synapse_configuration_extension: "{{ matrix_synapse_configuration_extensi
 # Holds the final Synapse configuration (a combination of the default and its extension).
 # You most likely don't need to touch this variable. Instead, see `matrix_synapse_configuration_yaml`.
 matrix_synapse_configuration: "{{ matrix_synapse_configuration_yaml|from_yaml|combine(matrix_synapse_configuration_extension, recursive=True) }}"
+
+
+# oidc / sso
+
+# Set to false if you need only SSO auth
+matrix_synapse_password_config_enabled: true
```
spantaleev, line 434: Maybe this variable should be moved closer to `matrix_synapse_password_config_localdb_enabled`, because they belong together. It may be nice to have a comment about it there, because it's a little confusing which one does what.

spantaleev (changes requested):
I think we can make the docs a little simpler and have more automation around things.

hungrymonkey:
Hello, I want to tell notify that upstream landed support for multiple ODIC provider support

https://github.com/matrix-org/synapse/pull/9110

spantaleev:
I think this makes it even more obvious that we shouldn't try to introduce more and more variables and complexity and just let the very few people who want OIDC to just use `matrix_synapse_configuration_extension_yaml` to inject whatever configuration they'd like.

tntclaus:
> I think this makes it even more obvious that we shouldn't try to introduce more and more variables and complexity and just let the very few people who want OIDC to just use `matrix_synapse_configuration_extension_yaml` to inject whatever configuration they'd like.

agree

PR #801, opened 2021-01-19
When validating LE certs, do not wait for a random time

pushytoxin:
While administering we will occasionally invoke this script interactively with the "non-interactive" switch still there, yet still sit at the desk waiting for 300 seconds for this timer to run out.

The systemd-timer already uses a 3h randomized delay for automatic renewals, which serves this purpose well.


spantaleev:
Great idea! Thanks! :+1:

PR #804, opened 2021-01-20
Self-hosted Etherpad

pushytoxin:
The role takes the official Etherpad docker image, exposes it at `https://dimension.<your-domain>/etherpad`

To save time, many lines were copied from the Dimension role.

TODO:

- [x] nginx proxy rules
- [x] ~~Dimension defaults to self-hosted Etherpad~~
- [x] use the matrix-postgres database
- [x] remove the redundant "Chat" plugin
- [x] verify Jitsi can actually embed a 'pad
- [x] ~~transmit the user alias~~ *Jitsi has some tricks to get this done, it's outside of the scope of this PR*
- [x] ~~arm self-build~~
- [x] ensure Dimension is enabled when installing Etherpad

### DIFF
--- a/roles/matrix-etherpad/defaults/main.yml
+++ b/roles/matrix-etherpad/defaults/main.yml
@@ -31,7 +31,7 @@ matrix_etherpad_base_url: "https://{{ matrix_server_fqn_dimension }}{{ matrix_et
 
 # Database-related configuration fields.
 #
-# Etherpad recommends using a dedicated database, and supports Sqliite only for development
+# Etherpad recommends using a dedicated database, and supports Sqlite only for development
 #
 # To use Postgres:
 # - change the engine (`matrix_etherpad_database_engine: 'postgres'`)
--- a/roles/matrix-etherpad/defaults/main.yml
+++ b/roles/matrix-etherpad/defaults/main.yml
@@ -2,7 +2,7 @@ matrix_etherpad_enabled: false
 
 matrix_etherpad_base_path: "{{ matrix_base_data_path }}/etherpad"
 
-matrix_etherpad_docker_image: "docker.io/etherpad/etherpad:latest"
+matrix_etherpad_docker_image: "docker.io/etherpad/etherpad:1.8.7"
 matrix_etherpad_docker_image_force_pull: "{{ matrix_etherpad_docker_image.endswith(':latest') }}"
 
 # List of systemd services that matrix-etherpad.service depends on.
@@ -53,9 +53,9 @@ matrix_etherpad_database_connection_string: 'postgres://{{ matrix_etherpad_datab
 matrix_etherpad_title: 'Etherpad'
 matrix_etherpad_default_pad_text: |
   Welcome to Etherpad!
-  
+
   This pad text is synchronized as you type, so that everyone viewing this page sees the same text. This allows you to collaborate seamlessly on documents!
-  
+
   Get involved with Etherpad at https://etherpad.org
 
 # Default Etherpad configuration template which covers the generic use case.
--- a/roles/matrix-etherpad/defaults/main.yml
+++ b/roles/matrix-etherpad/defaults/main.yml
@@ -1,0 +1,93 @@
+matrix_etherpad_enabled: false
+
+matrix_etherpad_base_path: "{{ matrix_base_data_path }}/etherpad"
+
+matrix_etherpad_docker_image: "docker.io/etherpad/etherpad:latest"
+matrix_etherpad_docker_image_force_pull: "{{ matrix_etherpad_docker_image.endswith(':latest') }}"
+
+# List of systemd services that matrix-etherpad.service depends on.
+matrix_etherpad_systemd_required_services_list: ['docker.service']
+
+# List of systemd services that matrix-etherpad.service wants
+matrix_etherpad_systemd_wanted_services_list: []
+
+# Container user has to be able to write to the source file directories until this bug is fixed:
+# https://github.com/ether/etherpad-lite/issues/2683
+matrix_etherpad_user_uid: '5001'
+matrix_etherpad_user_gid: '5001'
+
+# Controls whether the matrix-etherpad container exposes its HTTP port (tcp/9001 in the container).
+#
+# Takes an "<ip>:<port>" or "<port>" value (e.g. "127.0.0.1:9001"), or empty string to not expose.
+matrix_etherpad_container_http_host_bind_port: '9001'
+
+# A list of extra arguments to pass to the container
+matrix_etherpad_container_extra_arguments: []
+
+matrix_etherpad_public_endpoint: '/etherpad'
+
+# By default, the Etherpad app can be accessed within the Dimension domain
+matrix_etherpad_base_url: "https://{{ matrix_server_fqn_dimension }}{{ matrix_etherpad_public_endpoint }}"
+
+# Database-related configuration fields.
+#
+# Etherpad recommends using a dedicated database, and supports Sqliite only for development
+#
+# To use Postgres:
+# - change the engine (`matrix_etherpad_database_engine: 'postgres'`)
+# - adjust your database credentials via the `matrix_etherpad_postgres_*` variables
+matrix_etherpad_database_engine: 'sqlite'
+
+matrix_etherpad_sqlite_database_path_local: "{{ matrix_etherpad_base_path }}/etherpad.db"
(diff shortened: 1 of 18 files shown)


roles/matrix-etherpad/templates/settings.json.j2
```diff
@@ -0,0 +1,106 @@
...
+  "dbType": {{ matrix_etherpad_database_engine|to_json }},
+  "dbSettings": {
+    {% if matrix_etherpad_database_engine == 'sqlite' %}
+      "filename": {{ matrix_etherpad_sqlite_database_path_in_container|to_json }}
+    {% elif matrix_etherpad_database_engine == 'postgres' %}
+      "database": {{ matrix_etherpad_database_name|to_json }},
+      "host":     {{ matrix_etherpad_database_hostname|to_json }},
+      "password": {{ matrix_etherpad_database_password|to_json }},
+      "port":     {{ matrix_etherpad_database_port|to_json }},
+      "user":     {{ matrix_etherpad_database_username|to_json }},
```
spantaleev, line 18: This trailing comma may be problematic.

pushytoxin: Works fine, but I'll make it tidy.

spantaleev:
Wow, that's a lot of work! Thanks a lot for starting on this! :+1: 

Thinking about Etherpad, as something which integrates with both Dimension and/or Jitsi, it feels a little strange to host it under `dimension.DOMAIN`. I can see it being useful even without Dimension.

I'm starting to think we should have some auxiliary services domain (`aux.DOMAIN`, `aux.matrix.DOMAIN`?) where we can host such things. Other services (like synapse-admin) can also be moved to that domain. In the future, I'm sure we'll have more and more services like this (where would Grafana go for example?)

We can control whether the playbook enables this aux domain (and obtains SSL certificates for it) using some other variable. This variable's value can also be determined dynamically based on the installed services (if Etherpad or synapse-admin is installed -> enable the AUX domain). We can ask people to set this domain up in our **Configuring DNS**.

I'd be happy to develop this aux domain part and relocate synapse-admin there for starters.

roles/matrix-etherpad/templates/settings.json.j2
pushytoxin: The config file is preprocessed, you should see the original. It has C style comments in it

pushytoxin:
That's a decision I didn't want to occupy myself with, I leave it up to you. I put it behind the dimension domain since it needs dimension anyway to work, and it kind of makes a bit of sense to me. But I see how it will confuse some, especially when they open https://dimension.DOMAIN/, which is basically a placeholder

spantaleev:
Does it really need Dimension to work though? Can't one use Etherpad + Jitsi (at least via `https://jitsi.DOMAIN`) or using Jitsi embedded into a room manually (without Dimension's help).

pushytoxin:
aux.DOMAIN doesn't say much.
Consider the *.widget.DOMAIN wildcard, then the proxy can decide where to route it then. and the DNS zone won't balloon

pushytoxin:
> Does it really need Dimension to work though? Can't one use Etherpad + Jitsi (at least via `https://jitsi.DOMAIN`) or using Jitsi embedded into a room manually (without Dimension's help).

Right now the menu item doesn't even show up, I'll see about jitsi later

pushytoxin:
> Does it really need Dimension to work though? Can't one use Etherpad + Jitsi (at least via `https://jitsi.DOMAIN`) or using Jitsi embedded into a room manually (without Dimension's help).

Can you configure the widget settings with the default integrations manager? Can you set it up serverside? I haven't used it much

spantaleev:
I was thinking a single domain (perhaps `services.matrix.DOMAIN`) under which we'll serve various services. The first examples would be this Etherpad service, but also synapse-admin, and maybe others which I haven't thought of yet. Things to come as well.

This way, it's **just 1 additional domain** people need to set up during DNS configuration and it will handle various services that people may enable now or in the future. If we keep asking people to create additional domains for each service, that'd be more of a hassle.

I guess certain services which we may add in the future may be prominent enough to deserve their own dedicated domain. The idea was to bundle all others together in a simple way.

Perhaps this needs more thought.

T3chTobi:
I agree with Slavi, one subdomain would be the best solution. `services.matrix.DOMAIN` sound good!

roles/matrix-etherpad/defaults/main.yml
```diff
@@ -51,7 +51,12 @@ matrix_etherpad_database_connection_string: 'postgres://{{ matrix_etherpad_datab
 
 # Variables configuring the etherpad
 matrix_etherpad_title: 'Etherpad'
-matrix_etherpad_default_pad_text: 'Welcome to Etherpad!\n\nThis pad text is synchronized as you type, so that everyone viewing this page sees the same text. This allows you to collaborate seamlessly on documents!\n\nGet involved with Etherpad at https:\/\/etherpad.org\n'
+matrix_etherpad_default_pad_text: |
```
pushytoxin, line 54: Should we change the default text?

spantaleev: This default text seems nice and descriptive to me.

pushytoxin:
Jitsi working as intended:
image

pushytoxin:
I added a short guide on changing the default Etherpad URL instead of messing with the database from within Ansible. On top of being more work, more complex than what it is worth, it would override the user's settings.

Same goes with the "Chat" button. Jitsi removes it, but it is visible inside Element. The docs explain how it can be removed.

pushytoxin:
I am unable to test the self-built ARM Docker images. I conclude this PR

roles/matrix-etherpad/tasks/validate_config.yml
```diff
@@ -0,0 +1,7 @@
+- name: Fail if required Etherpad settings not defined
+  fail:
+    msg: >-
+      You need to define a required configuration setting (`{{ item }}`) for using Etherpad.
+  with_items:
+    -
```
spantaleev, line 6: Anything to add here?

spantaleev:
I haven't tested this, but it looks great! I think we can merge it.

Perhaps we can add some checks in `validate_config.yml` to ensure Dimension is really enabled. It's described in the prerequisites, but it would be even better if we catch common mistakes like that.

--------

I also wonder whether we should merge this as-is and worry about a migration path and breaking people's widgets later on, or if I should work on the `services.matrix.DOMAIN` idea now, so we that this can launch with that and not have to be touched. When it comes to this PR, I think it would only be some minor redoing around URLs and around which variable you inject the nginx configuration into.

roles/matrix-etherpad/tasks/validate_config.yml
pushytoxin: Good catch! That came by copypasting Dimension tasks.
That would be the right place to verify that Dimension is installed.

pushytoxin:
> I also wonder whether we should merge this as-is and worry about a migration path and breaking people's widgets later on, or if I should work on the `services.matrix.DOMAIN` idea now, so we that this can launch with that and not have to be touched. When it comes to this PR, I think it would only be some minor redoing around URLs and around which variable you inject the nginx configuration into.

Adding `services.matrix.DOMAIN` at this point is early, when there's only one "service".
The current setup mimics the way it works on the original default, scalar.vector.im (scalar.vector.im/etherpad). I argue it's fine as it is.
Give it a few days of thought! I can change it later.

spantaleev:
True. There aren't too many things to put at `services.matrix.DOMAIN` right now (just this and synapse-admin that I can think of right now).

Dimension may also belong there (`services.matrix.DOMAIN/dimension/`) as well..? It's usually not accessed directly by users, so it having its own dedicated friendly domain/URL is not too important.

Same for Jitsi.

If we go for that, we can just ask people to set up these 3 domains:
- `matrix.DOMAIN`
- `services.matrix.DOMAIN`
- `element.DOMAIN`

And then they'd be set "for life". Enabling Jitsi later on, no problem. Enabling Dimension, no problem. Etc.

We could do this later though. Just food for thought.

pushytoxin:
It looks good to me, but this is an intrusive change, that needs additional administration when updating. I wouldn't hurry with the decision.

spantaleev:
I've switched the container image tag from `latest` to `1.8.7` and have merged this.

Thanks a lot for bringing Etherpad support to the playbook and making it integrate nicely with both Dimension and Jitsi! :+1:

PR #807, opened 2021-01-22
Fix typos

thedanbob:
Thanks for the clear instructions by the way, making this change went without a hitch.


spantaleev:
Thank you! :+1:

PR #809, opened 2021-01-22
Use the correct variable name in Postgres password length check.

hardye:
This PR fixes a small error when verifying the maximum PostgreSQL password length. The check mistakenly looked at the variable `matrix_postgres_connection_hostname` instead of `matrix_postgres_connection_password`.


hardye:
Never mind. I see the problem was already fixed in [f9968b6981448f](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/f9968b6981448f21bf886d405dd0c4d32dc75164).

spantaleev:
Yep! Thanks anyway! :+1:

PR #819, opened 2021-01-24
small fix to CHANGELOG.md

jolly-jump:
propose explicit showing single quotes around the password, since I forgot to put them there.


spantaleev:
They're not really required (depending on the password) but it's safer having them there. Thanks! :+1:

PR #831, opened 2021-01-26
Correct some typos in FAQ

phirz:
Thanks for the nice FAQ. While reading it, I tried to correct what seemed to be typos to me. Feel free to just discard the changes if they are not helpful.


spantaleev:
Ohh.. I guess I wrote and pushed that FAQ way too quickly! Thank you for correcting all these! :+1:

Issue #833, opened 2021-01-28
rust-synapse-compress-state: "state_group_edges" table doesn't exist


pushytoxin:
Trying to run the "rust-synapse-compress-state" task fails with the following error:

```json
{
  "ansible_job_id": "788848754618.62521",
  "changed": true,
  "cmd": [
    "/usr/bin/env",
    "docker",
    "run",
    "--rm",
    "--name",
    "matrix-rust-synapse-compress-state-psql-import",
    "--user=991:991",
    "--cap-drop=ALL",
    "--network=matrix",
    "--env-file=/matrix/postgres/env-postgres-psql",
    "--mount",
    "type=bind,src=/matrix/rust-synapse-compress-state,dst=/work,ro",
    "--entrypoint=/bin/sh",
    "docker.io/postgres:13.1-alpine",
    "-c",
    "cat /work/state-compressor.sql | psql -v ON_ERROR_STOP=1 -h matrix-postgres"
  ],
  "delta": "0:00:01.371347",
  "end": "2021-01-28 17:20:05.751054",
  "finished": 1,
  "msg": "non-zero return code",
  "rc": 3,
  "start": "2021-01-28 17:20:04.379707",
  "stderr": "ERROR:  relation \"state_group_edges\" does not exist\nLINE 1: DELETE FROM state_group_edges WHERE state_group = 288465;\n                    ^",
  "stderr_lines": [
    "ERROR:  relation \"state_group_edges\" does not exist",
    "LINE 1: DELETE FROM state_group_edges WHERE state_group = 288465;",
    "                    ^"
  ],
  "stdout": "BEGIN",
  "stdout_lines": [
    "BEGIN"
  ]
}
```

pushytoxin:
What kind of magic made this run without errors prior to having the database name in the `psq` command?

spantaleev:
It previously worked, because the default user was `synapse` and the default database was `homeserver`, so it was going where it should have.

But after our large Postgres refactoring lately (95346f3117f2a3a67a5), the default user and database is `matrix` (Synapse's database is now named `synapse`).

PR #834, opened 2021-01-28
Name the Synapse database on state compression import

pushytoxin:
Fixes:
https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/833


spantaleev:
Thanks! :+1:

PR #840, opened 2021-01-29
Add mx-puppet-skype and mx-puppet-slack to README

aaronraimist:
I also moved matrix-sms-bridge up to match the order from container-images.md


spantaleev:
Thank you! :+1:

PR #841, opened 2021-01-30
Use Debian Buster Docker repo on Debian Bullseye

Peetz0r:
There is no Docker for Bullseye (Debian 11) yet, so I installed Docker for Buster (Debian 10) instead.

I have been running it for a few days now with most services enabled and haven't encountered any issues.

Future maintainer: check on https://docs.docker.com/engine/install/debian/ if Docker for Debian 11 is released, then undo this PR

### DIFF
--- a/roles/matrix-base/tasks/server_base/setup_debian.yml
+++ b/roles/matrix-base/tasks/server_base/setup_debian.yml
@@ -23,7 +23,14 @@
     repo: "deb [arch={{ matrix_debian_arch }}] https://download.docker.com/linux/{{ ansible_distribution|lower }} {{ ansible_distribution_release }} stable"
     state: present
     update_cache: yes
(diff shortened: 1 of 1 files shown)


spantaleev:
Overriding an Ansible-created fact seems a little dirty. I wonder if we should introduce some other variables for this instead (which default to those values and gets used in the repository "template"). You can then override those and avoid touching core facts.

Peetz0r:
Good point. Fixed :)

spantaleev:
Thanks! :+1:

RockSolidBrawler:
Leading to an error on updating mine installation on Ubuntu 18.04.

Is there any " ' " missing behind 'bullseye'  on lines 26 and 33?

gjelsas:
Same here, on Ubuntu 20.04!

Peetz0r:
@spantaleev People are waiting for you to merge #842 which fixes the above :)

PR #844, opened 2021-01-30
IPv6 support

Peetz0r:
`matrix_docker_ipv6_enabled` is true by default. I think IPv6 should be everywhere by now.

Nginx now always listens on ipv6.

When the setting is true, the `matrix_docker_network` gets created with ipv6 enabled. Because we are basically doing NATv6 between the host and the containers there is a service that sets up the ip6tables rule.

I also moved the `matrix-certbot` into the `matrix_docker_network` because in my ipv6-only tests it couldn't validate in the default `bridge` network.

### DIFF
--- a/roles/matrix-base/tasks/server_base/docker_ipv6.yml
+++ b/roles/matrix-base/tasks/server_base/docker_ipv6.yml
@@ -1,0 +1,47 @@
+---
+
+- block:
+  - name: Ensure matrix-ip6tables.service exists
+    template:
+      src: "{{ role_path }}/templates/{{ item }}.j2"
+      dest: "{{ matrix_systemd_path }}/{{ item }}"
+      owner: "root"
+      group: "root"
+      mode: 0644
+    with_items:
+      - matrix-ip6tables.service
+    register: matrix_ip6tables_systemd_service_result
+
+  - name: Ensure systemd reloaded after matrix-ip6tables.service installation
+    service:
+      daemon_reload: yes
+    when: "matrix_ip6tables_systemd_service_result.changed"
+  
+  - name: Ensure matrix-ip6tables.service is started and autoruns
+    service:
+      name: matrix-ip6tables
+      state: started
+      enabled: yes
+
+  when: "matrix_docker_ipv6_enabled|bool"
+
+
+- block:
+  - name: Check existence of matrix-ip6tables service
+    stat:
+      path: "{{ matrix_systemd_path }}/matrix-ip6tables.service"
+    register: matrix_ip6tables_service_stat
+  
+  - name: Ensure matrix-ip6tables.service doesn't exist
+    file:
+      path: "{{ matrix_systemd_path }}/matrix-ip6tables.service"
+      state: absent
+    when: "matrix_ip6tables_service_stat.stat.exists"
+  
+  - name: Ensure systemd reloaded after matrix-ip6tables.service removal
+    service:
+      daemon_reload: yes
+    when: "matrix_ip6tables_service_stat.stat.exists"
+    
+  when: "not matrix_docker_ipv6_enabled|bool"
+
--- a/roles/matrix-base/templates/matrix-ip6tables.service.j2
+++ b/roles/matrix-base/templates/matrix-ip6tables.service.j2
@@ -1,0 +1,16 @@
+#jinja2: lstrip_blocks: "True"
+[Unit]
+Description=Matrix ip6tables rule to enable IPv6 internet access from containers
+DefaultDependencies=no
+
+[Service]
+Type=oneshot
+Environment="HOME={{ matrix_systemd_unit_home_path }}"
+
+ExecStart={{ matrix_host_command_ip6tables }} -t nat -A POSTROUTING -s fd00::/80 ! -o docker0 -j MASQUERADE
+ExecStart={{ matrix_host_command_ip6tables }} -P FORWARD ACCEPT
+
+SyslogIdentifier=matrix-ip6tables
+
+[Install]
+WantedBy=multi-user.target
--- a/roles/matrix-base/defaults/main.yml
+++ b/roles/matrix-base/defaults/main.yml
@@ -76,6 +76,8 @@ matrix_host_command_fusermount: "/usr/bin/env fusermount"
 matrix_host_command_openssl: "/usr/bin/env openssl"
 matrix_host_command_systemctl: "/usr/bin/env systemctl"
 matrix_host_command_sh: "/usr/bin/env sh"
+matrix_host_command_iptables: "/usr/bin/env iptables"
+matrix_host_command_ip6tables: "/usr/bin/env ip6tables"
 
 matrix_ntpd_package: "ntp"
 matrix_ntpd_service: "{{ 'ntpd' if ansible_os_family == 'RedHat' or ansible_distribution == 'Archlinux' else 'ntp' }}"
@@ -114,6 +116,9 @@ matrix_client_element_e2ee_secure_backup_setup_methods: []
 
 # The Docker network that all services would be put into
 matrix_docker_network: "matrix"
+
+# Controls whether we'll enable IPv6 in docker
+matrix_docker_ipv6_enabled: true
 
 # Controls whether we'll preserve the vars.yml file on the Matrix server.
 # If you have a differently organized inventory, you may wish to disable this feature,
--- a/roles/matrix-base/tasks/server_base/setup.yml
+++ b/roles/matrix-base/tasks/server_base/setup.yml
@@ -27,6 +27,8 @@
 - include_tasks: "{{ role_path }}/tasks/server_base/setup_archlinux.yml"
   when: ansible_distribution == 'Archlinux'
 
+- include_tasks: "{{ role_path }}/tasks/server_base/docker_ipv6.yml"
+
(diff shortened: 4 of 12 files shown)


roles/matrix-base/tasks/server_base/docker_ipv6.yml
```diff
@@ -0,0 +1,52 @@
...
+      path: "{{ matrix_systemd_path }}/matrix-ip6tables.service"
+      state: absent
+    when: "matrix_ip6tables_service_stat.stat.exists"
+  
+  - name: Ensure systemd reloaded after matrix-ip6tables.service removal
+    service:
+      daemon_reload: yes
+    when: "matrix_ip6tables_service_stat.stat.exists"
+    
+  - name: Ensure /etc/docker/daemon.json doesn't exist
```
pushytoxin, line 46: Why are you deleting the users' docker configuration?

Peetz0r: Oops, that wasn't supposed to be there. That was mostly just an attempt at learning how this actually works.

pushytoxin:
You are running NAT6 in one direction, but I can't see FORWARD rules for the exposed sockets in the other. How does this even work?
Or does it rely on the userland proxy?

Peetz0r:
> You are running NAT6 in one direction, but I can't see FORWARD rules for the exposed sockets in the other. How does this even work?
> Or does it rely on the userland proxy?

Yes, it uses the userland proxy for incoming traffic, and it uses NAT6 for outgoing traffic.

As of a few minutes ago I am running this setup on glitch.im, so you see it working when you see [@peetz0r:glitch.im](https://matrix.to/#/@peetz0r:glitch.im) talking anywhere.

pushytoxin:
Last time I tried this, the nginx/coturn logs from IPv6 users and federated servers contained the userland proxy address, which is the same as the gateway. Is it still like that?

Peetz0r:
> Last time I tried this, the nginx/coturn logs from IPv6 users and federated servers contained nothing but the userland proxy address, which is the same as the gateway. Is it still like that?

Correct, they all get reported as `fd00::1`.

Peetz0r:
Also I did some more testing, and it's still not fully working. When I go inside a container I can reach only v4 servers on the internet somehow. This worked earlier and I have no clue what I am doing...

pushytoxin:
> > Last time I tried this, the nginx/coturn logs from IPv6 users and federated servers contained nothing but the userland proxy address, which is the same as the gateway. Is it still like that?
> 
> Correct, they all get reported as `fd00::1`.

That's kind of a bummer

roles/matrix-base/templates/matrix-ip6tables.service.j2
```diff
@@ -0,0 +1,15 @@
+#jinja2: lstrip_blocks: "True"
+[Unit]
+Description=Matrix ip6tables rule to enable IPv6 internet access from containers
+DefaultDependencies=no
+
+[Service]
+Type=oneshot
+Environment="HOME={{ matrix_systemd_unit_home_path }}"
+
+ExecStart=ip6tables -t nat -A POSTROUTING -s fd00::/80 ! -o docker0 -j MASQUERADE
```
aaronraimist, line 10:
```suggestion
ExecStart={{ matrix_host_command_sh }} -c "ip6tables -t nat -A POSTROUTING -s fd00::/80 ! -o docker0 -j MASQUERADE"
```

Changing this made it work for me. I don't know if this is the right solution.

Peetz0r: That'd be unnecessary shell invocation. It works, but it's not the best. Something involving an absolute path to `env` is probably the right solution. And I also forgot the second import `ip6tables` rule in there.

Like I said, this needs more testing before it's ready :p

Peetz0r:
@aaronraimist your ip6tables rules should now work.

> Also I did some more testing, and it's still not fully working. When I go inside a container I can reach only v4 servers on the internet somehow. This worked earlier and I have no clue what I am doing...

This should also be fixed now

Peetz0r:
> > > Last time I tried this, the nginx/coturn logs from IPv6 users and federated servers contained nothing but the userland proxy address, which is the same as the gateway. Is it still like that?
> > 
> > 
> > Correct, they all get reported as `fd00::1`.
> 
> That's kind of a bummer

I might spend some time next week on trying out https://github.com/robbertkl/docker-ipv6nat to fix this.

No guarantees, obviously.

roles/matrix-base/templates/matrix-ip6tables.service.j2
pushytoxin: It is recommended to put absolute paths into ExecStart=, in this case that would be `/usr/sbin/ip6tables`

aaronraimist:
Is there anything else that needs to happen here to merge this? It's been working for me. Seems like using https://github.com/robbertkl/docker-ipv6nat could happen in the future.

spantaleev:
`matrix_host_command_iptables` seems to be unused, so we can remove it.

I was wondering if we need to install `iptables` explicitly to ensure it's there, but it seems like it's pulled as a dependency for various things on various distros. On Archlinux, even `systemd` requires `iptables` to be there. On CentOS 7, `iptables` is also required by `docker-ce` one way or another. On Debian (well, Raspbian), `iptables` is some dependency for `containerd.io`. I guess we can safely assume that it's available.

aaronraimist:
Could we move this forward?

Peetz0r:
I guess I'm not gonna do the proper fix for the "everything is `fd00::1`" issue. But some support might still be better than none. Especially if it's disabled by default.

Having IPv6 disabled by default doesn't sit well with me, but having this PR sitting dormant for months is definitely worse.

Lemme mark as non-a-draft-anymore and get it into a state without conflicts.

roles/matrix-base/tasks/server_base/docker_ipv6.yml
```diff
@@ -0,0 +1,47 @@
...
+    with_items:
+      - matrix-ip6tables.service
+    register: matrix_ip6tables_systemd_service_result
+
+  - name: Ensure systemd reloaded after matrix-ip6tables.service installation
+    service:
+      daemon_reload: yes
+    when: "matrix_ip6tables_systemd_service_result.changed"
+  
+  - name: Ensure matrix-ip6tables.service is started and autoruns
```
spantaleev, line 20: We should probably inject this into `matrix_systemd_services_list` and start it as part of `--tags=start`.

On the other hand, maybe that makes it start too late? We probably wish to have working IPv6 connectivity for the purposes of obtaining SSL certificates, for example..? Is that why this is here?

CaptainMalu:
Any news about this topic?
I'm not good enough with docker and ansible to configure IPv6 support on my own.

I added this part into my vars.yaml to most likely only have to set enable ipv6 for the network creation part. But i'm not sure which dockernetworks need it and if it will work.

`docker_daemon_options:
  experimental:  true
  ip6tables:     true
  ipv6:          true
  fixed-cidr-v6: "fc01:1:1:1::/64"
  default-address-pools: [
    { "base": "172.17.0.0/16", "size": 16 },
    { "base": "172.18.0.0/16", "size": 16 },
    { "base": "172.19.0.0/16", "size": 16 },
    { "base": "172.20.0.0/14", "size": 16 },
    { "base": "172.24.0.0/14", "size": 16 },
    { "base": "172.28.0.0/14", "size": 16 },
    { "base": "192.168.0.0/16", "size": 20 },
    { "base": "fc01:1:1:2:1::/104", "size": 112 },
    { "base": "fc01:1:1:2:2::/104", "size": 112 },
    { "base": "fc01:1:1:2:3::/104", "size": 112 },
    { "base": "fc01:1:1:2:4::/104", "size": 112 },
    { "base": "fc01:1:1:2:5::/104", "size": 112 }
    ]`

ly1998117:
> Any news about this topic? I'm not good enough with docker and ansible to configure IPv6 support on my own.
> 
> I added this part into my vars.yaml to most likely only have to set enable ipv6 for the network creation part. But i'm not sure which dockernetworks need it and if it will work.
> 
> ```
> docker_daemon_options:
>   experimental:  true
>   ip6tables:     true
>   ipv6:          true
>   fixed-cidr-v6: "fc01:1:1:1::/64"
>   default-address-pools: [
>     { "base": "172.17.0.0/16", "size": 16 },
>     { "base": "172.18.0.0/16", "size": 16 },
>     { "base": "172.19.0.0/16", "size": 16 },
>     { "base": "172.20.0.0/14", "size": 16 },
>     { "base": "172.24.0.0/14", "size": 16 },
>     { "base": "172.28.0.0/14", "size": 16 },
>     { "base": "192.168.0.0/16", "size": 20 },
>     { "base": "fc01:1:1:2:1::/104", "size": 112 },
>     { "base": "fc01:1:1:2:2::/104", "size": 112 },
>     { "base": "fc01:1:1:2:3::/104", "size": 112 },
>     { "base": "fc01:1:1:2:4::/104", "size": 112 },
>     { "base": "fc01:1:1:2:5::/104", "size": 112 }
>     ]
> ```

I used ipv6nat to make the server work on my IPv6 VPS. According to my practice, the federation function works well. For more details, please refer to my repo [ipv6 support](https://github.com/ly1998117/matrix-docker-ansible-deploy-ipv6).

PR #851, opened 2021-02-03
Allow psql args to be given to matrix-postgres-cli

julianfoad:


spantaleev:
Convenient! Thanks! :+1:

PR #853, opened 2021-02-03
Remove note about federation tester not working with TLS 1.3

aaronraimist:
They finally deployed the 1 year old PR that fixed TLS 1.3 compatibility


spantaleev:
Great! :partying_face: Thank you!

Issue #849, opened 2021-02-01
Embedded etherpad complains can't be embedded


skepticalwaves:
I updated the dimension settings per the docs:
```
https://dimension.<your-domain>/etherpad/p/$roomId_$padName?showChat=false
```

However the etherpad when added as a widget complains:

(trimmed etherpad name)
![image](https://user-images.githubusercontent.com/66973897/106497332-e6c78800-648b-11eb-9a14-3f36f406a5eb.png)

skepticalwaves:
The room name it generates is a typical 3-word combo, not the format I set in the dimension settings. I can successfully access the pad directly from the name of the pad it does generate.

pushytoxin:
Make sure the padname is fewer than 50 characters
The bug is in Dimension (but also in the vector.scalar.im proprietary integrations manager that Dimension copies the template from)
https://github.com/turt2live/matrix-dimension/issues/395

skepticalwaves:
> Make sure the padname is fewer than 50 characters

Okay. This is a silly bug on etherpad's side. But how do I do that? I've done nothing to control the pad name anywhere, I have no idea what does that.

skepticalwaves:
Explanation found elsewhere, https://github.com/turt2live/matrix-dimension/issues/372 one can manually set URL and padname.

skepticalwaves:
This probably just needs an update to the documentation. I've managed to fix things by dropping the roomID

pushytoxin:
I'll add a line, but the real fix should be in Dimension (bc Etherpad rejects the change)

PR #859, opened 2021-02-06
Etherpad docs: Padname length bug

pushytoxin:
Warn users of the known bug https://github.com/turt2live/matrix-dimension/issues/395

Merging would close https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/849


spantaleev:
Thanks! :+1: 

Alternatively, we may even change the template above, so that we don't include `$roomId_` in it, leading to a working setup by default..?

pushytoxin:
We could inject the SQL statements I guess, the question is _when_ to do it. I wouldn't want to disrupt the user's preferences.

PR #862, opened 2021-02-07
Add additional domains for Let's Encrypt certificates to be obtained

s-thom:
This pull request adds a new variable (`matrix_ssl_additional_domains_to_obtain_certificates_for`) which is an array of extra domain names, specified in the user's `vars.yml`. These domains get included in the `matrix_ssl_domains_to_obtain_certificates_for` variable, which is what the playbook loops over to obtain/renew certificates.

It also adds documentation to `configuring-playbook-nginx.md` for using this new variable and the already existing `matrix_nginx_proxy_proxy_http_additional_server_configuration_blocks`. This should hopefully make it clear how the nginx proxy added in this playbook can be configured to act as a reverse proxy for other services as well, reducing the need to set up your own webserver.

---

I think future work could be done to make additional reverse proxy blocks easier to configure (I basically copied an entire template file into my vars.yml for my own configuration), but that's not really the concern of this PR, and, if wanted, should be discussed in its own issue.

### DIFF
--- a/docs/configuring-playbook-nginx.md
+++ b/docs/configuring-playbook-nginx.md
@@ -59,3 +59,17 @@ This will disable the access logging for nginx.
 ```yaml
 matrix_nginx_proxy_access_log_enabled: false
 ```
+
+## Additional configuration
+
+<!-- TODO: Introductory blurb -->
+
+<!-- TODO: Add section for including config files, however that's done -->
+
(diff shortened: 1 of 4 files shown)


spantaleev:
I've updated the documentation a bit more and merged this! Thank you! :+1:

PR #864, opened 2021-02-08
fix typo from skype to slack

yncyrydybyl:
The Link to the Slack Docker image was wrong in the README.


spantaleev:
Thanks! :+1:

PR #865, opened 2021-02-08
Fixes: a66a604e ("Selfbuild appservice-slack bridge")

yncyrydybyl:
It was missing the right variable name.


spantaleev:
Thank you! :+1:

PR #868, opened 2021-02-11
- Added a postgres-backup role

foxcris:
Added role to setup [prodrigestivill/docker-postgres-backup-local](https://github.com/prodrigestivill/docker-postgres-backup-local) container to backup all postgres databases.


docs/configuring-playbook-postgres-backup.md
```diff
@@ -0,0 +1,42 @@
+# Setting up postgres backup (optional)
+
+The playbook can install and configure [docker-postgres-backup-local](https://github.com/prodrigestivill/docker-postgres-backup-local) for you.
+
+## Adjusting the playbook configuration
+
+| Name                              | Default value                | Description                                                      |
+| :-------------------------------- | :--------------------------- | :--------------------------------------------------------------- |
+|matrix_postgres_backaup_enabled|false|Set to true to use [docker-postgres-backup-local](https://github.com/prodrigestivill/docker-postgres-backup-local) to create automatic database backups|
```
spantaleev, line 9: Typo.

`matrix_postgres_backaup_enabled` -> `matrix_postgres_backup_enabled`

docs/configuring-playbook-postgres-backup.md
```diff
@@ -0,0 +1,42 @@
...
+|matrix_postgres_backup_schedule| '@daily' |Cron-schedule specifying the interval between postgres backups.|
+|matrix_postgres_backup_keep_days|"7"|Number of daily backups to keep|
+|matrix_postgres_backup_keep_weeks|"4"|Number of weekly backups to keep|
+|matrix_postgres_backup_keep_months|"12"|Number of monthly backups to keep|
+|matrix_postgres_backup_path | "{{ matrix_base_data_path }}/postgres-backup" | Storagepath for the database backups|
+|matrix_postgres_backup_matrix_reminder_bot_enabled| false | Set to true to backup matrix_reminder_bot database. Per default matrix-reminder-bot uses an sqlite database. Only use this if you are using a postgres database for matrix-reminder-bot. |
```
spantaleev, line 15: Do we need the flexibility of being able to turn each of this on and off manually?

Wouldn't it be better if we back up all services by default?

roles/matrix-postgres-backup/defaults/main.yml
```diff
@@ -0,0 +1,140 @@
+matrix_postgres_backaup_enabled: false
```
spantaleev, line 1: Same typo.

roles/matrix-postgres-backup/defaults/main.yml
```diff
@@ -0,0 +1,140 @@
...
+
+matrix_postgres_connection_hostname: "matrix-postgres"
+matrix_postgres_connection_port: 5432
+matrix_postgres_connection_username: "matrix"
+matrix_postgres_connection_password: ""
```
spantaleev, line 6: These had better be named `matrix_postgres_backup_connection_*` (`matrix_postgres_backup_*` prefix).

Reasons:

1. You're redefining `matrix-postgres` roles variables here. It's confusing
2. Even if you're not really redefining variables from other roles, but rather using them, it's still not ideal. We try to make roles as independent of one another as possible, with `group_vars/matrix_servers` trying to tie their variables together.

That is to say, these variables should have some defaults (or even empty defaults, with `roles/matrix-postgres-backup/tasks/validate_config.yaml` complaining about them being blank). And then `group_vars/matrix_servers` will redefine these variables in term of other variables. Example (excerpt) from `group_vars/matrix_servers`:

```yaml
matrix_postgres_backup_connection_hostname: "{{ matrix_postgres_connection_hostname }}"
# Same for other variables
```

roles/matrix-postgres-backup/defaults/main.yml
```diff
@@ -0,0 +1,140 @@
...
+matrix_postgres_backup_keep_weeks: "4"
+matrix_postgres_backup_keep_months: "12"
+matrix_postgres_backup_healthcheck_port: "8080"
+matrix_postgres_backup_db_list: ""
+matrix_postgres_backup_path: "{{ matrix_base_data_path }}/postgres-backup"
+
+matrix_postgres_base_path: "{{ matrix_base_data_path }}/postgres"
+matrix_postgres_data_path: "{{ matrix_postgres_base_path }}/data"
+
+# uses sqlite per default
```
spantaleev, line 20: Instead of defining all of these enabled variables and `_database_name` variables and `matrix_postgres_backup_db_dict`, we may be able to redo this in a way that automatically enables backups for all of our databases.

Take a look at `matrix_postgres_additional_databases` (especially its definition in `group_vars/matrix_servers`).

You can forward that whole variable to `matrix_postgres_backup_databases` via `group_vars/matrix_servers` too, and then loop over it and back things up?

If we'd like to allow for more flexibility (people saying "don't back up this specific database" in the future), we may add some additional exclusion list feature (e.g. `matrix_postgres_backup_database_names_to_exclude: []`). But that can be added later, I think.

roles/matrix-postgres-backup/tasks/build_database_list.yml
```diff
@@ -0,0 +1,176 @@
+
+---
+
+# Detect wich databases have to backuped
```
spantaleev, line 4: All of these tasks can probable be deleted.

This dynamic "list building" can also be redone in terms of `matrix_postgres_backup_databases` / `matrix_postgres_additional_databases` (discussed in another comment in this review)

roles/matrix-postgres-backup/tasks/util/detect_existing_postgres_version.yml
```diff
@@ -0,0 +1,56 @@
+---
```
spantaleev, line 1: This file should probably be completely deleted.

foxcris: This file is used in setup_postgres_backup.yml to detect the currently used postgres version. The detected version is used to download/use the corresponding docker image for the prodrigestivill/docker-postgres-backup-local container.

docs/configuring-playbook-postgres-backup.md
foxcris: Done.

docs/configuring-playbook-postgres-backup.md
foxcris: I added an configuration option for each of them because the used container prodrigestivill/docker-postgres-backup-local requires a defined list of databases to backup. I did not know a solution to detect the used postgresdatabases during runtime thus i created the configuration options. Do you know a way to detect the used postgres databases?

roles/matrix-postgres-backup/defaults/main.yml
foxcris: Done

roles/matrix-postgres-backup/defaults/main.yml
foxcris: Thx. I resued the name by indent as the value should be the same as from the other role. I did not know of the machanism to use `group_vars/matrix_servers` to redefine these variables. I will change this.

roles/matrix-postgres-backup/defaults/main.yml
foxcris: Great idea. I will rework this.

docs/configuring-playbook-postgres-backup.md
foxcris: I reworked this to use the `matrix_postgres_additional_databases` variable in `group_vars/matrix_servers`

roles/matrix-postgres-backup/defaults/main.yml
foxcris: I reworked this to use the matrix_postgres_additional_databases variable in group_vars/matrix_servers

roles/matrix-postgres-backup/tasks/build_database_list.yml
foxcris: After the rework an use of group_vars we do not need this anymore. Removed.

foxcris:
The POSTGRES_DB is now set during building the template. No intermediate variable matrix_postgres_backup_db_list is used anymore.

spantaleev:
I think it's more natural and easy to use if there's an intermediate variable (`matrix_postgres_backup_db_list`) and all that `|map` and other complexity happens in `group_vars/matrix_servers`, like this:

```yaml
# the default matrix synapse databse is not always part of the matrix_postgres_additional_databases variable thus we have to add it if the default database is used
matrix_postgres_backup_db_list: |
  {{
    [
      matrix_synapse_database_database
    ] if (matrix_synapse_enabled and matrix_synapse_database_database == matrix_postgres_db_name and matrix_synapse_database_host == 'matrix-postgres') else [])
    +
    matrix_postgres_additional_databases|map(attribute='name')
  }}
```

.. and then it's just a simple list inside the role (it can be defined as `matrix_postgres_backup_db_list: []` in `roles/matrix-postgres-backup/defaults/main.yml`).

.. and then it can be used like this: `POSTGRES_DB={{ matrix_postgres_backup_databases|join(', ') }}`

People who wish to override it can do it easily (e.g.):

```yaml
matrix_postgres_backup_db_list: ['synapse', 'matrix_ma1sd']
```

without having to create a more complicated structure just to match some other syntax (`[{'name': 'synapse'}, {'name': 'matrix_ma1sd'}]`).

What do you think?

foxcris:
I liked the idea to provide a simple variable for the user to define the databases to use. However there is no need of an additional variable ```matrix_postgres_backup_db_list```. The same result can be achieved by directly using ```matrix_postgres_backup_databases```. 

From my side everything should be fine now and ready for a merge or are there any other open points for discussion that i missed?

spantaleev:
Looks good! :+1: 

I've updated the docs a bit in 37c089bb96 and 6526087c14.

Thanks a lot for contributing such an important and useful feature! :bow:

PR #876, opened 2021-02-15
Fix etherpad binding

marcquark:
The Etherpad container's port is exposed on all interfaces in the default configuration. That means that it's possible to access Etherpad in plaintext via http://matrix.[domain]:9001. While a user would have to stretch to actually find and deliberately use this insecure form of access, it's still a minor security flaw and should be adjusted.

This PR changes the default behaviour so that Etherpad
- does not bind to the host at all if the builtin reverse proxy is used
- only binds to a loopback port if the builtin reverse proxy is _not_ used
It's still possible for advanced users to override `matrix_etherpad_container_http_host_bind_port` and customize the behaviour to their liking.

Furthermore the PR instructs the Etherpad container to trust the reverse proxy by default, so that the logs contain correct information about origin requests

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -765,6 +765,8 @@ matrix_dimension_database_password: "{{ matrix_synapse_macaroon_secret_key | pas
 
 matrix_etherpad_enabled: false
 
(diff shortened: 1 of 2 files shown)


spantaleev:
Wonderful! Thank you! :+1:

Issue #874, opened 2021-02-14
squlite3 issue with Etherpad


ctwelve:
Hello,

On attempting to enable Etherpad, I run into an issue where the container complains about missing sqlite3:

```
Feb 14 19:46:28 matrix.arkmuse.org systemd[1]: Started Matrix Etherpad.
Feb 14 19:46:29 matrix.arkmuse.org matrix-etherpad[32447]: [2021-02-14 19:46:29.115] [DEBUG] console - Running on Node v10.23.0 (minimum required Node version: 10.13.0)
Feb 14 19:46:29 matrix.arkmuse.org matrix-etherpad[32447]: [2021-02-14 19:46:29.283] [INFO] console - All relative paths will be interpreted relative to the identified Etherpad base dir: /opt/etherpad-lite
Feb 14 19:46:29 matrix.arkmuse.org matrix-etherpad[32447]: [2021-02-14 19:46:29.284] [INFO] console - Random string used for versioning assets: ada4442b
Feb 14 19:46:29 matrix.arkmuse.org matrix-etherpad[32447]: [2021-02-14 19:46:29.287] [INFO] console - settings loaded from: /data/settings.json
Feb 14 19:46:29 matrix.arkmuse.org matrix-etherpad[32447]: [2021-02-14 19:46:29.288] [INFO] console - No credentials file found in /data/credentials.json. Ignoring.
Feb 14 19:46:29 matrix.arkmuse.org matrix-etherpad[32447]: Using skin "colibris" in dir: /opt/etherpad-lite/src/static/skins/colibris
Feb 14 19:46:29 matrix.arkmuse.org matrix-etherpad[32447]: Session key loaded from: /data/sessionkey.json
Feb 14 19:46:29 matrix.arkmuse.org matrix-etherpad[32447]: FATAL: The sqlite dependency could not be loaded. We removed it from the dependencies since it caused problems on several Platforms to compile it. If you still want to use sqlite, do a 'npm install sqlite3' in your etherpad-lite root folder
Feb 14 19:46:29 matrix.arkmuse.org systemd[1]: matrix-etherpad.service: Main process exited, code=exited, status=1/FAILURE
Feb 14 19:46:29 matrix.arkmuse.org systemd[1]: matrix-etherpad.service: Failed with result 'exit-code'.
```

I am not sure what is causing this, but my installation has an external PostgreSQL setup, if that helps...

spantaleev:
Interesting!

[...]

PR #883, opened 2021-02-18
Remove 'sqlite3' from the Etherpad role

pushytoxin:
PR #804 incorrectly assumed that `sqlite3` databases were supported by the official Etherpad Docker image, #874 reported this error.
To avoid any further confusion, I am removing all mention of sqlite3 in the matrix-etherpad role.
Even though it is possible to add sqlite3 support to the container, it is way more difficult to do than just connecting to the postgres database. I think the latter solution should be recommended.


spantaleev:
Great! Thanks for taking care of this! :+1:

PR #884, opened 2021-02-18
Make the proxy_pass address in Grafana's nginx config use a variable

pulmonarycosignerkindness:
This adjusts the proxy_pass port/IP so that it follows the `matrix_grafana_container_http_host_bind_port` variable from `/roles/matrix-grafana/defaults/main.yml` when using an external webserver. The code checks how long the variable is (if the user has set just a port, \<port\>, or the full address, \<ip\>:\<port\>), and sets the proxy_pass value accordingly. 

Has been tested with both a full address and just a port, and seems to work.


pushytoxin:
This makes no sense
`matrix_grafana_container_http_host_bind_port` is a number, why would you write `http://{{matrix_grafana_container_http_host_bind_port}};`

PR #885, opened 2021-02-18
fix wrong json parameter in webhook documentation

MarcProe:
fixes HTTP 400 Error: "{\"errcode\":\"M_MISSING_PARAM\",\"error\":\"Missing key 'avatar_url'\"}"

### DIFF
--- a/docs/configuring-playbook-bridge-appservice-webhooks.md
+++ b/docs/configuring-playbook-bridge-appservice-webhooks.md
@@ -45,7 +45,7 @@ matrix_appservice_webhooks_log_level: '<log_level>'
     "text": "Hello world!",
     "format": "plain",
     "displayName": "My Cool Webhook",
-    "avatarUrl": "http://i.imgur.com/IDOBtEJ.png"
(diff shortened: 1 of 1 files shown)


spantaleev:
Are you sure?

The README for the project (https://github.com/turt2live/matrix-appservice-webhooks) says `avatarUrl`.

Searching for that also yields some results: https://github.com/turt2live/matrix-appservice-webhooks/search?q=avatarUrl

It also makes sense. Who would name this `avatarUrl`, while also using `displayName` - that's inconsistent.

aaronraimist:
Related: https://github.com/turt2live/matrix-appservice-webhooks/issues/72

spantaleev:
Thanks for finding that!

It sounds like that bridge is quite broken and even with `avatar_url`, it may not work well.

MarcProe:
I use the bridge with IFTTT to forward SMS. It works quite decently (if you use the "correct" parameter).
image

spantaleev:
Alright! Let's merge this as it is, given that it works. Thank you! :+1:

PR #891, opened 2021-02-19
add support for mautrix-instagram

MarcProe:
This is basically copied from the facebook role, of course with some changes in the config.yml file according to the mautrix_instagram documentation.
This was tested and works on a test machine and on a live install.


MarcProe:
something went wrong.

PR #901, opened 2021-02-21
Recommend a better way of backing up postgres

marcquark:
The recommended way of backing up postgres has a couple of drawbacks:
- It uses a hardcoded Postgres version that's not guaranteed to be the same as the running one
- It spawns an extra container just to perform the dump

This PR proposes a slightly more elegant solution:
- Calls pg_dumpall inside the matrix-postgres container, ensuring that the version is always compatible. This also removes the need to download potentially another image
- Stores the backup file under /matrix so a backup of the whole folder (as recommended) will also contain a consistent DB dump
- Uses absolute paths for good measure, just in case something in the ENV is messed up


spantaleev:
It also solves one more issue having to do with ARM32. See ca223559107b20a7.

Thank you! :+1:

PR #902, opened 2021-02-22
Added paragraph about updating DNS to get stats

tctovsli:
This document didn't describe that it is necessary to have a DNS-entry for stats sub-domain.


spantaleev:
Thank you! 👍

PR #905, opened 2021-02-23
rust-compress-state debug: indent the chonky json

pushytoxin:
If the room list gets posted in the debug step, at least make it readable.

However something doesn't seem right to me.
Internally, `{{ matrix_synapse_rust_synapse_compress_state_eligible_rooms }}` is an array of objects, not a JSON string.
The **debug** step should have a sane default to pretty print the output, and not rely on us manually applying the filter.
If there is a way to configure what filter is used when Ansible is printing data to a terminal, let me know please.

### DIFF
--- a/roles/matrix-synapse/tasks/rust-synapse-compress-state/main.yml
+++ b/roles/matrix-synapse/tasks/rust-synapse-compress-state/main.yml
@@ -100,7 +100,7 @@
           The following rooms contain more than {{ matrix_synapse_rust_synapse_compress_state_min_state_groups_required }} state group rows
           (configurable via `matrix_synapse_rust_synapse_compress_state_min_state_groups_required`)
           and will be compressed:
-          {{ matrix_synapse_rust_synapse_compress_state_eligible_rooms }}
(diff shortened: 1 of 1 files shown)


pushytoxin:
Upon closer inspection, the the data that is printed by default isn't actually JSON

```
[{''room_id'': ''!iyIlasdfasdfRmRHFx:matrix.org'', ''count'': 8621734},
{''room_id'': ''!Oasdfasdfasdfasdfm:matrix.org'', ''count'': 6077025}...
]
```

Indentation is mine, but the `''` isn't.
This is too weird for me. I'm signing off this PR.

spantaleev:
So.. does this PR make things better or not? `|to_yaml` may be another possibility.

If we can make it more readable, that'd nice, but it's not too important. It's a `debug` step, after all ;) Mostly to help us (or the user) debug things, should they go wrong.

pushytoxin:
The message is informative, I think it's nice to have, it's just not very readable by default. Either YAML or JSON should work equally well.

PR #910, opened 2021-02-27
Run Let's Encrypt renewal checks daily instead of weekly.

hardye:
Currently we check for new certificates once a week. While this works well, this pull request ensures more timely updates of certifcates.

To be honest, I'm not sure this update is really needed, but following our recent discussion I promised to submit a pull request. I won't be offended if it doesn't get merged as one could easily argue that it's not a good idea to change something that already works well.

Also, we're needlessly restarting nginx every day around 6:30 a.m. with this update (and Coturn as well if enabled), although I doubt there is much going on at this time of day in most deployments ... :smile:

### DIFF
--- a/roles/matrix-coturn/templates/systemd/matrix-coturn-reload.timer.j2
+++ b/roles/matrix-coturn/templates/systemd/matrix-coturn-reload.timer.j2
@@ -3,8 +3,8 @@ Description=Reloads matrix-coturn periodically so that new SSL certificates can 
 
 [Timer]
 Unit=matrix-coturn-reload.service
-OnCalendar=Sunday *-*-* 13:00:00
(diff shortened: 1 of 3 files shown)


aaronraimist:
Well it is just reloading nginx, it isn't actually restarting it. There shouldn't be any downtime or dropped connections if I understand correctly.

hardye:
You are right, @aaronraimist. I looked up the nginx unit file and it defines `ExecReload` as `/usr/bin/env docker exec matrix-nginx-proxy /usr/sbin/nginx -s reload` which gracefully reloads nginx without dropping any open connections. Thanks for pointing this out.

spantaleev:
Thanks! :+1:

PR #915, opened 2021-03-02
Added nginx proxy worker configuration to template and defaults

SierraKiloBravo:
On my deployment's nginx server I get a lot of errors stating that the default amount (1024) of worker connections is not enough. This PR allows users to configure the amount of worker connections and worker processes. 
A PR (#544) was made implementing this before, but was not completed. I have implemented the comments made on that PR in this PR.


SierraKiloBravo:
I have also tested this on my homeserver, and it works as intended.

spantaleev:
Wonderful! Thank you! :+1:

PR #916, opened 2021-03-02
Adjust wait timeout

jokey2k:
During first setup postgres takes its time to get up and running, resulting in "postgres in startup" exceptions from synapse if you run without additional services that come in between. Hence suggesting increasing the time a bit to avoid having an error which heals itself and thus is hard to spot for newcomers.

### DIFF
--- a/roles/matrix-common-after/tasks/start.yml
+++ b/roles/matrix-common-after/tasks/start.yml
@@ -30,7 +30,7 @@
 # as we may run into systemd's automatic restart logic retrying the service.
 - name: Wait a bit, so that services can start (or fail)
   wait_for:
-    timeout: 5
+    timeout: 15
   delegate_to: 127.0.0.1
   become: false
 


spantaleev:
It'd be better if we made this configurable, I think.

Most of us aren't observing this problem, so having to wait +10 seconds more on every run is not necessary. However, 10 seconds is not that much compared to the total runtime anyway, and it's probably a safer default, so I'll merge this as it is. Thank you! :+1: 

People who want to optimize their runs can contribute support for making it configurable and we may even decrease the default then ;)

pushytoxin:
Delays? Is there _really_ no way to orchestrate container startup using Docker? Are we really back to having the issue that K&R fixed forty years ago by forking daemons in sysvinit?? Even nicer things have come along since then! How about sd_notify? How about socket activation?

A bit of looking around led me to this video, and I recommend all potential contributors (especially @spantaleev) to watch it in its entirety: https://www.redhat.com/en/about/videos/docker-versus-systemd

It basically shows a way to have a lot of systemd features that we are missing now, by having systemd as PID1 inside the container. The way to do it is to have fedora-based containers, if I understand correctly.

PR #920, opened 2021-03-05
Adding an initialization script

zevlee:
I created a basic bash script to streamline the initial process for creating the `vars.yml` and `hosts` files. Users can simply run the script and enter the prompted information. If the user desires to change the previously entered information, they can simply run the script again.

### DIFF
--- a/docs/configuring-playbook.md
+++ b/docs/configuring-playbook.md
@@ -6,7 +6,7 @@ To configure the playbook, you need to have done the following things:
 - [configured your DNS records](configuring-dns.md)
 - [retrieved the playbook's source code](getting-the-playbook.md) to your computer
 
-You can then initialize the the configuration files `vars.yml` and `hosts` by running `./initialize.sh`. If you need to change the variables, you can opt to run the script again.
+You can then initialize the configuration files `vars.yml` and `hosts` by running `./initialize.sh`. If you need to change the variables, you can opt to run the script again.
 
 For a basic Matrix installation, that's all you need.
 For a more custom setup, see the [Other configuration options](#other-configuration-options) below.
--- a/docs/configuring-playbook.md
+++ b/docs/configuring-playbook.md
@@ -6,18 +6,7 @@ To configure the playbook, you need to have done the following things:
 - [configured your DNS records](configuring-dns.md)
 - [retrieved the playbook's source code](getting-the-playbook.md) to your computer
 
-You can then follow these steps inside the playbook directory:
-
-- create a directory to hold your configuration (`mkdir inventory/host_vars/matrix.<your-domain>`)
-
-- copy the sample configuration file (`cp examples/vars.yml inventory/host_vars/matrix.<your-domain>/vars.yml`)
-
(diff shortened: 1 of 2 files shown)


spantaleev:
Hi and thank you for taking the time to improve the playbook! :bow: 

While this could really make things simpler for some people (it spares them from having to generate a bunch of annoying secrets manually), it also worries me a bit, because:

- this bash script may or may not be compatible with all operating systems we run it on (MacOS's `bash` being old, etc.). Perhaps `sh` should be used? Perhaps a more limited and more generally compatible script should be used? (although this one may very well be very compatible.. I can't say just by looking at it, nor have I tested it). Not sure if we have such playbook users, but some may be running [Ansible from within Windows / WSL](https://www.jeffgeerling.com/blog/running-ansible-within-windows). Not sure we should worry too much about this, but we should at least not actively hide the alternative way of setting things up manually.

- some other tools (like `openssl`) may not be available, causing the script to break. And while you could encounter a similar situation while generating secrets for `vars.yml` manually (our advice being to use `pwgen`), one could always generate the secret any other way, including manually entering it.

- "If you need to change the variables, you can opt to run the script again." sounds benign, but it can be quite destructive. Especially the Postgres password - that one you can't change subsequently without also logging into the server, connecting to Postgres and changing it there manually as well. You can easily lock yourself out of your Postgres server by running `initialize.sh` a 2nd time.

- the script appears to make some symlinks in the main directory, which can be problematic for people managing multiple Matrix hosts via the playbook (although, this is rare and such people probably know how to work around it).

- trying to use `initialize.sh` to initialize a 2nd host will destroy everything in `inventory/host_vars` and `inventory/hosts` as well. Ouch.

- using the script also prevents people from learning the right way of working with Ansible and likely makes them more incapable of troubleshooting things.

- it does not account for a relatively common situation (a remote server with non-root login + `sudo` elevation to `root`). Support could be added or we may put up a note to read and modify the `hosts` file, because additional setup may be required (not just related to this situation, but also others -- disabling SSH pipelining may be necessary in some instances for example, etc.)

Throughout all the docs, we attempt to strike a good balance between teaching (both Ansible things & Matrix things) and letting people easily set up a server. Setup could be way more straightforward (and done with much less reading), at the expense of teaching people how things work and why they work that way. Also, we have various people with various setup needs (reusing VPS servers for multiple things; running a different reverse-proxy; etc), so we can't dumb things down too much, without negatively affecting them.

Still, I think such an initialization script can be very useful to make things simpler, provided:

- we point out the alternative (manual) way of doing things

- we don't dumb things down for people too much, to the point of making them more incapable of troubleshooting. That is, we should probably avoid `vars.yml` and `hosts` files in the main directory and point people to the actual ones that got generated.

- not be so destructive. Likely avoid working for a given domain a 2nd time (until its `vars.yml` file is deleted). Also, avoid regenerating `inventory/hosts` and try to "append" new hosts to it. Or tell people "Not updating your inventory/hosts file, because there already is one there. Modify it manually".

- point people to the generated configs and encourage them to read/tweak them. We'd rather they verify thing and possibly learn something new (or account for a different situation - like that `sudo` thing) than to come here saying "I did things right, it says it can't connect".

In short -- try to be helpful. Don't be destructive. Don't encourage blindly doing things - it may come back to bite us later.

zevlee:
Thank you for providing a very in-depth response regarding the philosophy of this project. I have a much better understanding of the considerations that ought to be made.

It's clear that the solution I proposed would create far more problems than it would solve. With that in mind, I will close this pull request and resubmit one when I believe it aligns with the intended goal of the project.

PR #923, opened 2021-03-06
Make steps in configuring-playbook.md numbered

aaronraimist:
instead of bullet points which make it more difficult to tell whether or not you completed all of the steps


spantaleev:
Thank you! :+1:

PR #929, opened 2021-03-11
Added support for the Go-NEB bot

Zir0h:
Hello!

I added support for the [Go-NEB](https://github.com/matrix-org/go-neb) bot. I pretty much just copied the [matrix-reminder-bot](https://github.com/spantaleev/matrix-docker-ansible-deploy/tree/master/roles/matrix-bot-matrix-reminder-bot) and modified it to my needs.

Please let me know if I need to make changes. The configuration of the bot is quite dynamic, so I opted to keep the original config structure. Because of this the vars.yml file might become too large.

### DIFF
--- a/docs/configuring-playbook-bot-go-neb.md
+++ b/docs/configuring-playbook-bot-go-neb.md
@@ -1,0 +1,218 @@
+# Setting up Go-NEB (optional)
+
+The playbook can install and configure [Go-NEB](https://github.com/matrix-org/go-neb) for you.
+
+Go-NEB is a Matrix bot written in Go. It is the successor to Matrix-NEB, the original Matrix bot written in Python.
+
+See the project's [documentation](https://github.com/matrix-org/go-neb) to learn what it does and why it might be useful to you.
+
+
+## Registering the bot users
+
+The playbook does not automatically create users for you. The bot requires at least 1 access token to be able to connect to your homeserver.
+
+You **need to register the bot user manually** before setting up the bot.
+
+Choose a strong password for the bot. You can generate a good password with a command like this: `pwgen -s 64 1`.
+
+If you use curl, you can immediatly copy/paste the access token into the configuration file.
+
+```
+curl -X POST --header 'Content-Type: application/json' -d '{
+    "identifier": { "type": "m.id.user", "user": "bot.go-neb" },
+    "password": "a strong password",
+    "type": "m.login.password"
+}' 'https://matrix.YOURDOMAIN/_matrix/client/r0/login'
+```
+
+You can use also use the playbook to [register a new user](registering-users.md):
+
+```
+ansible-playbook -i inventory/hosts setup.yml --extra-vars='username=bot.go-neb password=PASSWORD_FOR_THE_BOT admin=no' --tags=register-user
+```
+
+And then log in via element, but doing so might lead to decryption problems. That warning comes from [here](https://github.com/matrix-org/go-neb#quick-start)
+
+## Adjusting the playbook configuration
+
+Add the following configuration to your `inventory/host_vars/matrix.DOMAIN/vars.yml` file (adapt to your needs):
+
+```yaml
+matrix_bot_go_neb_enabled: true
+
+# You need at least 1 client.
+matrix_bot_go_neb_clients:
+  - UserID: "@goneb:localhost"
+    AccessToken: "MDASDASJDIASDJASDAFGFRGER"
+    DeviceID: "DEVICE1"
+    HomeserverURL: "{{ matrix_homeserver_container_url }}"
+    Sync: true
+    AutoJoinRooms: true
+    DisplayName: "Go-NEB!"
+    AcceptVerificationFromUsers: [":localhost:8008"]
+
+  - UserID: "@another_goneb:localhost"
+    AccessToken: "MDASDASJDIASDJASDAFGFRGER"
+    DeviceID: "DEVICE2"
+    HomeserverURL: "{{ matrix_homeserver_container_url }}"
+    Sync: false
+    AutoJoinRooms: false
+    DisplayName: "Go-NEB!"
+    AcceptVerificationFromUsers: ["^@admin:localhost:8008$"]
+
+# Optional, for use with the github_cmd, github_webhooks or jira services
+matrix_bot_go_neb_realms:
+  - ID: "github_realm"
+    Type: "github"
+    Config: {} # No need for client ID or Secret as Go-NEB isn't generating OAuth URLs
+
+# Optional. The list of *authenticated* sessions which Go-NEB is aware of.
+matrix_bot_go_neb_sessions:
+  - SessionID: "your_github_session"
+    RealmID: "github_realm"
+    UserID: "@YOUR_USER_ID:localhost" # This needs to be the username of the person that's allowed to use the !github commands
+    Config:
+      # Populate these fields by generating a "Personal Access Token" on github.com
+      AccessToken: "YOUR_GITHUB_ACCESS_TOKEN"
+      Scopes: "admin:org_hook,admin:repo_hook,repo,user"
+
+# The list of services which Go-NEB is aware of.
+# Delete or modify this list as appropriate.
+# See the docs for /configureService for the full list of options:
+# https://matrix-org.github.io/go-neb/pkg/github.com/matrix-org/go-neb/api/index.html#ConfigureServiceRequest
+# You need at least 1 service.
+matrix_bot_go_neb_services:
+  - ID: "echo_service"
+    Type: "echo"
+    UserID: "@goneb:localhost"
+    Config: {}
+
+# Can be obtained from https://developers.giphy.com/dashboard/
+  - ID: "giphy_service"
+    Type: "giphy"
+    UserID: "@goneb:localhost" # requires a Syncing client
+    Config:
+      api_key: "qwg4672vsuyfsfe"
+      use_downsized: false
+
+# This service has been dead for over a year :/
+  - ID: "guggy_service"
+    Type: "guggy"
+    UserID: "@goneb:localhost" # requires a Syncing client
+    Config:
+      api_key: "2356saaqfhgfe"
+
+# API Key via https://developers.google.com/custom-search/v1/introduction
+# CX via http://www.google.com/cse/manage/all
+# https://stackoverflow.com/questions/6562125/getting-a-cx-id-for-custom-search-google-api-python
+# 'Search the entire web' and 'Image search' enabled for best results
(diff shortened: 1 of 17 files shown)


roles/matrix-bot-go-neb/templates/systemd/matrix-bot-go-neb.service.j2
```diff
@@ -0,0 +1,46 @@
+#jinja2: lstrip_blocks: "True"
+[Unit]
+Description=Matrix reminder bot
```
spantaleev, line 3: Typo

roles/matrix-bot-go-neb/templates/config.yaml.j2
```diff
@@ -0,0 +1,44 @@
...
+#   - /configureAuthRealm
+#   - /configureService
+#   - /requestAuthSession (redirects not supported)
+
+# The list of clients which Go-NEB is aware of.
+# Delete or modify this list as appropriate.
+# See the docs for /configureClient for the full list of options:
+# https://matrix-org.github.io/go-neb/pkg/github.com/matrix-org/go-neb/api/index.html#ClientConfig
+clients:
+  {{ matrix_bot_go_neb_clients }}
```
spantaleev, line 22: Shouldn't we use `|to_json` here and in other instances below?

docs/configuring-playbook-bot-go-neb.md
```diff
@@ -0,0 +1,218 @@
...
+# You need at least 1 client.
+matrix_bot_go_neb_clients:
+  - UserID: "@goneb:localhost"
+    AccessToken: "MDASDASJDIASDJASDAFGFRGER"
+    DeviceID: "DEVICE1"
+    HomeserverURL: "{{ matrix_homeserver_container_url }}"
+    Sync: true
+    AutoJoinRooms: true
+    DisplayName: "Go-NEB!"
+    AcceptVerificationFromUsers: [":localhost:8008"]
```
spantaleev, line 52: `:localhost:8008` is probably never correct in our use-case.

Maybe we should use something like `AcceptVerificationFromUsers: [":{{ matrix_domain }}"]`.
Would that work?

Likewise, for other invocations below (and possible in the `defaults.yml` file).

roles/matrix-bot-go-neb/defaults/main.yml
```diff
@@ -0,0 +1,225 @@
...
+matrix_bot_go_neb_data_store_path: "{{ matrix_bot_go_neb_data_path }}/store"
+
+# A list of extra arguments to pass to the container
+matrix_bot_go_neb_container_extra_arguments: []
+
+# List of systemd services that matrix-bot-go-neb.service depends on
+matrix_bot_go_neb_systemd_required_services_list: ['docker.service']
+
+# List of systemd services that matrix-bot-go-neb.service wants
+matrix_bot_go_neb_systemd_wanted_services_list: []
```
spantaleev, line 22: Adding a `matrix_bot_go_neb_container_http_host_bind_port` variable (with an empty value) may be of value here.

See `matrix_ma1sd_container_http_host_bind_port` in the `roles/matrix-ma1sd/defaults/main.yml` role for an example. It should mention `127.0.0.1:4050` in the variable's comment.

group_vars/matrix_servers
```diff
@@ -729,6 +729,27 @@ matrix_bot_matrix_reminder_bot_database_password: "{{ matrix_synapse_macaroon_se
...
+######################################################################
+
+# We don't enable bots by default.
+matrix_bot_go_neb_enabled: false
+
+matrix_bot_go_neb_systemd_required_services_list: |
+  {{
+    ['docker.service']
+  }}
+
```
spantaleev, line 745: Here would be a good place to redefine the value of `matrix_bot_go_neb_container_http_host_bind_port` like this:

```yaml
matrix_bot_go_neb_container_http_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:4050' }}"
```

roles/matrix-bot-go-neb/templates/systemd/matrix-bot-go-neb.service.j2
```diff
@@ -0,0 +1,46 @@
...
+Environment="HOME={{ matrix_systemd_unit_home_path }}"
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} kill matrix-bot-go-neb 2>/dev/null'
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} rm matrix-bot-go-neb 2>/dev/null'
+
+ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-bot-go-neb \
+			--log-driver=none \
+			--user={{ matrix_user_uid }}:{{ matrix_user_gid }} \
+			--cap-drop=ALL \
+			--read-only \
+			--network={{ matrix_docker_network }} \
```
spantaleev, line 24: Here would be a good place to make use of `matrix_bot_go_neb_container_http_host_bind_port`.

Example:

```
{% if matrix_bot_go_neb_container_http_host_bind_port %}
-p {{ matrix_bot_go_neb_container_http_host_bind_port }}:4050 \
{% endif %}
```

roles/matrix-bot-go-neb/defaults/main.yml
```diff
@@ -0,0 +1,225 @@
+# Go-NEB is a Matrix bot written in Go. It is the successor to Matrix-NEB, the original Matrix bot written in Python.
+# See: https://github.com/matrix-org/go-neb
+
+matrix_bot_go_neb_enabled: false
```
spantaleev, line 4: We usually default roles to enabling their services (which eases people who are including our roles manually in their own other playbook).

We then disable services in `group_vars/matrix_servers` as needed. Example line to put in `group_vars/matrix_servers`:

```yaml
matrix_bot_go_neb_enabled: false
``

**Review by spantaleev — COMMENTED** [review]
Thank you for your work on adding something so useful!

I've submitted some feedback about how this PR can be improved.

File: roles/matrix-bot-go-neb/templates/config.yaml.j2 (hunk 2 above)
Comment (Zir0h) [reply roles/matrix-bot-go-neb/templates/config.yaml.j2:22]: Hi! Thank you for the feedback!

The `{{ matrix_bot_go_neb_clients }}` contains the entire code block from the vars.yaml file.

For example:
```
matrix_bot_go_neb_clients: 
  - UserID: "@goneb:localhost"
    AccessToken: "MDASDASJDIASDJASDAFGFRGER"
    DeviceID: "DEVICE1"
    HomeserverURL: "http://localhost:8008"
    Sync: true
    AutoJoinRooms: true
    DisplayName: "Go-NEB!"
    AcceptVerificationFromUsers: [":localhost:8008"]
```

Then that block gets filled in as-is in the config.yaml from the bot. Like the [sample.config.yaml](https://github.com/matrix-org/go-neb/blob/master/config.sample.yaml#L21) from the upstream repo.

Adding a |to_json will break the bot, because it's expecting yaml.

Do you prefer a custom array/dict in the vars.yaml which we then can loop with with_items?
And then in the config.yaml.j2 something like:
```
clients:
    {% for client in matrix_bot_go_neb_clients %}
  - UserID: {{ client.user_id|to_json }}
    AccessToken: {{ client.access_token|to_json }}
    DeviceID: {{ client.device_id|to_json }}
etc
etc
    {% endfor %}
```

File: roles/matrix-bot-go-neb/templates/systemd/matrix-bot-go-neb.service.j2 (hunk 1 above)
Comment (Zir0h) [reply roles/matrix-bot-go-neb/templates/systemd/matrix-bot-go-neb.service.j2]: Good catch! Fixed with https://github.com/Zir0h/matrix-docker-ansible-deploy/commit/7d1d3b47bc1135311f253106b3dc73b6e53e64e7

File: roles/matrix-bot-go-neb/defaults/main.yml (hunk 7 above)
Comment (Zir0h) [reply roles/matrix-bot-go-neb/defaults/main.yml]: Adapted in https://github.com/Zir0h/matrix-docker-ansible-deploy/commit/ce14e3e8af7fbd884c4b7ab1b40b7e662a31548d

File: roles/matrix-bot-go-neb/templates/config.yaml.j2 (hunk 2 above)
Comment (spantaleev) [reply roles/matrix-bot-go-neb/templates/config.yaml.j2:22]: It reminds me of `matrix_appservice_irc_ircService_servers` that we use in the `matrix-appservice-irc` role.

Ansible parses that value into a native list + dict, and then you can use `|to_json` to serialize it as JSON (well, YAML).

Are your values being turned into a string somehow?

Comment (Zir0h) [reply roles/matrix-bot-go-neb/templates/config.yaml.j2:22]: Ah yes, it does look a lot like `matrix_appservice_irc_ircService_servers`. I might've misunderstood the workings of `|to_json`.

I will do some testing!

File: docs/configuring-playbook-bot-go-neb.md (hunk 3 above)
Comment (Zir0h) [reply docs/configuring-playbook-bot-go-neb.md:52]: That does work indeed, I've modified it in https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/929/commits/9689948e7340f1f33f45a4d5bd727191c6fc8839

File: roles/matrix-bot-go-neb/templates/config.yaml.j2 (hunk 2 above)
Comment (Zir0h) [reply roles/matrix-bot-go-neb/templates/config.yaml.j2]: Tested with |to_json and everything is working. So it was just a misunderstanding of that function on my part. Corrected in https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/929/commits/49028f1b0588bf5aefdd2985e5749958b043b4b7

File: docs/configuring-playbook-bot-go-neb.md [hunk 8]
```diff
@@ -0,0 +1,218 @@
...
+## Adjusting the playbook configuration
+
+Add the following configuration to your `inventory/host_vars/matrix.DOMAIN/vars.yml` file (adapt to your needs):
+
+```yaml
+matrix_bot_go_neb_enabled: true
+
+# You need at least 1 client.
+matrix_bot_go_neb_clients:
+  - UserID: "@goneb:localhost"
```
spantaleev, line 45: Looks like these examples still reference `localhost` in `UserID` and `AcceptVerificationFromUsers`

Zir0h: [Fixed](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/929/commits/bfe0ca6cf816191ff7e80a445288834215f0f8b5)

spantaleev:
Awesome! Thanks! Merged now! :+1: 

I've fixed it up a bit in 38b3fbd6f, 6cbfee539cd, b46ef46663.

PR #940, opened 2021-03-16
Another field with 'invalid input syntax for type smallint'

Zir0h:
Looks like this field was forgotten in https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/80c72615c7bfdcf47644d55f033e88f4e610cf25

SequelizeDatabaseError: invalid input syntax for type smallint: "false" error in the logs when adding a custom go-neb bot in dimension.


Zir0h:
> Thanks! +1
> 
> Isn't Go-NEB management "either via config or via its API, but not both"? If we are to manage it via Dimension, do we need to do something different (avoid passing a config to Go-NEB)?

That is what I'm experimenting with right now, but currently I'm not having much success with Dimension :/

PR #941, opened 2021-03-16
synapse: preload libjemalloc

pushytoxin:
Preload the jemalloc library supported by the upcoming 1.30.0 release


pushytoxin:
Enabled by default on the synapse docker image's next release:
https://github.com/matrix-org/synapse/pull/8553

PR #944, opened 2021-03-17
Also check if matrix_ssl_lets_encrypt_support_email is blank

aaronraimist:


spantaleev:
Good catch! Thanks for fixing it! :+1:

PR #945, opened 2021-03-17
matrix.{{ matrix_domain }} -> {{ matrix_server_fqn_matrix }}

pushytoxin:
In case anyone wanted to use a domain name other than matrix.DOMAIN


spantaleev:
Makes sense! Thank you! :+1:

PR #950, opened 2021-03-17
Set X-Forwarded-Proto on federation requests

aaronraimist:
Missed by https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/9b72384df7a7663079b41ad17768ff9e9b4b9320


spantaleev:
Thank you! :+1:

PR #955, opened 2021-03-20
GoMatrixHosting v0.3.2 hotfix

GoMatrixHosting:
Minor fix to install jq on AWX if monitor not installed.
Fix to allow include_tasks to work properly.
Fix to limit max filesize limit to 100MB.


GoMatrixHosting:
Scrap that, forgot something.

PR #957, opened 2021-03-22
Update IRC appservice

thomwiggers:
Update to 0.25; doesn't seem to change much.

https://github.com/matrix-org/matrix-appservice-irc/releases


spantaleev:
Thanks! :+1:

PR #962, opened 2021-03-24
Add mjolnir

### LINKED ISSUE #353: Add Mjolnir (moderation bot)
https://github.com/matrix-org/mjolnir is a moderation bot intended for use within individual communities. For those who spin up a homeserver for their community, Mjolnir would be a good fit.

This request is for the bot to be added, not the Synapse module (because that's complicated). Docker images are available on Docker Hub: https://hub.docker.com/r/matrixdotorg/mjolnir

aaronraimist:
Fixes #353

This is just the bot portion of mjolnir, not the synapse module.

Todo:
- [x] Write docs
- [x] Figure out why the first start doesn't work
  - It seems to start before Synapse so it crashes but after it automatically restarts then it works fine. I thought this is what the `..systemd_required_services_list` variable was for but I have Synapse in there and it doesn't seem to make a difference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,8 @@ Using this playbook, you can get the following services configured on your serve
 
 - (optional) [Go-NEB](https://github.com/matrix-org/go-neb) multi functional bot written in Go - see [docs/configuring-playbook-bot-go-neb.md](docs/configuring-playbook-bot-go-neb.md) for setup documentation
 
+- (optional) [Mjolnir](https://github.com/matrix-org/mjolnir), a moderation tool for Matrix - see [docs/configuring-playbook-bot-mjolnir.md](docs/configuring-playbook-bot-mjolnir.md) for setup documentation
+
 - (optional) [synapse-admin](https://github.com/Awesome-Technologies/synapse-admin), a web UI tool for administrating users and rooms on your Matrix server - see [docs/configuring-playbook-synapse-admin.md](docs/configuring-playbook-synapse-admin.md) for setup documentation
 
 - (optional) [matrix-registration](https://github.com/ZerataX/matrix-registration), a simple python application to have a token based matrix registration - see [docs/configuring-playbook-matrix-registration.md](docs/configuring-playbook-matrix-registration.md) for setup documentation
--- a/docs/configuring-playbook-bot-mjolnir.md
+++ b/docs/configuring-playbook-bot-mjolnir.md
@@ -21,7 +21,7 @@ You can use the playbook to [register a new user](registering-users.md):
 ansible-playbook -i inventory/hosts setup.yml --extra-vars='username=bot.mjolnir password=PASSWORD_FOR_THE_BOT admin=no' --tags=register-user
 ```
 
-If you would like Mjolnir to be able to deactivate users, move aliases, shutdown rooms, etc then it must be a server admin so you need to change `admin=no` to `admin=yes` in the  command above.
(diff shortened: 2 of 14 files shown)


spantaleev:
It'd be nice to have! Thanks for working on this!

----------

> Figure out why the first start doesn't work

Maybe the bot is too quick to start and too quick to try and hit Synapse (or die during this initial attempt?).

`..systemd_required_services_list` just makes one systemd service require another. systemd will make sure to start services in accordance, but this does not guarantee that services are actually fully running and usable.

A service can signal to systemd that it's started and usable using this [`sd_notify`](https://www.man7.org/linux/man-pages/man3/sd_notify.3.html) thing. However, Synapse needs to support it, and doing that from within a Docker container probably comes with quite some complication (I haven't looked into it lately).

--------

Speaking of Mjolnir, and looking at its README, I'm starting to wonder if we should work on a `matrix-pantalaimon` role as well.

aaronraimist:
I tried adding `ExecStartPre={{ matrix_host_command_sleep }} 10` (and 20 and 30) but that doesn't seem to help. The bot starts first (after waiting for nothing for 10 seconds). I'm not sure if misconfiguration or systemd bug or what. This is the service file that is generated:

```
[Unit]
Description=Matrix Mjolnir bot
Requires=docker.service
After=docker.service
Requires=matrix-synapse.service
After=matrix-synapse.service
Requires=matrix-postgres.service
After=matrix-postgres.service
Wants=docker.service
Wants=matrix-synapse.service
Wants=matrix-postgres.service
DefaultDependencies=no

[Service]
Type=simple
Environment="HOME=/root"
ExecStartPre=-/usr/bin/env sh -c '/usr/bin/env docker kill matrix-bot-mjolnir 2>/dev/null'
ExecStartPre=-/usr/bin/env sh -c '/usr/bin/env docker rm matrix-bot-mjolnir 2>/dev/null'

# Intentional delay, so that the homeserver (we likely depend on) can manage to start.
ExecStartPre=/usr/bin/env sleep 10

ExecStart=/usr/bin/env docker run --rm --name matrix-bot-mjolnir \
			--log-driver=none \
			--user=991:991 \
			--cap-drop=ALL \
			--read-only \
			--network=matrix \
			--mount type=bind,src=/matrix/mjolnir/config,dst=/data/config,ro \
			--mount type=bind,src=/matrix/mjolnir/data,dst=/data \
			docker.io/matrixdotorg/mjolnir:v0.1.17

ExecStop=-/usr/bin/env sh -c '/usr/bin/env docker kill matrix-bot-mjolnir 2>/dev/null'
ExecStop=-/usr/bin/env sh -c '/usr/bin/env docker rm matrix-bot-mjolnir 2>/dev/null'
Restart=always
RestartSec=30
SyslogIdentifier=matrix-bot-mjolnir

[Install]
WantedBy=multi-user.target
``` 

> Speaking of Mjolnir, and looking at its README, I'm starting to wonder if we should work on a matrix-pantalaimon role as well.

Wouldn't hurt for someone to work on that but I am not likely to work on that in the near future. Mjolnir works fine for what I need it to do without E2E.

aaronraimist:
Added a few more services to depend on and that seems to work OK. Just finishing up the docs and then this will be ready to merge./

aaronraimist:
Okay I think this is ready to go

spantaleev:
Thanks a lot! Happy to see the playbook supporting more and more bots! :+1:

PR #967, opened 2021-03-25
feat: push ephemeral events to appservices

jdreichmann:
This adds https://github.com/matrix-org/matrix-doc/pull/2409 to the
appservice registrations, enabling synapse to push EDUs to appservices.


spantaleev:
No harm in having this, I guess! Thank you! :+1:

PR #969, opened 2021-03-27
Make realms and sessions optional for go-neb config

Zir0h:
Go-NEB fails to start if the var is defined, but contains nothing:
```
matrix-bot-go-neb[16190]: time="2021-03-27T10:39:59Z" level=panic msg="Failed to load config file" config_file=/config/config.yaml error="Failed to convert to config file: json: cannot unmarshal object into Go struct field ConfigFile.Realms of type []api.ConfigureAuthRealmRequest"
```


spantaleev:
Leaving variables undefined is inconsistent with what we do in the rest of the playbook.

Defining them empty (e.g. `matrix_bot_go_neb_realms: {}`) and doing `{% if matrix_bot_go_neb_realms|length > 0 %}` (or should that be `{% if matrix_bot_go_neb_realms|keys|length > 0 %}`) checks may be better.

spantaleev:
Actually.. I just saw the actual error reported by the bot.

It appears that `matrix_bot_go_neb_realms` should be a list, not a dictionary. Likewise for the others.

I've fixed it in 77d598b.

Thanks for bringing this up! :+1:

PR #972, opened 2021-03-30
matrix-nginx-proxy: add custom nginx options to nginx.conf.j2

JohannesKleine:
To have the possibility to add some custom nginx configuration options to prevent timeouts.

### DIFF
--- a/roles/matrix-nginx-proxy/defaults/main.yml
+++ b/roles/matrix-nginx-proxy/defaults/main.yml
@@ -227,8 +227,17 @@ matrix_nginx_proxy_proxy_matrix_federation_api_ssl_certificate_key: "{{ matrix_s
 # The tmpfs at /tmp needs to be large enough to handle multiple concurrent file uploads.
 matrix_nginx_proxy_tmp_directory_size_mb: "{{ (matrix_nginx_proxy_proxy_matrix_federation_api_client_max_body_size_mb | int) * 50 }}"
 
+# A list of strings containing additional configuration blocks to add to the nginx server configuration (nginx.conf).
+# for big matrixservers to enlarge the number of open files to prevent timeouts
+# matrix_nginx_proxy_proxy_additional_server_configuration_blocks:
+#  - 'worker_rlimit_nofile 30000;'
+matrix_nginx_proxy_proxy_additional_server_configuration_blocks: []
+
 # A list of strings containing additional configuration blocks to add to the nginx http's server configuration (nginx-http.conf).
 matrix_nginx_proxy_proxy_http_additional_server_configuration_blocks: []
+
+# A list of strings containing additional configuration blocks to add to the nginx event server configuration (nginx.conf).
+matrix_nginx_proxy_proxy_event_additional_server_configuration_blocks: []
 
 # A list of strings containing additional configuration blocks to add to the base matrix server configuration (matrix-domain.conf).
 matrix_nginx_proxy_proxy_matrix_additional_server_configuration_blocks: []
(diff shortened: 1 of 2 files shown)


roles/matrix-nginx-proxy/defaults/main.yml
```diff
@@ -227,9 +227,18 @@ matrix_nginx_proxy_proxy_matrix_federation_api_ssl_certificate_key: "{{ matrix_s
 # The tmpfs at /tmp needs to be large enough to handle multiple concurrent file uploads.
 matrix_nginx_proxy_tmp_directory_size_mb: "{{ (matrix_nginx_proxy_proxy_matrix_federation_api_client_max_body_size_mb | int) * 50 }}"
 
+# A list of strings containing additional configuration blocks to add to the nginx server configuration (nginx.conf).
+# for big matrixservers to enlarge the number of open files to prevent timeouts
+# matrix_nginx_proxy_proxy_additional_server_configuration_blocks:
+#  - 'worker_rlimit_nofile 30000;'
+matrix_nginx_proxy_proxy_additional_server_configuration_blocks: []
```
spantaleev, line 234: Should we remove the `_server_` part from this variable's name?

It kind of implies this ends up in a `server {}` block? Or that this is for defining new `server {}` blocks?
Maybe:

- `matrix_nginx_proxy_proxy_additional_configuration_blocks`
- `matrix_nginx_proxy_proxy_event_additional_configuration_blocks`

JohannesKleine: Ok i change it. 
Is it possible to do something like this:
```
matrix_nginx_proxy_proxy_additional_configuration_blocks:
  - http { server { listen 80; server_name foo; rooot /foo} }
  ```
to add additional websites to nginx? I am not sure whether it can handle two http blocks. Otherwise i would add this feature in line 59 to nginx.conf.j2

Comment (spantaleev) [reply roles/matrix-nginx-proxy/defaults/main.yml:234]: Having not tested it, not sure if one can add a 2nd `http` block.

Comment (JohannesKleine) [reply roles/matrix-nginx-proxy/defaults/main.yml]: 2nd http block does not work, but `matrix_nginx_proxy_proxy_http_additional_server_configuration_blocks:` allready does this
```
matrix_nginx_proxy_proxy_http_additional_server_configuration_blocks::
  - http { server { listen 8080; server_name foo; root /foo} }
```
and a port  above 1024

File: roles/matrix-nginx-proxy/defaults/main.yml [hunk 2]
```diff
@@ -227,9 +227,18 @@ matrix_nginx_proxy_proxy_matrix_federation_api_ssl_certificate_key: "{{ matrix_s
...
+# for big matrixservers to enlarge the number of open files to prevent timeouts
+# matrix_nginx_proxy_proxy_additional_configuration_blocks:
+#  - 'worker_rlimit_nofile 30000;'
+matrix_nginx_proxy_proxy_additional_configuration_blocks: []
+
 # A list of strings containing additional configuration blocks to add to the nginx http's server configuration (nginx-http.conf).
 matrix_nginx_proxy_proxy_http_additional_server_configuration_blocks: []
 
+# A list of strings containing additional configuration blocks to add to the nginx event server configuration (nginx.conf).
+matrix_nginx_proxy_proxy_event_additional_server_configuration_blocks: []
```
spantaleev, line 240: Maybe `matrix_nginx_proxy_proxy_event_additional_configuration_blocks`? (without `_server_` here too)

JohannesKleine: Done. So `*_server_*` defines here only changes within the `server {}`block.

spantaleev: I think that was the intention, yes.

spantaleev:
Thank you! :+1:

PR #980, opened 2021-04-03
Couple of additional tweaks to the mjolnir docs

aaronraimist:
I always notice more things once it gets merged 😅


spantaleev:
Thanks! 👍

PR #981, opened 2021-04-03
Allow special characters in Grafana password and upgrade Grafana 

### LINKED ISSUE #971: Grafana password string should be protected
Per [Grafana's doc](https://grafana.com/docs/grafana/latest/administration/configuration/#password), matrix_grafana_default_admin_password should be protected by triple-quotes if the password string contains "#" or ";". That part of the doc talks about the database password, but it seems to also be true for admin_password later in the doc (kind of obvious if it's in the same file with the same syntax, but they dont actually say it).

I usually use a password generator (KeepassXC's) to generate any password in this playbook. For Grafana's, I got one with a "#" in the string, which means that only the left of the # was actually taken into Grafana. The login still worked after truncating the password accordingly.

aaronraimist:
Fixes #971

### DIFF
--- a/docs/configuring-playbook-prometheus-grafana.md
+++ b/docs/configuring-playbook-prometheus-grafana.md
@@ -17,11 +17,10 @@ matrix_grafana_anonymous_access: false
 
 # This has no relation to your Matrix user id. It can be any username you'd like.
 # Changing the username subsequently won't work.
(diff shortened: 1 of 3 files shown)


spantaleev:
Great! Thanks! :+1:

PR #985, opened 2021-04-06
Use synapse db host var instead of hard coding matrix-postgres

cvwright:
I found one place where the playbook is hardcoding "matrix-postgres" instead of using the variable `matrix_synapse_database_host`

### DIFF
--- a/roles/matrix-synapse/tasks/rust-synapse-compress-state/main.yml
+++ b/roles/matrix-synapse/tasks/rust-synapse-compress-state/main.yml
@@ -57,7 +57,7 @@
       --network={{ matrix_docker_network }}
       --env-file={{ matrix_postgres_base_path }}/env-postgres-psql
       {{ matrix_postgres_docker_image_latest }}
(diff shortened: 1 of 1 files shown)


pushytoxin:
It's hardcoded in many places, just grep for `matrix-postgres` and see. All the bridges hardcode it, for example.

I would like some input from people using an external pgsql database on how well the playbook behaves on their systems, and include any changes necessary to make that setup work seamlessly.

pushytoxin:
A side note, you don't need to sign off your commits, authorship is preserved in github pull requests.

spantaleev:
Does that even help? If you're not using `matrix-postgres`, then `--env-file={{ matrix_postgres_base_path }}/env-postgres-psql` would be pointing to a file that doesn't even exist (at least as it is right now).

cvwright:
Doh!  I thought a lot of the postgres support files were still created in order to have a place to run psql.  But I'm still pretty new to this.  If it doesn't matter, then we might as well close this for now.

spantaleev:
They're not, at least not yet. It might be a good idea to create them though, to allow tools like this (and others) to run.
I haven't looked much into it though.

Issue #520, opened 2020-05-27
Support podman as alternative to docker


ptman:
This is quite far in the future, as ubuntu 20.04 doesn't yet include podman, but:

https://packages.debian.org/bullseye/podman

Probably relates to #300 and #64

christianlupus:
Just a remark: If podman is supported (which is perfectly fine from my perspective), a parallel running instance of dockerd should not be broken. Otherwise all containers running on the machine will have to be migrated in one step. This might be impossible if other containers are running in parallel for different tasks.

spantaleev:
Just recording my last experience with Podman from some other tests I've done recently.

TLDR: Podman is just not good enough (for our use case) yet.

------------

On CentOS 8, a relatively old `podman` version is being shipped (`1.8.x`), with the latest being `2.2` right now.

Something as simple as:

```
podman network create --driver=bridge custom

podman run -it --rm --network=custom -p 80:80 docker.io/nginx:1.19.5-alpine
```

.. leads to the port not being exposed correctly. It's not accessible from the internet.

If the `--network=custom` call is removed and we use the default network, it works. Strange.

-------------

On Ubuntu 20.10, there's a `podman` package which is newer than the one in CentOS 8 (yay!), but packaging is broken.

Certain dependencies (`runc`, `iptables`) are not listed as such and need to be installed separately (`apt-get install -y runc iptables`).

`iptables`, okay. Maybe it's optional, as one may wish to run non-networked containers and be fine without it.

`runc` not being included as a dependency seems like a mistake. It doesn't inspire confidence.

After installing these dependencies manually, it seems to work and is not broken like the CentOS 8 package (likely because it's much newer - `2.0.x`, if I remember correctly).

-------------

For containers to talk to each other by name (something we rely on and quite enjoy to have available with Docker), we need the [dnsname](https://github.com/containers/dnsname) plugin.

Installing this plugin is a manual process which requires `git`, `golang` and then `make && make install ...`, along with modifying a network configuration file in `/etc/cni/..` to make it load the `dnsname` plugin for a given network. For the plugin to function `dnsmasq` also needs to be installed on the host.

I've verified that it works nicely though, but this kind of installation procedure is quite ugly.

--------------

An alternative to containers talking to each other by name (using the dnsname plugin) is making them explicitly pass IP addresses to one another.

Example `.service` file snippet:

```
[Unit]
Requires=matrix-postgres.service
After=matrix-postgres.service

[Service]

[...]

PR #986, opened 2021-04-07
Fix stuff and add Podman/Fedora support

penn5:


penn5:
The only change needed in vars.yml to use Podman is:
```
matrix_docker_dns_resolver_ip: 10.89.0.1
matrix_docker_package_name: podman-docker
```

penn5:
Closes #520

PR #990, opened 2021-04-10
use custom docker registry

haghighi-ahmad:
After [Docker Hub rate limitting](https://www.docker.com/increase-rate-limits) we need to use our OWN Docker registry mirrors OR using public ones -like Google ([GCR](https://mirror.gcr.io))- to bypass DH limitations.   
FYI: Hard coding `docker.io` prefix on all images tells `docker` daemon to pull images only from docker.io, 

Use-case one:
* `matrix_global_registry: ""` to use system (docker) configured Docker mirror

Use-case two:
* `matrix_global_registry: "our-docker-mirror.foo.bar/"` to pull images from `our-docker-mirror.foo.bar` Docker mirror.
* `matrix_global_registry: "docker.io/"` (Default Value)

Please feel free to change and adjust place/position/names :)


spantaleev:
Maybe it's better to:
- **either** leave the slash out of `matrix_global_registry` (e.g. `matrix_global_registry: "docker.io"`) and then hardcode it in each image variable (e.g. `matrix_bot_matrix_reminder_bot_docker_image: "{{ matrix_global_registry }}/anoa/matrix-reminder-bot:{{ matrix_bot_matrix_reminder_bot_version }}"`)
- **or** rename `matrix_global_registry` to something like `matrix_container_global_registry_prefix`

haghighi-ahmad:
I think the second method (rename to `matrix_container_global_registry_prefix`) is better.

Empty prefix (`matrix_container_global_registry_prefix: ""`) can be used in the case of setting a Local Mirror in Docker configs (`/etc/docker/daemon.json`), 
Hard codding a `/` in image URL prevents us to achieve that. 

Pull req. is ready @spantaleev

spantaleev:
Looks good! Thank you! :+1:

PR #998, opened 2021-04-15
GoMatrixHosting v0.4.2

GoMatrixHosting:
Lot's of minor fixes.
A new /doc/Installation_Minimal.md guide for performing a minimal installation.
Added a purge local/remote media function.

This system is finally at the point where I can imagine someone else playing with it without too much friction, so if you're curious about it go ahead and have a play with it. :)


spantaleev:
Thanks! :+1:

Issue #141, opened 2019-04-19
Trying to login from Riot Android app: "outdated TLS security protocol" error


jeromecc:
Hi!

Thanks for this amazing playbook. Worked like a charm.

I just installed on medica.im with fairly a standard configuration (almost all default) on Ubuntu 18.04. I could create a user, connect through the server's riot.medica.im web interface.

Trying to login from the Android app (downloaded from the Playstore) gave me this error.

>Your device is using an outdated TLS security protocol, vulnerable to attack, for your security you will not be able to connect

This is related to SSLHandshakeException in the riot-android code https://github.com/vector-im/riot-android/blob/a57581ca0d1e33a0a3cb9dbfce185fa50e9e6fc4/vector/src/main/java/im/vector/activity/LoginActivity.java

I will investigate further. I guess it has to do with the nginx TLS configuration.

I'm pretty sure it's not related to my phone because i reinstalled the app today and connection to matrix.org works seamlessly.

Thanks.

aaronraimist:
What version of Android? If you are using a fairly old version of Android it may not support modern TLS protocols.

jeromecc:
Thing is i can connect to my usual matrix.org account on this phone with the most recent version of the app. I am connected right now. Would it work with matrix.org and not with the server i just installed thanks to matrix-docker-ansible-deploy
 ?
Android 4.4.2

aaronraimist:
Yeah that is quite an old version of Android. Matrix.org may support that but by default the playbook doesn't. You'll have to do research and determine what modifications you need to make if you want to use it with that version of Android but as the error says it is going to be less secure. https://github.com/vector-im/riot-android/issues/1376#issuecomment-427761890 appears to have a working nginx config.

jeromecc:
https://github.com/vector-im/riot-android/issues/1376#issuecomment-427761890 we found the same comment ;-) thanks for your help i guess i will add a few ciphers

Maybe I could send a PR to update the docs to warn about this potential pitfall?

aaronraimist:
👍

jeromecc:
i realize my phone is old from a dev point of view but according to Google it still represents 7.6% of the Android user base.
4.4 | KitKat | 19 | 7.6%
https://developer.android.com/about/dashboards

riot android app advertises "Requires Android 4.1 and up" on the Playstore. That is 10.6 % of all Android users...

jeromecc:
[...]

PR #999, opened 2021-04-15
Disable TLS 1.0 and TLS 1.1 for coturn

sakkiii:
These old versions of TLS rely on MD5 and SHA-1, both now broken, and contain other flaws. TLS 1.0 is no longer PCI-DSS compliant and the TLS working group has adopted a document to deprecate TLS 1.0 and TLS 1.1.

### DIFF
--- a/roles/matrix-coturn/templates/turnserver.conf.j2
+++ b/roles/matrix-coturn/templates/turnserver.conf.j2
@@ -16,6 +16,8 @@ no-cli
 {% if matrix_coturn_tls_enabled %}
 cert={{ matrix_coturn_tls_cert_path }}
 pkey={{ matrix_coturn_tls_key_path }}
+no-tlsv1
+no-tlsv1_1
 {% else %}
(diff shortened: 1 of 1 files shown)


pushytoxin:
I suggest you title your commit to include that this relates to coturn.

aaronraimist:
It would be good if this could be configurable/follow the existing config option. https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/f18037ae42b3837ac14ab96f6868731069a5b6be

Otherwise this change will break things for people running old software like https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/141

spantaleev:
Thanks! :+1: 

It may be useful to have some other "more global" variable, which could control both coturn SSL and nginx SSL.

spantaleev:
I've made this configurable in fcb9e96 and announced it in the changelog, as it seems like it could be important to some people.

PR #1005, opened 2021-04-17
Improve security for grafana

sakkiii:
- removed duplicate X-Content-Type-Options
- removed duplicate X-Frame-Options header
- Referrer-Policy [Might consider adding variable]
- Secure flag with cookies
- matrix_grafana_content_security_policy variable for [Content Security Policy](https://grafana.com/docs/grafana/latest/administration/configuration/#content_security_policy)

### DIFF
--- a/roles/matrix-grafana/defaults/main.yml
+++ b/roles/matrix-grafana/defaults/main.yml
@@ -31,6 +31,12 @@ matrix_grafana_anonymous_access_org_name: 'Main Org.'
 matrix_grafana_default_admin_user: admin
 matrix_grafana_default_admin_password: admin
 
+# Set to true to add the Content-Security-Policy header to your requests.
+# CSP allows to control resources that the user agent can load and helps 
+# prevent XSS attacks.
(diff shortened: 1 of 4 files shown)


spantaleev:
Thanks! :+1:

aaronraimist:
@sakkiii does grafana actually work for you with the CSP enabled? It doesn't for me. Shows some error saying try refreshing the page and in the browser console is several CSP violations.

sakkiii:
@aaronraimist yeah it does work for me. i just checked with Firefox, Brave(based on the Chromium) & Safari for all working fine.
Can you share browser logs?

aaronraimist:
@sakkiii This is what I see in Safari:



Grafana does seem to work in Firefox but there are a bunch of things being blocked by the CSP

sakkiii:
@aaronraimist unable to regenerate similar issue but added backward compatible with older browsers.
Could you review this, may solve your problem #1036

aaronraimist:
@sakkiii interestingly as of today I cannot reproduce it with or without that patch. 🤷‍♂️

PR #1007, opened 2021-04-18
Don't expose nginx version with each response

teutat3s:


spantaleev:
Good idea! Thanks! :+1:

Issue #983, opened 2021-04-05
matrixdotorg/synapse:v1.30.1 taking up port 9100 required by matrix-prometheus-node-exporter


schwony:
I‘m unable to use „matrix_prometheus_node_exporter_enabled: true“

it will fail to start due to port 9100 already in use by the matrix-synapse docker image

```
journalctl -fu matrix-prometheus-node-exporter.service shows

„err="listen tcp :9100: bind: address already in use"“
```

```
docker ps | grep 9100                                 
864d1774f819   matrixdotorg/synapse:v1.30.1               "/start.py run -m sy…"   7 minutes ago   Up 7 minutes (healthy)    127.0.0.1:8008->8008/tcp, 8009/tcp, 127.0.0.1:8048->8048/tcp, 8448/tcp, 0.0.0.0:9100->9100/tcp                                         matrix-synapse
```

I‘ve tried binding to a different port via: 
`matrix_prometheus_node_exporter_container_http_host_bind_port: '9101'`
or even
`matrix_synapse_container_metrics_api_host_bind_port: '9101'`

in inventory/host_vars/matrix.DOMAIN/vars.yml to no avail.

any suggestions?

spantaleev:
This only happened when `matrix_nginx_proxy_enabled: false`.

The playbook was trying to expose both Synapse's metrics port and prometheus-node-exporter at `127.0.0.1:9100`.

With a1a0d91, we now expose prometheus-node-exporter at `127.0.0.1:9200`, should you need it.

If you're using the integrated prometheus service, it will scrape it over the local `matrix` Docker network and won't even use that `127.0.0.1:9200` port.

PR #1008, opened 2021-04-19
security** node-exporter data & port publicly exposed

sakkiii:
By default node-exporter docker container expose port 9100

```
**netstat -ntlp**
tcp6       0      0 :::9100                 :::*                    LISTEN      7577/node_exporter
```

### DIFF
--- a/roles/matrix-prometheus-node-exporter/defaults/main.yml
+++ b/roles/matrix-prometheus-node-exporter/defaults/main.yml
@@ -18,5 +18,5 @@ matrix_prometheus_node_exporter_systemd_wanted_services_list: []
 
 # Controls whether the matrix-prometheus container exposes its HTTP port (tcp/9100 in the container).
 #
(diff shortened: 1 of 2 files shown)


spantaleev:
Thanks a lot! :+1: 

This actually explains why a1a0d9160410ec3e1a was not an effective fix for #983.

When `--net=host` is used, `prometheus-node-exporter` was always binding to port 9100.

Unfortunately.. I don't think binding to `localhost:9100` is good enough for us, since such a port (on the loopback interface) can't be accessed by other containers (like `matrix-prometheus`). I'll work on some other fix for this now.

spantaleev:
As far as security is concerned, I'm not really sure it's a security problem.

I do see it listening on `tcp6       0      0 :::9100` for me as well, but I haven't been able to connect to it from outside (neither via IPv4, nor via IPv6). I may be doing something wrong though. My IPv6 connectivity seems to be fine when I'm attempting to connect to port 80/443 though.

It may have something to do with https://github.com/prometheus/node_exporter/issues/1653.

PR #1009, opened 2021-04-19
Switch to official coturn image

thedanbob:
Coturn got an official image recently (https://github.com/coturn/coturn/pull/746) from the maintainer of `instrumentisto/coturn`. As far as I can tell the original image isn't deprecated yet but probably will be soon, e.g. https://github.com/instrumentisto/coturn-docker-image/issues/37#issuecomment-767876958.

### DIFF
--- a/roles/matrix-coturn/defaults/main.yml
+++ b/roles/matrix-coturn/defaults/main.yml
@@ -1,10 +1,10 @@
 matrix_coturn_enabled: true
 
 matrix_coturn_container_image_self_build: false
-matrix_coturn_container_image_self_build_repo: "https://github.com/instrumentisto/coturn-docker-image.git"
+matrix_coturn_container_image_self_build_repo: "https://github.com/coturn/coturn/tree/master/docker/coturn/alpine.git"
 
 matrix_coturn_version: 4.5.2
-matrix_coturn_docker_image: "{{ matrix_coturn_docker_image_name_prefix }}instrumentisto/coturn:{{ matrix_coturn_version }}"
+matrix_coturn_docker_image: "{{ matrix_coturn_docker_image_name_prefix }}coturn/coturn:{{ matrix_coturn_version }}"
 matrix_coturn_docker_image_name_prefix: "{{ 'localhost/' if matrix_coturn_container_image_self_build else matrix_container_global_registry_prefix }}"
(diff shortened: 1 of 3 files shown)


spantaleev:
Thanks! :+1: 

I wonder if we need `NET_BIND_SERVICE` if we're not asking it to bind to privileged ports (`<=1023`). Have you tested if it works without this?

thedanbob:
I did, unfortunately it doesn't start due to this line in the Dockerfile: https://github.com/coturn/coturn/blob/b31b0049220ef26e87d784f65e3db286cbdc50a2/docker/coturn/alpine/Dockerfile#L120

spantaleev:
That's somewhat unfortunate, but it's no big deal.

Thanks for switching us to the official image! :+1:

roles/matrix-coturn/defaults/main.yml
```diff
@@ -1,10 +1,10 @@
 matrix_coturn_enabled: true
 
 matrix_coturn_container_image_self_build: false
-matrix_coturn_container_image_self_build_repo: "https://github.com/instrumentisto/coturn-docker-image.git"
+matrix_coturn_container_image_self_build_repo: "https://github.com/coturn/coturn/tree/master/docker/coturn/alpine.git"
 
 matrix_coturn_version: 4.5.2
-matrix_coturn_docker_image: "{{ matrix_coturn_docker_image_name_prefix }}instrumentisto/coturn:{{ matrix_coturn_version }}"
+matrix_coturn_docker_image: "{{ matrix_coturn_docker_image_name_prefix }}coturn/coturn:{{ matrix_coturn_version }}"
```
tyranron, line 7: @thedanbob @spantaleev despite specifying an `alpine` Dockerfile above in `matrix_coturn_container_image_self_build_repo`, you use `debian` image here. To use `alpine` Docker image the `-alpine` postfix is required now:
```yaml
matrix_coturn_docker_image: "{{ matrix_coturn_docker_image_name_prefix }}coturn/coturn:{{ matrix_coturn_version }}-alpine"
```

spantaleev: Good point! Fixed in 62c0587b6aaed84c

thedanbob: Good catch! I didn't notice that the alpine image wasn't the default for a given version tag. (Really should have, since I did notice the new image was much bigger than the old one.)

I've submitted #1019 to switch to the alpine image.

PR #1013, opened 2021-04-21
Update synapse version (1.31.0 -> 1.32.1)

rakshazi:
**NOTE**: Matrix.org now builds both amd64 and arm64 containers at the same time, may be it's time to remove separate arm64 version param?


spantaleev:
Thanks, but I've done it in a more complete manner in 1fb54a37cba126235b168 (includes configuration updates).

amd64 and arm64 builds are triggered at the same time, but they actually land at least an hour apart, so we can't do it together.
Unless we decide to always push in a delayed manner (both at the same time) - that's probably ideal and it's something I'm doing now anyway (waiting for the arm64 image to land).

PR #1024, opened 2021-04-24
HSTS preload + X-XSS enables 

sakkiii:
1. **HSTS Preloading:**
By default `matrix_nginx_proxy_hsts_preload_enable: false`
But its strongest and recommended form, the [HSTS policy](https://www.chromium.org/hsts) includes all subdomains, and indicates a willingness to be “preloaded” into browsers:
`Strict-Transport-Security: max-age=31536000; includeSubDomains; preload`

2. **X-Xss-Protection:**
`1; mode=block` which tells the browser to block the response if it detects an attack rather than sanitising the script.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,23 @@
+# 2021-05-21
+
+## Hydrogen support
+
+Thanks to [Aaron Raimist](https://github.com/aaronraimist), the playbook now supports [Hydrogen](https://github.com/vector-im/hydrogen-web) - a new lightweight matrix client with legacy and mobile browser support.
+
+By default, we still install Element, as Hydrogen is still not fully-featured. Still, people who'd like to try Hydrogen out can now install it via the playbook.
+
+Additional details are available in [Setting up Hydrogen](docs/configuring-playbook-client-hydrogen.md).
+
+
+# 2021-05-19
+
+## Heisenbridge support
+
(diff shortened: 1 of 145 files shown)


spantaleev:
I wonder if we should be setting `preload` by default. Also see here: https://hstspreload.org/#opt-in

>  If you maintain a project that provides HTTPS configuration advice or provides an option to enable HSTS, do not include the preload directive by default. We get regular emails from site operators who tried out HSTS this way, only to find themselves on the preload list by the time they find they need to remove HSTS to access certain subdomains. Removal tends to be slow and painful for those sites.
> 
> It's great to support HSTS preloading as a best practice, and for projects to provide a simple option to enable it. However, site operators who enable HSTS should know about the long-term consequences of preloading before they turn it on for a given domain. They should also be informed that they need to meet additional requirements and submit their site to hstspreload.org to ensure that it is successfully preloaded (i.e. to get the full protection of the intended configuration).

sakkiii:
@spantaleev you are right and i have add new variable matrix_nginx_proxy_hsts_preload_enable for `HSTS preload`

spantaleev:
Great! `matrix_nginx_proxy_hsts_preload_enabled` would be more consistent with the rest of our variables though.

spantaleev:
Still seems to be called `matrix_nginx_proxy_hsts_preload_enable` rather than `matrix_nginx_proxy_hsts_preload_enabled`?

sakkiii:
> 
> 
> Still seems to be called `matrix_nginx_proxy_hsts_preload_enable` rather than `matrix_nginx_proxy_hsts_preload_enabled`?

yeah done

spantaleev:
Great! Thanks! :+1:

PR #1036, opened 2021-05-05
Grafana csp template backward compatible with older browsers

sakkiii:
`matrix_grafana_content_security_policy_customized` to customized content security policy template.
- added 'unsafe-inline' (ignored by browsers supporting nonces/hashes) to be backward compatible with older browsers.
- added https: and http: url schemes (ignored by browsers supporting 'strict-dynamic') to be backward compatible with older browsers.

### DIFF
--- a/roles/matrix-grafana/templates/grafana.ini.j2
+++ b/roles/matrix-grafana/templates/grafana.ini.j2
@@ -9,7 +9,7 @@ admin_password = """{{ matrix_grafana_default_admin_password }}"""
 content_security_policy = "{{ matrix_grafana_content_security_policy }}"
 
 # specify content security policy template to customized template
-{% if matrix_synapse_metrics_enabled %}
+{% if matrix_grafana_content_security_policy_customized %}
 content_security_policy_template = """script-src http: https: 'unsafe-inline' 'unsafe-eval' 'strict-dynamic' $NONCE;object-src 'none';font-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data:;base-uri 'self';connect-src 'self' grafana.com;manifest-src 'self';media-src 'none';form-action 'self';"""
 {% else %}
 {% endif %}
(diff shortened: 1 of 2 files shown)


roles/matrix-grafana/templates/grafana.ini.j2
```diff
@@ -8,6 +8,12 @@ admin_password = """{{ matrix_grafana_default_admin_password }}"""
 # specify content_security_policy to add the Content-Security-Policy header to your requests
 content_security_policy = "{{ matrix_grafana_content_security_policy }}"
 
+# specify content security policy template to customized template
+{% if matrix_synapse_metrics_enabled %}
```
aaronraimist, line 12:
```suggestion
{% if matrix_grafana_content_security_policy_customized %}
```

Is this supposed to be `matrix_grafana_content_security_policy_customized`?

roles/matrix-grafana/templates/grafana.ini.j2
```diff
@@ -8,6 +8,12 @@ admin_password = """{{ matrix_grafana_default_admin_password }}"""
...
+content_security_policy_template = """script-src http: https: 'unsafe-inline' 'unsafe-eval' 'strict-dynamic' $NONCE;object-src 'none';font-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data:;base-uri 'self';connect-src 'self' grafana.com;manifest-src 'self';media-src 'none';form-action 'self';"""
+{% else %}
```
aaronraimist, line 14: You should be able to get rid of this `{% else %}`

sakkiii:
@aaronraimist thanks 👍

spantaleev:
A good way to make it even more flexible is to put the policy string into `matrix_grafana_content_security_policy_customized`.

Then people would be able to override it with their own other policy or to unset it to get rid of it.

Still, thanks for adding this, even as it is now! :+1:

thedanbob:
Grafana has been broken on iOS Safari the last couple of versions and this fixed it. Cheers!

PR #1055, opened 2021-05-15
Add Hydrogen

aaronraimist:
TODO:
- [x] Write docs
- [x] Figure out how to make custom `config.json` work
- [x] Check if `Sync failed because of TypeError: child is undefined` is a Hydrogen bug or something wrong with this setup
  - haven't seen this again so assuming Hydrogen bug
- [x] Make sure CSP makes sense, currently just copied from Element role


roles/matrix-client-hydrogen/tasks/setup.yml
```diff
@@ -0,0 +1,108 @@
...
+    dest: "{{ matrix_client_hydrogen_docker_src_files_path }}"
+    version: "{{ matrix_client_hydrogen_docker_image.split(':')[1] }}"
+    force: "yes"
+  register: matrix_client_hydrogen_git_pull_results
+  when: "matrix_client_hydrogen_enabled|bool and matrix_client_hydrogen_container_image_self_build|bool"
+
+- name: Ensure Hydrogen configuration installed
+  copy:
+    content: "{{ matrix_client_hydrogen_configuration|to_nice_json }}"
+    dest: "{{ matrix_client_hydrogen_docker_src_files_path }}/assets/config.json"
```
spantaleev, line 39: Shouldn't this be:

```yaml
dest: "{{ matrix_client_hydrogen_data_path }}/config.json"
```

aaronraimist: No. I tried that first but Hydrogen only looks for a `config.json` when it gets built. It doesn't work like Element. https://github.com/vector-im/hydrogen-web/issues/324

roles/matrix-client-hydrogen/templates/config.json.j2
```diff
@@ -0,0 +1,3 @@
+{
+	"defaultHomeServer": {{ matrix_client_hydrogen_default_hs_url|string|to_json }}
```
spantaleev, line 2: https://github.com/vector-im/hydrogen-web/blob/master/assets/config.json seems to use a domain name for `defaultHomeServer`, not a full URL like we currently are. Maybe both work? Maybe not.

aaronraimist: I've tested. Both work.

roles/matrix-client-hydrogen/tasks/setup.yml
spantaleev: Ah, I see. That's kind of misleading then.

So, in the general case (no self-building), we'll use the [default upstream configuration](https://github.com/vector-im/hydrogen-web/blob/master/assets/config.json), which points us to `matrix.org`. That's not really ideal and will hopefully be addressed in vector-im/hydrogen-web#324 (or some other issue), so that users would just need to enter their localpart (`.well-known` may be made to work in the future, but that's asking people to enter their full MXID is not ideal).

aaronraimist: Yeah. At the moment the pre-built images no not work so self building will be enabled by default.

aaronraimist:
I think this is ready

spantaleev:
Great job! Thank you, Aaron! :+1:

PR #1057, opened 2021-05-15
Enable OCSP Stapling

sakkiii:
`matrix_nginx_proxy_ocsp_stapling_enabled` Added. Default value:

```
# set false to disable OCSP Stapling
matrix_nginx_proxy_ocsp_stapling_enabled: true
```

- OCSP Stapling eliminating the need for clients to contact the CA, with the aim of improving both security and performance.
- OCSP stapling can provide a performance boost of up to 30%
- nginx web server supports OCSP stapling since version 1.3.7.

> **warning** Nginx is lazy loading OCSP responses, which means that for the first few web requests might unable to add the OCSP response.
  
Learn more about what it is here:
 - https://en.wikipedia.org/wiki/OCSP_stapling
 - https://blog.cloudflare.com/high-reliability-ocsp-stapling/
 - https://blog.mozilla.org/security/2013/07/29/ocsp-stapling-in-firefox/

### DIFF
--- a/roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-client-element.conf.j2
+++ b/roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-client-element.conf.j2
@@ -10,12 +10,6 @@
 	{% if matrix_nginx_proxy_floc_optout_enabled %}
 		add_header Permissions-Policy interest-cohort=() always;
 	{% endif %}
-	
-	{% if matrix_nginx_proxy_ocsp_stapling_enabled and matrix_ssl_retrieval_method in ["lets-encrypt", "manually-managed"] %}
-		ssl_stapling on;
-		ssl_stapling_verify on;
-		ssl_trusted_certificate {{ matrix_ssl_config_dir_path }}/live/{{ matrix_nginx_proxy_base_domain_hostname }}/chain.pem;
-	{% endif %}
 
 	{% for configuration_block in matrix_nginx_proxy_proxy_element_additional_server_configuration_blocks %}
 		{{- configuration_block }}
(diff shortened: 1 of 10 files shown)


roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-client-element.conf.j2
```diff
@@ -10,6 +10,12 @@
 	{% if matrix_nginx_proxy_floc_optout_enabled %}
 		add_header Permissions-Policy interest-cohort=() always;
 	{% endif %}
+	
+	{% if matrix_nginx_proxy_ocsp_stapling_enabled and matrix_ssl_retrieval_method in ["lets-encrypt", "manually-managed"] %}
+		ssl_stapling on;
+		ssl_stapling_verify on;
+		ssl_trusted_certificate {{ matrix_ssl_config_dir_path }}/live/{{ matrix_nginx_proxy_base_domain_hostname }}/chain.pem;
+	{% endif %}
```
spantaleev, line 18: This likely applies to all other vhosts too.

I see you've added this new configuration inside `{% macro render_vhost_directives() %}`.

This **usually** gets rendered into an SSL-enabled vhost, but it's not always the case.

If `matrix_nginx_proxy_https_enabled` is `false`, then `render_vhost_directives()` gets rendered into a plain-HTTP `server` block. See here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/e4dd933cf0f48a8722a43239b07a9335c85c4f14/roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-client-element.conf.j2#L42-L60

-------------

Not sure if having `ssl_*` directives in non-SSL-enabled `server` block causes critical nginx errors (maybe it does). Even if it doesn't, it's still not very clean to do it there.

I suggest all these get moved down to the SSL-enabled `server` block, next to all the other `ssl_*` directives.

sakkiii: yeah right thanks 👍

spantaleev:
Wonderful! Thank you! :+1: 

I've done some fixups and also added support for OCSP stapling for the federation vhost (`matrix.DOMAIN:8448`). I'm not sure sure if it's beneficial doing so, however.

PR #1061, opened 2021-05-18
Optimize SSL session

sakkiii:
**ssl_session_cache:**
- Creating a cache of TLS connection parameters reduces the number of handshakes 
  and thus can improve the performance of application.
- Default session cache is not optimal as it can be used by only one worker process
  and can cause memory fragmentation. It is much better to use shared cache.
- Learn More: https://nginx.org/en/docs/http/ngx_http_ssl_module.html

**ssl_session_timeout:**
- Nginx by default it is set to 5 minutes which is very low.
  should be like 4h or 1d but will require you to increase the size of cache.
- Learn More: 
    https://github.com/certbot/certbot/issues/6903
    https://github.com/mozilla/server-side-tls/issues/198

**ssl_session_tickets:**
- In case of session tickets, information about session is given to the client.
  Enabling this improve performance also make Perfect Forward Secrecy useless.
- If you would instead like to use ssl_session_tickets by yourself, you can set
  matrix_nginx_proxy_ssl_session_tickets_off false.
- Learn More: https://github.com/mozilla/server-side-tls/issues/135

**Note: Presets are taken from Mozilla's Server Side TLS Recommended configurations**

### DIFF
--- a/roles/matrix-nginx-proxy/defaults/main.yml
+++ b/roles/matrix-nginx-proxy/defaults/main.yml
@@ -386,6 +386,34 @@ matrix_ssl_log_dir_path: "{{ matrix_ssl_base_path }}/log"
 matrix_ssl_pre_obtaining_required_service_name: ~
 matrix_ssl_pre_obtaining_required_service_start_wait_time_seconds: 60
 
+# Nginx Optimize SSL Session
+#
+# ssl_session_cache:
+# - Creating a cache of TLS connection parameters reduces the number of handshakes 
+#   and thus can improve the performance of application.
+# - Default session cache is not optimal as it can be used by only one worker process
+#   and can cause memory fragmentation. It is much better to use shared cache.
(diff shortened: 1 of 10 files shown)


spantaleev:
How much security are we trading for performance with this? I wonder if these are good defaults.

sakkiii:
**Improve Performance:**
`ssl_session_cache `and `ssl_session_timeout `both improve performance of `TLS-Handshake` . As default nginx values/configuration are not optimal.

**Improve Security:**
By default Enabled `ssl_session_tickets` doesn't affect performance. Enabling this  should improve performance when used in conjunction with ssl_session_cache. But it is recommended not use because proper rotation of session ticket encryption key is not yet implemented in Nginx.

Overall this should improve performance and security both.

spantaleev:
Thanks for the clarification and for working on these improvements! I've merged it as it is! :bow:

PR #1067, opened 2021-05-19
Configure root_url in granafa.ini

theCalcaholic:
Because grafana is running behind a reverse proxy, it will in some cases (e.g. in notifications) show links to localhost:3000 instead the web interface address.

This PR fixes that.


spantaleev:
Thanks! :+1:

Issue #1073, opened 2021-05-21
Fix OCSP Stapling with self signed certs


aaronraimist:
https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/1057 broke using self signed certs because that PR uses a file called `chain.pem` but the instructions say to call the file `fullchain.pem`, see https://matrix.to/#/!YTvKGNlinIzlkMTVRl:matrix.org/$XrnMfun1xKC4xLtQ09A8Jnmpv2StWC3gyIeIK76cjWk?via=matrix.org&via=privacytools.io&via=mozilla.org

cc @sakkiii

aaronraimist:
I'm guessing the right solution may be to disable that when self signing is enabled

PR #1074, opened 2021-05-21
OCSP Stapling self-signed issue

sakkiii:
Fix https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/1073

### DIFF
--- a/docs/configuring-playbook-ssl-certificates.md
+++ b/docs/configuring-playbook-ssl-certificates.md
@@ -43,6 +43,7 @@ With such a configuration, the playbook would expect you to drop the SSL certifi
 
 - `<matrix_ssl_config_dir_path>/live/<domain>/fullchain.pem`
 - `<matrix_ssl_config_dir_path>/live/<domain>/privkey.pem`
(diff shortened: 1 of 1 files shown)


spantaleev:
I think we need to make changes to `roles/matrix-nginx-proxy/tasks/ssl/setup_ssl_self_signed_obtain_for_domain.yml` to create a `chain.pem` file.

We should also migrate over existing users.

aaronraimist:
Does OCSP Stapling even do anything for a self signed certificate?

spantaleev:
You're right! I guess it doesn't make sense. Disabled here: 9a8a81f

sakkiii:
OCSP Stapling work only with trusted certificates. it will work fine with lets-encrypt.

But again some concern with manual-certs as well

m-amir-ir:
Hello there,

I really agree with @sakkiii about that. 
but I think this line is better : 
```
matrix_nginx_proxy_ocsp_stapling_enabled: "{{ matrix_ssl_retrieval_method == 'lets-encrypt' }}"
```

PR #1037, opened 2021-05-05
Update Jitsi (5142 -> 5765-1)

pushytoxin:
Also closes #975 #1028

PR #1076, opened 2021-05-21
Allow Jitsi XMPP websocket support for users using own webserver.

Eagle-251:
Since the update to Jitsi from 5142 to 5765 (#1037) and the corresponding changes to the nginx template to allow for the XMPP websocket client support (#1037), people using their webservers might not have been able to create calls using the web client. This manifested in the "You have been disconnected" error when starting a call.

The above commit added and {{else}} for people that disabled the internal nginx-proxy that redirected ```/xmpp-websocket``` to ```http://127.0.0.1:5280;```.
The prosody containers that ```/xmpp-websocket``` needed to redirect to lacked the necessary port binding to allow this for people running their own webserver.

To fix this I have added:
 - Conditional localhost Port bindings for Jitsi Prosody systemd template
- Added variable to main.yml to allow overriding from vars.yml

Manually editing the systemd file for prodsody for my server seems to fix this and remove the "You have been disconnected" error.

I am sure there is a way to more intelligently set this when people set ```matrix_nginx_proxy_enabled: false``` but my knowledge of the playbook (and ansible) is not yet such to do that. I will keep thinking and make another PR when I work it out!

(This being my first PR, I may have done things wrong, apologies in advance if so!)

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1053,6 +1053,8 @@ matrix_jitsi_web_container_http_host_bind_port: "{{ '' if matrix_nginx_proxy_ena
 
 matrix_jitsi_jvb_container_colibri_ws_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:13090' }}"
 
+matrix_jitsi_prosody_container_http_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:5280' }}"
+
 matrix_jitsi_jibri_xmpp_password: "{{ matrix_synapse_macaroon_secret_key | password_hash('sha512', 'jibri') | to_uuid }}"
 matrix_jitsi_jicofo_auth_password: "{{ matrix_synapse_macaroon_secret_key | password_hash('sha512', 'jicofo') | to_uuid }}"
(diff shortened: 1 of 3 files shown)


spantaleev:
A better solution (and one we do for the other components) would be to define `matrix_jitsi_prosody_container_http_host_bind_port` as empty in the role (`roles/matrix-jitsi/defaults/main.yml`:

```yaml
matrix_jitsi_prosody_container_http_host_bind_port: ''
```

Then to have `group_vars/matrix_servers` redefine it like this:

```yaml
matrix_jitsi_prosody_container_http_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:5280' }}"
```

And then the systemd service file would do:

```
{% if matrix_jitsi_prosody_container_http_host_bind_port %}
-p {{ matrix_jitsi_prosody_container_http_host_bind_port }}:5280
{% endif %}
```

--------------

The reason for doing it this way is that we try to keep roles independent from one another as much as possible. Using `matrix_nginx_proxy_*` variables inside the `matrix-jitsi` role means crossing boundaries.

When we need to hook roles together, we use `group_vars/matrix_servers`.

Eagle-251:
Great! That makes more sense. Hadn't paid much attention to that file before, but upon reading I see it was the solution I was looking for. Thank you for explaining it!
I have made the changes, hopefully now it follows the convention of the playbook!
Tested this with my deployment (using host nginx) and seems to work.

spantaleev:
Great! Thanks! :+1:

Issue #1077, opened 2021-05-21
invalid mount config for type "bind": bind source path does not exist: /matrix/client-hydrogen/nginx.conf


ptman:
How have any of you managed to get hydrogen working?

ptman:
```
failed: [matrix.DOMAIN] (item=matrix-client-hydrogen.service) => {"ansible_loop_var": "item", "changed": false, "item": "matrix-client-hydrogen.service", "msg": "matrix-client-hydrogen.service was not detected to be running. It's possible that there's a configuration problem or another service on your server interferes with it (uses the same ports, etc.). Try running `systemctl status matrix-client-hydrogen.service` and `journalctl -fu matrix-client-hydrogen.service` on the server to investigate. If you're on a slow or overloaded server, it may be that services take a longer time to start and that this error is a false-positive. You can consider raising the value of the `matrix_common_after_systemd_service_start_wait_for_timeout_seconds` variable. See `roles/matrix-common-after/defaults/main.yml` for more details about that."}
```

Sirius-KiH:
It seems that the nginx configuration is missing. Also, the config.json is not installed in the right place and therefore cannot be mounted either.

I managed to get it to work by temporarily mounting the nginx config for element and fixing the path of the config.json by changing line 26 and 27 in matrix-client-hydrogen.service.j2 to this:

```
--mount type=bind,src={{ matrix_client_element_data_path }}/nginx.conf,dst=/etc/nginx/nginx.conf,ro \
--mount type=bind,src={{ matrix_client_hydrogen_docker_src_files_path }}/assets/config.json,dst=/app/config.json,ro \
```

Though this is kind of a hacky workaround and should probably be replaced with a proper fix.

ptman:
@aaronraimist any ideas?

aaronraimist:
Uh oops I see what I did. Try this PR and see if it works for you https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/1079

PR #1079, opened 2021-05-21
Fix hydrogen

aaronraimist:
Fixes https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/1077


spantaleev:
Thank you! :+1:

PR #1083, opened 2021-05-25
Synapse LDAP auth: add support for Active Directory

haghighi-ahmad:
Add two fields in synapse/homeserver.yaml.j2 to support AD,  
see https://github.com/matrix-org/matrix-synapse-ldap3#active-directory-forest-support

### DIFF
--- a/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
+++ b/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
@@ -2596,6 +2596,12 @@ password_providers:
       uri: {{ matrix_synapse_ext_password_provider_ldap_uri|string|to_json }}
       start_tls: {{ matrix_synapse_ext_password_provider_ldap_start_tls|to_json }}
       base: {{ matrix_synapse_ext_password_provider_ldap_base|string|to_json }}
+  {% if matrix_synapse_ext_password_provider_ldap_is_active_directory %}
+      active_directory: true
+  {% endif %}
+  {% if matrix_synapse_ext_password_provider_ldap_default_domain != '' %}
+      default_domain: {{ matrix_synapse_ext_password_provider_ldap_default_domain|string|to_json }}
+  {% endif %}
       attributes:
         uid: {{ matrix_synapse_ext_password_provider_ldap_attributes_uid|string|to_json }}
         mail: {{ matrix_synapse_ext_password_provider_ldap_attributes_mail|string|to_json }}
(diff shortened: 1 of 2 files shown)


roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -2596,6 +2596,12 @@ password_providers:
       uri: {{ matrix_synapse_ext_password_provider_ldap_uri|string|to_json }}
       start_tls: {{ matrix_synapse_ext_password_provider_ldap_start_tls|to_json }}
       base: {{ matrix_synapse_ext_password_provider_ldap_base|string|to_json }}
+  {% if matrix_synapse_ext_password_provider_ldap_is_active_directory %}
+      active_directory: true
```
spantaleev, line 2600: Calling this variable `matrix_synapse_ext_password_provider_ldap_active_directory` and giving it a default value of `false` is probably better.

This line can then change to:

```yaml
active_directory: "{{ matrix_synapse_ext_password_provider_ldap_active_directory|to_json }}"
```

Likewise for `default_domain`. Perhaps `''` is a good default value which can always be passed, avoiding the need for having `if` checks.

haghighi-ahmad: By default `active_directory` is `Flase` [1] and `default_domain` is ONLY used when Active Directory is enable (`active_directory: true`) [2,3]

Links:   
* [1] [ldap_auth_provider.py#L402](https://github.com/matrix-org/matrix-synapse-ldap3/blob/main/ldap_auth_provider.py#L402)   
* [2] [ldap_auth_provider.py#L403](https://github.com/matrix-org/matrix-synapse-ldap3/blob/main/ldap_auth_provider.py#L403)       
* [3] [ldap_auth_provider.py#L88](https://github.com/matrix-org/matrix-synapse-ldap3/blob/main/ldap_auth_provider.py#L88)  

I suggest:  

One:  
```yaml
  {% if matrix_synapse_ext_password_provider_ldap_is_active_directory %}
      active_directory: true
      default_domain: {{ matrix_synapse_ext_password_provider_ldap_default_domain|string|to_json }}
  {% else %}
      active_directory: false
  {% endif %}
```
  
Two (For non AD LDAPs `default_domain` exist in Synapse config but not used.):   
```yaml  
      active_directory: "{{ matrix_synapse_ext_password_provider_ldap_active_directory|to_json }}"
      default_domain: {{ matrix_synapse_ext_password_provider_ldap_default_domain|string|to_json }}
```

@spantaleev

spantaleev: I think the 2nd one is better. Avoiding `if`s also avoids potential problem with YAML indentation.

haghighi-ahmad:
It's ready @spantaleev

PR #1085, opened 2021-05-25
GoMatrixHosting v0.4.7

GoMatrixHosting:
- reinvented database purging section, seperated final shrinking that causes downtime.
- fixed AWX issue causing rust-synapse-compress-state section to not execute.
- added more reliable script method of generating the total room list.


spantaleev:
Thanks! :+1:

PR #1089, opened 2021-05-26
Update IRC appservice

thomwiggers:
I haven't tested this, but the [changelog](https://github.com/matrix-org/matrix-appservice-irc/releases/tag/0.26.0) isn't that exciting


spantaleev:
Thanks! :+1:

PR #1092, opened 2021-05-26
Server migration doc update

wol-win:
Clarify that migration as per this document is only applicable between servers of same architecture


spantaleev:
I've improved wording a bit. Thank you! :+1:

PR #1097, opened 2021-05-28
Fix local/bin scripts autocompletion by adding rx perms to everyone

pushytoxin:
It's mildly annoying when trying to execute these scripts while logged in as a regular user, as the missing execute permissions will hinder autocompletion even when trying to use with sudo.

These shell scripts don't contain secrets, but may fail when ran by a regular user. The failure is due to the lack of access to the /matrix directory, and does not result in any damage.


spantaleev:
Good point! Thank you! :+1:

PR #1099, opened 2021-05-31
Heisenbridge identd on unprivileged port

hifi:
Fixes running the container as an unprivileged user.


spantaleev:
Awesome! Thanks! :+1:

Issue #1029, opened 2021-05-03
Enable Time sync service


borgmanJeremy:
My signal bridge has been periodically breaking and I think I just traced it to the system time in the VM slowly "desyncronizing" from true time.

Are you open to a PR that adds/enables chrony or another time sync daemon?

spantaleev:
We already auto-install and run ntpd. Which distro are you on?

skepticalwaves:
From what I can tell the various ntp* packages are mostly unmaintained now, and systemd recommends its built-in daemon:
https://wiki.archlinux.org/title/Systemd-timesyncd

borgmanJeremy:
I was using Ubuntu 20.04 in a VM.... I'll see if I can replicate it

spantaleev:
`systemd-timesyncd` sounds good and simple enough. Looks like it's not supported on older distros though.

[This article](https://www.digitalocean.com/community/tutorials/how-to-set-up-time-synchronization-on-ubuntu-18-04) suggests that Ubuntu 18.04 has it, but I'm guessing 16.04 probably doesn't. CentOS 7's systemd is also quite old and doesn't have it.

It'd be great to add support for `systemd-timesyncd` for the distros that support it and leave ntp for those that don't, at least for a while longer. With time, we should be able to drop support for these old distros. Ubuntu 16.04 [reached EOL on 30th of April 2021](https://ubuntu.com/16-04), so we can probably start dropping support for it soon, as most people have probably migrated away anyway. CentOS 7 [receives some updates](https://en.wikipedia.org/wiki/CentOS#End-of-support_schedule) till 2024-06-30 though. My server is also still on CentOS 7 (especially with all the uncertainty around CentOS 8), so I'm definitely interested in keeping that going until we see what happens (Rockylinux, Almalinux, ..), and for some time beyond.

borgmanJeremy:
Ok cool, I'll look at adding systemd-timesyncd to the systems that support it. 

Thanks!

spantaleev:
This has been implemented. The playbook now includes a timesync role (roles/galaxy/timesync) which installs and enables systemd-timesyncd on distros that support it, falling back to ntpd on older ones. The service used can also be overridden via devture_timesync_ntpd_service (e.g. chronyd). Closing as done.

PR #1105, opened 2021-06-03
Add support for CentOS (Stream) 8

blaztinn:
With these changes I've successfully installed and have been using the server (with a few bridges) on the CentOS Stream 8. AFAIK, everything seems to be working as it should.

This PR is basically more of a request for comments. I don't know enough Ansible to replace `ntp`/`ntpd` with `chrony`/`chronyd` only for CentOS 8+.

Also, `docker-py` is installed through `pip` since I couldn't find it in any repos (checked CentOS and EPEL). Is that a good enough solution?

### DIFF
--- a/roles/matrix-base/defaults/main.yml
+++ b/roles/matrix-base/defaults/main.yml
@@ -80,8 +80,8 @@ matrix_host_command_openssl: "/usr/bin/env openssl"
 matrix_host_command_systemctl: "/usr/bin/env systemctl"
 matrix_host_command_sh: "/usr/bin/env sh"
 
-matrix_ntpd_package: "ntp"
-matrix_ntpd_service: "{{ 'ntpd' if ansible_os_family == 'RedHat' or ansible_distribution == 'Archlinux' else 'ntp' }}"
+matrix_ntpd_package: "{{ 'systemd-timesyncd' if ansible_distribution == 'CentOS' and ansible_distribution_major_version > '7' else 'ntp' }}"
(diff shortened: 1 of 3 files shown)


spantaleev:
Thanks for thinking about making the playbook run on newer distros!

Looks like this PR would break installation on older distros (which don't have chrony).

Since systemd is already a hard dependency for us, using something like systemd's timesyncd for such newer distros is better than using ntp/chrony. See our discussion in #1029.

It's also good to separate the CentOS 8-Stream setup tasks into another YAML file, and to make `setup_centos.yml` only run for older CentOS. You'd then make your changes in your new file (`setup_centos8_stream.yml`).

blaztinn:
Thanks, I'll look into it and update the PR :)

blaztinn:
I've updated the PR to use `systemd-timesyncd` and made a separate setup `setup_centos8.yml`.

Because checking for CentOS vs CentOS Stream needs quite recent Ansible (https://github.com/ansible/ansible/pull/73034) not yet packaged by my distro, I've made the setup available for all CentOS 8+. Looking at the packages mirror of CentOS 8 I believe that the setup should also work for that release but I haven't tested it. I have only tested it on CentOS 8 Stream.

For using the `systemd-timesyncd` I needed to add a task to install EPEL repos, I hope that's OK? Otherwise I'll change it back to `chrony`.

spantaleev:
I guess this (using pip, installing `epel-release`) is better than not having any CentOS 8 support at all, so I'm merging it as it is. Thank you! :+1: 

Should we find a way to improve it in the future, we can.

PR #1108, opened 2021-06-06
fix error when running with Ansible 2.7 in the hydrogen setup

tommes0815:
Without this the hydrogen task would fail for Ansible 2.7 even when hydrogen is not enabled.


spantaleev:
Thanks! :+1: 

This actually affects all our other roles as well

spantaleev:
Fixed for the other roles in fa76c1e

tommes0815:
a thanks for merging and fixing the rest. I did not see the other roles failing, probably because hydrogen is the only one which is by default self building

PR #1124, opened 2021-06-15
Update configuring-playbook-synapse.md

N07070:
Added an example for configuring the variables for the OpenID, that might fails because the variables are not defined for the playbook parsing, but Synapse's own config file parsing.

Based on advice given by @spantaleev on Matrix : https://matrix.to/#/!cNSQwPqhHKkIZdBnvt:devture.com/$Ob97GK84soL9TudCg7JujbWSjjEZT_P70OrtkbVgK88?via=devture.com&via=matrix.org&via=maunium.net


spantaleev:
Thank you! :+1:

PR #1131, opened 2021-06-20
according to discussion in channel not needed anymore

dreamflasher:


spantaleev:
So the bridge handles authentication later on, after you message it?

If so, we should also remove other references to these variables too. There are some in:

- `roles/matrix-bridge-mx-puppet-slack/defaults/main.yml`
- `roles/matrix-bridge-mx-puppet-slack/templates/config.yaml.j2`

dreamflasher:
Apparently, because the authentication is done per slack workspace/team with the bot itself. Shall I remove the other references as well?

spantaleev:
Yes, please. Let's remove them and get this merged

spantaleev:
Thanks! :+1:

PR #1133, opened 2021-06-20
Fix stuff and add Podman/Fedora support (#986 rebase)

penn5:


spantaleev:
This PR is huge and changes lots of unrelated things. A lot of these changes are dubious:
- `docker stop` instead of `docker kill`. I guess you did it because `docker kill` fails on Podman, but it may be more unreliable on Docker (untested)
- enabling Ansible retry files, which we'd rather not have
- unrelated changes to `mautrix-signal` in 33391860fe76c161
- various other unrelated changes in e8047bcb267c1bd

It'd be great to add support for Fedora and Podman (on Fedora or even on other systems), but these changes should be done in an isolated manner, without introducing a ton of other changes across the whole codebase.

penn5:
I'd be happy for you to pick individual commits from this PR as they get reviewed, if that makes things simpler. They would still need to be applied in the order given (though the retry files is not required, I just needed it for testing)

penn5:
e8047bc fixes ARM support
3339186 makes setting up mautrix-signal simpler but again obviously not needed for Fedora.
`docker kill` does work on podman (iirc, it's been a while), but leads to database corruption (well, it's recoverable, but still not ideal) because it sends SIGKILL instead of SIGINT. `docker stop` will send first SIGINT (or whatever signal is needed, SIGTERM or whatever) and wait some timeout before sending SIGKILL. So hopefully it has no risk of breaking things. It all worked for me.

ttuffin:
Is this PR still under consideration? @spantaleev would you consider opening up a new branch for the development of this feature?

spantaleev:
You can fork this project and make your own branch there.

This PR is very outdated now anyway and likely not a great way to approach things nowadays.

PR #1139, opened 2021-06-23
Jitsi Update -> stable-5963

sakkiii:


skepticalwaves:
Can this be made optional instead of being enabled by default when jitsi is enabled?

sakkiii:
Jibri recording issues fixed in stable-5870 but here it's still didn't worked. :disappointed:

sakkiii:
To make Jibri work need 80:80 and/or 443:443 port mapping, need lot more work. Any suggest or contribution are welcome.

@spantaleev I think we should test and review this PR as jitsi docker latest update stable-5963 and create PR for recording.

roles/matrix-jitsi/templates/prosody/env.j2
```diff
@@ -18,32 +18,35 @@ LDAP_TLS_CHECK_PEER={{ 1 if matrix_jitsi_ldap_tls_check_peer else 0 }}
...
-
 XMPP_MODULES={{ matrix_jitsi_xmpp_modules }}
 XMPP_MUC_MODULES=
 XMPP_INTERNAL_MUC_MODULES=
-
 XMPP_RECORDER_DOMAIN={{ matrix_jitsi_recorder_domain }}
-
-JICOFO_COMPONENT_SECRET={{ matrix_jitsi_jicofo_component_secret }}
+XMPP_CROSS_DOMAIN=true
+JICOFO_COMPONENT_SECRET
```
spantaleev, line 31: Why empty? What's happened to the jicofo component secret? Intentionally setting it to empty?

sakkiii: JICOFO_COMPONENT_SECRET removed in stable-5765 -> https://github.com/jitsi/docker-jitsi-meet/commit/953a4d2425df7128c35c4b1edebb88e8bfe28a97

spantaleev: Why not remove the reference completely?

sakkiii: yeah right done

spantaleev:
I'm not using Jitsi anymore, so I can't easily test this, but it looks good. Thank you for updating Jitsi for us! :+1:

PR #1143, opened 2021-06-25
Remove asterisks from configuring-dns.md

aaronraimist:
These previously denoted optional DNS records but now optional records are a separate table so they are unnecessary.


spantaleev:
Thank you! :+1:

PR #1145, opened 2021-06-28
Mention Let'sEncrypt issues with Element mobile clients

pushytoxin:
Let me know if you'd prefer to have a `matrix_coturn_enable_tls: true/false` variable instead

### DIFF
--- a/docs/configuring-playbook-turn.md
+++ b/docs/configuring-playbook-turn.md
@@ -40,3 +40,14 @@ matrix_jitsi_web_stun_servers:
 - stun:HOSTNAME_OR_IP:PORT
 ```
 You can put multiple host/port combinations if you like.
+
+## Disabling TLS support
+
+The Element Android and iOS clients are known not to support _Let's Encrypt_ certificates, leading to increased delays when making calls. ([Android issue](https://github.com/vector-im/element-android/issues/1533), [iOS issue](https://github.com/vector-im/element-ios/issues/2712))
+If you serve users with these clients you may want to disable `turns` URIs by adding the following snipplet to your configuration:
+
+```yaml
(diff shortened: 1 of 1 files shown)


spantaleev:
Maybe we should make this the default? If Element doesn't support that, what is there (as of right now) that justifies having it enabled?

pushytoxin:
Using TURNS there is a small benefit of the signaling channel not leaking the remote address during the WebRTC setup.

spantaleev:
Element doesn't support TURNS though, so we don't really get that benefit?

pushytoxin:
Element (mobile) does support TURNS, and even attempts to use it, the issue is specifically with the Let's Encrypt chain. The WebRTC lib uses a different CA root list than that of the operating system.
So if somebody uses external certs, they can enable TURNS. My guess is this is < 1% of those who deploy the playbook.

pushytoxin:
Turning it off by default would leave some people with questions, don't you think? "Why are we reducing the (wrongly perceived) security of the homeservers."

There might be those who currently using the playbook with manually installed certs too. They don't expect changes like this.

I propose we change the default to turn off turns in case the built-in Let'sEncrypt support is disabled.

spantaleev:
> I propose we change the default to turn off turns in case the built-in Let'sEncrypt support is disabled.

Sounds reasonable and likely better than updating some documentation page that no one will read. I've done that in 8b146f083ef3b.

pushytoxin:
I've been away the past week and couldn't work on this, thanks for completing the job! :+1:

pushytoxin:
For the record, these are the issue pages in for the mobile clients:

https://github.com/vector-im/element-android/issues/1533
https://github.com/vector-im/element-ios/issues/2712

PR #1147, opened 2021-06-28
Make federation domain customizable

oxmie:
**tl;dr:** Setup option to allow matrix federation fqn to be different to matrix client fqn

### Description
In some cases it might be useful to have a different domain for your matrix federation API than for your client communications.
e.g.:
- you are running the matrix-docker-ansible-deploy ecosystem behind a webproxy, which only allows `:80` and `:443` to be configured. By making the federation fqn different to client communications, you can handle the federation requests within _matrix-nginx-proxy_ separately with the nginx `server_name` key.

**Attention:** I did not checked every role and most roles does not uses federation API anyway. Therefore I am happy about advises, where these code changes to the federation domain could cause problems.

### DIFF
--- a/roles/matrix-base/defaults/main.yml
+++ b/roles/matrix-base/defaults/main.yml
@@ -11,6 +11,9 @@ matrix_domain: ~
 # This and the Element FQN (see below) are expected to be on the same server.
 matrix_server_fqn_matrix: "matrix.{{ matrix_domain }}"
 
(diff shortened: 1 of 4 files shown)


spantaleev:
Thank you! :+1: 

I'm not sure whether this covers everything, but I can't think of anything else to be done about it right now. I guess people who use a different hostname will find it sooner or later.

PR #1150, opened 2021-06-29
Update IRC bridge

thomwiggers:
https://github.com/matrix-org/matrix-appservice-irc/releases/tag/0.27.0

The default settings should probably be updated, but I think it's fine for now.


spantaleev:
Thank you! :+1:

PR #1153, opened 2021-06-30
Add worker metrics to prometheus exporter

Cadair:
I tried a few ways of doing this in various levels of insanity. This was the only way I could come up with that actually worked, but I appreciate it's using a synapse variable in the prometheus role.


spantaleev:
It'd be nice to make this not use `matrix_synapse_` variables.

We can probably pass this `matrix_synapse_workers_enabled_list` variable as-is via `group_vars/matrix_servers`.

Cadair:
So just put in `group_vars/matrix_servers` something like

```
matrix_prometheus_synapse_workers_enabled_list: "{{ matrix_synapse_workers_enabled_list }}"
```
?

That's easy enough.

spantaleev:
It's not pretty (there may be a cleaner way to do it), but it works.

Cadair:
done :)

spantaleev:
Thank you! :+1:

PR #1156, opened 2021-07-02
Upgrade mjolnir (0.1.17 -> 0.1.18) and implement self building 

aaronraimist:


roles/matrix-bridge-mx-puppet-twitter/tasks/init.yml
```diff
@@ -68,10 +68,3 @@
       URL endpoint to the matrix-mx-puppet-twitter container.
       You can expose the container's port using the `matrix_mx_puppet_twitter_container_http_host_bind_port` variable.
   when: "matrix_mx_puppet_twitter_enabled|bool and matrix_nginx_proxy_enabled is not defined"
-
```
spantaleev, line 71: Why does this PR remove these tasks from other roles?

aaronraimist: They are redundant. There are two duplicate checks in each of those files.

spantaleev: Oh, interesting! :laughing:

spantaleev:
Thank you! :+1:

PR #1160, opened 2021-07-02
Skip importing validate_config task when Synapse is disabled

dotneutron:
When running the playbook for a host with `matrix_synapse_enabled: false`, it would fail on task `Fail if required Synapse settings not defined`.

The easiest fix I see is checking for this condition when importing the task itself.


spantaleev:
True! That's what we do for other roles too. Thank you! :+1:

PR #1162, opened 2021-07-05
Fix docs template syntax error in OIDC SSO example

hauau:
Example uses `{%raw}` while proper syntax is `{% raw %}`. This example doesn't work.

https://docs.ansible.com/ansible/latest/user_guide/playbooks_advanced_syntax.html#unsafe-or-raw-strings


spantaleev:
Thanks! :+1:

PR #1171, opened 2021-07-10
Updating to ma1sd v2.5.0

BackInBash:
Version 2.5.0 did not have an architecture tag.
https://hub.docker.com/r/ma1uta/ma1sd/tags?page=1&ordering=last_updated


spantaleev:
`matrix_ma1sd_architecture` also gets set via `group_vars/matrix_servers`. It's not enough to unset it in the role, as it gets repopulated again.

Seems like `matrix_ma1sd_container_image_self_build` in `group_vars/matrix_servers` also needs to be adjusted.

We may consider removing the `matrix_ma1sd_architecture` variable as well (with the relevant change in `roles/matrix-ma1sd/tasks/validate_config.yml` that tells users who use it about that)

spantaleev:
I've implemented the aforementioned fixes and merged this! Thanks for getting things started! :+1:

PR #1174, opened 2021-07-13
Update synapse to v1.38.0

juusujanar:
There are no breaking changes, but there is a database index rebuild needed, that can consume approximately 4 times more space.

Release notes: https://github.com/matrix-org/synapse/releases/tag/v1.38.0


spantaleev:
Thank you! 👍

PR #1175, opened 2021-07-14
Added selfbuild functionality to mautrix-signal bridge

Toorero:


spantaleev:
Thank you!

It would be nice to update `group_vars/matrix_servers` to pass a better value to the self-build variable, depending on `matrix_architecture`

Toorero:
There you go. Should now work automatically as wished.

spantaleev:
Thanks! :+1: 

I just wonder whether `:latest` (what we use by default when not self-building) really works on both `amd64` and `arm64` or if we need some other tag for `arm64`.

Toorero:
Ok I think you might be right. Also, I encountered the issue that I forgot to add self-build functionality for the signald docker image, which is ONLY amd64.

Toorero:
Perhaps an even better approach in general would be to just try to pull the image with the specified `matrix_architecture` and this should simply fail if the architecture is not listed in the manifest. I find it weird in the first place that the ansible docker pull doesn't error out if the pattform the current plattform isn't in the manifest and simply just uses what it can get.

PR #1180, opened 2021-07-16
Fixed self-build functionality for mautrix-signal

Toorero:
I now autodetect compatible platforms of the mautrix-signal docker images via the manifest and added overlooked self-build functionality for the signal daemon.

Follow-up to  #1175

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -374,7 +374,8 @@ matrix_mautrix_signal_login_shared_secret: "{{ matrix_synapse_ext_password_provi
 matrix_mautrix_signal_database_engine: 'postgres'
 matrix_mautrix_signal_database_password: "{{ matrix_synapse_macaroon_secret_key | password_hash('sha512', 'mau.signal.db') | to_uuid }}"
 
-matrix_mautrix_signal_daemon_container_self_build: "{{ matrix_architecture != 'amd64' }}" # sadly not automatic detectable because no manifest
+matrix_mautrix_signal_container_self_build: "{{ matrix_architecture not in ['amd64', 'arm64'] }}"
+matrix_mautrix_signal_daemon_container_self_build: "{{ matrix_architecture != 'amd64' }}"
 
 ######################################################################
 #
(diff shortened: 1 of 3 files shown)


Toorero:
Also, I think it would be cool to append a notice to docs/configuring-playbook-bridge-mautrix-signal.md about a possible occurring self-build issue of signald which just spits out a cryptic Gradle error and is just a result of to low memory 🤷🏻

spantaleev:
Looks like you're currently not setting `matrix_mautrix_signal_container_self_build` via `group_vars/matrix_servers`? You're only seeing the signal daemon self-build variable.

Toorero:
Yes because the signal container itself has a manifest which describes for which architecture it is compatible. Therefore, the docker pull will fail automatically and I start the self build process.

Toorero:
OK I'm not sure how the review thing works and how I can cite code but just look up to the comments I left at the concerning lines.

spantaleev:
So you're trying to pull and then fall back to building if it fails?

Sounds like this is more fragile than just pulling for platforms where we know pulling works.

Right now, `docker_image`-pulling that fails for another reason (temporary network error, etc.) will probably also trigger self-building.

Toorero:
That's actually right, but I think it's a better approach because if somehow he drops support for the arm64 build or adds another platform to the manifest, it just works out of the box. Also, the pulling of the source code will fail if there is a network error.

Perhaps I can just ignore fails if the platform is not equal to amd64 or arm64 but that would diminish the whole point.

If you don't like the approach, you can just revert to the old approach I implemented in #1175 . I totally understand your reasoning, but don't like the hard coded platform approach.

spantaleev:
Yes, I think an approach where we hardcode the the known-working platforms to be more stable and predictable. It's also more in line with what we do in all other roles.

Toorero:
Ok, I reverted to the manual approach you proposed. I hope everything is now to your liking 😅

Toorero:
uff you're right I overlooked some stuff

Toorero:
thx 🙏🏻

spantaleev:
I've removed some references that were left from before.

Looks good! Thank you for fixing self-building! :+1:

PR #1182, opened 2021-07-16
Added missing X-Forwarded-Proto header

Toorero:
Added missing X-Forwarded-Proto header according to the example here: https://github.com/matrix-org/synapse/blob/3fffb71254d052c54d7a6eabae8534480f021adc/docs/reverse_proxy.md


spantaleev:
Thank you! :+1:

PR #1183, opened 2021-07-16
Fixed mautrix-telegram selfbuild not working on non amd64 platforms

Toorero:
I added this changes because the docker [build file](https://mau.dev/tulir/mautrix-telegram/-/blob/c02f67e0d1dcabde598d65d346c63097e309d410/Dockerfile) contains an architecture constraint to only work for amd64 because the FROM image is amd64 only. Therefore, I mask the constraint and compile the [underlying image](https://mau.dev/tulir/lottieconverter.git) by myself and then compile the constraint mautrix-telegram image.

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -393,8 +393,8 @@ matrix_mautrix_telegram_enabled: false
 
 # Images are multi-arch (amd64 and arm64, but not arm32).
 matrix_mautrix_telegram_container_self_build: "{{ matrix_architecture not in ['arm64', 'amd64'] }}"
-matrix_lottieconverter_container_self_build: "{{ matrix_architecture not in ['arm64', 'amd64'] }}"
-matrix_lottieconverter_container_self_build_mask_arch: "{{ matrix_architecture != 'amd64' }}"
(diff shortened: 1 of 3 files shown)


spantaleev:
Can you rename the variables (`mautrix_lottieconverter_*` -> `mautrix_telegram_lottieconverter_*`)?

Maybe also add some comments somewhere, so we (and others) can figure out why it's being done:

> Self-building mautrix-telegram requires that:
> - we remove (mask) a hardcoded `TARGETARCH=amd64` from its Dockerfile (https://mau.dev/tulir/mautrix-telegram/-/blob/c02f67e0d1dcabde598d65d346c63097e309d410/Dockerfile)
> - we build its base image lottieconverter (https://mau.dev/tulir/lottieconverter) first

By the way, it's probably possible to mask that `TARGETARCH` thing by passing [build/args](https://docs.ansible.com/ansible/latest/collections/community/docker/docker_image_module.html#parameter-build/args) to the `docker_image` Ansible module. That would save us some `sed` hacks on each and every run.

Toorero:
Ah I didn't know that I'm going to look into that into and sure I can rename it. No problem.

spantaleev:
Thank you! :+1:

PR #1184, opened 2021-07-16
Added proxy config for synapse-admin

Toorero:
Added some optional reverse proxy passthroughs for synapse-admin and the admin api.


spantaleev:
Thank you! :+1:

PR #1185, opened 2021-07-16
Docs - Change link to the correct Step

JokerGermany:


spantaleev:
Makes sense! Thank you! :+1:

PR #1189, opened 2021-07-19
root path for the base domain is wrong

JokerGermany:
`root /nginx-data/matrix-domain;` should be root `/matrix/nginx-proxy/data/matrix-domain;`


spantaleev:
When running in a container, `matrix_nginx_proxy_data_path` (`/matrix/nginx-proxy/data`) on the host is mounted inside the container at `matrix_nginx_proxy_data_path_in_container` (`/nginx-data`), as seen here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/70cb27138c78ad1ee9fb49323d7b160c5e5bbe7b/roles/matrix-nginx-proxy/templates/systemd/matrix-nginx-proxy.service.j2#L36

So the configuration (`root /nginx-data/matrix-domain;`) is correct in the general case (when running in a container).

I've fixed up your patch, so that it works both in container mode and not.

PR #1190, opened 2021-07-19
Bug - docker resolver is used when matrix_nginx_proxy_enabled is false

JokerGermany:


roles/matrix-nginx-proxy/templates/nginx/conf.d/nginx-http.conf.j2
```diff
@@ -5,7 +5,7 @@
 # Thus, we ensure a larger bucket size value is used.
 server_names_hash_bucket_size 64;
 
-{% if matrix_nginx_proxy_http_level_resolver %}
+{% if matrix_nginx_proxy_enabled %}
```
aaronraimist, line 8:
```suggestion
{% if matrix_nginx_proxy_http_level_resolver and matrix_nginx_proxy_enabled %}
```

spantaleev:
`matrix_nginx_proxy_http_level_resolver` is set to empty when matrix-nginx-proxy is disabled. See: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/51b94d9987c89922e859e05133b196b0a07f3cf5/roles/matrix-nginx-proxy/defaults/main.yml#L370

Due to this and to us using a `{% if matrix_nginx_proxy_http_level_resolver %}` check, I believe that there's no issue to be solved.

JokerGermany:
hmm, you are right, it's working now, but i don't know why i had this problem, why i had to delete the resolver line with `matrix_nginx_proxy_enabled: false`

PR #1192, opened 2021-07-19
migrate from ntp to systemd-timesyncd for ubuntu & Archlinux

sakkiii:
Facing time sync issue in ubuntu 20.04 https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/1029

- chrony availability and compatibility with mostly all OS.
- systemd-timesyncd is light weight and chrony more accurate, full featured.

### DIFF
--- a/roles/matrix-base/defaults/main.yml
+++ b/roles/matrix-base/defaults/main.yml
@@ -80,8 +80,8 @@ matrix_host_command_openssl: "/usr/bin/env openssl"
 matrix_host_command_systemctl: "/usr/bin/env systemctl"
 matrix_host_command_sh: "/usr/bin/env sh"
 
-matrix_ntpd_package: "{{ 'systemd-timesyncd' if ansible_distribution == 'CentOS' and ansible_distribution_major_version > '7' else 'ntp' }}"
-matrix_ntpd_service: "{{ 'systemd-timesyncd' if ansible_distribution == 'CentOS' and ansible_distribution_major_version > '7' else ('ntpd' if ansible_os_family == 'RedHat' or ansible_distribution == 'Archlinux' else 'ntp') }}"
+matrix_ntpd_package: "{{ 'chrony' if ansible_distribution == 'CentOS' or ansible_distribution == 'Ubuntu' else 'ntp' }}"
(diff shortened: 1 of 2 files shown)


skepticalwaves:
I don't see any need at all for Chrony, which is intended for a modern ntp sync + server, whereas the goal here is to ensure the system time doesn't drift.

Ref: https://unix.stackexchange.com/questions/504381/chrony-vs-systemd-timesyncd-what-are-the-differences-and-use-cases-as-ntp-cli

spantaleev:
Agreed! `systemd-timesyncd` sounds like some lightweight way to keep time in sync (without all the extra features that ntp/chrony support). And since we rely on systemd anyway, it's a good bet that `systemd-timesyncd` is available to install on all supported distros (going forward, at least.. some of the old distros we support could stick to what we have now - `ntp`)

spantaleev:
Looks good!

Maybe we should go with systemd-timesyncd on Archlinux as well?

Seems like `systemd-timesyncd` is provided by the base `systemd` package, so we can set `matrix_ntpd_package` to `systemd`.
Alternatively, we can set it to empty and adjust `roles/matrix-base/tasks/server_base/setup_archlinux.yml` to avoid installing it. However, that makes it impossible to switch to something else (e.g. `matrix_ntpd_package: ntp` and `matrix_ntpd_service: ntpd`), so setting `matrix_ntpd_package` is probably better.

skepticalwaves:
> Seems like systemd-timesyncd is provided by the base systemd package, so we can set matrix_ntpd_package to systemd.

I think the arch people might be very unhappy if you force a systemd install on them, unless its not possible to use this playbook any other way anyways?

spantaleev:
The playbook installs `systemd` services and that's how the various components are started.

Docker and systemd are our only hard dependencies, pretty much.

sakkiii:
@spantaleev Can you please Review/Test this PR?

spantaleev:
Seems to work well! Thanks! :+1: 

Removing `ntp` from existing installations which will now use `systemd-timesyncd` may also be a good idea, but doing it is not very straightforward.

PR #1203, opened 2021-07-25
Restore authentication for Jitsi Meet

hardye:
Commit 2ca92b76b6c accidentally removed the check for authentication and resulted in Jitsi Meet configured for internal authentication suddenly becoming an open server. This PR restores authentication and updates the documentation for registering a new user in Prosody.


spantaleev:
Thank you for fixing this regression! :+1:

hardye:
You are very welcome, Slavi. I think it's possible that this fix won't be enough for some people to restore Jitsi authentication if they had it enabled previously. Here is why:

Pull request #1139 changed the internal server names of two Jitsi components. See the specific diff [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/30339cd313717bf18471f39ee27db76ec3ba2ddf#diff-56250465ef1e1476a8b42dd4c5f2daa027ad71aa54bc5c916054f53d6dc343f3L35). If I understand the [command for registering a Jitsi user in Prosody](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-jitsi.md#required-if-configuring-jitsi-with-internal-authentication-register-new-users) correctly, the registered user is bound to a server name. So if `user1` was registered before the PR was merged (using the previous server names), what happens if the Jitsi servers get different names assigned when running the updated playbook? It's likely that Jitsi authentication won't work anymore and people will probably have to register their Jitsi users again.

Whether or not this will open all playbook-managed Jitsi servers to the world is something I don't know. Unfortunately I could not test this scenario myself because I deleted the `/matrix/jitsi`directory with the previous user data in the course of trying to fix the problem.

So, if this really becomes an issue, I suppose people will start complaining soon ...

PR #1209, opened 2021-07-28
Configurable session_lifetime

saboteurkid:


spantaleev:
1. You've used a `matrix_synapse_session_lifetime` variable, but haven't defined it.
2. I suppose your `|length > 0` condition is supposed to work around that "variable is not defined" case, but I doubt it would. You probably need `{% if matrix_synapse_session_lifetime is defined %}`. In any case, that's not a pretty way to do things
3. Defining `matrix_synapse_session_lifetime: 24h` in `roles/matrix-synapse/defaults/main.yml` and just using that is much better (no `if` condition needed at all).
4. Still, we'd rather not define every possible Synapse configuration setting as a playbook variable (hundreds of variables). We have `matrix_synapse_configuration_extension_yaml` which lets people extend the default configuration easily.

saboteurkid:
`matrix_synapse_configuration_extension_yaml` to the rescue. Don't understand why I didn't find it before.

Issue #12, opened 2018-10-13
Cannot create a new user on the Matrix server (Error: No such container: matrix-synapse) 


anadahz:
I followed the install [documentation](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/README.md) and successfully run the role with the tags `setup-all` and `start` by issuing the following commands:
```
ansible-playbook -v -i inventory/hosts setup.yml --tags=setup-all
ansible-playbook -v -i inventory/hosts setup.yml --tags=start
```
However when I run the role with the `register-user` tag I get the error : `Error: No such container: matrix-synapse`.

Following verbose output:
```
ansible-playbook -vvvv -i inventory/hosts setup.yml --extra-vars='username=user password=password admin=yes' --tags=register-user
```
                                                                             
```
TASK [matrix-server : Register user] **********************************************************************************************************************************
task path: /home/user/matrix-docker-ansible-deploy/roles/matrix-server/tasks/register_user.yml:23
fatal: [matrix.domain.tld]: FAILED! => {                                            
    "changed": true,                                                                
    "cmd": "/usr/local/bin/matrix-synapse-register-user user password 1",           
    "delta": "0:00:00.103034",                                                      
    "end": "2018-10-13 13:24:39.089012",                                            
    "invocation": {                                                                 
        "module_args": {                                                            
            "_raw_params": "/usr/local/bin/matrix-synapse-register-user user password 1", 
            "_uses_shell": true,                                                    
            "argv": null,                                                           
            "chdir": null,                                                          
            "creates": null,                                                        
            "executable": null,                                                     
            "removes": null,                                                        
            "stdin": null,                                                          
            "warn": true                                                            
        }                                                                           
    },                                                                              
    "msg": "non-zero return code",                                                  
    "rc": 1,                                                                        
    "start": "2018-10-13 13:24:38.985978",                                          
    "stderr": "Error: No such container: matrix-synapse",                           
    "stderr_lines": [                                                               
        "Error: No such container: matrix-synapse"                                  
    ],                                                                              
    "stdout": "",                                                                   
    "stdout_lines": []                                                              
}                                                                                   
                                                                                    
PLAY RECAP **********************************************************************************************************************************
matrix.domain.tld        : ok=2    changed=0    unreachable=0    failed=1           
```

Any ideas to help debug and solve this issue are greatly appreciated.

spantaleev:
Seems like Synapse is not running.

[...]

Issue #10, opened 2018-09-15
Add .well-known


aaronraimist:
https://matrix.org/docs/spec/client_server/r0.4.0.html#get-well-known-matrix-client

Example: https://matrix.org/.well-known/matrix/client

spantaleev:
This playbook sets up the Matrix server (at `matrix.<domain>` and `riot.<domain>`) and not on the actual `<domain>` server.

As per the specification, the `/.well-known/matrix/client` is supposed to be on the actual `<domain>` server, which is not managed by this playbook.

Because of this, this playbook cannot really set it up for you. Still, we should add documentation somewhere that tells people about it, so they can set it up (similar to how we advise people to set up SRV DNS records for `<domain>` in order to point the Matrix service at `matrix.<domain>`).

-------

To make things simpler to set up though, I think the playbook can generate and serve the appropriate file at `https://matrix.<domain>/.well-known/matrix/client` (even though that's the wrong domain name to set it up at!).

And then, we can tell people to set up `https://<domain>/.well-known/matrix/client` by:

1) simply copying the static file configured by the playbook at `https://matrix.<domain>/.well-known/matrix/client`. This is easy to do and people's only choice if `<domain>` is just some static website server that they can't control

2) or: setting it up more reliably by configuring proxying for that file

We can provide proxying examples. For nginx, it would be something like:

```nginx
server {
	server_name <domain>;

	location /.well-known/matrix {
		proxy_pass https://matrix.<domain>;
	}

	# other configuration
}
```

Method 1) may be easier and a better choice in certain situations, while method 2) is less prone to breakage in case the playbook ever generates the `/.well-known/matrix/client` file differently for some reason (new configuration keys, etc.)

spantaleev:
As of 0d0ccde286ac39, this is now implemented!

The playbook generates the file, as described above.
You're then supposed to either copy it over to the base domain's server or to set up reverse-proxying there.
The documentation describes it in detail, along with example configuration for both nginx and Apache.

PR #1224, opened 2021-08-10
GoMatrixHosting v0.5.7

GoMatrixHosting:
GoMatrixHosting v0.5.7

- Prevent Provision stage re-writing variables it should only write on the first run, see [#12](https://gitlab.com/GoMatrixHosting/ansible-provision-server/-/issues/12).
- Fix SFTP website upload, see [#10](https://gitlab.com/GoMatrixHosting/gomatrixhosting-matrix-docker-ansible-deploy/-/issues/10).

If there's anything problematic about this release please let me know, we can revise it.  :+1:


spantaleev:
Referring to `matrix_awx_enabled` inside the `matrix-nginx-proxy` role is not very clean.

Maybe a new `_enabled` variable should be created for these tasks, which would then default to on (`true`), but be toggled to off via `group_vars/matrix_servers`.

Or maybe some other way..?

GoMatrixHosting:
There we go, i've removed all reference to matrix_awx_enabled from the matrix-nginx-proxy role. What do you think?

PR #1228, opened 2021-08-12
Update matrix-mautrix-signal config to 0.2.0 to enable relay mode

Jaffex:
I updated the config-template to the latest version of the project's [example-config.yaml](https://github.com/mautrix/signal/blob/master/mautrix_signal/example-config.yaml)

I changed the example-config's default in two places:
```
'{{ matrix_mautrix_signal_homeserver_domain }}': relay
```
to only relay messages of users of your own homeserver (is this a desired default?) and
```
 relay:
        enabled: true
```
to be able to enable relay mode for rooms via `!signal set-relay`


spantaleev:
Thank you! 👍

PR #1229, opened 2021-08-13
Feature/signal config

wol-win:
The recent changes for the relay bot functionality has been enabled within the template file for the signal bridge. These changes now exposes the relay bot enable/disable as a configuration item.

Additionally, the permissions are set for the base domain users depending on whether the relay bot is switched on or off. 

Documentation augmented in case admins need a more detailed permissions setup.

### DIFF
--- a/docs/configuring-playbook-bridge-mautrix-signal.md
+++ b/docs/configuring-playbook-bridge-mautrix-signal.md
@@ -27,8 +27,8 @@ matrix_mautrix_signal_configuration_extension_yaml: |
   bridge:
     permissions:
       '@YOUR_USERNAME:YOUR_DOMAIN': admin
-      '*': user
-      YOUR_DOMAIN: relay
+      YOUR_DOMAIN: user
+      '*': relay
 ```
 
 You may wish to look at `roles/matrix-bridge-mautrix-signal/templates/config.yaml.j2` to find more information on the permissions settings and other options you would like to configure.
--- a/docs/configuring-playbook-bridge-mautrix-signal.md
+++ b/docs/configuring-playbook-bridge-mautrix-signal.md
@@ -23,14 +23,36 @@ Use `!signal unset-relay` to deactivate.
 By default, any user on your homeserver will be able to use the bridge.
 If you enable the relay bot functionality, it will relay every user's messages in a portal room - no matter which homeserver they're from.
 
-If you would like to have a more specific setting of the permissions you can set the permissions as follows (example). For more details see also [mautrix-bridge documentation](https://docs.mau.fi/bridges/python/signal/relay-mode.html)
+Different levels of permission can be granted to users:
+
+* relay - Allowed to be relayed through the bridge, no access to commands;
+* user - Use the bridge with puppeting;
+* admin - Use and administer the bridge.
+
+The permissions are following the sequence: nothing < relay < user < admin.
+
+The default permissions are set as follows:
+```yaml
+permissions:
+  '*': relay
+  YOUR_DOMAIN: user
+```
+
+If you want to augment the preset permissions, you might want to set the additional permissions with the following settings in your `vars.yml` file:
 ```yaml
 matrix_mautrix_signal_configuration_extension_yaml: |
   bridge:
     permissions:
       '@YOUR_USERNAME:YOUR_DOMAIN': admin
-      YOUR_DOMAIN: user
-      '*': relay
+```
+
+This will add the admin permission to the specific user, while keepting the default permissions.
+
+In case you want to replace the default permissions settings **completely**, populate the following item within your `vars.yml` file:
+```yaml
+matrix_mautrix_signal_bridge_permissions: |
+  '@ADMIN:YOUR_DOMAIN': admin
+  '@USER:YOUR_DOMAIN' : user
 ```
 
 You may wish to look at `roles/matrix-bridge-mautrix-signal/templates/config.yaml.j2` to find more information on the permissions settings and other options you would like to configure.
--- a/docs/configuring-playbook-bridge-mautrix-signal.md
+++ b/docs/configuring-playbook-bridge-mautrix-signal.md
@@ -20,7 +20,8 @@ matrix_mautrix_signal_relaybot_enabled: true
 ```
 If you want to activate the relay bot in a room, use `!signal set-relay`.
 Use `!signal unset-relay` to deactivate.
-Additionally the permissions for the bridge grant user rights to all base domain users in case the relay bot is disabled, or relay rights in case the relay bot is enabled.
+By default, any user on your homeserver will be able to use the bridge.
+If you enable the relay bot functionality, it will relay every user's messages in a portal room - no matter which homeserver they're from.
 
 If you would like to have a more specific setting of the permissions you can set the permissions as follows (example). For more details see also [mautrix-bridge documentation](https://docs.mau.fi/bridges/python/signal/relay-mode.html)
 ```yaml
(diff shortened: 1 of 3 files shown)


docs/configuring-playbook-bridge-mautrix-signal.md
```diff
@@ -12,6 +12,27 @@ Use the following playbook configuration:
...
+Additionally the permissions for the bridge grant user rights to all base domain users in case the relay bot is disabled, or relay rights in case the relay bot is enabled.
+
+If you would like to have a more specific setting of the permissions you can set the permissions as follows (example). For more details see also [mautrix-bridge documentation](https://docs.mau.fi/bridges/python/signal/relay-mode.html)
+```yaml
+matrix_mautrix_signal_configuration_extension_yaml: |
+  bridge:
+    permissions:
+      '@YOUR_USERNAME:YOUR_DOMAIN': admin
+      '*': user
+      YOUR_DOMAIN: relay
```
tulir, line 31:
```suggestion
      YOUR_DOMAIN: user
      '*': relay
```

wol-win:
@tulir - thanks for the quick review.

I have just done another test and saw a strange behaviour. The base setting from the template file was still setting YOUR_DOMAIN: user although I set in my vars.yml the permissions to only one user as admin.

It seems like the settings within `matrix_mautrix_signal_configuration_extension_yaml` is not replacing but augmenting the setup in the template.

I remove the the permission settings in the template file completely and set it only through extension and then it works.

roles/matrix-bridge-mautrix-signal/defaults/main.yml
```diff
@@ -93,6 +96,21 @@ matrix_mautrix_signal_configuration_extension_yaml: |
...
+  #      admin - Use and administrate the bridge.
+  # Permitted keys:
+  #        * - All Matrix users
+  #   domain - All users on that homeserver
+  #     mxid - Specific user
+  #
+  bridge:
+    permissions:
+      {{ matrix_mautrix_signal_homeserver_domain }}: "{{ "relay" if matrix_mautrix_signal_relaybot_enabled else "user" }}"
+
```
Jaffex, line 113: I think relay and user are not exclusive.

I'd suggest moving this part back to config.yaml.j2 and change it (see below)

```suggestion
```

roles/matrix-bridge-mautrix-signal/templates/config.yaml.j2
```diff
@@ -188,14 +188,15 @@ bridge:
     #        * - All Matrix users
     #   domain - All users on that homeserver
     #     mxid - Specific user
-    permissions:
-      '{{ matrix_mautrix_signal_homeserver_domain }}': relay
-      '{{ matrix_mautrix_signal_homeserver_domain }}': user
+    #permissions: 
+    #
+    # Remark: permissions will be set in the defaults/main.yml file of this role 
+    # (see matrix_mautrix_signal_configuration_extension_yaml)
```
Jaffex, line 194:
```suggestion
    permissions: 
        *: relay
        '{{ matrix_mautrix_signal_homeserver_domain }}': user
```
This way we'd enable the bridge to be used by users on the homeserver and the relay to relay any room member's messages (no matter where they're from - which I think would be an understandable default)

docs/configuring-playbook-bridge-mautrix-signal.md
```diff
@@ -12,6 +12,27 @@ Use the following playbook configuration:
...
 ```
 
+There are some additional things you may wish to configure about the bridge before you continue.
+
+The relay bot functionality is off by default. If you would like to enable the relay bot, add the following to your `vars.yml` file:
+```yaml
+matrix_mautrix_signal_relaybot_enabled: true
+```
+
+Additionally the permissions for the bridge grant user rights to all base domain users in case the relay bot is disabled, or relay rights in case the relay bot is enabled.
```
Jaffex, line 22:
```suggestion
By default, any user on your homeserver will be able to use the bridge.
If you enable the relay bot functionality, it will relay every user's messages in a portal room - no matter which homeserver they're from.
```
(This would be in line with the suggestions below)

docs/configuring-playbook-bridge-mautrix-signal.md
```diff
@@ -12,6 +12,27 @@ Use the following playbook configuration:
 matrix_mautrix_signal_enabled: true
 ```
 
+There are some additional things you may wish to configure about the bridge before you continue.
+
+The relay bot functionality is off by default. If you would like to enable the relay bot, add the following to your `vars.yml` file:
+```yaml
+matrix_mautrix_signal_relaybot_enabled: true
+```
+
```
Jaffex, line 21:
```suggestion
If you want to activate the relay bot in a room, use `!signal set-relay`.
Use `!signal unset-relay` to deactivate.
```

wol-win:
@Jaffex - I think we need to take this back and investigate - with moving the pieces back into the template file, these lines are always populated. Any change through the configuration_extension_yaml will augment those parts that are not available, but will not completely replace it. 

I tested it with my configuration; can someone confirm this?

To reproduce:
```
matrix_mautrix_signal_configuration_extension_yaml: |
  bridge:
    permissions:
      '@user:MY_DOMAIN': admin
      MY_DOMAIN: user
```
Results in 
```
    permissions:
        '*': relay
        '@user:MY_DOMAIN': admin
        MY_DOMAIN: user
```
but my intention was to get rid of the relay permissions.

Jaffex:
The result you see is what I'd expect, but I think always having the relay settings in doesn't do any harm, as the only relevant setting regarding the relay is enabling or disabling it as a whole.
I think having the relay permissions in without having the relay enabled would not affect anything.

In fact, I believe the relay permission will be `'*': relay` by default, even if it's not specifically written in the config.yaml - I deployed the bridge already with the current state of the repo (the one where you noticed that the relay-permissions will not get written at all) and it behaves just like it would if the setting was `'*':relay`

That means getting rid of that part of the config would be purely cosmetical.

@tulir maybe you can confirm this?

wol-win:
@Jaffex - There is one aspect still to be considered. Equally for the relay, all domain users will always be granted user permissions. There might be case where u might want to limit the user permission to only a subset of the domain users. I wonder how this would be achieved.

wol-win (commented):
Probably the case that not all domain user should have access to the bridge as user is an edge case and will rarely occur. In line with the presets for the other mautrix bridges (e.g. facebook) the default setting of the permissions is OK.

The PR could be implemented.

tulir:
> I believe the relay permission will be `'*': relay` by default, even if it's not specifically written in the config.yaml

The default is no permissions at all

Jaffex:
> The default is no permissions at all

That surprises me. I have the bridge running and the config has no entry for `: relay` at all.
While users from foreign homeservers will not get relayed, users from my own homeserver do get relayed.
If it was no permissions at all, I'd expect nobodies messages to be relayed.

tulir:
Permissions are linear, nothing < `relay` < `user` < `admin`

wol-win:
@Jaffex - but then we probably should also go back to my initial version, where nothing was set in the template file. Because anything that is written will be in the config file later, regardless what is speficied in the vars.yaml file.

So, the user line in the template will not be overwritten in any way but only augmented. What is your view?

wol-win (commented):
@Jaffex @tulir - would you mind to review the latest changes? Could someone then approve the pull request, too, so that it gets merged? Thx.

Jaffex (approved):
Looks great!
Also deployed it with the new variables, worked like a charm!

tulir (approved):

spantaleev:
Thanks a lot for your patience in polishing this! :+1:

Issue #1239, opened 2021-08-19
Playbook installation fails, while ensuring Mautrix Telegram image is pulled


N07070:
On an already installed server with this playbook, I've update it by git-pulling, and launching the playbook. It fails as such ; 

```
TASK [matrix-bridge-mautrix-telegram : Ensure Mautrix Telegram image is pulled] ************************
fatal: [matrix.domain]: FAILED! => changed=false 
  msg: 'Error pulling image dock.mau.dev/mautrix/telegram:v0.9.0 - 404 Client Error: Not Found ("manifest for dock.mau.dev/mautrix/telegram:v0.9.0 not found: manifest unknown: manifest unknown")'

PLAY RECAP *********************************************************************************************
domain         : ok=82   changed=9    unreachable=0    failed=1    skipped=429  rescued=0    ignored=0   
```

Is there anything I should do ?

I've looked for similar issues in this repo.

N07070:
Seems like it's trying to pull a docker image from a dockler repo maintained by mau.dev. This docker repo seems offline, which might be the issue.

spantaleev:
This seems to be a regression caused by #1237.

There's no image tagged with `v0.9.0` in the `dock.mau.dev/mautrix/telegram` container registry. Seems like there are no version-tagged releases there are all. We should either switch to `:latest` for now (likely not a good idea), or revert to the old registry (`dock.mau.dev/tulir/mautrix-telegram`) until a version-tagged release is added.

tulir:
Oh right, I intended to make new releases for most of the bridges, but forgot to release mautrix-telegram v0.10.1

N07070:
Created a pull request to fix this issue. If you want to merge it @spantaleev , I've run tests on my own server.

spantaleev:
Thanks for fixing it!

PR #1240, opened 2021-08-19
Update the docker image version for mautrix-telegram

N07070:
See #1239  . Just tested, and the installation works without issues.


spantaleev:
Thanks! :+1:

PR #1243, opened 2021-08-22
Postgres Minor Updates

sakkiii:
https://www.postgresql.org/about/news/postgresql-134-128-1113-1018-9623-and-14-beta-3-released-2277/


spantaleev:
Thank you! :+1:

PR #1247, opened 2021-08-24
Update Synapse from 1.40.0 to 1.41.0

FSG-Cat:


FSG-Cat:
Just wanted to add a quick comment that the upgrade notes from 1.41.0 has a note that says 

> Note that when using workers the /_synapse/admin/v1/users/{userId}/media must now be handled by media workers. See the upgrade notes for more information.

This is probably something that should be adressed before this PR is merged.

wol-win:
As far as I can see, there is one line to be added in:
'roles/matrix-synapse/vars/workers.yml' after line 255.
It should be amended with '^/_synapse/admin/v1/users/.*/media$'

spantaleev:
We have a bash script in `roles/matrix-synapse/files` which can update the workers configuration.

wol-win:
@spantaleev - this script is cool. I just tried it locally, and it will modify the workers.yml at two locations: one indeed in regard to the aforementioned line, the second change is in client API requests regarding the createRooms endpoint.

Is this shell script automatically run within the ansible playbook, or is this a manual action each time?

spantaleev:
Thank you! :+1:

PR #1248, opened 2021-08-24
Fix appservice-irc version

webpigeon:
Fix incorrect docker version tag for matrix-appservice-irc

This is the fix I mentioned in the matrix channel.

Looks like there is a bug in the irc bridge docker image tag, just tried to update and it complained the tag didn't exist: Error pulling image docker.io/matrixdotorg/matrix-appservice-irc:release-v0.30.0

On docker hub, the image tag looks to me release-0.30.0 (no v)
https://hub.docker.com/r/matrixdotorg/matrix-appservice-irc/tags?page=1&ordering=last_updated


spantaleev:
Thank you! 👍

PR #1250, opened 2021-08-24
Endpoint changes for Client and media API due to migration to 1.41.0

wol-win:
Changes after executing the `workers-doc-to-yaml.sh` bash script to retrieve new endpoints


spantaleev:
Thank you! :+1:

PR #1258, opened 2021-08-30
Jitsi Update (stable-5963 -> stable-6173)

sakkiii:
- add support for A/V Moderation [matrix_jitsi_enable_av_moderation default True]
- add ability to configure external TURN server


spantaleev:
Nice! Thank you! :+1:

PR #1265, opened 2021-09-03
Added variable matrix_nginx_proxy_request_timeout

### LINKED ISSUE #1262: Hope add timeout param for nginx proxy
Now I use element login to the matrix by a user has lots of rooms, but it always timeout. Can add a timeout param, incase a large amount of rooms user.

sakkiii:
Fix #1262


spantaleev:
We should probable define a few more variables (one for each thing we configure), instead of using `matrix_nginx_proxy_request_timeout` everywhere.

spantaleev:
Thanks! Need to also make use of the new variables inside `roles/matrix-nginx-proxy/templates/nginx/nginx.conf.j2` though

PR #1272, opened 2021-09-12
Revert "Move some related tasks closer together in matrix-client-hydrogen"

aaronraimist:
With https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/e913347fe15d4d1c6db0d5de8cc1026f335a02c2 the config file will never be read so Hydrogen will use matrix.org as the default server. I've added a comment explaining why it was put in this order.


spantaleev:
Oh, it makes sense now! Thanks for fixing what I broke! 👍

Issue #1277, opened 2021-09-16
Ubuntu 16.04 does not provide docker-py 1.10.0


andresterba:
Sadly I am still running my matrix server on a Ubuntu 16.04 machine, that was working until now.
The documentation states that this setup should work on Ubuntu 16.04 but it seems like the version of 'python-docker' is too old.

```sh
TASK [matrix-base : Ensure Matrix network is created in Docker] ****************
fatal: [matrix.my.server.com]: FAILED! => changed=false
  msg: 'Error: Docker SDK for Python version is 1.9.0 (hosts''s Python /usr/bin/python3). Minimum version required is 1.10.0.Use `pip install --upgrade docker-py` to upgrade. Hint: if you do not need Python 2.6 support, try `pip uninstall docker-py` instead, followed by `pip install docker`.'
```

Thanks for all the good work on `matrix-docker-ansible-deploy` I'm a happy user since a few years!

JokerGermany:
I hope you use Ubuntu ESM, otherwise this is your smallest problem...

andresterba:
> I hope you use Ubuntu ESM, otherwise this is your smallest problem...

The fix for the problem described above was to migrate to a new VM with a newer Ubuntu version :sweat_smile: (this was also a chance to test the backups^^)

PR #1282, opened 2021-09-18
Bump minimum supported Ubuntu version

aaronraimist:
Ubuntu ended support for 16.04 in April and it looks like Docker no longer supports 16.04 either according to https://docs.docker.com/engine/install/ubuntu/#os-requirements

Fixes https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/1277


spantaleev:
Sounds fair, thanks! 👍 

I wonder if we should bump the Debian version to 10 too. I don't see Debian listed here, but it might still work nevertheless: https://docs.docker.com/engine/install/debian/

aaronraimist:
Yeah looks like it was intentionally removed since Stretch got to it's EOL (not counting LTS) https://github.com/docker/docker.github.io/commit/c876e5cba7998db70c9d00b85f1a2c47df1b16b2

PR #1287, opened 2021-09-21
Update Synapse version (1.42.0 -> 1.43.0)

wol-win:
Update Synapse version (1.42.0 -> 1.43.0)

PR #1288, opened 2021-09-21
Update Synapse version from 1.42.0 to 1.43.0

FSG-Cat:
This upgrade should not need any special attention as the only addition that has an upgrade note is that Workers can now handle some spaces related tasks. This PR does not worry about making these changes. 

Making V9 default for Restricted rooms is a lovely upgrade from 1.43.0


FSG-Cat:
Closed due to this being a duplicate of #1287

PR #1295, opened 2021-09-24
Support trusting the upstream server when it says the protocol is HTTPS

### LINKED ISSUE #1132: OIDC Login not working
I am trying to get OIDC login working, using a self-hosted keycloak instance as identity provider following these guides:

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/1e46785a0d213d17524dfd4657d08a2785147f60/docs/configuring-playbook-synapse.md#synapse--openid-connect-for-single-sign-on

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/1fb54a37cba126235b1688bb5e96226a65b4fe80/docs/configuring-playbook-nginx.md#synapse--openid-connect-for-single-sign-on

https://matrix-org.github.io/synapse/develop/openid.html

However, trying to login or register with keycloak redirects me to` /_matrix/client/r0/login/sso/redirect/oidc-keycloak?redirectUrl=https%3A%2F%2Fapp.element.io%2F`, which does not correctly redirect to keycloak, but redirects unto itself so that it ends up failing with too many redirects.

Is there some configuration that I missed?

nogweii:
This replaces `$scheme` with the contents of the `X-Forwarded-Proto` when enabled.

Not 100% certain on the variables. Creating `matrix_nginx_proxy_trust_forwarded_proto` means that you can have a completely HTTP matrix server, which I don't know if that's supported or not. Or I can move `matrix_nginx_proxy_x_forwarded_proto_value` to `roles/matrix-nginx-proxy/tasks/init.yml` and have it key off `matrix_nginx_proxy_https_enabled` instead?

I think this resolves #1132 but I'm not 100% certain people's problems there are the same as mine.

### DIFF
--- a/docs/configuring-playbook-own-webserver.md
+++ b/docs/configuring-playbook-own-webserver.md
@@ -108,6 +108,9 @@ matrix_nginx_proxy_container_federation_host_bind_port: '127.0.0.1:8449'
 # Since we don't obtain any certificates (`matrix_ssl_retrieval_method: none` above), it won't work by default.
 # An alternative is to tweak some of: `matrix_coturn_tls_enabled`, `matrix_coturn_tls_cert_path` and `matrix_coturn_tls_key_path`.
 matrix_coturn_enabled: false
+
+# Trust the reverse proxy to send the correct `X-Forwarded-Proto` header as it is handling the SSL connection.
+matrix_nginx_proxy_trust_forwarded_proto: true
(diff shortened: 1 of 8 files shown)


spantaleev:
It seems like it solves your problem and likely the problem in #1132, so it looks good as it is. Thank you! :+1:

PR #1297, opened 2021-09-25
Fix hydrogen OCSP typo

aaronraimist:


spantaleev:
Good catch! Thanks! 👍

PR #1309, opened 2021-10-03
doc: how to export node and postgres sans grafana

HarHarLinks:
i set it up like this. thoughts?

could also go in an extra document, i'm not sure how commonly this is needed vs how advanced it is.


spantaleev:
Can we put this into another section (`## Exporting metrics without enabling Grafana and Prometheus (advanced)`) to separate it from the other stuff?

HarHarLinks:
I added it to a subsection since you're not going to do the advanced thing without doing the basic thing. In fact some of the basic settings are needed for the advanced to work, like the htpasswd.

spantaleev:
Thank you! :+1:

PR #1312, opened 2021-10-04
add how to generate htpasswd

### LINKED ISSUE #1308: clarify matrix_nginx_proxy_proxy_synapse_metrics_basic_auth_key
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/b5d84447643ab10c8b66428a1fc9dee75ffaff9c/roles/matrix-nginx-proxy/defaults/main.yml#L186

This ends up in the htpasswd file and thus needs to be formatted fittingly (`man 1 htpasswd`), e.g. modified MD5 by default.
Ansible can handle this: https://docs.ansible.com/ansible/latest/collections/community/general/htpasswd_module.html
If that is for some reason not desireable, e.g. perhaps the hashed version is preferable to keep in the your vars.yml instead of the plain password, then a comment about how to encrypt the pw should be added here.

HarHarLinks:
for `matrix_nginx_proxy_proxy_synapse_metrics_basic_auth_key`
resolves #1308


spantaleev:
Thank you! Much more obvious how to use it now! :+1:

Issue #1289, opened 2021-09-22
Google Chat


ljlongwing:
Is there any plans to retire the use of Google Hangouts and replace it with Google Chat?
https://github.com/mautrix/googlechat
I have hangouts configured and working well, but according to Google, Hangouts is 'going away' and being replaced by Chat....they currently work in tandem with each other, but eventually I expect Hangouts will stop working.  Mautrix has a Google Chat bridge that looks to be a replacement for hangouts, so I was hoping you had plans to do this replacement :)

aaronraimist:
As far as I know it is the same bridge that is already in the playbook. It has just been renamed and updated. https://github.com/mautrix/googlechat/commit/483d3a42ff81f2f895914447d599e681e763d5dc

ljlongwing:
hmmmm....so, it's just an alias for the same project?  Ok...if that's the case then no need to 'replace' it because it's the same thing I guess....but if he's renamed his, should it be renamed in this project?

aaronraimist:
I think the documentation should definitely be updated but I'm not sure about all of the variables. If they were just straight up changed then it would break the config of everyone currently using the bridge. I guess there would need to be a bunch of variables that translate the name from the old one to the new one.

pszypowicz:
They indeed just had a 'search and replace' sort of commit, but currently they keep both old prebuild docker images as mautrix-hangouts, as well as new 'mautrix-googlechat'

Do I understand code properly that in master we still point to the old repo:
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/b5d84447643ab10c8b66428a1fc9dee75ffaff9c/roles/matrix-bridge-mautrix-hangouts/defaults/main.yml#L11

This would mean that we will not benefit automatically from future updates?

aaronraimist:
That appears to be correct. You can submit a PR to change it. I don't use that bridge so I can't easily test it and don't want to change something that will break things.

pszypowicz:
I think it would make sense to introduce a new role - mautrix-googlechat - and it would be basically a similar approach as they did, by simply modifying the keywords and links to new image. We could then inform that mautrix-hangouts is deprecated and anyone could just move to googlechat.

What should the upgrade procedure be:
1. Removing manually old bridge?
2. Detect if both old and new bridge is used at the same time and fail?

spantaleev:
There's probably no need to maintain both roles operational simultaneously and accumulate cruft and cause confusion.

[...]

PR #1323, opened 2021-10-10
Replace hangouts with googlechat

mochman:
Mostly a 1 for 1 swap from hangouts -> googlechat.

Googlechat doesn't use sqlalchemy/alembic anymore so that was removed from the systemd template.

I tested this on a fresh database, I didn't try copying over the hangouts database.

This PR is for Issue #1289

### DIFF
--- a/docs/configuring-playbook-dimension.md
+++ b/docs/configuring-playbook-dimension.md
@@ -3,14 +3,12 @@
 **[Dimension](https://dimension.t2bot.io) can only be installed after Matrix services are installed and running.**
 If you're just installing Matrix services for the first time, please continue with the [Configuration](configuring-playbook.md) / [Installation](installing.md) flow and come back here later.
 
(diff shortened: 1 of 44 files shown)


mochman:
Looking through my hangouts_db and my googlechat_db, I'm not even sure they are compatible.  Some tables exist in one but not the other, and a bunch of columns in the matching tables have been renamed or added.  
I tried looking through the [bridge's db code](https://github.com/mautrix/googlechat/blob/master/mautrix_googlechat/db/upgrade/v01_initial_revision.py), but I don't see any place where it alters the old hangouts DB to use the new schema.
I don't think migration is going to work between hangouts and chat.

Do you want me to change the pull request to just add googlechat and leave hangouts where it is?  I could modify the README for hangouts to talk about how it is no longer being updated. That googlechat is available, but that all the previous conversations won't migrate over.

spantaleev:
I also took a quick look and didn't see any other DB changes, so I'm not sure why you're seeing very different tables.

Perhaps @tulir could point us in the right direction as to what should our migration path be.

PR #1327, opened 2021-10-12
Update Synapse default room version (6 -> 9)

altsalt:
From the [Synapse 1.43.0 release highlights](https://matrix.org/blog/2021/09/21/synapse-1-43-0-released):
> Asks clients to prefer [room version 9](https://github.com/matrix-org/matrix-doc/pull/3375) when creating restricted rooms ([#10772](https://github.com/matrix-org/synapse/issues/10772)), via the API defined in [MSC3244: room version capabilities](https://github.com/matrix-org/matrix-doc/pull/3244).


altsalt:
I haven't forgotten about this and have been keeping it open to see if movement can be made upstream in regards to the latter thought.

That said, there has been more discussion about defaulting to room v9: https://github.com/matrix-org/matrix-doc/pull/3589

aaronraimist:
Yeah it should finally happen in the next week or two.

aaronraimist:
This can be merged now. The spec now says v9 is the default and the next version of Synapse will use v9 as the default.

spantaleev:
Thanks for pinging us!

It may have been better to wait an extra week, until Synapse makes the switch, but.. I've merged it now.

I've been using room version 9 as the default for months. I'm sure others have done the same, so it doesn't seem risky switching.

PR #1328, opened 2021-10-12
Added Mautrix Google Chat

mochman:
This adds support for the [Mautrix Google Chat](https://github.com/mautrix/googlechat) bridge.  It is based on the Mautrix Hangouts bridge.


spantaleev:
Thank ouThanks a lot for bringing us up to date! +1

I've updated the changelog to mention this there as well.

PR #1329, opened 2021-10-12
Updated Hangouts configuration readme.

mochman:
Also indicates that the Mautrix googlechat bridge is a replacement for hangouts.


spantaleev:
Thanks a lot for bringing us up to date! :+1:

I've updated the changelog to mention this there as well.

PR #1333, opened 2021-10-13
Add support for suse linux

Samonitari:
Not tested with many openSUSE versions, but should at least work on Leap 15.2 and Tumbleweed.


spantaleev:
Looks good! Thanks! :+1:

PR #1335, opened 2021-10-15
Mention suse linux support in docs

ChrisKar96:
Add a note in the docs about the suse linux support that was added with #1333

### DIFF
--- a/docs/prerequisites.md
+++ b/docs/prerequisites.md
@@ -7,6 +7,7 @@ To install Matrix services using this Ansible playbook, you need:
   - **Debian** (10/Buster or newer)
   - **Ubuntu** (18.04 or newer, although [20.04 may be problematic](ansible.md#supported-ansible-versions))
   - **Archlinux**
+  - **Suse** (Leap 15.2 and Tumbleweed)
 
 Generally, newer is better. We only strive to support released stable versions of distributions, not betas or pre-releases. This playbook can take over your whole server or co-exist with other services that you have there.
 


spantaleev:
Don't we need to adjust `roles/matrix-base/tasks/server_base/setup.yml` as well, to install various dependencies?

ChrisKar96:
That's true, let me look into it. Making those changes means the title of the PR should change from just "mention in docs" to "add support", so I'm going to close this one and come back with another.

Samonitari:
Sorry for the sloppy PR of #1333 
I have been reapplying my local stash to `roles/matrix-base/defaults/main.yml` since almost two years ago, and there were some occasional conflict, so I made it "official".
I did realize later that I missed the docs, but package dependencies (and possible other preconditions) eluded me, as they were installed before I first run this ansible setup - matrix-related were not the first docker containers on my server.

I'll create a new PR in sometime which will:
- add `matrix-docker-ansible-deploy/roles/matrix-base/tasks/server_base/setup_suse.yml` (zypper packages `"{{ matrix_ntpd_package }}"`, `docker`, `python3-docker` )
- make `matrix-docker-ansible-deploy/roles/matrix-base/tasks/server_base/setup.yml` include it if os is SUSE
- mention SUSE support in the docs, with exactly this change

Thank @spantaleev and @ChrisKar96

spantaleev:
Thanks! Looking forward to having such complete SUSE support!

PR #1366, opened 2021-10-30
fixes #1365

### LINKED ISSUE #1362: Postgres container configuration has insufficient shm size for some queries (and auto-VACUUM)
I was playing chicken with my disk space and deicded to try a partial VACUUM and ANALYZE on my server.

When I pulled up a postgres cli and executed it, it quickly failed with:

```
ERROR:  could not resize shared memory segment "/PostgreSQL.<numbers>" to <numbers>  bytes: No space left on device
```

A quick bit of googling told me that docker's default shm filesystem (tempfiles in memory) size is too small for how postgres uses those as memory-maps. So I added:

```
      --shm-size="256m" \
```

To `/etc/systemd/system/matrix-postgres.service` and restarted postgres, and then I was able to execute a VACUUM and ANALYZE.

I suspect I haven't had an auto VACUUM/ANALYZE work in a while due to this bug.

### LINKED ISSUE #1365: #1362 broke postgres container on CentOS 7
the commit fc593bc8ab58c0430eea75da4df157ece5451bc9 breaks the postgres container on CentOS 7

```text
matrix-postgres[…]: invalid argument "\"256m\"" for "--shm-size" flag: invalid size: '"256m"'
```

As @vrutkovs [pointed out quite rightly](https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/1362#issuecomment-955173443), removing the quotes makes it work again.

MR for this trivial adjustment coming up shortly, had to change it anyway on my checkout, might as well push to a remote.

pcfe:
docker-ce does not like quotes around the shm-size value

### DIFF
--- a/roles/matrix-postgres/templates/systemd/matrix-postgres.service.j2
+++ b/roles/matrix-postgres/templates/systemd/matrix-postgres.service.j2
@@ -22,7 +22,7 @@ ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-postgres \
 			--read-only \
 			--tmpfs=/tmp:rw,noexec,nosuid,size=100m \
 			--tmpfs=/run/postgresql:rw,noexec,nosuid,size=100m \
-			--shm-size="256m" \
+			--shm-size=256m \
 			--network={{ matrix_docker_network }} \
 			{% if matrix_postgres_container_postgres_bind_port %}
 			-p {{ matrix_postgres_container_postgres_bind_port }}:5432 \


pcfe:
see also #1362

pcfe:
apologies for the force push noise, initially pasted wrong issue number in the commit msg. doh!

spantaleev:
Thanks! 👍 Interesting.. I supposedly did `--tags=setup-postgres,start` before pushing this change and did not observe my Postgres dying, but.. I'll have to check again why later.

PR #1369, opened 2021-10-31
change port 8090 to matrix_ma1sd_default_port

borisrunakov:
Changed hard coded port 8090 to `matrix_ma1sd_default_port` wherever possible.


spantaleev:
Thanks! :+1: 

We do hard code ports like that a lot and it's nice to move away from it.

borisrunakov:
I have overseen one part though, will fix it now. Are there any tests I can run before the PR ? i.e. doing the self test will only test functionality based on settings on `vars.yaml`. Do you have something to test every possible use case ?

spantaleev:
The self-test is meant for users to check if some important ports on their server are open (which indicates important services are running). It's not meant as a unit/integration test for us while developing. Unfortunately, we don't have such a thing (yet).

borisrunakov:
Good to know! Will try to think of some unit / integration tests.

PR #1376, opened 2021-11-03
chore: adding some explanations about setting prometheus password

ldicarlo:
So obviously I am submitting this PR because I put the password in plain text first, and at some point I found the explanation in https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/4586efebd7b7feb9d1936485168d84d25ff250dc/roles/matrix-nginx-proxy/defaults/main.yml#L187.


spantaleev:
Thank you! :+1:

PR #1380, opened 2021-11-05
add server_name to matrix-synapse.conf only if matrix_nginx_proxy_enabled

borisrunakov:
nginx matrix-synapse.conf generated by the playbook, should have `server_name` only if `matrix_nginx_proxy_enabled`

### DIFF
--- a/roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2
+++ b/roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2
@@ -62,7 +62,9 @@
 
 server {
 	listen 12080;
-	server_name {{ matrix_nginx_proxy_proxy_synapse_hostname }};
+	{% if matrix_nginx_proxy_enabled %}
+		server_name {{ matrix_nginx_proxy_proxy_synapse_hostname }};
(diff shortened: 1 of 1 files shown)


spantaleev:
Thanks! :+1: 

--------

To give some backstory:

When `matrix_nginx_proxy_enabled: false`, `roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2` sends the `matrix.DOMAIN` traffic to `matrix_nginx_proxy_proxy_synapse_client_api_addr_sans_container`.

`matrix_nginx_proxy_proxy_synapse_client_api_addr_sans_container`, which is `127.0.0.1:12080`.

These 2 vhosts define a `server_name` of `matrix-nginx-proxy` (at least before this PR). They are the only ones that listen on port `12080`, so there's no confusion about where traffic should go. The problem is that when an external nginx server is used, a `server_name` of `matrix-nginx-proxy` is not appropriate, because that nginx webserver cannot resolve `matrix-nginx-proxy` and crashes on startup. We either need to replace it with something else which would resolve (`localhost`, etc.), or drop the `server_name` definition entirely.

This PR goes for the latter.

borisrunakov:
Appreciate !

PR #1382, opened 2021-11-06
mautrix signal - allow double puppet discovery

borisrunakov:


borisrunakov:
Ah very nice ! Didn't know of this feature. Thank you !

PR #1383, opened 2021-11-06
set homeserver_url to matrix_domain when nginx-proxy disabled and workers enabled

borisrunakov:
Trying to set mautrix-signal bridge, I could not `link` to a new device. This was because the bridge could not connect to the homeserver endpoint  with `matrix-nginx-proxy` disabled and `workers` enabled. 
Setting the `homeserver.address` to `https:matrix_domain` solved my issue

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -21,7 +21,7 @@ matrix_identity_server_url: "{{ ('https://' + matrix_server_fqn_matrix) if matri
 
 # If Synapse workers are enabled and matrix-nginx-proxy is disabled, certain APIs may not work over 'http://matrix-synapse:8008'.
 # This is because we explicitly disable them for the main Synapse process.
-matrix_homeserver_container_url: "{{ 'http://matrix-nginx-proxy:12080' if matrix_nginx_proxy_enabled else 'http://matrix-synapse:8008' }}"
+matrix_homeserver_container_url: "{{ 'http://matrix-nginx-proxy:12080' if matrix_nginx_proxy_enabled else 'https://' + matrix_domain }}"
 
 ######################################################################
 #


spantaleev:
Seems like this change will affect lots of things. It makes it so that when the nginx proxy is disabled, all communication would go via the public interface (`https://matrix.DOMAIN`), which is not ideal. If communication can happen locally (especially inside the `matrix` Docker network), it should.

Maybe it's just the Signal bridge that is doing something weird with this `matrix_homeserver_container_url` value (`http://matrix-synapse:8008` in your case). Perhaps it's trying to use it for public things too. Maybe it's just this bridge that should be fixed to use some other value, instead of making all playbook services default to communicating over the public network interfaces.

borisrunakov:
The problem I got with mautrix-signal was : 
```
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/mautrix/bridge/commands/handler.py", line 406, in handle
    await self._run_handler(handler, evt)
  File "/usr/lib/python3.9/site-packages/mautrix/bridge/commands/handler.py", line 291, in __call__
    return await self._handler(evt)
  File "/usr/lib/python3.9/site-packages/mautrix_signal/commands/auth.py", line 65, in link
    content = await make_qr(evt.az.intent, sess.uri)
  File "/usr/lib/python3.9/site-packages/mautrix_signal/commands/auth.py", line 45, in make_qr
    mxc = await intent.upload_media(qr, "image/png", "qr.png", len(qr))
  File "/usr/lib/python3.9/site-packages/mautrix/appservice/api/intent.py", line 83, in wrapper
    return await __method(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/mautrix/client/api/modules/media_repository.py", line 58, in upload_media
    resp = await self.api.request(Method.POST, MediaPath.upload, content=data,
  File "/usr/lib/python3.9/site-packages/mautrix/api.py", line 259, in request
    return await self._send(method, full_url, content, query_params, headers or {})
  File "/usr/lib/python3.9/site-packages/mautrix/api.py", line 180, in _send
    raise make_request_error(http_status=response.status,
mautrix.errors.request.MatrixUnknownRequestError: 404: 
<html>
  <head><title>404 - No Such Resource</title></head>
  <body>
    <h1>No Such Resource</h1>
    <p>Sorry. No luck finding that resource.</p>
  </body>
</html>

```
which regarding to tulir, it seems that the media repo isn't reachable. 
With nginx proxy disabled and workers enabled the bridge has to reach port `12080` in matrix-synapse.conf:
```
server {
        listen 12080;

        server_tokens off;
        root /dev/null;

        gzip on;
        gzip_types text/plain application/json;
...
        location ~ ^/_matrix/media/ {
                                        proxy_pass http://media_repository_upstream$request_uri;
                                        proxy_set_header Host $host;
        
                                        client_body_buffer_size 25M;
                                        client_max_body_size 50M;
                                        proxy_max_temp_file_size 0;
...
        location / {
                        proxy_pass http://127.0.0.1:8008;

                proxy_set_header Host $host;

                client_body_buffer_size 25M;
                client_max_body_size 50M;
                proxy_max_temp_file_size 0;
        }
}
```
but `http://127.0.0.1:12080` is proxied through matrix-domain.conf:
```
server {
        listen 443 ssl http2;
        listen [::]:443 ssl http2;
        server_name matix_domain;
...
        location ~* ^(/_matrix|/_synapse/client) {
                                proxy_pass http://127.0.0.1:12080;
        
                        proxy_set_header Host $host;
                        proxy_set_header X-Forwarded-Proto $scheme;
        
                        client_body_buffer_size 25M;
                        client_max_body_size 50M;
                        proxy_max_temp_file_size 0;
                }
}

```
I agree that traffic should be kept locally, but pointing `matrix_homeserver_container_url`  to `matrix-domain` seemed to me a consistent solution.

spantaleev:
What if you point `matrix_homeserver_container_url` to `http://localhost:12080`? Does it fix your problem?

borisrunakov:
> What if you point `matrix_homeserver_container_url` to `http://localhost:12080`? Does it fix your problem?

Sorry I missed that part. However the same happened to me today, while trying to set up `matrix_mautrix_whatsapp`. 
With `login` I got `/media` related errors so I had to override `matrix_mautrix_whatsapp_homeserver_address` to `https://{{ matrix-domain }}` to make this work

spantaleev:
I think that the best solution is to forbid disabling `matrix-nginx-proxy` when Synapse workers are enabled.

Instead of disabling `matrix-nginx-proxy`, one could leave it enabled, but only make it expose ports locally (and not retrieve SSL certificates): https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-own-webserver.md#method-2-fronting-the-integrated-nginx-reverse-proxy-webserver-with-another-reverse-proxy

This way, everything can continue functioning as it is, and bridges don't need to communicate with the homeserver over the public network (like you're proposing in this PR).

----------

A potential easy workaround for you is to redefine `matrix_mautrix_signal_homeserver_address`, so that it doesn't use the `matrix_homeserver_container_url` variable. Example:

```yaml
matrix_mautrix_signal_homeserver_address: "https://{{ matrix_server_fqn_matrix }}"
```

borisrunakov:
> A potential easy workaround for you is to redefine `matrix_mautrix_signal_homeserver_address`, so that it doesn't use the `matrix_homeserver_container_url` variable. Example:
> 
> ```yaml
> matrix_mautrix_signal_homeserver_address: "https://{{ matrix_server_fqn_matrix }}"
> ```
This is what I am using at the moment. However forbidding disabling `matrix-nginx-proxy`, gives a big disadvantage to the playbook imo, as there are many use cases where the server is used already with an nginx proxy.

spantaleev:
Do you use the playbook with `matrix_nginx_proxy_enabled: false` and your own nginx running on the host?

borisrunakov:
Yeap

PR #1394, opened 2021-11-13
Port 80 for IPv6

JokerGermany:
If you use a external certbot because of an external reverse proxy, the server has to be reached via ipv6 on port80.


spantaleev:
Looks good! Thank you! :+1:

PR #1397, opened 2021-11-14
Synapse container federation port

borisrunakov:
Make port `8048` configurable

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1047,6 +1047,8 @@ matrix_dimension_enabled: false
 # the Dimension HTTP port to the local host.
 matrix_dimension_container_http_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:8184' }}"
 
+matrix_dimension_homeserver_federationUrl: "http://matrix-synapse:{{matrix_synapse_container_federation_port|string}}"
+
 matrix_integration_manager_rest_url: "{{ matrix_dimension_integrations_rest_url if matrix_dimension_enabled else None }}"
 matrix_integration_manager_ui_url: "{{ matrix_dimension_integrations_ui_url if matrix_dimension_enabled else None }}"
 
@@ -1305,6 +1307,9 @@ matrix_nginx_proxy_proxy_matrix_federation_api_addr_sans_container: "127.0.0.1:1
 
 # Settings controlling matrix-synapse-proxy.conf
 matrix_nginx_proxy_proxy_synapse_enabled: "{{ matrix_synapse_enabled }}"
+
+matrix_nginx_proxy_proxy_synapse_federation_api_addr_with_container: "matrix-synapse:{{matrix_synapse_container_federation_port|string}}"
+matrix_nginx_proxy_proxy_synapse_federation_api_addr_sans_container: "localhost:{{matrix_synapse_container_federation_port|string}}"
 
 # When matrix-nginx-proxy is disabled, the actual port number that the vhost uses may begin to matter.
 matrix_nginx_proxy_proxy_matrix_federation_port: "{{ matrix_federation_public_port }}"
@@ -1720,8 +1725,7 @@ matrix_synapse_account_threepid_delegates_msisdn: "{{ 'http://matrix-ma1sd:' + m
 matrix_synapse_container_client_api_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:8008' }}"
 #
 # For exposing the Matrix Federation API's plain port (plain HTTP) to the local host.
-matrix_synapse_container_default_federation_port: 8048
-matrix_synapse_container_federation_api_plain_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:' + matrix_synapse_container_default_federation_port|string }}"
+matrix_synapse_container_federation_api_plain_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:' + matrix_synapse_container_federation_port|string }}"
 #
 # For exposing the Matrix Federation API's TLS port (HTTPS) to the internet on all network interfaces.
 matrix_synapse_container_federation_api_tls_host_bind_port: "{{ matrix_federation_public_port if (matrix_synapse_federation_enabled and matrix_synapse_tls_federation_listener_enabled) else '' }}"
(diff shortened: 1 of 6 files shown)


group_vars/matrix_servers
```diff
@@ -1720,7 +1720,8 @@ matrix_synapse_account_threepid_delegates_msisdn: "{{ 'http://matrix-ma1sd:' + m
 matrix_synapse_container_client_api_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:8008' }}"
 #
 # For exposing the Matrix Federation API's plain port (plain HTTP) to the local host.
-matrix_synapse_container_federation_api_plain_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:8048' }}"
+matrix_synapse_container_default_federation_port: 8048
```
spantaleev, line 1723: This is the first definition of this variable. As such, it should be part of the `matrix-synapse` role, not in this hook-all-roles-together `group_vars/matrix_servers` file.

Relocating this to `roles/matrix-synapse/defaults/main.yml` is better. Renaming it to `matrix_synapse_container_federation_port` (without `_default`) is better.

roles/matrix-dimension/defaults/main.yml
```diff
@@ -39,7 +39,7 @@ matrix_dimension_integrations_rest_url: "https://{{ matrix_server_fqn_dimension
 matrix_dimension_integrations_widgets_urls: ["https://{{ matrix_server_fqn_dimension }}/widgets"]
 matrix_dimension_integrations_jitsi_widget_url: "https://{{ matrix_server_fqn_dimension }}/widgets/jitsi"
 
-matrix_dimension_homeserver_federationUrl: "http://matrix-synapse:8048"
+matrix_dimension_homeserver_federationUrl: "http://matrix-synapse:{{matrix_synapse_container_default_federation_port|string}}"
```
spantaleev, line 42: Leaving this default variable in `roles/matrix-dimension/defaults/main.yml` is probably OK. This way, we won't have to refer to `matrix-synapse` variables inside the `matrix-dimension` role. We usually try to avoid using variables across roles like this and prefer to hook roles together using the `group_vars/matrix_servers` file.

It may even make sense to define `matrix_dimension_homeserver_federationUrl` as empty in the `matrix-dimension` role and use `roles/matrix-dimension/tasks/validate_config.yml` to validate that it's defined.

You can redefine `matrix_dimension_homeserver_federationUrl` via `group_vars/matrix_servers`:

```
matrix_dimension_homeserver_federationUrl: "http://matrix-synapse:{{matrix_synapse_container_default_federation_port|string}}"
```

roles/matrix-nginx-proxy/defaults/main.yml
```diff
@@ -116,8 +116,8 @@ matrix_nginx_proxy_proxy_synapse_enabled: false
 matrix_nginx_proxy_proxy_synapse_hostname: "matrix-nginx-proxy"
 matrix_nginx_proxy_proxy_synapse_federation_api_enabled: "{{ matrix_nginx_proxy_proxy_matrix_federation_api_enabled }}"
 # The addresses where the Federation API is, when using Synapse.
-matrix_nginx_proxy_proxy_synapse_federation_api_addr_with_container: "matrix-synapse:8048"
-matrix_nginx_proxy_proxy_synapse_federation_api_addr_sans_container: "localhost:8048"
+matrix_nginx_proxy_proxy_synapse_federation_api_addr_with_container: "matrix-synapse:{{matrix_synapse_container_default_federation_port|string}}"
```
spantaleev, line 119: Same. This should not reference `matrix-synapse` variables from within this `matrix-nginx-proxy` role. This should be kept as it is and then redefined in `group_vars/matrix_servers`.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -302,7 +302,7 @@ listeners:
 {% if matrix_synapse_federation_port_enabled %}
   # Unsecure HTTP listener (Federation API): for when matrix traffic passes through a reverse proxy
   # that unwraps TLS.
-  - port: 8048
+  - port: {{ matrix_synapse_container_default_federation_port }}
```
spantaleev, line 305: `{{ matrix_synapse_container_default_federation_port|to_json }}` would be better.

spantaleev (changes requested):
Thank you for making the playbook more flexible!

I've submitted some suggestions for how to make this PR better.

-----------------

However, I also don't understand what the benefit of using this PR is. Usually, the `8048` port is just an internal container port and it doesn't hurt (conflict with) anyone. I'm guessing you'd like to use `matrix_nginx_proxy_enabled: false`, which potentially exposes this internal port as `127.0.0.1:8048` and you'd like to use another port.

You may be able to manually change this to something else by redefining `matrix_synapse_container_federation_api_plain_host_bind_port` along with potential changes to `matrix_nginx_proxy_proxy_synapse_federation_api_addr_sans_container`. Example:

```yaml
matrix_synapse_container_federation_api_plain_host_bind_port: '127.0.0.1:55555'
matrix_nginx_proxy_proxy_synapse_federation_api_addr_sans_container: 'localhost:55555'
```

This way, your external nginx server can reach Synapse via the the new external port.

All other variables can be kept as-is, as the `8048` internal port number for Synapse still remains the same and operational inside the `matrix` container network. Thus, Dimension and other internal services can still connect to `matrix-synapse:8048` just the same.

group_vars/matrix_servers
borisrunakov: I agree, you are right with `matrix_synapse_container_federation_port`

roles/matrix-dimension/defaults/main.yml
borisrunakov: Ah nice ! Yes leaving `matrix_dimension_homeserver_federationUrl` in `matrix-dimension` role as empty and define it in `group_vars/matrix_servers` is better.

roles/matrix-nginx-proxy/defaults/main.yml
borisrunakov: So `matrix_nginx_proxy_proxy_synapse_federation_api_addr_with_container` should be empty here too ?

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
borisrunakov: I agree

borisrunakov:
In general I am trying to remove any hard coded ports from the playbook. The goal is to end up with a playbook that can be installed in parallel with other instances on the same server.

roles/matrix-nginx-proxy/defaults/main.yml
spantaleev: It can be. But it doesn't hurt to leave some default value (which doesn't use a variable).

spantaleev:
I've renamed this variable in c1bc7b9f934a1 to prevent confusion with the one from #1399. Thank you! :+1:

PR #1399, opened 2021-11-14
make 8448 configurable instead of hard coded

borisrunakov:


roles/matrix-synapse/defaults/main.yml
```diff
@@ -27,6 +27,8 @@ matrix_synapse_storage_path: "{{ matrix_synapse_base_path }}/storage"
 matrix_synapse_media_store_path: "{{ matrix_synapse_storage_path }}/media-store"
 matrix_synapse_ext_path: "{{ matrix_synapse_base_path }}/ext"
 
+ matrix_synapse_container_default_federation_api_port: 8448
```
spantaleev, line 30: Using `_default` in the name doesn't make much sense.

`matrix_synapse_container_federation_api_port` seems better.

This variable can default to `8448` here in the `matrix-synapse` role, and can be redefined in terms of the `matrix_federation_public_port` variable via `group_vars/matrix_servers` (although it's not really necessary to make the public port and the internal port match).

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -276,7 +276,7 @@ listeners:
 
 {% if matrix_synapse_federation_port_enabled and matrix_synapse_tls_federation_listener_enabled %}
   # TLS-enabled listener: for when matrix traffic is sent directly to synapse.
-  - port: 8448
+  - port: {{ matrix_federation_public_port }}
```
spantaleev, line 279: This seems to be confusing `matrix_federation_public_port` and `matrix_synapse_container_default_federation_api_port` (well, `matrix_synapse_container_federation_api_port` if the `_default` part of the name is dropped).

`matrix_federation_public_port` is the public port number. `matrix_synapse_container_federation_api_port` is the port inside the Synapse container. They don't really need to match.

spantaleev (changes requested):
I've submitted some minor suggestions.

I'm also interested in what your use-case for this is?

roles/matrix-synapse/defaults/main.yml
borisrunakov: Will change it to `matrix_synapse_container_federation_api_port`. Is there any naming convention that you follow ? Also what is the better place to discuss development ideas ?

spantaleev: Well, in this case, "default" sounds like a default only. It sounds like a value that should be defined and used as a reference without being redefined.

It may make sense to use it like this, actually:

```yaml
# This is some internal variable with a default value.
default_value: 5555

# Users can redefine this.
actual_value: "{{ default_value }}"

something_else: "https://localhost:{{ actual_value }}"
```

------

In your case, you intend to have this variable redefined by users. Redefining `matrix_synapse_container_default_federation_api_port: 55555` sounds strange. You've actually redefined the federation API port and many other variables to use 55555. The default for Synapse is still 8448 (as per spec and natural consensus).

borisrunakov: I agree. Really appreciate your explanation!

spantaleev:
I've renamed this variable in c1bc7b9f934a1 to prevent confusion with the one from #1397.

PR #1400, opened 2021-11-14
Updated Mautrix-WhatsApp config

IUCCA:
Only works with multidevice beta, but i need some help to enable MSC2716. 
I tried this in synapse homeserver.yml, but it doesn't work.
With this version chat history backfill won't work.


spantaleev:
This PR does quite a lot in so many places via a new global `matrix_bridges_backfill_enabled` variable, so I don't think it's great to merge it as it is.

By the way, you can inject custom Synapse configuration via `matrix_synapse_configuration_extension_yaml`:

```yaml
matrix_synapse_configuration_extension_yaml: |
  Experimental:
    msc2716_enabled: true
```

Similarly, you can inject Whatsapp bridge configuration via `matrix_mautrix_whatsapp_configuration_extension_yaml`, so you can definitely enable backfilling without any playbook modifications.

----------

As for the other parts of the PR (updating the Whatsapp configuration with newer configuration keys) is great and I'd like to merge it!

IUCCA:
Now it is working with history backfill and only with changes in the bridge config template and bridge documentation

IUCCA:
Should we note in the change file that whatsapp multidevice beta is necessary

roles/matrix-bridge-mautrix-whatsapp/templates/config.yaml.j2
```diff
@@ -51,79 +52,146 @@ bridge:
...
+    # Note that updating the m.direct event is not atomic (except with mautrix-asmux)
+    # and is therefore prone to race conditions.
+    sync_direct_chat_list: false
+    # When double puppeting is enabled, users can use `!wa toggle` to change whether
+    # presence and read receipts are bridged. These settings set the default values.
+    # Existing users won't be affected when these are changed.
+    default_bridge_receipts: true
+    default_bridge_presence: true
+    # Servers to always allow double puppeting from
+    double_puppet_server_map: "{{ matrix_domain }}"
```
spantaleev, line 108: Looking at the [example configuration](https://mau.dev/mautrix/whatsapp/-/blob/master/example-config.yaml#L136), this needs to be a map/dictionary (domain name => URL).

As an example, see the `matrix_mautrix_instagram_bridge_login_shared_secret_map` map definition. That one is a (domain => shared secret), however. I'm guessing you need `matrix_mautrix_whatsapp_homeserver_domain` and `matrix_mautrix_whatsapp_homeserver_address`:

```yaml
matrix_mautrix_whatsapp_brdige_double_puppet_server_map: "{{ {matrix_mautrix_whatsapp_homeserver_domain: matrix_mautrix_whatsapp_homeserver_address} }}"
```
or this (if it works syntax-wise):

```yaml
matrix_mautrix_whatsapp_brdige_double_puppet_server_map:
  {{ matrix_mautrix_whatsapp_homeserver_domain }}: {{ matrix_mautrix_whatsapp_homeserver_address }}
```

roles/matrix-bridge-mautrix-whatsapp/templates/config.yaml.j2
```diff
@@ -51,79 +52,146 @@ bridge:
...
-    # chat portals when backfilling if needed.
-    # This always uses the default puppet instead of custom puppets due to
-    # rate limits and timestamp massaging.
-    invite_own_puppet_for_backfilling: true
-    # Whether or not to explicitly set the avatar and room name for private
-    # chat portal rooms. This can be useful if the previous field works fine,
-    # but causes room avatar/name bugs.
-    private_chat_portal_meta: false
+    login_shared_secret_map:
+        "{{ matrix_mautrix_whatsapp_homeserver_domain }}": {{ matrix_mautrix_whatsapp_login_shared_secret|to_json }}
```
spantaleev, line 118: It's probably better if this `login_shared_secret_map` thing is moved to a `matrix_mautrix_whatsapp_bridge_login_shared_secret_map` variable like we've done for the mautrix-instagram bridge:

```yaml
matrix_mautrix_whatsapp_bridge_login_shared_secret_map: "{{ {matrix_mautrix_whatsapp_homeserver_domain: matrix_mautrix_whatsapp_login_shared_secret} if matrix_mautrix_whatsapp_login_shared_secret else {} }}"
```

If the (optional) shared secret is not defined, it wouldn't be enabled.

The configuration then becomes:

```yaml
  login_shared_secret_map: {{ matrix_mautrix_whatsapp_bridge_login_shared_secret_map|to_json }}
```

roles/matrix-bridge-mautrix-whatsapp/templates/config.yaml.j2
```diff
@@ -105,19 +95,16 @@ bridge:
     default_bridge_receipts: true
     default_bridge_presence: true
     # Servers to always allow double puppeting from
-    double_puppet_server_map: "{{ matrix_domain }}"
-
+    double_puppet_server_map:
```
spantaleev, line 98: Shouldn't we push this to a `matrix_mautrix_whatsapp_bridge_double_puppet_server_map` variable?

roles/matrix-bridge-mautrix-whatsapp/templates/config.yaml.j2
```diff
@@ -105,19 +95,16 @@ bridge:
...
     # Shared secrets for https://github.com/devture/matrix-synapse-shared-secret-auth
     #
     # If set, double puppeting will be enabled automatically for local users
     # instead of users having to find an access token and run `login-matrix`
     # manually.
-    login_shared_secret_map:
-        "{{ matrix_mautrix_whatsapp_homeserver_domain }}": {{ matrix_mautrix_whatsapp_login_shared_secret|to_json }}
-
-
+    login_shared_secret_map: {{ matrix_mautrix_whatsapp_login_shared_secret_map|to_json }}
```
spantaleev, line 107: Given that this is in the `bridge` section of the configuration, it should probably be named `matrix_mautrix_whatsapp_bridge_login_shared_secret_map` (not `matrix_mautrix_whatsapp_login_shared_secret_map`).

IUCCA: Okay, schould i rename `matrix_mautrix_whatsapp_login_shared_secret:  ' '` also?

spantaleev: I think `matrix_mautrix_whatsapp_login_shared_secret` is OK and it's an existing variable that people are already using. I suppose we assign to it from `group_vars/matrix_servers` too.

spantaleev:
Looks good! Thank you! :+1:

PR #1403, opened 2021-11-16
remove default from variable name

borisrunakov:


spantaleev:
Thanks! :+1:

Looks like this is a relatively new variable, so renaming it like this will probably not break things for many people.
Still, I've added a warning to `roles/matrix-ma1sd/tasks/validate_config.yml` in 75cb4ce3b02ba0c

PR #1404, opened 2021-11-17
Allow workers to serve new v3 APIs

aaronraimist:
Support for these APIs will probably be in Synapse 1.48.0 but this can get merged at any time.

### DIFF
--- a/roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2
+++ b/roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2
@@ -120,7 +120,7 @@ server {
 			{% endfor %}
 			{% if matrix_nginx_proxy_synapse_presence_disabled %}
 			# FIXME: keep in sync with synapse workers documentation manually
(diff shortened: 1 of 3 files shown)


spantaleev:
We'll probably lose the changes to `roles/matrix-synapse/vars/workers.yml` the next time we run `roles/matrix-synapse/files/workers-doc-to-yaml.sh`.

It doesn't seem like it hurts to merge this, however.

Does the `develop` branch of Synapse (and some `:latest` Docker image?) already handle any v3 endpoints? I'm not sure what the `:latest` Docker image actually is - latest stable or latest `develop`? Seems like there is a `:develop` Docker image, but it hasn't been updated in 2 years.

aaronraimist:
Ah I didn't know about `workers-doc-to-yaml.sh`.

Develop does serve v3 APIs as of https://github.com/matrix-org/synapse/pull/11318

I believe the `latest` tag on docker is latest stable version.

spantaleev:
I see that you've also updated Synapse's `docs/workers.md` in (https://github.com/matrix-org/synapse/pull/11371, etc.), so our auto-generated `workers.yml` should not lose too much. Still, it'd be nice to test it and fix it up when the time comes.

I'm merging this as-is. Thank you! :+1:

PR #1406, opened 2021-11-17
expose smtp_user and smtp_pass to ansible configs (role: matrix-synapse)

rakshazi:
Hello,

currently to configure synapse email sending you need to add smtp username and password through synapse additional configuration var/block, but all other smtp connection-related vars already exposed via role vars.

This PR adds missing `smtp_user` and `smtp_pass` to synapse role defaults

### DIFF
--- a/roles/matrix-synapse/defaults/main.yml
+++ b/roles/matrix-synapse/defaults/main.yml
@@ -473,6 +473,8 @@ matrix_synapse_turn_allow_guests: False
 matrix_synapse_email_enabled: false
 matrix_synapse_email_smtp_host: ""
 matrix_synapse_email_smtp_port: 587
+matrix_synapse_email_smtp_user: ""
+matrix_synapse_email_smtp_pass: ""
 matrix_synapse_email_smtp_require_transport_security: false
 matrix_synapse_email_notif_from: "Matrix <matrix@{{ matrix_domain }}>"
 matrix_synapse_email_client_base_url: "https://{{ matrix_server_fqn_element }}"
--- a/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
+++ b/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
@@ -2338,9 +2338,8 @@ email:
 
   # Username/password for authentication to the SMTP server. By default, no
   # authentication is attempted.
-  #
-  #smtp_user: "exampleusername"
-  #smtp_pass: "examplepassword"
+  smtp_user: {{ matrix_synapse_email_smtp_user|string|to_json }}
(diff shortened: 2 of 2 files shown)


skepticalwaves:
How does this relate to https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-email.md ?

rakshazi:
@skepticalwaves it doesn't. The doc you linked explains how to work with exim (role: matrix-mailer), but this PR exposes 2 missed vars of synapse (role: matrix-synapse).

Both exim and synapse can send emails and the default way for playbook is to use exim. This PR allows to use synapse directly to send emails, without need of exim. Of couse, that's totally optional and will not change current flow anyhow, but in my case I don't have identity server, don't actually need exim and totally fine with synapse emails.

skepticalwaves:
If I understand then, this conflicts with the mailer config? Can that be documented?

spantaleev:
Thanks! :+1: 

It doesn't hurt adding these variables for completeness.

Basically, people have 2 ways to go about configuring email:

1) Globally, for all services via `matrix-mailer`. By default, `matrix-mailer` is enabled and Synapse and other services send emails through it. By default, `matrix-mailer` tries to send emails directly which may or may not work for some people. If people want to relay emails via a custom SMTP server, they can follow [docs/configuring-playbook-email.md](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-email.md) to reconfigure `matrix-mailer`. When `matrix-mailer` is used, the playbook doesn't need to set `smtp_user` and `smtp_pass` in Synapse's configuration, because  it's all unauthenticated private communication over the `matrix` container network.

2)  Locally, just for Synapse. Some people may wish to disable `matrix-mailer` or otherwise make Synapse send email directly via an SMTP server (without email going through `matrix-mailer`). If an external SMTP server is to be used directly by Synapse, it's most likely necessary to authenticate against it using a username and password. Having `smtp_user` and `smtp_pass` variables makes that easier -- otherwise one would need to set `matrix_synapse_email_smtp_host`, `matrix_synapse_email_smtp_port`, etc. via convenient variables, and set the SMTP username/password via `matrix_synapse_configuration_extension_yaml`.

--------

[docs/configuring-playbook-email.md](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-email.md) can probably be adjusted to mention this **global** configuration (`matrix-mailer`, which is used by all services), and the possibility for using **component-specific configuration** (`matrix_synapse_email_*` variables; `matrix_ma1sd_threepid_medium_email_connectors_*` variables; etc.)

skepticalwaves:
Thanks for the clarification @spantaleev

PR #1409, opened 2021-11-20
changed ExecStartPre in the systemd file(--name of docker container contained a typo) of mautrix-telegram

yoitcc:
removed "-db" from the third ExecStartPre= command (--name of the docker container), because this creates a container that is not getting removed by the previous ExecStartPre commands ^^

thanks for all the great work, signal and whatsapp bridges have been blazingly simple to install ;)

created this error:
```
$ sudo journalctl -u matrix-mautrix-telegram.service -f

Nov 20 19:01:20 webphp systemd[1]: Starting Matrix Mautrix Telegram bridge...
Nov 20 19:01:20 webphp matrix-mautrix-telegram[360889]: docker: Error response from daemon: Conflict. The container name "/matrix-mautrix-telegram-db" is already in use by container "e01ca...c5f8840d67986". You have to remove (or rename) that container to be able to reuse that name.
Nov 20 19:01:20 webphp matrix-mautrix-telegram[360889]: See 'docker run --help'.
Nov 20 19:01:20 webphp systemd[1]: matrix-mautrix-telegram.service: Control process exited, code=exited, status=125/n/a
Nov 20 19:01:20 webphp systemd[1]: matrix-mautrix-telegram.service: Failed with result 'exit-code'.
Nov 20 19:01:20 webphp systemd[1]: Failed to start Matrix Mautrix Telegram bridge.
Nov 20 19:01:50 webphp systemd[1]: matrix-mautrix-telegram.service: Scheduled restart job, restart counter is at 41.
```

### DIFF
--- a/roles/matrix-bridge-mautrix-telegram/templates/systemd/matrix-mautrix-telegram.service.j2
+++ b/roles/matrix-bridge-mautrix-telegram/templates/systemd/matrix-mautrix-telegram.service.j2
@@ -15,7 +15,7 @@ Type=simple
 Environment="HOME={{ matrix_systemd_unit_home_path }}"
 ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} kill matrix-mautrix-telegram 2>/dev/null'
 ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} rm matrix-mautrix-telegram 2>/dev/null'
-ExecStartPre={{ matrix_host_command_docker }} run --rm --name matrix-mautrix-telegram-db \
+ExecStartPre={{ matrix_host_command_docker }} run --rm --name matrix-mautrix-telegram \
 			--log-driver=none \
 			--user={{ matrix_user_uid }}:{{ matrix_user_gid }} \
 			--cap-drop=ALL \


spantaleev:
This is not a typo. It's intentional.

We `kill` / `rm` the "long-running container" to make sure restarting the systemd service happens without hitting issues. Normally when you stop (or restart) a systemd service, systemd sends a signal to the service to tell it to exit. If it doesn't within a certain time interval, systemd forcefully kills it. The way that `docker run` works, the actual container service is not really spawned by `docker run` itself, but by the Docker daemon. So forcefully killing the `docker run` command may, in some cases, leave a running container somewhere. For this reason, we try to clean up such zombie containers (from previous runs) before starting the current run.

The database migrations command is less likely to get stuck and run forever, so we don't try to do similar cleanup for it. We can, but it makes the systemd service file even dirtier and has no practical reason.

Because the database container is less likely to get stuck, the name we choose for it likely doesn't matter. It probably eases debugging if people can see a `mautrix-telegram-db` container running or getting stuck, instead of a `mautrix-telegram` container.

yoitcc:
Oh, I am very sorry, made that mistake with blazing guns :D 
Compared it to the whatsapp-bridge and blindly changed it, maybe something else even moved me past the error (restarting/rebooting).

Thank you for the clear explanation, the service file is now much clearer to me ^^ 
Especially that this container run migrates the database, as far as I understand the error was shown when the migrations were still running and systemd (might have been manually) tried to restart the service?
It is working now, is it save to keep it without "-db" in the case where the migrations simply might take longer than usual? Or do you mean the migrations will eventually finish and then it will proceed?

I understand and agree with your point of nameing the db-migrations container differently. will close the Pull with this comment (hopefully that is the right thing to do on github ^^)
Thank you for the great work (and also the speedy reply), it is simply magnificent!
Oswald

spantaleev:
All `ExecStartPre` lines are executed one by one. It's only then that it proceeds to `ExecStart`. So migrations will definitely have finished by the time it tries to start the actual bridge program.

PR #1412, opened 2021-11-23
mx_puppet_slack: Add support for stripHomeservers config option

moenzuel:
This option helps prevent continuous notification spam as slack users mention each other while the bridge, user, and puppeted slack user are on the same homeserver.

### DIFF
--- a/roles/matrix-bridge-mx-puppet-slack/defaults/main.yml
+++ b/roles/matrix-bridge-mx-puppet-slack/defaults/main.yml
@@ -111,3 +111,5 @@ matrix_mx_puppet_slack_registration_yaml: |
   de.sorunome.msc2409.push_ephemeral: true
 
 matrix_mx_puppet_slack_registration: "{{ matrix_mx_puppet_slack_registration_yaml|from_yaml }}"
+
+matrix_mx_puppet_slack_stripHomeservers: ''
--- a/roles/matrix-bridge-mx-puppet-slack/templates/config.yaml.j2
+++ b/roles/matrix-bridge-mx-puppet-slack/templates/config.yaml.j2
@@ -13,7 +13,12 @@ bridge:
   loginSharedSecretMap:
     {{ matrix_domain }}: {{ matrix_mx_puppet_slack_login_shared_secret }}
   {% endif %}
-
+  {% if matrix_mx_puppet_slack_stripHomeservers != '' %}
+  stripHomeservers:
+    {% for stripHomeserver in matrix_mx_puppet_slack_stripHomeservers %}
+    - {{ stripHomeserver }};
+    {% endfor %}
+  {% endif %}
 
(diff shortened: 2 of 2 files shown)


spantaleev:
This seems confusing. The default value of `matrix_mx_puppet_slack_stripHomeservers` seems to make it look like a string is expected.

Later on, you have an `if` check which compares it to a string as well. Then again, even later, you do a `for` on it, as if it's a list of strings.

------

From your description above, I fail to understand what this option really does and what one should set it to. Given that I'm not using this bridge, that may be on me though.

Neither the [sample config for mx-puppet-slack](https://github.com/Sorunome/mx-puppet-slack/blob/master/sample.config.yaml), nor [its source code](https://github.com/Sorunome/mx-puppet-slack/search?q=stripHomeservers) mention a `stripHomeservers` option, so one can't easily learn more about it there.

Are you sure there really is such a configuration option for this bridge?

moenzuel:
The option comes from the shared bridge base https://github.com/Sorunome/mx-puppet-bridge/blob/b8c95df7d9e07664ef8c59c26e49da568b73a80e/sample.config.yaml#L28

Though slack is the only puppet bridge where I have needed this option (tried slack, discord, and gchat)

It probably is indeed confusing that the default is set to an empty string. What do you think would be a better default for an optional list? Should one not even provide a default? I am admittedly an ansible beginner.

spantaleev:
Ah, I see.

We can have a default like this:

```yaml
# A list of homeserver domain names for mx-puppet-bridge's stirpHomeservers option.
#
# Learn more about it here: https://github.com/Sorunome/mx-puppet-bridge/blob/b8c95df7d9e07664ef8c59c26e49da568b73a80e/sample.config.yaml#L26-L28
matrix_mx_puppet_slack_stripHomeservers: []
```

------

Then, the `roles/matrix-bridge-mx-puppet-slack/templates/config.yaml.j2` template could be simplified like this:

```
stripHomeservers: {{ matrix_mx_puppet_slack_stripHomeservers|to_json }}
```

There's no need for `if` statements or `for` loops.

--------

Even without this PR, you can apply this configuration by yourself like this:

```
matrix_mx_puppet_slack_configuration_extension_yaml: |
  bridge:
    stripHomeservers: ['example.org', 'another.example.org']
```

We generally refrain from creating a dedicated variable for each and every setting exposed by components, because we will end up with hundreds (or thousands) of variables for all sorts of esoteric options. Each component provides a `matrix_*_configuration_extension_*` option that gives us the flexibility to fine-tune the configuration, without having a dedicated variable for each and every option.

moenzuel:
Oh `to_json` for yaml is something I'd never guess. Thank you for the insights.

With the ability to use the built in method you provided I'll go ahead and close out this PR as it is not needed.

Issue #1363, opened 2021-10-29
Random docker_network error


InterFelix:
I am using this chart to build my own GitLab CI/CD config to deploy a matrix server, and after committing a minor change only affecting the GitLab side of things (no config change of the ansible playbook itself), this strange error occurs every time I try to run the playbook setup:
```
ERROR! couldn't resolve module/action 'docker_network'. This often indicates a misspelling, missing collection, or incorrect module path.
The error appears to be in '/builds/buendischer-matrix-server/matrix-docker-ansible-deploy/roles/matrix-base/tasks/setup_matrix_base.yml': line 22, column 3, but may
be elsewhere in the file depending on the exact syntax problem.
The offending line appears to be:
- name: Ensure Matrix network is created in Docker
  ^ here
  ```
  
The last time the pipeline ran the job went through, and there has not been any config change at all.
I already tried overriding `matrix_docker_network` in my vars.yml, to no avail.
As a sidenote, I am using willhallonline/ansible:2.10-ubuntu20.04 as the docker base image from which the playbook is executed.

InterFelix:
I just noticed something:
The last time the job went through, the ansible version being run was 2.10.12, now it is 2.10.15. The docker image seems to have been upgraded in the last few hours with a bump to a new ansible version, which has now broken this playbook.
Sadly, there are no immutable tags for this docker image to fall back to.

spantaleev:
Why aren't you using the [devture/ansible](https://hub.docker.com/r/devture/ansible/) Docker image, like we recommend on our [Ansible docs page](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/ansible.md#using-ansible-via-docker)?

That one seems a bit more dated though, but at least there's a fixed tag that you can use.

InterFelix:
I considered this before switching to willhallonline's image, but my CI/CD pipeline relies on the base image being Ubuntu currently. I'll consider this, though, shouldn't be too much work to port it over to alpine.

spantaleev:
Your problem with Ansible 2.10 seems to stem from the fact that on some distros Ansible 2.10+ is split into multiple system packages ("Ansible Core" and "Ansible Collections"). All these modules (like `docker_network`, etc.) are probably part of some "collection" package, which is not installed along with the core `ansible` package.

[...]

PR #1419, opened 2021-11-24
Ansible 2.10 & AWX 19.4 compatibility

swarnat:
I know there is an extended AWX implementation within this repository, but we use the repository since beginning in AWX without any extra functions.

Unfortunately we was forced to upgrade to latest, k8s only, AWX 19.4 and run into some problems, because I suppose a latest ansible version is used in images. In other issues this also was mentioned, related to ansible-base: (example #1363)

**Our problem:**
> ERROR! couldn't resolve module/action 'docker_network'. This often indicates a misspelling, missing collection, or incorrect module path.

Is it possible to merge this collections/requirements.yaml file into the repository, so latest Ansible and also AWX come back to compatibility without adjustments? 
I'm not fully sure if this create problems with existing AWX integration, but I think not. Because also their this collections are needed.

Also I'm not fully sure if "community.general" is really needed, because "community.docker" should be extracted from general collection, but it also not hurts.

### DIFF
--- a/collections/requirements.yml
+++ b/collections/requirements.yml
@@ -1,0 +1,3 @@
+---
+collections:
+  - name: community.general.docker_network
--- a/collections/requirements.yml
+++ b/collections/requirements.yml
@@ -1,3 +1,4 @@
 ---
 collections:
   - name: community.general
+  - name: community.docker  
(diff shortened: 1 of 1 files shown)


spantaleev:
Thank you! :+1:

PR #1420, opened 2021-11-26
Support encrypted matrix_synapse_macaroon_secret_key

ChrisKar96:
This change forces ansible to decrypt the variable before using it, if encrypted with ansible-vault, to avoid the error `{"msg": "Unexpected templating type error occurred on ({{ matrix_synapse_macaroon_secret_key | password_hash('sha512') }}): secret must be unicode or bytes, not ansible.parsing.yaml.objects.AnsibleVaultEncryptedUnicode"}`
Every other variable in the playbook was found to have no problems with encryption.
The change has no negative impact on non-encrypted matrix_synapse_macaroon_secret_key variable.


spantaleev:
Thank you for making the playbook work better with Ansible Vault! :+1:

PR #1422, opened 2021-11-27
Upgrade Mjolnir from 1.1.20 to version 1.2.1

AnonyPla:
https://hub.docker.com/r/matrixdotorg/mjolnir/tags

using the "latest" tag seems inefficient as it doesn't actually redirect to the latest release

In any case, the latest release is now 1.2.1

docker pull matrixdotorg/mjolnir:v1.2.1

### DIFF
--- a/roles/matrix-bot-mjolnir/defaults/main.yml
+++ b/roles/matrix-bot-mjolnir/defaults/main.yml
@@ -10,8 +10,7 @@ matrix_bot_mjolnir_container_image_self_build_repo: "https://github.com/matrix-o
 
 matrix_bot_mjolnir_docker_image: "{{ matrix_bot_mjolnir_docker_image_name_prefix }}matrixdotorg/mjolnir:{{ matrix_bot_mjolnir_version }}"
 matrix_bot_mjolnir_docker_image_name_prefix: "{{ 'localhost/' if matrix_bot_mjolnir_container_image_self_build else matrix_container_global_registry_prefix }}"
-
-matrix_bot_mjolnir_docker_image_force_pull: "{{ matrix_bot_mjolnir_docker_image.endswith(':v1.2.1') }}"
+matrix_bot_mjolnir_docker_image_force_pull: "{{ matrix_bot_mjolnir_docker_image.endswith(':latest') }}"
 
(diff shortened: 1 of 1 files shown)


roles/matrix-bot-mjolnir/defaults/main.yml
```diff
@@ -3,15 +3,15 @@
...
-matrix_bot_mjolnir_version: "v1.1.20"
+matrix_bot_mjolnir_version: "v1.2.1"
 
 matrix_bot_mjolnir_container_image_self_build: false
 matrix_bot_mjolnir_container_image_self_build_repo: "https://github.com/matrix-org/mjolnir.git"
 
 matrix_bot_mjolnir_docker_image: "{{ matrix_bot_mjolnir_docker_image_name_prefix }}matrixdotorg/mjolnir:{{ matrix_bot_mjolnir_version }}"
 matrix_bot_mjolnir_docker_image_name_prefix: "{{ 'localhost/' if matrix_bot_mjolnir_container_image_self_build else matrix_container_global_registry_prefix }}"
 
-matrix_bot_mjolnir_docker_image_force_pull: "{{ matrix_bot_mjolnir_docker_image.endswith(':latest') }}"
```
spantaleev, line 14: This is incorrect.

The purpose of the boolean variable `matrix_bot_mjolnir_docker_image_force_pull` is to indicate whether the current tag (as defined in `matrix_bot_mjolnir_docker_image`) is a tag that gets re-pointed once in a while, and as such needs to be "force-pulled" to get the updates.

Ansible's `docker_image` task, which we use to pull images (the equivalent of the `docker pull` CLI command) only pulls an image if it doesn't have the image pulled already. For versioned images (e.g. `docker.io/matrixdotorg/mjolnir:v1.2.1`), we don't need to force-pull, as the expectation is that `v1.2.1` remains static - pulling it over and over again is not supposed to give us a new image and is thus wasteful. If someone has set `matrix_bot_mjolnir_version` to `latest`, however (for some reason), we'd like the playbook to force-pull the new `latest` every time the playbook runs. That's what `matrix_bot_mjolnir_docker_image_force_pull` indicates.

spantaleev:
Thanks! :+1:

PR #1429, opened 2021-11-30
Update importing-postgres.md

felixx9:
Typos an suggestion to expand search/replace statement to not replace on wrong places.
I'm not done yet, with my migration form "handmade" installation to ansible. But I'm testing all steps as good as possible and would have found this helpful...


spantaleev:
Thank you for exercising all the steps and finding/fixing issues with it! :+1:

PR #1434, opened 2021-12-02
Update jitsi/web ( stable-6173 -> stable-6433)

wol-win:


spantaleev:
Jitsi upgrades most often than not require changes to configuration files, etc. Have you tested that the new Jitsi version works?

wol-win:
Not yet! Wasn't aware of configuration change requirements.

apmechev:
I was just about to open a PR to do this as well. I've been running 6433 successfully for two weeks or so, so updating should not be a problem

PR #1438, opened 2021-12-04
Update signal bridge configuration template

Jaffex:
Updates the signal bridge configuration to the possible options in [version 0.2.1](https://github.com/mautrix/signal/blob/v0.2.1/mautrix_signal/example-config.yaml)

Since the ansible role uses the "latest" container, this is the currently installed verison if the signal bridge is activated.

Also adjusted some comments and moved the provisioning block around to make future comparisons via diff easier.


spantaleev:
Thank you for updating us! :+1:

PR #1443, opened 2021-12-08
Run `workers-doc-to-yaml.sh`

aaronraimist:
1472958e25c729f3fb9f6c018c2df947bcae97aa reverted some of the v3 changes. I'm not sure why. Running the `workers-doc-to-yaml.sh` script right now puts them back 🤷‍♂️.


spantaleev:
The `workers-doc-to-yaml.sh` script reverted more than the ones you see in 1472958e25c7, but I manually patched up `workers.yaml` to retain your changes made in #1404.

Good to hear that running the script yields the correct result now!

Thank you!

PR #1444, opened 2021-12-08
Fix comments about `_database_*` variables

aaronraimist:
Was renamed in 087dbe4ddc80ba6308e7ee98391ea475354e8860

It is unclear to me if there is anything you actually need to adjust with these variables. It looks like that is done automatically in `matrix_servers`.


spantaleev:
Thank you! :+1:

PR #1452, opened 2021-12-11
Updated signald docker image path

laszabine:
Recently, a vulnerability in log4j was discovered. Signald uses log4j. This PR updates the path of the signald docker image from the previously recommended, but now outdated docker image at mau.dev to the [now by @tulir recommended](https://docs.mau.fi/bridges/python/signal/setup-docker.html) docker image at docker.io.

Please merge this ASAP since there is a vulnerability on all servers running signald until the fixed image gets pulled.


spantaleev:
Thanks! 👍

PR #1464, opened 2021-12-15
Update Jitsi to stable-6726-1

sakkiii:
Ref: https://github.com/jitsi/docker-jitsi-meet/releases/tag/stable-6726-1


roles/matrix-jitsi/templates/web/matrix-jitsi-web.service.j2
```diff
@@ -23,6 +23,7 @@ ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-jitsi-web \
 			{% endif %}
 			--mount type=bind,src={{ matrix_jitsi_web_config_path }},dst=/config \
 			--mount type=bind,src={{ matrix_jitsi_web_transcripts_path }},dst=/usr/share/jitsi-meet/transcripts \
+			--mount type=bind,src={{ matrix_jitsi_web_crontabs_path }},dst=/var/spool/cron/crontabs \
```
spantaleev, line 26: What creates this crontabs file? Perhaps something is missing from `tasks/setup_install.yml`?

sakkiii: I think this is for lets-encrypt renew certs crontab which will be manged by it's docker itself.

spantaleev: As far as I know, you can't use `--mount` with a non-existing file (`/matrix/jitsi/web/crontabs` in this case).

sakkiii: yes, mount directory should be an valid path. Thanks 👍

spantaleev:
Thanks! :+1:

spantaleev:
Possibly fixed in 60592fd6a8c3

PR #1469, opened 2021-12-16
ansible lint workflow

sakkiii:
avoiding common pitfalls that can easily lead to bugs or make code harder to maintain

### DIFF
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -1,0 +1,19 @@
+name: Ansible Lint
+
+on: [push, pull_request]
+
+jobs:
+  build:
+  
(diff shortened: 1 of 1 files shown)


spantaleev:
Let's give it a try ;) Thanks! :+1: 

------

I wonder if we should fix the Ubuntu version and the [ansible-lint action](https://github.com/ansible/ansible-lint-action) version. 

About the latter, it seems like the newest ansible-lint action version is 4.1.0, which is quite old (I'm seeing 5.3.1 on my Arch linux install). Seems like they had trouble with newer versions (https://github.com/ansible/ansible-lint-action/issues/45). Using `@master` seems dangerous.

PR #1470, opened 2021-12-17
Remove a space from XMPP_INTERNAL_MUC_DOMAIN

alexbers:
The domain is invalid with space, so it shouldn't be here


spantaleev:
Seems like a regression from https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/1464. Good catch! Thanks! :+1:

PR #1425, opened 2021-11-29
Update matrix-synapse-admin version

skepticalwaves:
Not yet tested.

PR #1472, opened 2021-12-17
Update synapse-admin to newest tagged release

skepticalwaves:


skepticalwaves:
Dockerhub entry exists and is tagged. Not yet tested.

spantaleev:
I haven't tested it too, but given that this is a minor version upgrade and Aaron has reported 0.8.4 to be OK [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/1425#issuecomment-997006113) (even though his version is self-built), I'm trusting that all is fine and merging this. Thank you both! :+1:

PR #1476, opened 2021-12-20
added .well-known path to Caddy2 example, closes #1442

### LINKED ISSUE #1442: Caddyfile for Caddy2 in Examples doesn't contain configration for .well-known
### Problem

I'm using my own web server (Caddy2 instead of the default Nginx) and following [this concept](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/ae45254d83387b349c4ce4daa142c41c6252bcf3/docs/configuring-playbook-own-webserver.md) in docs.   
After I copied [the example Caddyfile](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/ae45254d83387b349c4ce4daa142c41c6252bcf3/examples/caddy2/Caddyfile) into my `/etc/caddy/Caddyfile`, things worked fine but I access to `https://matrix.DOMAIN/.well-known/matrix/server` it returned 404. 

As said in [configuring-well-known docs](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-well-known.md) but it doesn't exist:

> To make things easy for you to set up, this playbook generates and hosts 2 well-known files on the Matrix domain's server (e.g. `https://matrix.example.com/.well-known/matrix/server` and `https://matrix.example.com/.well-known/matrix/client`), even though this is the wrong place to host them.

### Solution

So the problem is caused by the example Caddyfile because it has no `path` set for well-known. After I added a handle for the path everything seems worked fine.

```shell
matrix.DOMAIN {
  @wellknown {
        path /.well-known/matrix/*
  }

  # ...

  handle @wellknown {
        encode zstd gzip
        root * /matrix/static-files
	header Cache-Control max-age=14400
        header Content-Type application/json
        header Access-Control-Allow-Origin *
        file_server
  }

  # ...
}
```

Also, if I'm also hosting the base domain, here's how to config in the same Caddyfile:

```shell
DOMAIN {
    @wellknown {
          path /.well-known/matrix/*
    }

    handle @wellknown {
          reverse_proxy https://matrix.meta.io {
          	header_up Host {http.reverse_proxy.upstream.hostport}
          }
    }
    # Configration for the base domain goes here
    # handle {
    #    header -Server
    #     encode zstd gzip
    #    reverse_proxy localhost:4020
    # }
}
```

Hope this can help someone who's facing the same problem as mine. Or could someone updates the document and the example that would be better.

altsalt:


spantaleev:
Thank you! 👍

PR #1478, opened 2021-12-21
Twitter Mautrix from mattcen

shreyasajj:


shreyasajj:
I think I fixed the one problem that he seems not have fixed

roles/matrix-bridge-mautrix-twitter/templates/config.yaml.j2
```diff
@@ -124,10 +124,10 @@ bridge:
         invite_own_puppet: true
         # Maximum number of messages to backfill initially.
         # Set to 0 to disable backfilling when creating portal.
-        initial_limit: 0
+        initial_limit: 100
```
spantaleev, line 127: Why `100`?

The default bridge config says `0`: https://github.com/mautrix/twitter/blob/e287a08b35fc36f03ecff4f5931b943644e76b6c/mautrix_twitter/example-config.yaml#L147

shreyasajj: Oh I forgot that I had an open pull request, I was just testing something on my local branch. I will roll it back for the last two  commitment. However it seems a change from `disable_notifications: true` on instagram for backfilling would be nice as a default. Would you like me to keep that in.

spantaleev: We usually try to stick to upstream defaults and only rarely deviate from them, so I guess it makes sense to restore to that.

shreyasajj:
@spantaleev Fixed it

spantaleev:
Thanks for getting this ready! :+1: 

I've updated the changelog to announce this new bridge.

mattcen:
Thanks @shreyasajj and @spantaleev! 😁

shreyasajj:
@mattcen you basically did 95% I did 5% just got it over the finish line

PR #1479, opened 2021-12-21
Update Synapse from 1.49.0 to 1.49.1

FSG-Cat:
This update fixes a sync issue that would cause some users to experience sync issues that could cause initial sync to completely fail.

https://github.com/matrix-org/synapse/releases/tag/v1.49.1


spantaleev:
The `arm64` image still hasn't been published to Docker Hub. It's why I haven't pushed an update yet.

spantaleev:
Now it's OK. Thank you! :+1:

PR #1481, opened 2021-12-21
Update Synapse Version from 1.49.1 to 1.49.2

FSG-Cat:
This upgrade is technically not needed due to 1.49.1 and 1.49.2 being identical with a lone fix to Debian packaging being the only change. 

Still some might want us to be on the absolutely latest version even tho these 2 are practically identical.

ARM64 has yet to be built so this has to wait for that before merge.


spantaleev:
Thanks! 👍

PR #1489, opened 2021-12-27
Add encryption_enabled_by_default_for_room_type

xlejo:
This commit simply add encryption_enabled_by_default_for_room_type
variable.

Signed-off-by: Alejo Diaz <xlejo@protonmail.com>


spantaleev:
One can use `matrix_synapse_configuration_extension_yaml` for setting any setting without the need for explicitly defining a variable. Because of this, we usually reject such PRs, as we'd rather not end up with hundreds of variables unnecessarily.

In this case, however, this seems like an important security-related setting that many may wish to change, so having a dedicated variable for it probably makes sense.

Thank you! :+1:

PR #1490, opened 2021-12-28
Upgrade Etherpad to 1.8.16

pushytoxin:
Extend the config template with the lines added in the past year


spantaleev:
Thanks! :+1:

Issue #1486, opened 2021-12-25
Add matrix-hookshot (formerly matrix-github)


HarHarLinks:
https://github.com/Half-Shot/matrix-hookshot

the author of matrix-appservice-webhooks [suggests that their project might get archived](https://matrix.to/#/!MRDYgvSSrLVHFrWIuX:t2l.io/$1640274154420ocmpT:t2l.io?via=t2l.io&via=matrix.org) and looking at hookshot instead, which sounds like it has better and general webhook support (I need to take a closer look to confirm to what extent). Supposedly hookshot integrates really well with dimension.

The project provides a docker image at https://hub.docker.com/r/halfshot/matrix-hookshot

PR #1491, opened 2021-12-28
switch appservice-webhooks fork

HarHarLinks:
turt2live/matrix-appservice-webhooks -> redoonetworks/matrix-appservice-webhooks

the original turt2live project is almost abandoned and archived at this point. redoonetworks have applied a few essential fixes and provide a docker image.

I have been testing this for 20 days now, works as expected.

in the long run, webhooks should probably migrate to hookshot: #1486


spantaleev:
Thanks for switching us to a more maintained version! :+1: 

Another benefit might be us being able to pin the version (in `matrix_appservice_webhooks_version`) to `v1.0.2-01`. Would you test it out?

HarHarLinks:
would that just be `matrix_appservice_webhooks_version: v1.0.2-01`? its the same as current image, so all we need to test is whether deployment works at all, no?

spantaleev:
Yes, that's all we need to set to pin it. I think it's the same as the current image, so it should work flawlessly. Still, it's nice to have someone confirm that.

PR #1493, opened 2021-12-28
Allow editing of MXID/dispname templates for mautrix-telegram users

SierraKiloBravo:
Added defaults and edited config template for mautrix-telegram, allowing editing of MXID/displayname templates for bridged users.

This is primarily useful for removing the `(Telegram)` suffix from displaynames.

I have added the following config variables (defaults also shown):

```
matrix_mautrix_telegram_username_template: 'telegram_{userid}'
matrix_mautrix_telegram_alias_template: 'telegram_{groupname}'
matrix_mautrix_telegram_displayname_template: '{displayname} (Telegram)'
```

The defaults are the same as the current (hardcoded) config.

I have tested this on my deployment and can confirm that it works as intended.


spantaleev:
Thank you! :+1:

PR #1500, opened 2022-01-02
Remove Jitsi web interface_config.js in favour of custom-interface_config.js

cnvandijk:
As of docker-jitsi-meet stable-6433 [1], `/config/interface_config.js`
is regenerated on every boot. The correct way to modify the interface
config is now via `/config/custom-interface_config.js`, which is
appended to a default copy of `interface_config.js` by
`/etc/cont-init.d/10-config` on every boot of the docker image.

Given that `interface_config.js` is considered deprecated by upstream
(all options will eventually be moved to `config.js`), we also deprecate
the `matrix_jitsi_web_interface_config_*` variables in favour of
`matrix_jitsi_web_custom_interface_config_extension`.

[1] https://github.com/jitsi/docker-jitsi-meet/blob/stable-6433/CHANGELOG.md#stable-6433


cnvandijk:
If we want to maintain backwards compatibility and keep the `matrix_jitsi_web_interface_config_*` variables, we could do something like this in `custom-interface_config.js.j2`:

```
interfaceConfig.LANG_DETECTION = {{ matrix_jitsi_web_interface_config_lang_detection|to_json }};
interfaceConfig.SHOW_JITSI_WATERMARK = {{ -matrix_jitsi_web_interface_config_show_jitsi_watermark|to_json }};
...

{{ matrix_jitsi_web_custom_interface_config_extension }}
```


Let me know if anybody prefers to do this and I will update the PR. IMHO however this feels clunky, and as I wrote in the commit message `interface_config.js` is deprecated by upstream.

spantaleev:
Awesome! Thanks for keeping us up to date! :+1: 

I think you're right that it feels clunky maintaining backward compatibility like that. We'd better keep our part smaller.

PR #1504, opened 2022-01-03
s/plabook/playbook

HarHarLinks:
actual command was `find ./ -type f -exec sed -i -e 's/plabook/playbook/g' {} \;`


spantaleev:
Thanks! :+1:

PR #1507, opened 2022-01-04
Updated: Mautrix Telegram to v0.11.0 & alpine 3.15

GoliathLabs:
Hello,
I have updated mautrix-telegram to version v0.11.0. I also changed the lottieconverter docker image to use the alpine-3.15 tag, since the mautrix-telegram dockerfile currently does this.

Feel free to leave some feedback 😄

### DIFF
--- a/roles/matrix-bridge-mautrix-telegram/defaults/main.yml
+++ b/roles/matrix-bridge-mautrix-telegram/defaults/main.yml
@@ -7,13 +7,13 @@ matrix_telegram_lottieconverter_container_self_build: false
 matrix_telegram_lottieconverter_container_self_build_mask_arch: false
 matrix_telegram_lottieconverter_docker_repo: "https://mau.dev/tulir/lottieconverter.git"
 matrix_telegram_lottieconverter_docker_src_files_path: "{{ matrix_base_data_path }}/lotticonverter/docker-src"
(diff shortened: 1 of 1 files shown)


spantaleev:
Looks like [v0.11.0](https://mau.dev/mautrix/telegram/-/blob/master/CHANGELOG.md#v0110-2021-12-28) changes a few other things, among which getting rid of alembic.

At the very least, we should modify `roles/matrix-bridge-mautrix-telegram/templates/systemd/matrix-mautrix-telegram.service.j2` and get rid of the `alembic`-related `ExecStartPre` call.

This has been discussed in `#matrix-docker-ansible-deploy:devture.com` recently and @chagai95 (unless I'm forgetting) said that the new version doesn't work as expected. Likely because the `lottieconverter` change hadn't been done. Perhaps it would work now.

Could you give it a try and confirm?

chagai95:
I think the issue was related to pgloader

Here is what tulir said about this:

https://matrix.to/#/!FDTDciUzgMWGJcYnwn:maunium.net/$dbJJdAY9I7QorxKgCVUhPTkg1-Pychgs_tY8lyx6mnc?via=matrix.org&via=maunium.net&via=tchncs.de




> I'd guess your database was corrupted by the postgres migration in the ansible playbook, not entirely sure what to do about that yet
> it probably always happens when using tools like pgloader

And there is a branch that tulir made to try fix it. I was meaning to try out but I'm stuck with getting the config right.

https://github.com/mautrix/telegram/commit/b13da925204aac62694600e4ecfeace5efe4002f

PR #1509, opened 2022-01-05
added matrix-client-cinny

### LINKED ISSUE #1212: Add new web client Cinny
Maybe this ticket is more of a placeholder, but still. 

https://cinny.in

https://github.com/ajbura/cinny

rakshazi:
Hi there,

That PR adds matrix-client-cinny role

closes #1212

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1771,6 +1771,8 @@ matrix_client_hydrogen_self_check_validate_certificates: "{{ false if matrix_ssl
 
 matrix_client_cinny_enabled: false
 
+matrix_client_cinny_container_image_self_build: "{{ matrix_architecture != 'amd64' }}"
+
 # Normally, matrix-nginx-proxy is enabled and nginx can reach Cinny over the container network.
 # If matrix-nginx-proxy is not enabled, or you otherwise have a need for it, you can expose
 # the HTTP port to the local host.
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1332,6 +1332,7 @@ matrix_nginx_proxy_proxy_matrix_client_redirect_root_uri_to_domain: "{{ matrix_s
 matrix_nginx_proxy_proxy_matrix_enabled: true
 matrix_nginx_proxy_proxy_element_enabled: "{{ matrix_client_element_enabled }}"
 matrix_nginx_proxy_proxy_hydrogen_enabled: "{{ matrix_client_hydrogen_enabled }}"
+matrix_nginx_proxy_proxy_cinny_enabled: "{{ matrix_client_cinny_enabled }}"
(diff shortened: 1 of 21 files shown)


aaronraimist:
You'll need to put the closes #1212 in the description for it to work

GoliathLabs (approved):

group_vars/matrix_servers
```diff
@@ -1760,6 +1763,31 @@ matrix_client_hydrogen_self_check_validate_certificates: "{{ false if matrix_ssl
 
 
 
+######################################################################
+#
+# matrix-client-cinny
+#
+######################################################################
+
+matrix_client_cinny_enabled: false
```
spantaleev, line 1772: Seems like we need to add this here to make Cinny work on non-`amd64` platforms via self-building:

```yaml
matrix_client_cinny_container_image_self_build: "{{ matrix_architecture != 'amd64' }}"
```

roles/matrix-client-cinny/tasks/validate_config.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Fail if required Cinny settings not defined
+  fail:
+    msg: >
+      You need to define a required configuration setting (`{{ item }}`) to use Cinny.
+  when: "(vars[item] == '' or vars[item] is none) and matrix_client_cinny_container_image_self_build|bool"
```
spantaleev, line 6: Why do we have `and matrix_client_cinny_container_image_self_build|bool` here?

spantaleev (changes requested):
Looks great! Thank you for working on this! :+1: 

I've added some minor comments we should address before merging.

roles/matrix-client-cinny/tasks/validate_config.yml
rakshazi: I used `matrix-client-hydrogen` as template and seems that condition was added because hydrogen can be only self-built, but that's not true for cinny, so removed it

spantaleev:
Thank you! :+1: 

I've updated the changelog to mention this.

Joshndroid:
I just updated my instance and seem to see this... are you seeing this error as well

```
    The task includes an option with an undefined variable. The error was: {{ matrix_client_cinny_docker_image_name_prefix }}ajbura/cinny:{{ matrix_client_cinny_version }}: {{ 'localhost/' if matrix_client_cinny_container_image_self_build }}: the inline if-expression on line 1 evaluated to false and no else section was defined.
  
    The error appears to be in '/home/matrixserver/matrix-docker-ansible-deploy/roles/matrix-client-cinny/tasks/setup_uninstall.yml': line 32, column 3, but may
  
    - name: Ensure Cinny Docker image doesn't exist
      ^ here
```

spantaleev:
Thanks for the report, @Joshndroid! Fixed in 7dfb68b

Joshndroid:
Love your work @spantaleev :)

PR #1514, opened 2022-01-07
matrix-bot-honoroit: fix self-build

rakshazi:


spantaleev:
I think `_container_image_self_build` is better naming than `_container_self_build`.

The playbook currently uses both, unfortunately, but the former is more common:

```
rg container_image_self_build | wc -l
311

rg container_self_build | wc -l
58
```

Issue #1513, opened 2022-01-07
Latest commit broke playbook


MatthiasGrandl:
```
TASK [matrix-bot-honoroit : Ensure honoroit Docker image doesn't exist] ***************************************************************************
fatal: [matrix.bsr.mn]: FAILED! =>
  msg: |-
    The task includes an option with an undefined variable. The error was: {{ matrix_bot_honoroit_docker_image_name_prefix }}honoroit:{{ matrix_bot_honoroit_version }}: {{ 'localhost/' if matrix_bot_honoroit_container_image_self_build else 'registry.gitlab.com/etke.cc/' }}: 'matrix_bot_honoroit_container_image_self_build' is undefined

    The error appears to be in '/Users/matthias/matrix-docker-ansible-deploy/roles/matrix-bot-honoroit/tasks/setup_uninstall.yml': line 33, column 3, but may
    be elsewhere in the file depending on the exact syntax problem.

    The offending line appears to be:


    - name: Ensure honoroit Docker image doesn't exist
      ^ here
```

I even tried disabling honoroit specifically in my vars.yml, but this error persists.

spantaleev:
Fixed by #1514

PR #1515, opened 2022-01-07
self-build consistency

rakshazi:
Hi there,
that PR fixes consistency of self-build vars, to avoid #1513 in future


spantaleev:
Thanks for fixing this consistency trouble! :+1:

PR #1520, opened 2022-01-09
rename variable for matrix-awx, update documentation

GoMatrixHosting:
hopefully this patches things for the variable name change, thanks :+1:


GoMatrixHosting:
apparently this needs to be another variable entirely!

Issue #1518, opened 2022-01-08
matrix-bot-honoroit failed


ghost:
```
TASK [matrix-bot-honoroit : Ensure honoroit image is built] ************************************************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: docker.errors.APIError: 500 Server Error: Internal Server Error ("Cannot locate specified Dockerfile: docker/Dockerfile")
fatal: [matrix.domain.tld]: FAILED! => changed=false 
  msg: 'An unexpected docker error occurred: 500 Server Error: Internal Server Error ("Cannot locate specified Dockerfile: docker/Dockerfile")'
```
Latest commit on ARM64

spantaleev:
Hopefully fixed by 817d6833bfc0. Please let me know if that's not enough!

rdaries:
```  msg: 'Error building localhost/honoroit - code: 1, message: The command ''/bin/sh -c make build'' returned a non-zero code: 1, logs: [''Step 1/11 : FROM registry.gitlab.com/etke.cc/base AS builder'', '''', ''Pulling from etke.cc/base'', ''Digest: sha256:765ab5657cd2d2c98816afc1765777bb08762cc88898976b78ec8609dd188b64'', ''Status: Image is up to date for registry.gitlab.com/etke.cc/base:latest'', '' ---> b0dc7e5127f0'', ''Step 2/11 : WORKDIR /honoroit'', '''', '' ---> Using cache'', '' ---> f6770db95365'', ''Step 3/11 : COPY . .'', '''', '' ---> 808645eeb0ba'', ''Step 4/11 : RUN make build'', '''', " ---> [Warning] The requested image''s platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested", '' ---> Running in 5c16c0d0b3ce'', ''\x1b[91mstandard_init_linux.go:228: exec user process caused: exec format error'', ''\x1b[0m'', ''Removing intermediate container 5c16c0d0b3ce'']'```

my vars:
```
# Honoroit
matrix_bot_honoroit_enabled: true
#matrix_bot_honoroit_container_self_build: true
matrix_bot_honoroit_password: PASS
matrix_bot_honoroit_roomid: "!room:domain.tld"
```

spantaleev:
[...]

PR #1522, opened 2022-01-09
matrix-bot-honoroit: disable self-build by default, update to v0.9.1

etkecc:
Hi there,

To mitigate #1518 and prevent such issues in future, multiarch support was added both to the base image and to the honoroit image.

Both images support arm32, arm64 and amd64 architectures now, so no need for automatic self-building (of course, that's still possible, but default value changed to `false`).

These changes already available for the `latest` image tag, but as it works as expected, pinned release added as well.

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -986,7 +986,7 @@ matrix_bot_honoroit_systemd_required_services_list: |
 # Postgres is the default, except if not using `matrix_postgres` (internal postgres)
 matrix_bot_honoroit_database_engine: "{{ 'postgres' if matrix_postgres_enabled else 'sqlite' }}"
 matrix_bot_honoroit_database_password: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'honoroit.bot.db') | to_uuid }}"
-matrix_bot_honoroit_container_image_self_build: false
+matrix_bot_honoroit_container_image_self_build: "{{ matrix_architecture not in ['amd64', 'arm32', 'arm64'] }}"
 
 ######################################################################
 #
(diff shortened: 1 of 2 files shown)


etkecc:
one more thing: closes #1518

group_vars/matrix_servers
```diff
@@ -986,7 +986,7 @@ matrix_bot_honoroit_systemd_required_services_list: |
 # Postgres is the default, except if not using `matrix_postgres` (internal postgres)
 matrix_bot_honoroit_database_engine: "{{ 'postgres' if matrix_postgres_enabled else 'sqlite' }}"
 matrix_bot_honoroit_database_password: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'honoroit.bot.db') | to_uuid }}"
-matrix_bot_honoroit_container_image_self_build: "{{ matrix_architecture != 'amd64' }}"
+matrix_bot_honoroit_container_image_self_build: false
```
spantaleev, line 989: Probably best to change this to:

```yaml
matrix_bot_honoroit_container_image_self_build: "{{ matrix_architecture not in ['amd64', 'arm32', 'arm64'] }}"
```

Reasons:

- in case some other architecture comes up in the future (well, unlikely, but..), self-building will kick in. Well, it will probably fail as described in #1518, but..
- doing that (instead of hardcoding it to `false`) seems more consistent with what we do for other components in `group_vars/matrix_servers`.

Alternatively, if we'd be setting this to `false`, we may as well not define it in `group_vars/matrix_servers` at all. The default value is `false` anyway (coming from the role's `defaults/main.yml` file).

etkecc:
seems like CI lint is broken, but local lint doesn't show any issues

spantaleev:
We should get rid of that (well, or make it not fragile). Not sure why it's acting up when we've supposedly pinned it to a specific commit and versionz

etkecc:
The problem is not with ansible-lint itself, but with one of subpackages. Seems like it has dependencies defined without pin to minor and/or patch versions:

```
Collecting enrich>=1.2.6
  Downloading enrich-1.2.6-py3-none-any.whl (8.6 kB)
```

Could you merge the PR, please?

spantaleev:
Merged! Thanks! :+1:

PR #1531, opened 2022-01-14
Added writable /tmp dir for Mautrix Signal Bridge 

IUCCA:
Voice Messages work only if the bridge has an writable /tmp dir. The bridge needs the /tmp dir for transcoding with ffmpeg


spantaleev:
Thanks for fixing this! 👍

PR #1548, opened 2022-01-15
Updated: webhooks image tag to v1.0.3-01

### LINKED ISSUE #1533: [TASK] Update matrix-appservice-webhooks to v1.0.3-01
Hello,
I opened this issue because the version of matrix-appservice-webhooks currently in use is outdated. It might help to keep an eye on the update and discuss below if something needs to be done before creating a PR for the new update.

### Changelog
- Add ARM64 to build

GoliathLabs:
Updates the appservice-webhooks image tag to v1.0.3-01 which adds arm64 builds.
Closes #1533


spantaleev:
Thanks! :+1: 

Looks like the [v1.0.3-01](https://hub.docker.com/r/redoonetworks/matrix-appservice-webhooks/tags?name=v1.0.3-01) image is amd64-only on Docker Hub, so "adds arm64 builds" probably means it can be self-built on arm64, I suppose.

PR #1555, opened 2022-01-16
Added Instagram version tag instead of latest

GoliathLabs:
Hello,
I have added the version tag for this bridge as there has been a new release yesterday. This ensures that we don't break things in the future by using the latest tag, but can check if anything on our site needs to be changed.

https://github.com/mautrix/instagram/releases/tag/v0.1.2

### DIFF
--- a/roles/matrix-bridge-mautrix-instagram/defaults/main.yml
+++ b/roles/matrix-bridge-mautrix-instagram/defaults/main.yml
@@ -6,7 +6,7 @@ matrix_mautrix_instagram_enabled: true
 matrix_mautrix_instagram_container_image_self_build: false
 matrix_mautrix_instagram_container_image_self_build_repo: "https://github.com/mautrix/instagram.git"
 
-matrix_mautrix_instagram_version: latest
+matrix_mautrix_instagram_version: v0.1.2
 # See: https://mau.dev/tulir/mautrix-instagram/container_registry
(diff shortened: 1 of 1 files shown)


spantaleev:
Great! Thanks! :+1:

GoliathLabs:
@spantaleev we have a few other images using the latest tag (e.g. the googlechat one). Should we change over to tagged versions as well? https://github.com/mautrix/googlechat/releases/tag/v0.3.0

spantaleev:
It's sometimes dangerous to switch an image from `:latest` to a tagged release, especially when that image hasn't had a release recently.

.. because it could be that the `:latest` release is potentially much newer than the tagged release and thus using a newer database schema, etc.

This not only depends on when the release has been, but also on when users will actually pull and update their Matrix server though, so we can't really control how to do it safely.

Ideally, we'd switch to a tagged release as soon as new release comes out, so that the potential damage is smallest.

If you'd like, please go ahead and switch us to a tagged release where it seems safe.. or well, we can rip the bandaid and just do it everywhere.

GoliathLabs:
@spantaleev I have also thought about this . E.g. after the release of googlechat v0.3.0 there was a change in the structure of the database used (new primary key) and probably shouldn't update the image tag right now. Instead we should closely monitor the releases of the components (appservices/bridges) used and add the tagged releases as soon as there is a new version.

spantaleev:
Sounds good!

PR #1556, opened 2022-01-16
Updated Honoroit with fallback reply-to mode

etkecc:
Honoroit updated to v0.9.2

By default Element doesn't support threads yet. To mitigate that, fallback reply-to mode was added, so now you can interract with honoroit just replying to previous messages, to avoid [such frustrations](https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/1518#issuecomment-1008263296). That method considered as temporary workaround, so performance will be pretty poor on large reply-to chains, but that's better than nothing.


spantaleev:
Great! Why WIP?

etkecc:
[CI job is in progress](https://gitlab.com/etke.cc/honoroit/-/jobs/1980121490), so there is no actual docker image yet.
Will mark as ready when it will be done

etkecc:
@spantaleev merge it, please

spantaleev:
Merged! Thanks! :+1:

PR #1557, opened 2022-01-16
Updated: telegram to v0.11.1

GoliathLabs:
Hi,
I've updated mautrix-telegram to v0.11.1 which seems to fix the problem regarding the database corruption when upgrading to v0.11.0 (`Fixed migrating pre-v0.11 legacy databases if the database schema had been corrupted (e.g. by using 3rd party tools for SQLite -> Postgres migration`).

I've also removed the deduplication settings from the config as they have been removed in the example config upstream as well. Lottieconverter has been updated to 3.15 so it matches the requirement of telegram.


GoliathLabs:
I've also removed the `alembic`-related `ExecStartPre` call in the system unit template

spantaleev:
Thank you! :+1:

PR #1558, opened 2022-01-17
Added: mautrix-signal 0.2.2 & signald 0.16.1

GoliathLabs:
Hi,
mautrix-signal had an update two days ago. There has been only a small commit since then which doesn't break things for latest tag users.

https://github.com/mautrix/signal/releases/tag/v0.2.2
https://gitlab.com/signald/signald/-/releases/0.16.1


spantaleev:
Nice! Thanks! :+1:

PR #1561, opened 2022-01-17
Updated: Certbot to v1.22.0

GoliathLabs:
I've updated certbot to v1.22.0. Everything should behave as usual.


GoliathLabs:
I've also updated the nginx worker processes to auto, so the available cpu cores are used and the load of nginx is more balanced

spantaleev:
Thanks! :+1:

PR #1563, opened 2022-01-17
matrix_mailer default false

sakkiii:
- disable default (already an optional component)
- User must configure matrix_mailer_enabled along with other required parameters

### DIFF
--- a/roles/matrix-mailer/defaults/main.yml
+++ b/roles/matrix-mailer/defaults/main.yml
@@ -1,4 +1,4 @@
-matrix_mailer_enabled: true
+matrix_mailer_enabled: false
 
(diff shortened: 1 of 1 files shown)


spantaleev:
All roles intentionally define `enabled` as `true`. This way, people who use these roles as part of other playbooks only need to include the role and it would work right away.

The playbook, as a distribution, enables/disables components to fit its needs. This happens via `group_vars/matrix_servers`

etkecc:
The ma1sd (default component) depends on mailer, so it's pretty dangerous change.

You always can disable it inside your vars.yml (and ma1sd, too)

GoliathLabs:
@spantaleev this can be closed I guess?

PR #1505, opened 2022-01-03
add matrix-hookshot

HarHarLinks:
closes #1486 

I really want to try it out, why not go all the way and write the complete role? Why do something manually in 10 minutes when you can spend 10 <del>hours</del> days automating it?

This started based off the heisenbridge role but quickly became frankensteined together from different parts of different things. I tried to stick to the style of the rest of the app and things, but please feel free to comment, review, edit, contribute, or take over this effort.

todo:

- [x] actually deploy a test
  - [ ] I don't have a spare server to test stuff as significant as this in all its configurations, so you're welcome to give your feedback :)
  - [ ] github works
    - [x] personal notifications
    - [ ] repo notifications (webhooks)
    - [ ] repo bridging (issues, discussions as rooms, etc)
  - [ ] gitlab works
  - [ ] jira works
    - [ ] looking for testers
  - [ ] figma works
    - [ ] looking for testers
  - [x] generic works
  - [ ] `matrix_nginx_proxy_enabled = false`
    - [ ] looking for testers
  - [ ] prometheus/grafana
    - [ ] builtin: looking for testers
    - [ ] external
- [x] metrics in builtin prom/grafana?
- [x] add password protection for metrics if enabled publicly and add docs in https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-prometheus-grafana.md#collecting-metrics-to-an-external-prometheus-server
- [x] mention dimension provisioning integration must be manually configured
- [ ] <del>uninstall cleanup</del> (see review below)
- [x] <del>default to everything on a single port while retaining the ability to split things up?</del> split vhost by endpoints for each listener as supplied in the updated docs
- [x] does all the jinja work?
- [x] fail if config validation fails https://github.com/Half-Shot/matrix-hookshot/pull/125 https://github.com/Half-Shot/matrix-hookshot/pull/150/commits/e603b22db85f49f7ce3995cea024713ca4a46641
- [x] add permissions config https://github.com/Half-Shot/matrix-hookshot/pull/167/
- [x] handle github private key more properly, see discussion below
- [x] install openssl https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/b33ea48803b0414dced910ebcec520c1eaaa4c99 and use `matrix_host_command_openssl`

out of scope:
- self-building and testing the image for/on non-amd64 platforms, as I have little experience and no such platform to test

PR #1570, opened 2022-01-20
only enable openssl if necessary, closes #1510 

### LINKED ISSUE #1510: Ensure OpenSSL installed and use matrix_host_command_openssl
I need openssl for #1505 and searched the repo for it and found these:

1. [matrix-nginx-proxy](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/b1b4ba501fdfaa434dac8a981cec9ac0311ca4ef/roles/matrix-nginx-proxy/tasks/ssl/setup_ssl_self_signed.yml) has a task to install it if using self-signed certs
2. some jitsi script wants to use it
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/30339cd313717bf18471f39ee27db76ec3ba2ddf/inventory/scripts/jitsi-generate-passwords.sh#L6
3. there is a global variable with the path to it
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/61391647e9681968f65096223507c6645ad03acb/roles/matrix-base/defaults/main.yml#L82
4. which is used by matrix-bridge-appservice-irc
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/1ab507349c752042d26def3e95884f6df8886b74/roles/matrix-bridge-appservice-irc/tasks/setup_install.yml#L100

What needs to be done:
1. extract the installation from (1) so that all roles depending on it can install it
2. jitsi (2) should use the variable (3)
3. use both of these features for #1505

altsalt:
Changes not applied to [hookshot PR](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/1505).

### DIFF
--- a/roles/matrix-base/tasks/util/ensure_openssl_installed.yml
+++ b/roles/matrix-base/tasks/util/ensure_openssl_installed.yml
@@ -8,14 +8,14 @@
   when: ansible_distribution == 'CentOS'
 
 # This is for both Debian and Raspbian
-- name: Ensure fuse installed (Debian/Raspbian)
+- name: Ensure openssl installed (Debian/Raspbian)
   apt:
     name:
       - openssl
     state: latest
   when: ansible_os_family == 'Debian'
 
-- name: Ensure fuse installed (Archlinux)
+- name: Ensure openssl installed (Archlinux)
   pacman:
     name:
       - openssl
--- a/roles/matrix-base/tasks/util/ensure_openssl_installed.yml
+++ b/roles/matrix-base/tasks/util/ensure_openssl_installed.yml
@@ -1,0 +1,23 @@
+
+# This is for both CentOS 7 and 8
+- name: Ensure openssl installed (CentOS)
+  yum:
+    name:
+      - openssl
+    state: latest
+  when: ansible_distribution == 'CentOS'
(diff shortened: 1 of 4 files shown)


spantaleev:
Nice! Thanks! :+1:

HarHarLinks:
shouldn't this also set `matrix_host_command_openssl`?

altsalt:
Where? Sorry if something was missed, I was mostly following the suggested PR.

HarHarLinks:
currently it gets set by
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5e2f4564bbfc4572832068f564a858f14a1406a4/roles/matrix-base/defaults/main.yml#L97

currently it gets used by
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/b33ea48803b0414dced910ebcec520c1eaaa4c99/roles/matrix-bridge-appservice-irc/tasks/setup_install.yml#L102

If it exists, shouldn't everyone use it via the variable? Though I'm not firm enough with ansible to tell what advantages using `"/usr/bin/env openssl"` might have over a plain call. I suppose the idea is you can switch out the openssl binary used by overriding the variable?

Issue #1540, opened 2022-01-14
[TASK] Update prometheus postgres exporter to v0.10.1


GoliathLabs:
Hello,
I opened this issue because the version of the postgres-exporter currently in use is outdated. It might help to keep an eye on the update and discuss below if something needs to be done before creating a PR for the new update.

### Changelog
- https://github.com/prometheus-community/postgres_exporter/releases/tag/v0.10.1

PR #1571, opened 2022-01-20
Updated: prom-postgres-exporter to v0.10.1

GoliathLabs:
Hi,
I've updated the postgres exporter to v0.10.1. As this update is only a minor update which only changes the logging behavior, I do not expect any breakage.
`[BUGFIX] Fix broken log-level for values other than debug. `

Closes https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/1540


spantaleev:
Thanks! :+1:

Issue #1517, opened 2022-01-08
review service start order


HarHarLinks:
this is kind of pedantic as it generally works out alright, but might potentially speed up upgrade-restarts.

starting the playbook, some services fail to start a couple times before eventually working. the most obvious one is nginx: it depends on all other containers to exist that are mentioned in its config, e.g.:

```
2022/01/08 17:11:01 [emerg] 1#1: host not found in upstream "matrix-prometheus-postgres-exporter" in /etc/nginx/conf.d/matrix-grafana.conf:63
```

Other services depend on synapse and might start it earlier than the playbook order implies.

Services should obtain more After=, Requires=, etc config conditionally based on what is enabled and what isnt.

Is there a tool that can graphically render service relations from a bunch of .service files? Would be helpful.

spantaleev:
We specify some wanted services in `matrix_nginx_proxy_systemd_wanted_services_list` (which go to `Wanted=` in the `matrix-nginx-proxy.service` file) in `group_vars/matrix_servers`: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/4e4fb98a65474fd058c61a838db6ac312a09e7df/group_vars/matrix_servers#L1462-L1471

We might define additional *wanted* services there.

----------

We have `matrix_nginx_proxy_systemd_required_services_list` as well, but we avoid hardcoding such hard-dependencies, because.. If `matrix-nginx-proxy` lists some random service (say `matrix-client-element`) as required for `matrix-nginx-proxy` to run, then restarting `matrix-client-element` for whatever reason (it being restarted manually or dying and getting restarted) has the negative side-effect of bringing down `matrix-nginx-proxy` as well.

If some random service (say `matrix-dimension`) fails to start for whatever reason (misconfiguration, container image bug, etc.), we also don't want that to prevent `matrix-nginx-proxy` from starting.

We don't want random services failing to start or being restarted to bring down **everything else**.

This is why we only use a "wanted services" list.

--------

nginx insisting on resolving the DNS names for all defined upstreams when it starts is kind of bad, but.. from what I remember, there's no way around it, except for using their "nginx plus" offering.

spantaleev:
Adding this to the `matrix_nginx_proxy_systemd_wanted_services_list` list in `group_vars/matrix_servers` may solve this particular problem:

```diff
matrix_nginx_proxy_systemd_wanted_services_list: |
  {{
[...]

PR #1573, opened 2022-01-22
add etherpad to nginx wanted services

HarHarLinks:
it's required if enabled by the dimension config here:
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/6eb8fb83925af055483ca797e31b5d803135e61f/roles/matrix-etherpad/tasks/init.yml#L42-L49

see also #1517


spantaleev:
Seems like it makes sense to do this, yes.

> should it only be added if both etherpad and dimension are enabled?

Makes sense!

spantaleev:
Great! Thanks! :+1:

PR #1575, opened 2022-01-23
Add missing slash in url scheme for mediaUrl

Thracky:
The scheme portion of the URL for `matrix_mx_puppet_discord_bridge_mediaUrl` in `main.yml` for matrix-bridge-mx-puppet-discord is missing a slash and causes messages to fail when sending to discord if the matrix user has a profile picture configured.


spantaleev:
Interestingly, this was not causing such an issue for me. I've got a profile photo both on the Matrix side and on the Discord (bot) side, and message sending worked.

Still, something else might have been failing, so it's good to have it fixed.

Good catch! Thank you! :+1:

Thracky:
Yeah I didn't dig *super* deep on this but the Discord API was complaining about the scheme being invalid. I imagine it just depended on the method being used by the bot to send to Discord. Regardless, you're welcome for the fix 🙂

PR #1576, opened 2022-01-24
Update Synapse from 1.50.1 to 1.50.2

FSG-Cat:
Fixes a issue with room version 1.

Merging has to wait until ARM images are built ofc.


spantaleev:
Thanks! :+1:

PR #1580, opened 2022-01-28
feat(mx-puppet-slack): add support for OAuth client ID/secret

jimeh:
The OAuth credentials method seems to be the only viable way to
configure the mx-puppet-bridge now. Legacy tokens can no longer be
created, and the other methods (xoxs and xoxc tokens) come with warnings
about them being against Slack's terms of service.


spantaleev:
`matrix_mx_puppet_slack_oauth_client_id` and `matrix_mx_puppet_slack_oauth_client_secret` are not yet strictly required?

If they are now or at some point, we should add validation to `tasks/validate_config.yaml`.

-----

Thank you! :+1:

jimeh:
They're not strictly required no, as you could have a legacy token from before they stopped letting you generate new ones. Or you could use a xoxc or xoxs token if you so choose, but as both are against Slack's ToS, it's likely not a good idea.

PR #1581, opened 2022-01-29
Update honoroit (v0.9.3 -> v0.9.4)

etkecc:
New version brings in-memory cache that significantly decreases amount of requests to resolve threads relations


spantaleev:
Nice! Thank you! :+1:

PR #1590, opened 2022-02-01
Fixing 404 error in links to main.yml

downeymj:
Fixes a few small 404 errors in the docs due to a missing leading slash.


spantaleev:
Thanks! We usually use relative paths like `../roles`, but absolute also works, I guess. I just hope using an absolute path is not troublesome for some Markdown renderers

PR #1595, opened 2022-02-02
Fixed not being able to start a jitsi conference on webinterface or element

aleld:
I am accessing my jitsi domain via webbrowser and in the preview screen everything seems fine, but when I click "join conference" nothing happens. After a few seconds I get an error saying "your connection was terminated. A new attempt will be started in 30 seconds".

Fix
https://community.jitsi.org/t/jitsi-meet-in-docker-container-websocket-error/88703/15

However I have no idea what this is actually doing. Additionally the last comment pointed out, that one should set 

ENABLE_SCTP=1
ENABLE_COLIBRI_WEBSOCKET=0
ENABLE_XMPP_WEBSOCKET=0

for chromium 97.0.4692.71 web browser in Ubuntu 18.04 to work.

So someone should double check this before merging.


aleld:
The deeper issue was that my config was running behind a nginx reverse proxy and not a bug in the playbook.

PR #1599, opened 2022-02-05
Updated: Dendrite to v0.6.2

GoliathLabs:
Updated: Dendrite to v0.6.2

### DIFF
--- a/roles/matrix-dendrite/defaults/main.yml
+++ b/roles/matrix-dendrite/defaults/main.yml
@@ -5,7 +5,7 @@ matrix_dendrite_enabled: true
 
 matrix_dendrite_docker_image: "{{ matrix_dendrite_docker_image_name_prefix }}matrixdotorg/dendrite-monolith:{{ matrix_dendrite_docker_image_tag }}"
 matrix_dendrite_docker_image_name_prefix: "docker.io/"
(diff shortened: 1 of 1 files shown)


spantaleev:
Looks like Dendrite switched from Kafka / [Naffka](https://github.com/matrix-org/naffka) to [NATS JetStreat](https://docs.nats.io/nats-concepts/jetstream) between 0.5 and 0.6.x (see [this commit](https://github.com/matrix-org/dendrite/commit/161f14517669410d3e8207dc41eea5c9695f7e17)).

This also affects the Dendrite configuration (see [here](https://github.com/matrix-org/dendrite/commit/161f14517669410d3e8207dc41eea5c9695f7e17#diff-17c70e3d06e6f8e2d2c69a7aca56da863891eed8f71ba58058737362cc1bda03)).

Have you tested that 0.6 works with our configuration?

Even if it does work, it's probably better to upgrade to the 0.6-way of doing things.

GoliathLabs:
You're right, I'll change that accordingly. Sadly I currently run with synapse and note dendrite. So we might need somebody else to test the changes though

spantaleev:
More details about upgrading here: https://github.com/matrix-org/dendrite/pull/1866#issuecomment-1006703519

PR #1602, opened 2022-02-05
Updated: jitsi to stable-6865

GoliathLabs:
Updated: jitsi to stable-6865


spantaleev:
Has this been tested? No configuration upgrades between these Jitsi versions?

PR #1614, opened 2022-02-09
add basic fedora support

Wunderharke:
7330992b2098db3d96d12d1259cd628effac6789 fixed the major_version compare and made Fedora deployments to use setup_redhat8.yml. This however broke the script on fedora as there is no EPEL package. This commit adds very basic fedora support by:
1. Separates task into own file (for RedHat OS with major version > 30)
2. removes EPEL task for fedora
3. replaces docker-ce repo details

Not sure whether docker-py is necessary for fedora.


spantaleev:
I removed some misleading `with_items` statements and merged this. Thanks! :+1: 

If you think the playbook works well on Fedora now, we can also update `docs/prerequisites.md` and mention it.

----

> Not sure whether docker-py is necessary for fedora.

You can try yum removing it on your system, removing it from `roles/matrix-base/tasks/server_base/setup_fedora.yml` and re-running the playbook. If nothing breaks, it's likely not necessary (well, or you're not hitting any tasks that require it).

As per the [`docker_image_module` requirements](https://docs.ansible.com/ansible/latest/collections/community/docker/docker_image_module.html#requirements), it seems necessary:

> Docker SDK for Python: Please note that the [docker-py](https://pypi.org/project/docker-py/) Python module has been superseded by [docker](https://pypi.org/project/docker/) (see [here](https://github.com/docker/docker-py/issues/1310) for details). For Python 2.6, docker-py must be used. Otherwise, it is recommended to install the docker Python module. Note that both modules should not be installed at the same time. Also note that when both modules are installed and one of them is uninstalled, the other might no longer function and a reinstall of it is required.

Perhaps using the `docker` Python module instead of `docker-py` is better.

I think you may be able to:
- stop installing `python3-pip`
- stop installing `docker-py` via `pip`
- install `python3-docker` using `yum` instead

To clean up your installation, you probably need to do `pip remove docker-py` and `dnf remove python3-pip`.

PR #1617, opened 2022-02-11
Github action to lint with ansible-later

skepticalwaves:


skepticalwaves:
That's a big list of failures. :-/

Will see how much effort it is to clean these up.

GoliathLabs:
@skepticalwaves the good thing is that most of these errors are repetitive and can be fixed with a more or less simple regex

PR #1618, opened 2022-02-11
Add missing endpoints for SSL certificates

ZzMzaw:
While working on support for Let's Encrypt DNS challenges, I just noticed that some of the endpoints the playbook retrieves certificates for are not listed in the doc.

I was not able to figure out any expected contribution format so feel free to tell me which one to use if I missed it.

Thanks in advance.


spantaleev:
Thank you! :+1:

PR #1622, opened 2022-02-11
Update Hydrogen from v0.2.25 to v0.2.26

FSG-Cat:
This PR should be able to merge essentially instantly due to that all platforms self build hydrogen therefore we do not need to worry about docker image delay plus the release is 7 hours old at time of writing.


spantaleev:
Thanks! :+1:

PR #1630, opened 2022-02-16
Update faq.md

iambeingtracked:
It had an extra dot, which resulted in a question mark and a dot after each other


spantaleev:
Thanks! :+1:

PR #1631, opened 2022-02-16
Postgres Minor Updates (14.2, 13.6, 12.10, 11.15, 10.20)

aaronraimist:
https://www.postgresql.org/about/news/postgresql-142-136-1210-1115-and-1020-released-2402/


spantaleev:
Thank you! :+1:

PR #1634, opened 2022-02-16
Changed to: mautrix-whatsapp v0.2.4 tag

### LINKED ISSUE #1633: WhatsApp bridge broken after upgrade
#1559 breaks WhatsApp bridge when upgrading the playbook services.

See: https://github.com/mautrix/whatsapp/issues/440

pucilpet:
This fixes #1633 (at least for now). Probably should figure out a better solution or mention this in the changelog as a breaking change.


spantaleev:
Thank you! :+1: 

I believe that what's described in #1633 will remain a problem, for as long as people stay on on an old version of the playbook (which uses the `:latest` tag) and then migrate to whatever we've pinned here. It's rare though -- only if they keep re-running the playbook and thus re-pulling `:latest`.

PR #1635, opened 2022-02-16
Update hookshot to respect protocol and custom ports

HarHarLinks:
ref: https://github.com/Half-Shot/matrix-hookshot/issues/206

kindly requesting for @Half-Shot to glance over this before merging

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -695,6 +695,10 @@ matrix_hookshot_provisioning_enabled: "{{ matrix_hookshot_provisioning_secret an
 matrix_hookshot_proxy_metrics: "{{ matrix_nginx_proxy_proxy_synapse_metrics }}"
 matrix_hookshot_proxy_metrics_basic_auth_enabled: "{{ matrix_nginx_proxy_proxy_synapse_metrics_basic_auth_enabled }}"
 
+matrix_hookshot_generic_urlprefix_port_enabled: "{{ matrix_nginx_proxy_container_https_host_bind_port == 443 if matrix_nginx_proxy_https_enabled else matrix_nginx_proxy_container_https_host_bind_port == 80 }}"
+matrix_hookshot_generic_urlprefix_port: ":{{ matrix_nginx_proxy_container_https_host_bind_port if matrix_nginx_proxy_https_enabled else matrix_nginx_proxy_container_http_host_bind_port }}"
+matrix_hookshot_generic_urlprefix: "http{{ 's' if matrix_nginx_proxy_https_enabled }}://{{ matrix_server_fqn_matrix }}{{ matrix_hookshot_generic_urlprefix_port if matrix_hookshot_generic_urlprefix_port_enabled }}{{ matrix_hookshot_generic_endpoint }}"
+
 ######################################################################
 #
 # /matrix-bridge-hookshot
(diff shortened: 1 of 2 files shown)


group_vars/matrix_servers
```diff
@@ -695,6 +695,10 @@ matrix_hookshot_provisioning_enabled: "{{ matrix_hookshot_provisioning_secret an
 matrix_hookshot_proxy_metrics: "{{ matrix_nginx_proxy_proxy_synapse_metrics }}"
 matrix_hookshot_proxy_metrics_basic_auth_enabled: "{{ matrix_nginx_proxy_proxy_synapse_metrics_basic_auth_enabled }}"
 
+matrix_hookshot_generic_urlprefix_port_enabled: "{{ matrix_nginx_proxy_container_https_host_bind_port == 443 if matrix_nginx_proxy_https_enabled else matrix_nginx_proxy_container_https_host_bind_port == 80 }}"
+matrix_hookshot_generic_urlprefix_port: ":{{ matrix_nginx_proxy_container_https_host_bind_port if matrix_nginx_proxy_https_enabled else matrix_nginx_proxy_container_http_host_bind_port }}"
+matrix_hookshot_generic_urlprefix: "http{{ 's' if matrix_nginx_proxy_https_enabled }}://{{ matrix_server_fqn_matrix }}{{ matrix_hookshot_generic_urlprefix_port if matrix_hookshot_generic_urlprefix_port_enabled }}{{ matrix_hookshot_generic_endpoint }}"
```
Procuria, line 700: Seems as there is an `else` missing. Playbook throws following error 
````
fatal: [matrix.computerlie.be]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: {{ matrix_hookshot_configuration_yaml|from_yaml|combine(matrix_hookshot_configuration_extension, recursive=True) }}: {{ lookup('template', 'templates/config.yml.j2') }}: http{{ 's' if matrix_nginx_proxy_https_enabled }}://{{ matrix_server_fqn_matrix }}{{ matrix_hookshot_generic_urlprefix_port if matrix_hookshot_generic_urlprefix_port_enabled }}{{ matrix_hookshot_generic_endpoint }}: the inline if-expression on line 1 evaluated to false and no else section was defined.
````

Comment (HarHarLinks) [reply group_vars/matrix_servers:700]: Interesting, I don't get this error. Can you try the following?
```suggestion
matrix_hookshot_generic_urlprefix: "http{{ 's' if matrix_nginx_proxy_https_enabled else '' }}://{{ matrix_server_fqn_matrix }}{{ matrix_hookshot_generic_urlprefix_port if matrix_hookshot_generic_urlprefix_port_enabled }}{{ matrix_hookshot_generic_endpoint }}"
```

Are you using a different ansible version than I am?
ansible-playbook [core 2.12.2]
python version = 3.10.2 (main, Jan 15 2022, 19:56:27) [GCC 11.1.0]
jinja version = 3.0.3

spantaleev: There's one more `if` statement without an `else` on that line: `{{ matrix_hookshot_generic_urlprefix_port if matrix_hookshot_generic_urlprefix_port_enabled }}`.

Procuria: @HarHarLinks: I have 
ansible-playbook [core 2.12.2]
python version = 3.8.10 (default, Nov 26 2021, 20:14:08) [GCC 9.3.0]
jinja version = 2.10.1

So partly pretty outdated stuff on my machine.

HarHarLinks: @Procuria indeed, probably a difference in our jinja versions. either way since it's desirable to make this work with any jinja version reasonable in the wild, can you confirm the suggested change in my comment above works for you?

Procuria: @HarHarLinks once again, problem solved - it works now. This is also very awesome, because without this fix only a **http** address is displayed. If you just use this displayed address in a **https** environment a POST becomes a GET and the message body is then empty.

spantaleev:
Thanks! :+1:

PR #1640, opened 2022-02-19
Update for changing the federation port into the documentation

AnonyPla:
This is a proposed to add to the documentation information about how to change the Synapse federation port using the ansible playbook in order to allow  use with third party services such as CDNs not accepting port 8448.


spantaleev:
Thank you for documenting this! :+1:

PR #1651, opened 2022-02-22
Updated: jitsi to stable-6865

GoliathLabs:
It looks like we don't need to change anything on the configurations on our side. But I would be happy if someone could double check this


spantaleev:
It would be nice if someone who's using Jitsi could give this a test.

GoliathLabs:
@spantaleev can you assign a "help wanted" label or similar to this PR? That way others can better participate in the review of this PR.

etkecc:
It works!
Tested on etke.cc server.

spantaleev:
Thanks for confirming! Merged! :+1:

PR #1655, opened 2022-02-24
updated honoroit 0.9.4 -> 0.9.5

etkecc:
Hi there,

new version of Honoroit with following changes:
* changed cache algorithm to LRU
* added cache size config var
* deeper sentry integration


spantaleev:
Thanks! :+1:

PR #1660, opened 2022-02-26
Optimize signal bridge startup order

HarHarLinks:
bridge requires the daemon, so start it first


spantaleev:
Looks good! Thanks! :+1:

PR #1667, opened 2022-03-02
Add support for obtain ECDSA keys

xlejo:


spantaleev:
Thank you for making the playbook better! :bow: 

Maybe a `matrix_ssl_lets_encrypt_key_type` variable with a default value of `rsa` is better way to do this compared to `matrix_ssl_lets_encrypt_use_ecdsa_keys`?

That way:

- we don't need to have all these `if` statements. We can just use `--key-type={{ matrix_ssl_lets_encrypt_key_type }}`
- existing installations and new installations remain on `rsa` by default
- people can still easily change to ECDSA by specifying `matrix_ssl_lets_encrypt_key_type: ecdsa` in their `vars.yml` file

----

Linking to `https://eff-certbot.readthedocs.io/en/stable/using.html#using-ecdsa-keys` from `docs/configuring-playbook-ssl-certificates.md` may also be beneficial. This link has already been mentioned in `defaults/main.yml`, but that's less discoverable for people reading the docs.

roles/matrix-nginx-proxy/tasks/ssl/main.yml
```diff
@@ -5,6 +5,11 @@
     msg: "The `matrix_ssl_retrieval_method` variable contains an unsupported value"
   when: "matrix_ssl_retrieval_method not in ['lets-encrypt', 'self-signed', 'manually-managed', 'none']"
 
+- name: Fail if using unsupported private key type
+  fail:
+    msg: "The `matrix_ssl_lets_encrypt_key_type` variable contains an unsupported value"
+  when: "matrix_ssl_lets_encrypt_key_type not in ['rsa', 'ecdsa']"
+
```
spantaleev, line 12: I was also thinking we should have something like this. It's great that you had the same idea!

spantaleev:
Looks good! Thanks! :+1:

PR #1669, opened 2022-03-02
avoid namespace collisions between webhook bridges

HarHarLinks:
I have this deployed and my appservice-webhooks webhooks keep working as before, it does not even create new ones, so it appears this upgrade path is clear.


spantaleev:
Thanks! :+1: 

It's nice to have bridges that don't clash with one another by default.

PR #1672, opened 2022-03-03
Fix empty body for GET requests in hookshot-bridge 

Procuria:
add `$is_args$args` to proxy pass URL params in GET requests 
following the NGINX docs https://nginx.org/en/docs/http/ngx_http_core_module.html#var_args 

based on discussions in https://github.com/matrix-org/matrix-hookshot/issues/208


luismanson:
Confirmed it is working. Thanks

spantaleev:
Thanks! :+1:

PR #1674, opened 2022-03-04
fix hookshot provisioning url in nginx

luismanson:
Hello, this PR should fix Hookshot container not receiving API calls with the right path because nginx proxy was stripping the /v1 from it.

Thanks

### DIFF
--- a/roles/matrix-bridge-hookshot/tasks/init.yml
+++ b/roles/matrix-bridge-hookshot/tasks/init.yml
@@ -55,10 +55,10 @@
               {# Use the embedded DNS resolver in Docker containers to discover the service #}
               resolver 127.0.0.11 valid=5s;
               set $backend "{{ matrix_hookshot_container_url }}:{{ matrix_hookshot_provisioning_port }}";
-              proxy_pass http://$backend/$1;
+              proxy_pass http://$backend/v1/$1;
             {% else %}
               {# Generic configuration for use outside of our container setup #}
-              proxy_pass http://127.0.0.1:{{ matrix_hookshot_provisioning_port }}/$1;
+              proxy_pass http://127.0.0.1:{{ matrix_hookshot_provisioning_port }}/v1/$1;
             {% endif %}
             proxy_set_header Host $host;
           }
--- a/roles/matrix-bridge-hookshot/tasks/init.yml
+++ b/roles/matrix-bridge-hookshot/tasks/init.yml
@@ -55,10 +55,10 @@
               {# Use the embedded DNS resolver in Docker containers to discover the service #}
               resolver 127.0.0.11 valid=5s;
               set $backend "{{ matrix_hookshot_container_url }}:{{ matrix_hookshot_provisioning_port }}";
-              proxy_pass http://$backend/v1/$1;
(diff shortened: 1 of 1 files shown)


roles/matrix-bridge-hookshot/tasks/init.yml
```diff
@@ -55,10 +55,10 @@
               {# Use the embedded DNS resolver in Docker containers to discover the service #}
               resolver 127.0.0.11 valid=5s;
               set $backend "{{ matrix_hookshot_container_url }}:{{ matrix_hookshot_provisioning_port }}";
-              proxy_pass http://$backend/$1;
+              proxy_pass http://$backend/v1/$1;
             {% else %}
               {# Generic configuration for use outside of our container setup #}
-              proxy_pass http://127.0.0.1:{{ matrix_hookshot_provisioning_port }}/$1;
+              proxy_pass http://127.0.0.1:{{ matrix_hookshot_provisioning_port }}/v1/$1;
```
HarHarLinks, line 61: you were talking about GET args earlier, does it not need the `$is_args$args` fix?

HarHarLinks (commented):
thanks for investigating this

spantaleev:
Does doing what we do for all other vhosts (`proxy_pass http://$backend;`) not work?

Example for `jitsi.DOMAIN`: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/780bc36c78feedb1b9334e58a8ba2d88c9b572fb/roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-jitsi.conf.j2#L21-L34

roles/matrix-bridge-hookshot/tasks/init.yml
luismanson: Yes, It's needed, PR fixed.

HarHarLinks:
@spantaleev: can't match `location /` since we need/want to split by endpoint or at least that is how I did it. I tried around a lot with not arguments following after `$backend` but this is the only way I could get working. If anyone is more proficient with nginx syntax, feel free to refactor to something nicer.

spantaleev:
Oh, right.. we don't have a dedicated vhost for this, but are using `matrix.DOMAIN` and capturing some specific path.

If you're both confirming that this works, I can merge this as it is. It seems a little odd that you're capturing some `/_matrix/app` path and forwarding all that to `/v1` though. What happens when v2 comes around? Or am I overlooking something?

HarHarLinks:
one solution would be to then add v2 to that capturing regex like `^/hookshot/v[1-2]/(.*)`

alternatively we could add in another segment to the path so the URL externally becomes `matrix.host/hookshot/provisioning/v1/...` - i didn't do that originally in order to stay closer to hookshot itself and "compatible" to its own docs.

spantaleev:
Merged as is! Thanks! :+1:

Half-Shot:
>one solution would be to then add v2 to that capturing regex like ^/hookshot/v[1-2]/(.*)

Would be my suggestion. I don't think a v2 is on the horizon though, so don't worry just yet :)

PR #1676, opened 2022-03-05
"Federate rooms" switch for mautrix bridges

MrAnno:
This PR makes it configurable whether portal rooms created by mautrix bridges will be federatable or not.

Note: Rooms cannot transition from unfederated to federated after creation.

### DIFF
--- a/roles/matrix-bridge-mautrix-facebook/defaults/main.yml
+++ b/roles/matrix-bridge-mautrix-facebook/defaults/main.yml
@@ -33,6 +33,9 @@ matrix_mautrix_facebook_systemd_wanted_services_list: []
 matrix_mautrix_facebook_appservice_token: ''
 matrix_mautrix_facebook_homeserver_token: ''
 
+# Whether or not created rooms should have federation enabled.
+# If false, created portal rooms will never be federated.
+matrix_mautrix_facebook_federate_rooms: true
 
 # Database-related configuration fields.
 #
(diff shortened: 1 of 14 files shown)


spantaleev:
Adding `|to_json` for good measure is probably a good idea.

So the default is "let them be federateable"? Given that for most bridges you're not an administrator in the room and that gaining invite-privileges and inviting others is of limited purpose, I wonder why the upstream default is not to prevent federation.

MrAnno:
Thank you for the review.

I'm pretty new to Matrix, so I think I won't have a good answer to this question.
I believe the "default on" policy may have something to do with the limitation that non-federatable rooms can not be converted to federatable later.

It seems there is no server-side default federation switch at all (at least in Synapse); but for example, Element has its own frontend option:
https://github.com/vector-im/element-web/blob/f975d85806263c9f7b1750e53de8da0274dfb02c/config.sample.json#L30

A silly question: why do we use `|to_json` when formatting mautrix templates (YAML files)?

spantaleev:
> A silly question: why do we use `|to_json` when formatting mautrix templates (YAML files)?

YAML is a superset of JSON, so we put data into YAML configuration by using `|to_json`. We can use `|to_yaml `too to achieve the same, I suppose -- although YAML usually has 10 different ways to represent the same thing and it's more unpredictable what `|to_yaml` would do.

To get back as to "why", it's because values (especially `string` values) can contain special characters that need to be escaped when embedding them into YAML (JSON) text.

MrAnno:
@spantaleev Thanks for the explanation.
I've updated the pull request.

PR #1678, opened 2022-03-06
feat(appservice-discord): add disablePortalBridging bridge option

jimeh:
I've also expanded appservice-discord documentation to detail the different types of bridging available, and how to use them.


spantaleev:
Thanks! :+1:

PR #1685, opened 2022-03-08
bridge-mautrix-facebook: add support for web-based login

MrAnno:
This PR adds web login support for the Facebook bridge.
By default, it is on, as it does not seem to be a breaking change (the web login authentication method is optional, you can choose between this and the previously available "in-Matrix" login method when communicating with the bridge bot).

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -299,6 +299,10 @@ matrix_mautrix_facebook_appservice_token: "{{ '%s' | format(matrix_homeserver_ge
 
 matrix_mautrix_facebook_homeserver_token: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'fb.hs.token') | to_uuid }}"
 
+matrix_mautrix_facebook_public_endpoint: "/{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'facebook') | to_uuid }}"
+
+matrix_mautrix_facebook_container_http_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:9008' }}"
+
 matrix_mautrix_facebook_login_shared_secret: "{{ matrix_synapse_ext_password_provider_shared_secret_auth_shared_secret if matrix_synapse_ext_password_provider_shared_secret_auth_enabled else '' }}"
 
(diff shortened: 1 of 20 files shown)


spantaleev:
Maybe it makes sense to make this configurable via some `matrix_mautrix_facebook_appservice_public_enabled` or similar variable? It can influence:
- the `appservice.public.enabled` configuration option
- whether configuration will be injected into nginx
- whether ports will be exposed

When disabled, the bridge will not expose its HTTP endpoints publicly and will not occupy network ports (when `matrix_nginx_proxy_enabled: false`).

spantaleev:
Maybe it makes sense to make this configurable via some `matrix_mautrix_facebook_appservice_public_enabled` or similar variable? It can influence:
- the `appservice.public.enabled` configuration option
- whether configuration will be injected into nginx
- whether ports will be exposed

When disabled, the bridge will not expose its HTTP endpoints publicly and will not occupy network ports (when `matrix_nginx_proxy_enabled: false`).

MrAnno:
Sounds good. Can I add this switch to all the other mautrix bridges as well? Could we use "enabled" as a default for all of them, or should I disable them (for backward compatibility) in places where the "public" feature was not available previously?

I'm asking these questions because I've started thinking about ways to maybe unify bridge role options, or at least to keep bridge config files up-to-date (by generating and showing a diff automatically against the upstream default when bumping their version).
It looks like, for example, mautrix bridges slowly converge to a unified config, but this is not done atomically with each release, so the configurations in the `matrix-docker-ansible-deploy` repo are slowly diverging from each other and from the upstream default too.

I believe this will eventually be handled by upstream, for example, they might create a monorepo as Beeper did with [`mx-puppet-monorepo`](https://gitlab.com/beeper/mx-puppet-monorepo).

spantaleev:
Thinking about it.. enabling it by default is probably what most users expect - a simpler (and more secure?) login experience.

It makes sense to have a separate variable for each mautrix bridge though.

----

Keeping up with config updates better sounds good! There has been discussion in the past for using a single `mautrix` role that supports installing/uninstalling all `mautrix` bridges.. But such a role will probably be more difficult to develop and maintain. There have been discussions about support of multiple instances of each bridge, but it's also something that will be complex.

spantaleev:
Looks good! Thank you! :+1:

PR #1693, opened 2022-03-14
mautrix-facebook, synapse: minor improvements

MrAnno:
- bridge-mautrix-facebook: disable temporary disconnect notices
  mautrix/facebook#215

- synapse: do not expose plain federation port when it's disabled
  matrix_synapse_federation_port_enabled can be disabled by users, for example, when one wants to use the same port for client and federation requests (docs/configuring-playbook-federation.md).


spantaleev:
Thanks! :+1:

PR #1694, opened 2022-03-14
add default for matrix_prometheus_scraper_hookshot_enabled

HarHarLinks:
addresses this complaint https://matrix.to/#/!cNSQwPqhHKkIZdBnvt:devture.com/$M8bwywhf6ywXPahT08Ntq4g9MP6CyAfMMCqSAV9ak3k?via=devture.com&via=matrix.org&via=tu-dresden.de

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -2106,7 +2106,7 @@ matrix_prometheus_scraper_node_targets: "{{ ['matrix-prometheus-node-exporter:91
 matrix_prometheus_scraper_postgres_enabled: "{{ matrix_prometheus_postgres_exporter_enabled }}"
 matrix_prometheus_scraper_postgres_targets: "{{ ['matrix-prometheus-postgres-exporter:'+ matrix_prometheus_postgres_exporter_port|string] if matrix_prometheus_scraper_postgres_enabled else [] }}"
 
-matrix_prometheus_scraper_hookshot_enabled: "{{ matrix_hookshot_metrics_enabled|bool }}"
+matrix_prometheus_scraper_hookshot_enabled: "{{ matrix_hookshot_metrics_enabled|default(false) }}"
 matrix_prometheus_scraper_hookshot_targets: "{{ [matrix_hookshot_container_url|string +':'+ matrix_hookshot_metrics_port|string] if matrix_hookshot_metrics_enabled else [] }}"
 
 ######################################################################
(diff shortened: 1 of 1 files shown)


spantaleev:
Isn't the original complaint about an undefined variable?

I suppose that @Cadair is using `group_vars/matrix_servers` as is, without including all the roles. That way, he's passing variables to the `matrix-nginx-proxy` role (`matrix_prometheus_scraper_hookshot_enabled`), which are defined in terms of variables (`matrix_hookshot_metrics_enabled`) defined in a role (`matrix-bridge-hookshot`) which hasn't been included:

```yaml
matrix_prometheus_scraper_hookshot_enabled: "{{ matrix_hookshot_metrics_enabled }}"
```

----------

I believe that the right way to fix this in the way you're attempting to, would be to do something like this:

```diff
-matrix_prometheus_scraper_hookshot_enabled: "{{ matrix_hookshot_metrics_enabled }}"
+matrix_prometheus_scraper_hookshot_enabled: "{{ matrix_hookshot_metrics_enabled|default(false) }}"
```

-------

However.. this is likely not the only instance where we have such a problem and is not the only workaround we need.

We try to make roles work independently and to have each role not directly reference variables defined in other roles. Roles influence each other **in the context of the playbook distribution** via the `group_vars/matrix_servers` file. You're only *expected* to wire roles with that file when you're including **all** roles distributed by the playbook repository.

We don't aim to support cherry-picking roles and using all of `group_vars/matrix_servers` as it is. That file is for wiring all our roles together and not for arbitrary mixing of roles.

If we want arbitrary mixing and using `group_vars/matrix_servers` as it is that works with whatever arbitrary role mixing you've chosen, we'll need to have a lot of `|default(..)` calls to `group_vars/matrix_servers`. I'm not sure if we want to get into that.

HarHarLinks:
I see. Using https://j2live.ttl255.com in ansible mode, it renders as intended (empty value = falsey => becomes false).
However this may not apply to all ansible/jinja versions you try to support, nor as legible as `default()`.

Cadair:
> We don't aim to support cherry-picking roles and using all of group_vars/matrix_servers as it is. That file is for wiring all our roles together and not for arbitrary mixing of roles.

Wait what, I thought that was the whole point? How can having too many glue variables defined be a problem?

spantaleev:
I see. Good to hear that it pretty much works for this use case already. I guess you could report the other problems or submit PRs for fixing them.

I've fixed this PR up a bit and merged it. Thank you! :+1:

PR #1697, opened 2022-03-15
Add check for ma1sd repo changes on image self build

Sweeny42:
Adding when clause to ma1sd image build task to avoid image being built if not needed.


spantaleev:
Makes sense! Thanks! :+1:

PR #1698, opened 2022-03-16
fix: avoid yaml being wrapped at column 80 via to_nice_yaml

jimeh:
The `to_nice_yaml` helper will by default wrap any string YAML values on the
first space after column 80. This can in worst case yield invalid YAML syntax.
More details in Ansible's documentation here:

https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#formatting-data-yaml-and-json

In short, you need to explicitly provide a custom width argument of a high
number of some kind to avoid the line wrapping.

---

I noticed this via the mautrix-whatsapp config file, which has the following
default value:

    displayname_template: '{{if .PushName}}{{.PushName}}{{else if .BusinessName}}{{.BusinessName}}{{else}}{{.JID}}{{end}} (WA)'

But it was being output into config.yaml as:

    displayname_template: '{{if .PushName}}{{.PushName}}{{else if .BusinessName}}{{.BusinessName}}{{else}}{{.JID}}{{end}}
        (WA)'

### DIFF
--- a/roles/matrix-bot-go-neb/tasks/setup_install.yml
+++ b/roles/matrix-bot-go-neb/tasks/setup_install.yml
@@ -25,7 +25,7 @@
 
 - name: Ensure go-neb config installed
   copy:
-    content: "{{ matrix_bot_go_neb_configuration|to_nice_yaml }}"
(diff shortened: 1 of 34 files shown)


spantaleev:
Wow, that is some horrible default behavior! Thank you for finding and fixing it! :+1:

PR #1701, opened 2022-03-17
Added retries to the docker pulls

etkecc:
Each docker pull will retry on failed attempt 10 times with delay of 10 seconds between each attempt.


spantaleev:
Great! Thanks a lot for making the playbook more resilient to temporary container registry outages! :+1:

PR #1705, opened 2022-03-17
Fix index in external_prometheus.yml.example.j2

HarHarLinks:
For an unknown reason prometheus ignored the given "numeric" index and replaced it by 1. This made it not work properly, plus multiple workers of same types were not differentiable. With a "string" index, it works as intended.

before:
image

after:
image


spantaleev:
Thanks for investigating and finding a workaround! :+1:

PR #1706, opened 2022-03-18
Don't recommend that the Dimension admin interface be opened in a web browser

turt2live:
It is not meant to be opened in this way.


spantaleev:
Thank you for clearing it up! 👍

PR #1709, opened 2022-03-22
add detailed docs about hookshot's URLs

HarHarLinks:
there was a user that missed the hint to look at some yaml previously in there so I made it slightly more obvious


spantaleev:
Thanks! :+1:

PR #1713, opened 2022-03-23
Update mjolnir antispam commit hash

pulmonarycosignerkindness:
Changed the commit hash in matrix_synapse_ext_spam_checker_mjolnir_antispam_git_version to latest. Fixes a python ImportError in Synapse with mjolnir v1.4.1 leading to a Synapse crash-loop.


spantaleev:
Thanks! :+1:

PR #1714, opened 2022-03-24
Synapse 1.55.0 -> Synapse 1.55.2

joecool1029:
This is a minor hotfix, needs to be bumped though or new deploys will break (existing deploys not affected by bug).


spantaleev:
Thanks! 👍

PR #1715, opened 2022-03-26
Updated: matrix-googlechat to v0.3.1

GoliathLabs:
Since the release is only a little over a week old, we should still merge this, especially since there have been no commits made after this release was published (https://github.com/mautrix/googlechat/releases/tag/v0.3.1)


spantaleev:
Thanks! Good to move away from `:latest`!

PR #1717, opened 2022-03-29
Fix Facebook presence config key

ofalvai:
The presence config key has been changed in mautrix-facebook from `presence` to `presence_from_facebook`: https://github.com/mautrix/facebook/blame/master/mautrix_facebook/example-config.yaml#L157

Relevant upstream commit: https://github.com/mautrix/facebook/commit/a8e6f57bc94a17068086b4020477db508ea1639b

This PR updates the config key in the template.


spantaleev:
Thanks! :+1: 

I was wondering whether we should use a new `matrix_mautrix_facebook_bridge_presence_from_facebook` variable and put a deprecation task in `roles/matrix-bridge-mautrix-facebook/tasks/validate_config.yml`, but.. it's probably fine the way it is as well.

PR #1718, opened 2022-03-29
Add note about Prometheus retention policy

ofalvai:


spantaleev:
Thanks for clearing this up! :+1:

PR #1723, opened 2022-03-31
hookshot webhook listener moved to /webhook

### LINKED ISSUE #1681: Hookshot - Generic webhooks will be routed to /webhook
Currently hookshot generic webhooks get routed to `/:hookId` on the webhooks listener, but https://github.com/matrix-org/matrix-hookshot/pull/227 is chaging things so we now prefer them to be routed to `/webhook/:hookId`. While previous setups still work, it's deprecated behavior.

HarHarLinks:
fixes #1681, introduced in https://github.com/matrix-org/matrix-hookshot/pull/227

### DIFF
--- a/roles/matrix-bridge-hookshot/tasks/init.yml
+++ b/roles/matrix-bridge-hookshot/tasks/init.yml
@@ -68,7 +68,7 @@
               {# Use the embedded DNS resolver in Docker containers to discover the service #}
               resolver 127.0.0.11 valid=5s;
               set $backend "{{ matrix_hookshot_container_url }}:{{ matrix_hookshot_webhook_port }}";
-              proxy_pass http://$backend/$1$is_args$args;
+              proxy_pass http://$backend/webhook/$1$is_args$args;
             {% else %}
               {# Generic configuration for use outside of our container setup #}
               proxy_pass http://127.0.0.1:{{ matrix_hookshot_webhook_port }}/$1$is_args$args;


HarHarLinks:
Upstream hookshot now prefers to listen at `/webhook` instead of `/`.
The playbook default is currently to listen on (some common prefix plus) `/webhooks`, see https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/09412e34ea9db07f00e614d0a99c6d880a935fcf/roles/matrix-bridge-hookshot/defaults/main.yml#L33

I think it would be preferable to update our proxied endpoint to match upstream to avoid confusion for new users, but it would be necessary to also be backward-compatible to our old setup in my opinion so admins don't have to manually intervene (set that variable to the old default, or update all their configured services).

This is not a big problem to do with nginx locations, we can simply update the regex to `webhooks?` to match the `s` optionally and thus proxy both to the new `/webhook` endpoint on hookshot side.

However I am not certain on the hookshot side: this variable is used by the playbook to define some config like the following
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/09412e34ea9db07f00e614d0a99c6d880a935fcf/roles/matrix-bridge-hookshot/defaults/main.yml#L54-L55
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/09412e34ea9db07f00e614d0a99c6d880a935fcf/roles/matrix-bridge-hookshot/templates/config.yml.j2#L29

@Half-Shot please advice: I fear that if the playbooks "canonical" default were to be `/webhook` and this hookshot configuration done with this uri, then even though the nginx proxy accepts `/webhooks/oauth` and forwards it to hookshot, hookshot may not accept it, but only accept `/webhook/oauth`.

HarHarLinks:
I may have misunderstood the upstream change. Are GitHub, GitLab, oauth, etc webhooks supposed to go to `/webhook` now, or still go to `/` while **only generics** go to `/webhook`?

PR #1727, opened 2022-04-03
add borg backup

etkecc:
Note: the used docker image [supports multiple architectures](https://gitlab.com/etke.cc/borgmatic/-/blob/main/.gitlab-ci.yml#L15) and builds weekly

Another note: we use [rsync.net borg](https://www.rsync.net/products/borg.html) and that PR was developed and tested on it, but other vendors (eg: [hetzner storage box](https://www.hetzner.com/storage/storage-box)) work in the same way, so there is no vendor-lock

### DIFF
--- a/docs/configuring-playbook-backup-borg.md
+++ b/docs/configuring-playbook-backup-borg.md
@@ -28,10 +28,10 @@ Minimal working configuration (`inventory/host_vars/matrix.DOMAIN/vars.yml`) to 
 ```yaml
 matrix_backup_borg_enabled: true
 matrix_backup_borg_repositories:
-	- USER@HOST:REPO
+ - USER@HOST:REPO
 matrix_backup_borg_encryption_passphrase: "PASSPHRASE"
 matrix_backup_borg_ssh_key_private: |
-	PRIVATE KEY
+  PRIVATE KEY
 ```
 
 where:
--- a/docs/configuring-playbook-backup-borg.md
+++ b/docs/configuring-playbook-backup-borg.md
@@ -1,0 +1,52 @@
+# Setting up borg backup (optional)
+
+The playbook can install and configure [borgbackup](https://www.borgbackup.org/) with [borgmatic](https://torsion.org/borgmatic/) for you.
+The backup will run based on `matrix_backup_borg_schedule` var (systemd timer calendar), default: 4am every day
+
+## Prerequisites
+
+1. Create ssh key:
+
+```bash
+ssh-keygen -t ed25519 -N '' -C matrix
+```
+
+2. Add public part of that ssh key to your borg provider / server:
+
+```bash
+# example to append the new PUBKEY contents, where:
+# PUBKEY is path to the public key,
+# USER is a ssh user on a provider / server
+# HOST is a ssh host of a provider / server
+cat PUBKEY | ssh USER@HOST 'dd of=.ssh/authorized_keys oflag=append conv=notrunc'
+```
+
+## Adjusting the playbook configuration
+
+Minimal working configuration (`inventory/host_vars/matrix.DOMAIN/vars.yml`) to enable borg backup:
+
+```yaml
+matrix_backup_borg_enabled: true
+matrix_backup_borg_repository: "USER@HOST:REPO"
+matrix_backup_borg_passphrase: "PASSPHRASE"
+matrix_backup_borg_ssh_key: |
+	PRIVATE KEY
+```
+
+where:
+
+* USER - ssh user of a provider / server
+* HOST - ssh host of a provider / server
+* REPO - borg repository name, it will be initialized on backup start, eg: `matrix`
+* PASSPHRASE - super-secret borg passphrase, you may generate it with `pwgen -s 64 1` or use any password manager
+* PRIVATE KEY - the content of the public part of the ssh key you created before
+
+Check the `roles/matrix-backup-borg/defaults/main.yml` for the full list of available options
+
+## Installing
+
+After configuring the playbook, run the [installation](installing.md) command again:
+
+```
+ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start
+```
--- a/docs/configuring-playbook-backup-borg.md
+++ b/docs/configuring-playbook-backup-borg.md
@@ -2,7 +2,7 @@
 
 The playbook can install and configure [borgbackup](https://www.borgbackup.org/) with [borgmatic](https://torsion.org/borgmatic/) for you.
 BorgBackup is a deduplicating backup program with optional compression and encryption.
-That means your daily incremental backups can be stored in a fraction of the space and is safe weather you store it at home or a cloud service.
+That means your daily incremental backups can be stored in a fraction of the space and is safe wether you store it at home or on a cloud service.
 
 The backup will run based on `matrix_backup_borg_schedule` var (systemd timer calendar), default: 4am every day
 
--- a/docs/configuring-playbook-backup-borg.md
+++ b/docs/configuring-playbook-backup-borg.md
@@ -1,6 +1,9 @@
 # Setting up borg backup (optional)
 
 The playbook can install and configure [borgbackup](https://www.borgbackup.org/) with [borgmatic](https://torsion.org/borgmatic/) for you.
+BorgBackup is a deduplicating backup program with optional compression and encryption.
+That means your daily incremental backups can be stored in a fraction of the space and is safe weather you store it at home or a cloud service.
+
 The backup will run based on `matrix_backup_borg_schedule` var (systemd timer calendar), default: 4am every day
 
 ## Prerequisites
@@ -27,11 +30,11 @@ Minimal working configuration (`inventory/host_vars/matrix.DOMAIN/vars.yml`) to 
 
 ```yaml
 matrix_backup_borg_enabled: true
-matrix_backup_borg_repositories:
+matrix_backup_borg_location_repositories:
  - USER@HOST:REPO
-matrix_backup_borg_encryption_passphrase: "PASSPHRASE"
+matrix_backup_borg_storage_encryption_passphrase: "PASSPHRASE"
 matrix_backup_borg_ssh_key_private: |
-  PRIVATE KEY
+	PRIVATE KEY
 ```
 
 where:
--- a/docs/configuring-playbook-backup-borg.md
+++ b/docs/configuring-playbook-backup-borg.md
@@ -5,10 +5,10 @@ The backup will run based on `matrix_backup_borg_schedule` var (systemd timer ca
 
 ## Prerequisites
 
-1. Create ssh key:
+1. Create ssh key on any machine:
 
 ```bash
-ssh-keygen -t ed25519 -N '' -C matrix
+ssh-keygen -t ed25519 -N '' -f matrix-borg-backup -C matrix
 ```
 
 2. Add public part of that ssh key to your borg provider / server:
@@ -27,9 +27,10 @@ Minimal working configuration (`inventory/host_vars/matrix.DOMAIN/vars.yml`) to 
 
 ```yaml
 matrix_backup_borg_enabled: true
-matrix_backup_borg_repository: "USER@HOST:REPO"
-matrix_backup_borg_passphrase: "PASSPHRASE"
-matrix_backup_borg_ssh_key: |
+matrix_backup_borg_repositories:
+	- USER@HOST:REPO
+matrix_backup_borg_encryption_passphrase: "PASSPHRASE"
+matrix_backup_borg_ssh_key_private: |
 	PRIVATE KEY
 ```
 
--- a/docs/configuring-playbook-backup-borg.md
+++ b/docs/configuring-playbook-backup-borg.md
@@ -2,7 +2,7 @@
 
 The playbook can install and configure [borgbackup](https://www.borgbackup.org/) with [borgmatic](https://torsion.org/borgmatic/) for you.
 BorgBackup is a deduplicating backup program with optional compression and encryption.
-That means your daily incremental backups can be stored in a fraction of the space and is safe wether you store it at home or on a cloud service.
+That means your daily incremental backups can be stored in a fraction of the space and is safe whether you store it at home or on a cloud service.
 
 The backup will run based on `matrix_backup_borg_schedule` var (systemd timer calendar), default: 4am every day
 
--- a/roles/matrix-backup-borg/templates/config.yaml.j2
+++ b/roles/matrix-backup-borg/templates/config.yaml.j2
@@ -1,0 +1,33 @@
+#jinja2: lstrip_blocks: "True", trim_blocks: "True"
+
+location:
+  source_directories:
+  - /matrix
+  repositories:
+  - {{ matrix_backup_borg_repository }}
+  one_file_system: true
+
+storage:
+  compression: {{ matrix_backup_borg_compression }}
+  ssh_command: {{ matrix_backup_borg_ssh_command }}
+  archive_name_format: '{{ matrix_backup_borg_name }}'
+  encryption_passphrase: {{ matrix_backup_borg_passphrase }}
+
+retention:
+  keep_hourly: {{ matrix_backup_borg_hourly }}
+  keep_daily: {{ matrix_backup_borg_daily }}
+  keep_weekly: {{ matrix_backup_borg_weekly }}
+  keep_monthly: {{ matrix_backup_borg_monthly }}
+  keep_yearly: {{ matrix_backup_borg_yearly }}
+  prefix: '{{ matrix_backup_borg_prefix }}'
+
+consistency:
+  checks:
+    - repository
+    - archives
+
+hooks:
+  after_backup:
+    - echo "Backup created."
+  on_error:
+    - echo "Error while creating a backup."
--- a/roles/matrix-backup-borg/templates/config.yaml.j2
+++ b/roles/matrix-backup-borg/templates/config.yaml.j2
@@ -1,16 +1,16 @@
 #jinja2: lstrip_blocks: "True", trim_blocks: "True"
 
 location:
-  source_directories: {{ matrix_backup_borg_source_directories|to_json }}
-  repositories: {{ matrix_backup_borg_repositories|to_json }}
+  source_directories: {{ matrix_backup_borg_location_source_directories|to_json }}
+  repositories: {{ matrix_backup_borg_location_repositories|to_json }}
   one_file_system: true
-  exclude_patterns: {{ matrix_backup_borg_exclude_patterns|to_json }}
+  exclude_patterns: {{ matrix_backup_borg_location_exclude_patterns|to_json }}
 
 storage:
-  compression: {{ matrix_backup_borg_compression }}
-  ssh_command: {{ matrix_backup_borg_ssh_command }}
-  archive_name_format: '{{ matrix_backup_borg_archive_name_format }}'
-  encryption_passphrase: {{ matrix_backup_borg_encryption_passphrase }}
+  compression: {{ matrix_backup_borg_storage_compression }}
+  ssh_command: {{ matrix_backup_borg_storage_ssh_command }}
+  archive_name_format: '{{ matrix_backup_borg_storage_archive_name_format }}'
+  encryption_passphrase: {{ matrix_backup_borg_storage_encryption_passphrase }}
 
 retention:
   keep_hourly: {{ matrix_backup_borg_retention_keep_hourly }}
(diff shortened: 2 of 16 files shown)


HarHarLinks:
In my fork of the playbook I use [m3nu.ansible_role_borgbackup](https://github.com/borgbase/ansible-role-borgbackup). For the DB dump to work I needed to also install a psql-client that matches the container and make the container accessible from the host.

In my experience, trying to back up all of `/matrix` fails with errors such as `/matrix/postgres/data/pg_wal/0000000100000059000000B5: stat: [Errno 2] No such file or directory: '/matrix/postgres/data/pg_wal/0000000100000059000000B5'`. Also, my deduplicated archive size shrunk from ~2GB when including `/matrix/postgres/data` to <200MB after excluding it and instead backing up the DB dump.

I have added in mine the ability to post info about running backups and logs to matrix, and - in case they fail - also to email via the email proxy already included with the playbook via [apprise](https://github.com/caronc/apprise).
```yaml
      borgmatic_failure_command:
        - "\"{{ apprise_docker_call }} --title '[{{ hostname }}] Borgmatic Backup Error' --body '{% raw %}{{app_title}}{% endraw %}<br>@room: there was an error creating a backup!' --tag=error\""
      borgmatic_before_backup_command:
        - "\"{{ apprise_docker_call }} --title '[{{ hostname }}] Borgmatic Backup' --body '{% raw %}{{app_title}}{% endraw %}<br>starting...' --tag=backup\""
      borgmatic_after_backup_command:
        - "\"{{ apprise_docker_call }} --title '[{{ hostname }}] Borgmatic Backup Complete' --tag=backup\""
```

My cron jobs are roughly
```yaml
borgmatic_cron_file: "/etc/cron.d/borgmatic"
borgmatic_command: "borgmatic -c /etc/borgmatic/config.yaml --no-color"
borgmatic_log: "--log-file {{ borgmatic_log_file }}"
apprise_send_borgmatic_log: "{{ apprise_docker_call }} -e --body \"<pre><code>$(cat /borgmatic.log)\\n</code></pre>\\n\" --tag=backup"

# backup job
{{ borgmatic_command }} --create --stats > {{ borgmatic_log_file }} 2>&1; {{ apprise_send_borgmatic_log }}
# check job
rm {{ borgmatic_log_file }}; {{ borgmatic_command }} --check --log-file-verbosity 1 {{ borgmatic_log }}; {{ apprise_send_borgmatic_log }}
```

There are some more paths that are not necessary to include in a backup since they are just cached remote files, here are all I set:
```yaml
      borg_exclude_patterns:
        - /matrix/synapse/storage/media-store/remote_content
        - /matrix/synapse/storage/media-store/remote_thumbnail
        - /matrix/synapse/storage/media-store/url_cache
        - /matrix/synapse/storage/media-store/url_cache_thumbnails
        - /matrix/postgres/data
```
(note that one ends in `thumbnail` while the other uses `thumbnails` due to an inconsistency in synapse)

etkecc:
Here is the actual log output of that PR without the #1726 fix:

```
Apr 03 10:43:00 matrix systemd[1]: Starting Matrix Borg Backup...
Apr 03 10:43:03 matrix matrix-backup-borg[6305]: Remote: Warning: Permanently added 'REDACTED' (ED25519) to the list of known hosts.
Apr 03 10:43:09 matrix matrix-backup-borg[6488]: Remote: Warning: Permanently added 'REDACTED' (ED25519) to the list of known hosts.
Apr 03 10:43:10 matrix matrix-backup-borg[6488]: ------------------------------------------------------------------------------
Apr 03 10:43:10 matrix matrix-backup-borg[6488]:                        Original size      Compressed size    Deduplicated size
Apr 03 10:43:10 matrix matrix-backup-borg[6488]: Deleted data:                    0 B                  0 B                  0 B
Apr 03 10:43:10 matrix matrix-backup-borg[6488]: All archives:               28.23 GB             15.10 GB              8.10 GB
Apr 03 10:43:10 matrix matrix-backup-borg[6488]:                        Unique chunks         Total chunks
Apr 03 10:43:10 matrix matrix-backup-borg[6488]: Chunk index:                   21440                49005
Apr 03 10:43:10 matrix matrix-backup-borg[6488]: ------------------------------------------------------------------------------
Apr 03 10:43:13 matrix matrix-backup-borg[6488]: /matrix/postgres/env-postgres-psql: open: [Errno 13] Permission denied: '/matrix/postgres/env-postgres-psql'
Apr 03 10:43:13 matrix matrix-backup-borg[6488]: /matrix/postgres/env-postgres-server: open: [Errno 13] Permission denied: '/matrix/postgres/env-postgres-server'
Apr 03 10:45:50 matrix matrix-backup-borg[6488]: /matrix/jitsi/web/env: open: [Errno 13] Permission denied: '/matrix/jitsi/web/env'
Apr 03 10:45:50 matrix matrix-backup-borg[6488]: /matrix/jitsi/web/config/keys/cert.key: open: [Errno 13] Permission denied: '/matrix/jitsi/web/config/keys/cert.key'
Apr 03 10:45:50 matrix matrix-backup-borg[6488]: /matrix/jitsi/prosody/env: open: [Errno 13] Permission denied: '/matrix/jitsi/prosody/env'
Apr 03 10:45:50 matrix matrix-backup-borg[6488]: /matrix/jitsi/prosody/config/data: scandir: [Errno 13] Permission denied: '/matrix/jitsi/prosody/config/data'
Apr 03 10:45:50 matrix matrix-backup-borg[6488]: /matrix/jitsi/prosody/config/certs/meet.jitsi.key: open: [Errno 13] Permission denied: '/matrix/jitsi/prosody/config/certs/meet.jitsi.key'
Apr 03 10:45:50 matrix matrix-backup-borg[6488]: /matrix/jitsi/prosody/config/certs/meet.jitsi.crt: open: [Errno 13] Permission denied: '/matrix/jitsi/prosody/config/certs/meet.jitsi.crt'
Apr 03 10:45:50 matrix matrix-backup-borg[6488]: /matrix/jitsi/prosody/config/certs/auth.meet.jitsi.key: open: [Errno 13] Permission denied: '/matrix/jitsi/prosody/config/certs/auth.meet.jitsi.key'
Apr 03 10:45:50 matrix matrix-backup-borg[6488]: /matrix/jitsi/prosody/config/certs/auth.meet.jitsi.crt: open: [Errno 13] Permission denied: '/matrix/jitsi/prosody/config/certs/auth.meet.jitsi.crt'
Apr 03 10:45:50 matrix matrix-backup-borg[6488]: /matrix/jitsi/jicofo/env: open: [Errno 13] Permission denied: '/matrix/jitsi/jicofo/env'
Apr 03 10:45:50 matrix matrix-backup-borg[6488]: /matrix/jitsi/jvb/env: open: [Errno 13] Permission denied: '/matrix/jitsi/jvb/env'
Apr 03 10:45:50 matrix matrix-backup-borg[6488]: /matrix/honoroit/config/env: open: [Errno 13] Permission denied: '/matrix/honoroit/config/env'
Apr 03 10:45:52 matrix matrix-backup-borg[6488]: ------------------------------------------------------------------------------
Apr 03 10:45:53 matrix matrix-backup-borg[6488]: Archive name: matrix-2022-04-03-084311
Apr 03 10:45:53 matrix matrix-backup-borg[6488]: Archive fingerprint: REDACTED
Apr 03 10:45:53 matrix matrix-backup-borg[6488]: Time (start): Sun, 2022-04-03 08:43:13
Apr 03 10:45:53 matrix matrix-backup-borg[6488]: Time (end):   Sun, 2022-04-03 08:45:50
Apr 03 10:45:53 matrix matrix-backup-borg[6488]: Duration: 2 minutes 36.38 seconds
Apr 03 10:45:53 matrix matrix-backup-borg[6488]: Number of files: 24708
Apr 03 10:45:53 matrix matrix-backup-borg[6488]: Utilization of max. archive size: 0%
Apr 03 10:45:53 matrix matrix-backup-borg[6488]: ------------------------------------------------------------------------------
Apr 03 10:45:53 matrix matrix-backup-borg[6488]:                        Original size      Compressed size    Deduplicated size
Apr 03 10:45:53 matrix matrix-backup-borg[6488]: This archive:               14.13 GB              7.56 GB            774.33 MB
Apr 03 10:45:53 matrix matrix-backup-borg[6488]: All archives:               42.36 GB             22.66 GB              8.87 GB
Apr 03 10:47:46 matrix matrix-backup-borg[6488]: Backup created.
```

As it was mentioned in https://matrix.to/#/!cNSQwPqhHKkIZdBnvt:devture.com/$hXtfV8Y7oYav2VW6tH3jsBVV9hodtUHakssKGq0T75M?via=devture.com&via=matrix.org&via=tu-dresden.de the problem you mentioned may be caused by the user mismatch.

Please, keep in mind that it's "mvp" version and it should be enhanced further.

HarHarLinks:
> may be caused by the user mismatch

From your log it looks like it, the matrix user wouldn't have access to some of these files. In my case the job is simply running as root (idk how much worse this is over your solution, but that isn't relevant to the issue at hand) and should not have any user mismatch trouble, instead I believe it is cased by the flow of
1. borg collects a list of paths and starts working
2. in the meantime postgres keeps running and the files get deleted etc
3. borg fails to find a file that does not exist anymore, and might miss new ones. note that this is prevented by backing up a dump instead as that gets created instantaneously

I should mention that this error did not happen immediately, but only after some months of operation. I only caught it thanks to reading the daily logs (being posted to matrix).

> "mvp" version

I understand. Some of my hints were given towards later improvements, but the error I experienced certainly shouldn't happen in an mvp.

etkecc:
I'm prepairing a new commit with exclude patterns

etkecc:
now you can add anything to the ignore patterns

roles/matrix-backup-borg/templates/config.yaml.j2
```diff
@@ -0,0 +1,37 @@
+#jinja2: lstrip_blocks: "True", trim_blocks: "True"
+
+location:
+  source_directories:
+  - {{ matrix_backup_borg_source }}
```
spantaleev, line 5: `matrix_backup_borg_source` -> `matrix_backup_borg_source_directories`?

roles/matrix-backup-borg/templates/config.yaml.j2
```diff
@@ -0,0 +1,37 @@
...
+  repositories:
+  - {{ matrix_backup_borg_repository }}
```
spantaleev, line 7: Perhaps `repositories: {{ matrix_backup_borg_repositories|to_json }}` is more flexible here?

You can define:

```yaml
matrix_backup_borg_repositories: ["{{ matrix_backup_borg_repository }}"]
```

Or just get rid of `matrix_backup_borg_repository` and ask people to configure the `matrix_backup_borg_repositories` list?

roles/matrix-backup-borg/templates/config.yaml.j2
```diff
@@ -0,0 +1,37 @@
...
+  one_file_system: true
+  exclude_patterns:
```
spantaleev, line 9: Perhaps: `exclude_patterns: {{ matrix_backup_borg_exclude_patterns|to_json }}` is better and safer:

- using `|to_json` to spew out the whole list in properly escaped form. Even if the list is empty (`matrix_backup_borg_exclude_patterns: []`), `|to_json` will still do the right thing. The `for` loop will not (although YAML may be crazy enough to still take an unterminated `exclude_patterns:` to be valid syntax..)

roles/matrix-backup-borg/templates/config.yaml.j2
```diff
@@ -0,0 +1,37 @@
...
+  {% for pattern in matrix_backup_borg_exclude %}
+  - {{ pattern }}
+  {% endfor %}
+
+storage:
+  compression: {{ matrix_backup_borg_compression }}
+  ssh_command: {{ matrix_backup_borg_ssh_command }}
+  archive_name_format: '{{ matrix_backup_borg_name }}'
```
spantaleev, line 17: `matrix_backup_borg_name` -> `matrix_backup_borg_archive_name_format`?

roles/matrix-backup-borg/templates/config.yaml.j2
```diff
@@ -0,0 +1,37 @@
...
+  encryption_passphrase: {{ matrix_backup_borg_passphrase }}
```
spantaleev, line 18: `matrix_backup_borg_passphrase` -> `matrix_backup_borg_encryption_passphrase`?

roles/matrix-backup-borg/templates/config.yaml.j2
```diff
@@ -0,0 +1,37 @@
...
+
+retention:
+  keep_hourly: {{ matrix_backup_borg_hourly }}
```
spantaleev, line 21: `matrix_backup_borg_hourly` -> `matrix_backup_borg_retention_keep_hourly`?

Likewise for the others?

docs/configuring-playbook-backup-borg.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Adjusting the playbook configuration
+
+Minimal working configuration (`inventory/host_vars/matrix.DOMAIN/vars.yml`) to enable borg backup:
+
+```yaml
+matrix_backup_borg_enabled: true
+matrix_backup_borg_repository: "USER@HOST:REPO"
+matrix_backup_borg_passphrase: "PASSPHRASE"
+matrix_backup_borg_ssh_key: |
```
spantaleev, line 32: Should we put `_private` in the name somewhere, to make it more explicit that this is the private key?

Should we have a variable for the public key as well? If not for anything else, than at least for keeping a backup as part of `vars.yml`?

docs/configuring-playbook-backup-borg.md
```diff
@@ -0,0 +1,52 @@
...
+
+The playbook can install and configure [borgbackup](https://www.borgbackup.org/) with [borgmatic](https://torsion.org/borgmatic/) for you.
+The backup will run based on `matrix_backup_borg_schedule` var (systemd timer calendar), default: 4am every day
+
+## Prerequisites
+
+1. Create ssh key:
+
+```bash
+ssh-keygen -t ed25519 -N '' -C matrix
```
spantaleev, line 11: Maybe it should be pointed out that you can run this command on any machine, not necessarily on the Matrix server.

Adding `-f matrix-borg-backup` may make sense. Otherwise, someone might override their existing `~/.ssh/id_ed25519{,.pub}` files.

docs/configuring-playbook-backup-borg.md
```diff
@@ -0,0 +1,52 @@
+# Setting up borg backup (optional)
+
+The playbook can install and configure [borgbackup](https://www.borgbackup.org/) with [borgmatic](https://torsion.org/borgmatic/) for you.
+The backup will run based on `matrix_backup_borg_schedule` var (systemd timer calendar), default: 4am every day
+
+## Prerequisites
```
spantaleev, line 6: Should we say anything about what Borg providers one may use? Can any SSH-exposed server of your own serve as a Borg backup provider? What does it need for that (`borg` installed on the server? `rsync` installed on the server?)? We can let people discover this on the borgbackup site somewhere, but if there is some borgbackup documentation page about this, it'd be nice if we pointed to it directly from here.

spantaleev (changes requested):
Thank you for adding another backup option! We can never have enough of those ;)

I've submitted some comments - mostly about variables that could be named in a more consistent manner. We can merge this as it is, but if we'd like to rename them later, it'd be more trouble (adding deprecation notices in `valdiate_config.yml`), so it's better if we do it earlier.

roles/matrix-backup-borg/templates/config.yaml.j2
etkecc: It's done with 1 var to mount the whole /matrix dir into container. With multiple vars the human factor will be involved, so you accidentally may add something outside of mounted directory and your dirs will not be backuped.

Are you sure you want to change it?
To be honest, I look more into configuration extensions in the next iteration(-s), like it's done in other roles (eg synapse)

spantaleev: Oh, I see! You're right. Adding multiple sources will complicate things for mounting as well.

I think we can have both though?

```yaml
# A list of all directories that will be backed up.
# We only automatically mount `matrix_backup_borg_source_directory` into the borgbackup container.
# Should you override this and add additional directories to `matrix_backup_borg_source_directories`,
# you'll need to ensure these volumes are mounted into the borgbackup container.
# To mount additional volumes there, see `matrix_backup_borg_additional_volumes`.
matrix_backup_borg_source_directories: ["{{ matrix_backup_borg_source_directory }}"]

matrix_backup_borg_source_directory: "/matrix"
```

---------

Alternatively, we can have a `for` loop in the systemd `.service` file. There may be other problems with using multiple source directories though.

If you think this overly complicates things for no good reason, we can keep things as they are.

etkecc:
Will apply the suggestions a bit later today

docs/configuring-playbook-backup-borg.md
etkecc: Only private key is needed and I don't thing we need to add a public key var for backup purposes - you can restore public part from the private, IIRC

docs/configuring-playbook-backup-borg.md
etkecc: borg doesn't have vendor-lock, so any provider can be used. I'm not sure the configuring-**playbook**-backup-borg should have a copy of borg docs - it already has a link to the official site of the tool, where people can find docs and examples.

Some examples from borg docs (borgbackup.org -> click on docs): https://borgbackup.readthedocs.io/en/stable/usage/general.html#repository-urls

HarHarLinks: > Can any SSH-exposed server of your own serve as a Borg backup provider? What does it need for that (`borg` installed on the server?

yep, ssh and borg on the backup server should suffice. i think borg also comes packaged for some NAS distros etc which is nice

etkecc: moreover, you can backup to a local dir (usually some mounted volume), but this PR focuses on off-site backups

HarHarLinks: > any provider can be used. I'm not sure the configuring-playbook-backup-borg should have a copy of borg docs

I don't think we should put a list of providers, but I think it is reasonable to sum up in 1-2 sentences what it is about in the introduction/abstract paragraph for anyone not familiar, similar to 
> The playbook can install and configure [Honoroit](https://gitlab.com/etke.cc/honoroit) for you.
It's a bot you can use to setup your own helpdesk on matrix. 

or 

> Hookshot can bridge [Webhooks](https://en.wikipedia.org/wiki/Webhook) from software project management services such as GitHub, GitLab, JIRA, and Figma, as well as generic webhooks.

The essentials in my opinion are:
- "No vendor-lock" doesn't imply the same thing as "back up to your raspberry pi with a USB HDD at home" or "your NAS might already support receiving backups"
- short introduction with these key points: "borg is a deduplicating backup program with optional compression and encryption. That means your daily incremental backups can be stored in a fraction of the space and is safe weather you store it at home or a cloud service."

spantaleev (changes requested):
I found a few other potential issues. Sorry for the back and forth!

docs/configuring-playbook-backup-borg.md
```diff
@@ -0,0 +1,56 @@
+# Setting up borg backup (optional)
+
+The playbook can install and configure [borgbackup](https://www.borgbackup.org/) with [borgmatic](https://torsion.org/borgmatic/) for you.
+BorgBackup is a deduplicating backup program with optional compression and encryption.
+That means your daily incremental backups can be stored in a fraction of the space and is safe weather you store it at home or a cloud service.
```
HarHarLinks, line 5: Sorry, I think I typo'd this in my original comment
```suggestion
That means your daily incremental backups can be stored in a fraction of the space and are safe whether you store it at home or on a cloud service.
```

docs/configuring-playbook-backup-borg.md
```diff
@@ -2,7 +2,7 @@
 
 The playbook can install and configure [borgbackup](https://www.borgbackup.org/) with [borgmatic](https://torsion.org/borgmatic/) for you.
 BorgBackup is a deduplicating backup program with optional compression and encryption.
-That means your daily incremental backups can be stored in a fraction of the space and is safe weather you store it at home or a cloud service.
+That means your daily incremental backups can be stored in a fraction of the space and is safe wether you store it at home or on a cloud service.
```
HarHarLinks, line 5:
```suggestion
That means your daily incremental backups can be stored in a fraction of the space and is safe whether you store it at home or on a cloud service.
```
Nope, sorry, it's with an `h` in this case: https://www.merriam-webster.com/dictionary/whether

spantaleev:
Thanks a lot for adding this and for your patience fixing it up! :bow:

axiopaladin:
This might be a stupid question, but I couldn't find it obviously explained in any documentation: what timezone is the `backup_borg_schedule` variable supposed to be in? The default is "04:00", but is that in the matrix server's timezone? The remote server's timezone? UTC?

PR #1729, opened 2022-04-03
[STEAM] Move Docker source repository to a fixed version

RoiArthurB:
Current icewind1991 source is broken (the main problem is a package name which changed since the latest commit in Alpine's repo). A [PR is waiting from tilosp](https://github.com/icewind1991/mx-puppet-steam/pull/27) and fix that error.

A great side effect is that this new Dockerfile built and works on arm64 (tested for months now on my RP4) :)

This PR is bringing this 2 features (change build source + self-buildable for arm64)

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -908,7 +908,7 @@ matrix_mx_puppet_discord_database_password: "{{ '%s' | format(matrix_homeserver_
 # We don't enable bridges by default.
 matrix_mx_puppet_steam_enabled: false
 
-matrix_mx_puppet_steam_container_image_self_build: "{{ matrix_architecture != 'amd64' }}"
+matrix_mx_puppet_steam_container_image_self_build: "{{ matrix_architecture not in ['arm64', 'amd64'] }}"
 
 matrix_mx_puppet_steam_systemd_required_services_list: |
   {{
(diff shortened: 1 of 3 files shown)


group_vars/matrix_servers
```diff
@@ -1552,11 +1552,11 @@ matrix_ssl_domains_to_obtain_certificates_for: |
...
-	{
-		'amd64': 'amd64',
-		'arm32': 'arm32v6',
-		'arm64': 'arm64v8',
-	}[matrix_architecture]
+  {
+    'amd64': 'amd64',
+    'arm32': 'arm32v6',
+    'arm64': 'arm64v8',
+  }[matrix_architecture]
```
spantaleev, line 1559: Probably no need to change this?

RoiArthurB: Oups yes, wrongly commited, I'll fix it 🙈🙈🙈

RoiArthurB: Fixed 👍

spantaleev:
Thanks! :+1: 

Looks like `icewind1991/mx-puppet-steam` is unmaintained, given that @tilosp's fix hasn't been merged for a few months.

Upstream using unspecified (~`latest`) container images as the source (`FROM node:alpine AS builder`) is indeed fragile and shouldn't be done. Thankfully, the new repository does things better (`FROM node:14-alpine AS builder`)!

Issue #1625, opened 2022-02-13
dendrite has replaced naffka with nats.


corburn:
https://github.com/matrix-org/dendrite/blob/9ac27cabc5f624938264ddccf5500478d7d38bf3/CHANGES.md#dendrite-060-2022-01-28

spantaleev:
We know, it's the reason #1599 didn't get merged. We need to adapt the configuration and test it out.

Xemorph:
``Hello all,
for a first time installation it is relatively easy to customize the configurations, but for existing installations it might be difficult to upgrade. I refer here to a text section from the link of the first comment:

> Consolidated the federation API, federation sender and signing key server into a single component
If multiple databases are used, tables for the federation sender and signing key server should be merged into the federation API database (table names have not changed)

With version 0.6.x some databases were removed and replaced by a new one. In the config file `dendrite-config.yaml` the new database is titled as `federationapi`. During the upgrade the following should happen:
- Transfer table/s from DB `dendrite_federationsender` to the new DB `dendrite_federationapi`
- Transfer table/s from DB `dendrite_sigingkeyserver` to the new DB `dendrite_federationapi`
- Delete the following databases: `dendrite_naffka`, `dendrite_federationsender`, `dendrite_sigingkeyserver`, `dendrite_device`

**Note:** I don't know where the data from the database `dendrite_device` are migrated to!

---

Subsequently, other things would have to be adjusted, including the configuration file for Dendrite, the Docker command must be extended. The command could then look something like this:
```
ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-dendrite \
			--log-driver=none \
			--user={{ matrix_user_uid }}:{{ matrix_user_gid }} \
			--cap-drop=ALL \
			--read-only \
			--tmpfs=/tmp:rw,noexec,nosuid,size={{ matrix_dendrite_tmp_directory_size_mb }}m \
			--network={{ matrix_docker_network }} \
			{% if matrix_dendrite_container_http_host_bind_address and matrix_dendrite_http_bind_port %}
			-p {{ matrix_dendrite_container_http_host_bind_address }}:{{ matrix_dendrite_http_bind_port }} \
			{% endif %}
			{% if matrix_dendrite_container_https_host_bind_address and matrix_dendrite_https_bind_port %}
			-p {{ matrix_dendrite_container_https_host_bind_address }}:{{ matrix_dendrite_https_bind_port }} \
			{% endif %}
[...]

PR #1730, opened 2022-04-05
Dendrite 0.7.0

SaltireSoul:
An initial pass at dendrite 0.7.0 support

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1592,15 +1592,11 @@ matrix_postgres_additional_databases: |
     }] if (matrix_synapse_enabled and matrix_synapse_database_database != matrix_postgres_db_name and matrix_synapse_database_host == 'matrix-postgres') else [])
     +
     ([{
-      'name': matrix_dendrite_naffka_database,
-      'username': matrix_dendrite_database_user,
-      'password': matrix_dendrite_database_password,
-    },{
       'name': matrix_dendrite_appservice_database,
       'username': matrix_dendrite_database_user,
       'password': matrix_dendrite_database_password,
     },{
-      'name': matrix_dendrite_federationsender_database,
+      'name': matrix_dendrite_federationapi_database,
       'username': matrix_dendrite_database_user,
       'password': matrix_dendrite_database_password,
     },{
@@ -1616,19 +1612,15 @@ matrix_postgres_additional_databases: |
       'username': matrix_dendrite_database_user,
       'password': matrix_dendrite_database_password,
     },{
-      'name': matrix_dendrite_singingkeyserver_database,
-      'username': matrix_dendrite_database_user,
-      'password': matrix_dendrite_database_password,
-    },{
       'name': matrix_dendrite_syncapi_database,
       'username': matrix_dendrite_database_user,
       'password': matrix_dendrite_database_password,
     },{
-      'name': matrix_dendrite_account_database,
+      'name': matrix_dendrite_userapi_database,
       'username': matrix_dendrite_database_user,
       'password': matrix_dendrite_database_password,
     },{
-      'name': matrix_dendrite_device_database,
+      'name': matrix_dendrite_pushserver_database,
       'username': matrix_dendrite_database_user,
       'password': matrix_dendrite_database_password,
     },{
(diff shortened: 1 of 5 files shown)


spantaleev:
Thank you for bringing us up to date! :+1: 

This fixes #1625.

I wonder if it's a painless upgrade for users on pre-`0.7`.

Xemorph:
Most things look correct so far, but the configuration does not match the `dendrite-config.yml` from the ` v0.7.0` tag. In version `0.7.0` there is still an EDU server & the push server does not exist at that time. See [dendrite-config.yml](https://github.com/matrix-org/dendrite/blob/v0.7.0/dendrite-config.yamll) from tag `v0.7.0`.

@spantaleev 

> I wonder if it's a painless upgrade for users on pre-`0.7`.

I had started to update from version `0.6.5` to version `0.7.0` myself and had several problems with the databases. I was forced to reinstall the server completely. I'm not sure if it was a special case with me, but the commit should be tested first in my opinion.

spantaleev:
While I generally agree that we should be more careful and test things out before merging, it appears that a long time has passed and that we've been stuck on an old Dendrite version (0.5.1) for a long time. There doesn't seem to be anyone in the community that's using Dendrite 0.5.1 and caring to upgrade us to 0.6.0/0.7.0.

It's been a long time (2-3 months since [0.6.0 got released](https://github.com/matrix-org/dendrite/releases/tag/v0.6.0)) and no one has stepped up, so we can probably assume that we can either:

- go ahead and break things for existing users - there are either none of them or none who care much to actually do the migration in a way that works. Perhaps it's Dendrite's fault as well for not providing a cleaner migration path.
- at least be able to continue forward for new users (of who there will be a bigger number) as Dendrite moves forward and becomes more and more feasible. Becoming stuck on 0.5.1 for who knows how long more is not good
- breaking things for existing 0.5.1 users may be some necessary evil (sorry :disappointed:) that needs to be done to have some of them step up and actually fix up the playbook, so that upgrading works.

Xemorph:
Makes sense, then we should create a new base with Dendrite v0.7.0 & migrate reasonably from there. Only one thing worries me a lot, it might happen again soon that the database structure changes (Regarding the Push Server).

But we should revise this for Dendrite again:
```yaml
matrix_dendrite_base_path: "{{ matrix_base_data_path }}/dendrite"
matrix_dendrite_config_dir_path: "{{ matrix_dendrite_base_path }}/config"
matrix_dendrite_storage_path: "{{ matrix_dendrite_base_path }}/storage"
matrix_dendrite_media_store_path: "{{ matrix_dendrite_storage_path }}/media-store"
matrix_dendrite_nats_storage_path: "{{ matrix_dendrite_base_path }}/nats"
matrix_dendrite_ext_path: "{{ matrix_dendrite_base_path }}/ext"
```
because @spantaleev you said the following in the issue #1625 
> A `/matrix-jetstream-parent` (parent, child) kind of thing is not necessary to do here. We do it for the media store, because we're using Goofys for mounting S3 volumes and it's a workaround for some issue with that. Well, this is some `matrix-synapse` legacy. We don't really support Goofys/S3 for Dendrite anyway, so such parent/child hacks are unnecessary. We should probably simplify things.
> 
> For Jetstream, you can just mount a directory directly. We should probably introduce multiple storage paths. Right now we use `matrix_dendrite_storage_path`, but that's just the media store. We should have something like this:
> 
> * `matrix_dendrite_storage_base_path` - as a base path (instead of `matrix_dendrite_storage_path`, which we should deprecate)
> * `matrix_dendrite_storage_jetstream_path` - for Jetstream
> * `matrix_dendrite_storage_media_store_path` - for the media store

---

My suggested solution, off the top of my head:
```yaml
matrix_dendrite_base_path: "{{ matrix_base_data_path }}/dendrite"
matrix_dendrite_config_dir_path: "{{ matrix_dendrite_base_path }}/config"
matrix_dendrite_ext_path: "{{ matrix_dendrite_base_path }}/ext"
matrix_dendrite_storage_base_path: "{{ matrix_dendrite_base_path }}/storage"
matrix_dendrite_storage_jetstream_path: "{{ matrix_dendrite_storage_base_path }}/jetstream"
matrix_dendrite_storage_media_store_path: "{{ matrix_dendrite_storage_base_path }}/media-store"
```
Jetstream & Nats are the same. For the path we should rather go with `nats`, because inside this folder a separate folder `jetstream` is created.

PR #1738, opened 2022-04-09
Dendrite 0.8.1

SaltireSoul:
Minor changes to support Dendrite 0.8.1

Cleaned up config yaml based on [dendrite-config.yaml](https://github.com/matrix-org/dendrite/blob/main/dendrite-config.yaml) instead of [build/docker/config/dendrite.yaml](https://github.com/matrix-org/dendrite/blob/main/build/docker/config/dendrite.yaml)


spantaleev:
Seems like we're getting rid of the `push_server` definition then?

SaltireSoul:
I tested it last night it working without the push_server but in retrospect it is a newish feature [PR](https://github.com/matrix-org/dendrite/pull/1842) not an old one.  Looks like the default [dendrite-config.yaml](https://github.com/matrix-org/dendrite/blob/main/dendrite-config.yaml) just hasn't been updated for it

spantaleev:
Thanks! :+1:

PR #1741, opened 2022-04-10
Multi-IP coturn

heftyzauk:
Add support for multiple external turn IP addresses, this allows for better comptability with dualstack ipv4/ipv6 hosts, and is supported as per the documentation (point 6 here: https://matrix-org.github.io/synapse/latest/turn-howto.html#configuration)

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1144,7 +1144,7 @@ matrix_coturn_enabled: true
 
 matrix_coturn_container_image_self_build: "{{ matrix_architecture != 'amd64' }}"
 
-matrix_coturn_turn_external_ip_address: "{{ ansible_host }}"
+matrix_coturn_turn_external_ip_address: ["{{ ansible_host }}"]
 
 matrix_coturn_turn_static_auth_secret: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'coturn.sas') | to_uuid }}"
 
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1145,7 +1145,6 @@ matrix_coturn_enabled: true
 matrix_coturn_container_image_self_build: "{{ matrix_architecture != 'amd64' }}"
 
 matrix_coturn_turn_external_ip_address: "{{ ansible_host }}"
-matrix_coturn_turn_external_ip_addresses: [ "{{ matrix_coturn_turn_external_ip_address }}" ]
 
 matrix_coturn_turn_static_auth_secret: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'coturn.sas') | to_uuid }}"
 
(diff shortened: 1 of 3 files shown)


spantaleev:
Your approach breaks backward compatibility without any warning. Existing users that set `matrix_coturn_turn_external_ip_address` manually (e.g. `matrix_coturn_turn_external_ip_address: 1.2.3.4`) will find that the following Coturn configuration is generated if this PR were merged (unless looping over a string causes some error instead):

```
external-ip=1
external-ip=.
external-ip=2
external-ip=.
external-ip=3
external-ip=.
external-ip=4
```

It's also odd that a variable called `address` contains a list of `address`es and not a single one.

--------

A better way to do it would be to leave `matrix_coturn_turn_external_ip_address` as it is, and then introduce a new `matrix_coturn_turn_external_ip_addresses` list variable like this:

```yaml
matrix_coturn_turn_external_ip_addresses: ["{{ matrix_coturn_turn_external_ip_address }}"]
```

You can then do a `for` loop over it, as you do in your PR.

--------

Should we wish to not have 2 variables for this, we can also decide to deprecate `matrix_coturn_turn_external_ip_address` by adding a warning about it in `roles/matrix-coturn/tasks/validate_config.yml` and define `matrix_coturn_turn_external_ip_addresses` like this:

```yaml
matrix_coturn_turn_external_ip_addresses: ["{{ ansible_host }}"]
```

That warning code will look something like this: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5e00f0135ef53cad220b6e0a36ebc456f805c342/roles/matrix-corporal/tasks/validate_config.yml#L19-L27

I think having 2 variables is fine though. Existing users can continue using what they use now. New users that need to define a single IP address, can also continue using `matrix_coturn_turn_external_ip_address`. It's easier and more convenient not to deprecate `matrix_coturn_turn_external_ip_address`

heftyzauk:
Thanks for the review, I went a bit one track minded! 

Have updated with your suggestions, don't see any harm in keeping the "address" singular as you mentioned too.

group_vars/matrix_servers
```diff
@@ -1145,6 +1145,7 @@ matrix_coturn_enabled: true
 matrix_coturn_container_image_self_build: "{{ matrix_architecture != 'amd64' }}"
 
 matrix_coturn_turn_external_ip_address: "{{ ansible_host }}"
+matrix_coturn_turn_external_ip_addresses: [ "{{ matrix_coturn_turn_external_ip_address }}" ]
```
spantaleev, line 1148: This default should go (as-is) into `roles/matrix-coturn/defaults/main.yml`, because it applies not just to the playbook distribution, but also to the `matrix-coturn` role when used standalone (without the playbook overrides coming from `group_vars/matrix_servers`).

heftyzauk: Resolved

spantaleev:
I've merged this manually, because I had to fix it up a bit in [f4ba995](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/f4ba995d9b85b527205bcb32c266f5e98f4ef9d3).

Thank you for your effort! :+1:

PR #1748, opened 2022-04-12
Improve borg backup instructions

ofalvai:
I set up borg backup using the instructions and BorgBase as a hosted provider, but I found the documentation a bit unclear and incomplete.

This PR adds more context and explanation.

### DIFF
--- a/docs/configuring-playbook-backup-borg.md
+++ b/docs/configuring-playbook-backup-borg.md
@@ -4,17 +4,23 @@ The playbook can install and configure [borgbackup](https://www.borgbackup.org/)
 BorgBackup is a deduplicating backup program with optional compression and encryption.
 That means your daily incremental backups can be stored in a fraction of the space and is safe whether you store it at home or on a cloud service.
 
-The backup will run based on `matrix_backup_borg_schedule` var (systemd timer calendar), default: 4am every day
+You will need a remote server where borg will store the backups. There are hosted, borg compatible solutions available, such as [BorgBase](https://www.borgbase.com).
+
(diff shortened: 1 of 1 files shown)


docs/configuring-playbook-backup-borg.md
```diff
@@ -35,17 +41,21 @@ matrix_backup_borg_location_repositories:
 matrix_backup_borg_storage_encryption_passphrase: "PASSPHRASE"
 matrix_backup_borg_ssh_key_private: |
 	PRIVATE KEY
+matrix_backup_borg_location_source_directories:
+  - "{{ matrix_base_data_path }}"
```
spantaleev, line 45: This is the default as per `group_vars/matrix_servers` already, isn't it?

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/ce675b0e795ffacc39c37d2ca26aa4659509d9e2/group_vars/matrix_servers#L1084-L1085

We may mention the `matrix_backup_borg_location_source_directories` variable, as you're doing below, but asking people to redefine whatever is the default doesn't seem necessary.

ofalvai: My bad, I only looked at `main.yml` and saw the empty variable. Please check the updated version now.

spantaleev:
Thanks! :+1: 

----

As for `matrix_backup_borg_location_exclude_patterns`, we already define a default in `group_vars/matrix_servers`. The current documentation makes it sound like we don't, doesn't it?

On a related note, we should probably exclude `/matrix/postgres/data` somehow, but I think @rakshazi said he'd work on proper Postgres dumping and excluding the raw files in a separate PR.

PR #1750, opened 2022-04-13
Change borg backup service file to type=simple

Eagle-251:
This change is to prevent the playbook hanging on: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/aef2c4c32e8fb8249bde15066df6075cfaf95454/roles/matrix-common-after/tasks/start.yml#L18

It seems that setting the service type to `oneshot` caused the unit to sit at activating and never reach started which then won't satisfy the check in the start tasks.
The [systemd.service man](https://www.freedesktop.org/software/systemd/man/systemd.service.html) seems to imply with oneshot it would:
> directly transition from "activating" to "deactivating" or "dead" 
which I imagine would cause it fail after completing the backup.

`Type=simple` seems to be the choice in other systemd files in the playbook, so this would be following that.

### DIFF
--- a/roles/matrix-backup-borg/templates/systemd/matrix-backup-borg.service.j2
+++ b/roles/matrix-backup-borg/templates/systemd/matrix-backup-borg.service.j2
@@ -11,7 +11,7 @@ Wants={{ service }}
 DefaultDependencies=no
 
 [Service]
-Type=oneshot
+Type=simple
 Environment="HOME={{ matrix_systemd_unit_home_path }}"
 ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} kill matrix-backup-borg 2>/dev/null || true'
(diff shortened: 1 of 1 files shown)


Eagle-251:
That seems like a sensible way to run it!

Apologies if I am misunderstanding things but isn't that exactly what the [init.yml](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/aef2c4c32e8fb8249bde15066df6075cfaf95454/roles/matrix-backup-borg/tasks/init.yml) file in the borg backup role does?

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/aef2c4c32e8fb8249bde15066df6075cfaf95454/roles/matrix-backup-borg/tasks/init.yml#L3

spantaleev:
Yes, you're right. I was looking at the wrong place and didn't notice it. This `.service` probably shouldn't be there.

PR #1753, opened 2022-04-14
Upgrade Hookshot (1.4.0 -> 1.5.0)

HarHarLinks:
No API or config changes = no playbook changes, just a simple version bump

https://github.com/matrix-org/matrix-hookshot/releases/tag/1.5.0


HarHarLinks:
This is fine to merge, but at some point ability of renaming the ghost users should be added properly because these need to go into the registration.yml template as well

spantaleev:
Thanks! 👍 

What is this ghost user renaming that you're referring to?

HarHarLinks:
> ghost user

github hooks (may) get a user with mxid `@_github_something:example.org`, same for gitlab, etc. this used to be hardcoded but now seems to be optionally customizable. in which case you need to update the registration too, to allow the appservice to use that namespace.

PR #1754, opened 2022-04-14
matrix-backup-borg: add ability to backup to unencrypted repositories

thebiblelover7:


thebiblelover7:
This PR could be resolved by #1755 , so I don't know what should be done.

etkecc:
Yes, you can do that with #1755 with configuration extension

spantaleev:
How common are unencrypted Borg repositories? Perhaps backing up to such repositories is much quicker on low-power hardware, because it skips having to encrypt everything.

If it's a common use case, regardless of #1755, we can keep your PR to have a dedicated option for easy usage. Is setting this "ok" flag to `true` enough to do use an unencrypted repository? Or is something else necessary? It seems like if you leave `matrix_backup_borg_storage_encryption_passphrase` empty, `roles/matrix-backup-borg/tasks/validate_config.yml` will choke. We should probably fix this, and potentially update the docs too.

-------

You've also named the variable `matrix_backup_borg_unknown_unencrypted_access`, which is inconsistent with how we name variables: `matrix_(SERVICE_NAME)_(ORIGINAL_NAME_OF_CONFIGURATION_SETTING)`.

It should be called `matrix_backup_borg_unknown_unencrypted_repo_access_is_ok` instead.

--------

Another good practice is to use [`|to_json`](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#formatting-data-yaml-and-json) when spewing variables into JSON/YAML configuration files.

```diff
-unknown_unencrypted_repo_access_is_ok: {{ matrix_backup_borg_unknown_unencrypted_access }}
+unknown_unencrypted_repo_access_is_ok: {{ matrix_backup_borg_unknown_unencrypted_repo_access_is_ok|to_json }}
```

thebiblelover7:
@spantaleev I am currently backing up to my raspberry pi, and I don't need encryption which is why I disabled it on my repo. 

I will try to fix those things soon.

etkecc:
The #1755 has been merged. Could you try to use configuration extension for your use case and post feedback here, please?

It works the same as in other roles (eg synapse)

thebiblelover7:
> The #1755 has been merged. Could you try to use configuration extension for your use case and post feedback here, please?
> 
> It works the same as in other roles (eg synapse)

What I've seen is that it fails because you don't pass a password to `matrix_backup_borg_storage_encryption_passphrase`. As long as you pass it anything, it doesn't fail. I'm currently trying to find a way to have a conditional `with_items` so that it can only be verified when `matrix_backup_borg_unknown_unencrypted_repo_access_is_ok` is false.

I'm not very good with ansible, so any help is appreciated!

spantaleev:
> I'm currently trying to find a way to have a conditional `with_items` so that it can only be verified when `matrix_backup_borg_unknown_unencrypted_repo_access_is_ok` is false.

You can see some conditional `with_items` usage here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5c1ee66e8dab15485b8bcb69e0a153012d6d40cf/roles/matrix-client-element/tasks/setup_install.yml#L3-L13

Alternatively, you can use another `fail` task for `matrix_backup_borg_storage_encryption_passphrase` which is "hidden" behind `when: not matrix_backup_borg_unknown_unencrypted_repo_access_is_ok`. Example:

```yaml
- name: Fail if encryption enabled and encryption passphrase empty
  fail:
    msg: >-
      You need to define a required configuration setting (`matrix_backup_borg_storage_encryption_passphrase`).
  when: "matrix_backup_borg_storage_encryption_passphrase == '' and not matrix_backup_borg_unknown_unencrypted_repo_access_is_ok"
```

etkecc:
I suppose you need to update validate_config as well, because it checks for encryption passphrase and some other things.
As @spantaleev suggested - add a new fail when no passphrase is given and unencrypted backups disabled in the validate_config file

Btw, don't forget to pull new changes first

roles/matrix-backup-borg/tasks/validate_config.yml
```diff
@@ -7,4 +7,12 @@
   with_items:
     - "matrix_backup_borg_ssh_key_private"
     - "matrix_backup_borg_location_repositories"
-    - "matrix_backup_borg_storage_encryption_passphrase"
+#    - "matrix_backup_borg_storage_encryption_passphrase"
+#      when: "matrix_backup_borg_unknown_unencrypted_repo_access_is_ok == false"
+#    - "matrix_backup_borg_storage_encryption_passphrase"
```
spantaleev, line 12: You can get rid of this. There's no need for keeping commented-out code.

thebiblelover7:
@spantaleev Seems to me like this is ready to be merged. Let me know if there is anything else I should change!

I've tested it on my system and it seems to work, too.

roles/matrix-backup-borg/tasks/validate_config.yml
```diff
@@ -7,4 +7,9 @@
   with_items:
     - "matrix_backup_borg_ssh_key_private"
     - "matrix_backup_borg_location_repositories"
-    - "matrix_backup_borg_storage_encryption_passphrase"
+
+- name: Fail if encryption passphrase is undefined unless repository is unencrypted
+  fail:
+    msg: >-
+      You need to define a required passphrase.
+  when: "matrix_backup_borg_storage_encryption_passphrase == '' and not matrix_backup_borg_encryption == 'none'"
```
spantaleev, line 15: So, to enable the usage of non-encrypted repositories, one will need both `matrix_backup_borg_unknown_unencrypted_repo_access_is_ok: true` and `matrix_backup_borg_encryption: 'none'`?

This makes the comment above the `matrix_backup_borg_encryption` definition in `defaults/main.yml` incorrect:

> # borg encryption mode, only repokey-* is supported
> matrix_backup_borg_encryption: repokey-blake2

If so, I suppose we can change the default definition for `matrix_backup_borg_unknown_unencrypted_repo_access_is_ok` to be like this:

```yaml
matrix_backup_borg_unknown_unencrypted_repo_access_is_ok: "{{ matrix_backup_borg_encryption == 'none' }}"
```

Then, to enable unencrypted backup usage, we'll only need `matrix_backup_borg_encryption: 'none'`.

-----

Additionally, this `when` statement here (`not matrix_backup_borg_encryption == 'none'`) is confusing and can be done like this `matrix_backup_borg_encryption != 'none'`.

------

Also, `'none'` (a string) can be confused with `none` (`null`).

Perhaps what you mean is `matrix_backup_borg_encryption is none` or `matrix_backup_borg_encryption is not none`, instead of comparing to a string.

thebiblelover7: > Also, `'none'` (a string) can be confused with `none` (`null`).
> 
> Perhaps what you mean is `matrix_backup_borg_encryption is none` or `matrix_backup_borg_encryption is not none`, instead of comparing to a string.

The borgmatic config uses this term, should I change it to something else?

spantaleev:
Thank for contributing this and for the patience! :+1:

PR #1755, opened 2022-04-14
matrix-backup-borg: integrate postgres backups, add extended borgmatic configuration

etkecc:


roles/matrix-backup-borg/tasks/setup_install.yml
```diff
@@ -1,4 +1,7 @@
 ---
+- import_tasks: "{{ role_path }}/tasks/util/detect_existing_postgres_version.yml"
```
spantaleev, line 2: Instead of duplicating this detection code, why not not import `roles/matrix-postgres/tasks/util/detect_existing_postgres_version.yml`?

I don't know why the `matrix-postgres-backup` role duplicates the detection code as well. Perhaps it was easier to do? Perhaps including from the `matrix-postgres` role did not work or some changes are required to make it work?

Supposing that it works, after this `import_tasks` call, instead of  accepting all results, you may add another `fail` task for versions that you don't support.

```yaml
- block:
  - import_tasks: "{{ role_path }}/../matrix-postgres/tasks/util/detect_existing_postgres_version.yml"
    when: matrix_backup_borg_postgresql_enabled|bool
  
  - name: Fail if detected Postgres version is unsupported
    fail:
      msg: "You cannot use borg backup with such an old version ({{ matrix_postgres_detected_version }}) of Postgres. Consider upgrading - link to docs for upgrading Postgres here"
    when: "matrix_postgres_detected_version not in matrix_backup_borg_supported_postgres_versions"
  
  - name: Set the correct borg backup version to use
    set_fact:
       matrix_backup_borg_version: "{{ matrix_postgres_detected_version }}"
  when: "matrix_backup_borg_version != '' and matrix_backup_borg_postgresql_databases_hostname == 'matrix-postgres'"

# Do something for when matrix_backup_borg_postgresql_databases_hostname is not 'matrix-postgres' (external Postgres server).
```

`matrix_backup_borg_supported_postgres_versions` can be defined as `['12', '13', `'14']` in `roles/matrix-backup-borg/defaults/main.yml`.

roles/matrix-backup-borg/tasks/setup_install.yml
```diff
@@ -1,4 +1,7 @@
...
+  when: 'matrix_backup_borg_enabled|bool and matrix_backup_borg_postgresql_enabled|bool and matrix_postgres_backup_postgres_data_path != ""'
```
spantaleev, line 3: `matrix_backup_borg_enabled|bool` should be unnecessary in `tasks/setup_install.yml`, because `tasks/setup_install.yml` only gets included (by `tasks/main.yml`) with the same condition (`matrix_backup_borg_enabled|bool`).

roles/matrix-backup-borg/templates/config.yaml.j2
```diff
@@ -26,6 +26,16 @@ consistency:
     - archives
 
 hooks:
+{% if matrix_backup_borg_postgresql_enabled %}
```
spantaleev, line 29: Adding ` and matrix_backup_borg_postgresql_databases|length > 0` here may help prevent generate an empty `postgresql_databases:` list under some weird conditions (Postgres backups enabled, but list of database is empty for whatever strange reason).

roles/matrix-backup-borg/templates/config.yaml.j2
```diff
@@ -26,6 +26,16 @@ consistency:
...
+  postgresql_databases:
+  {% for database in matrix_backup_borg_postgresql_databases %}
+  - name: {{ database }}
+    hostname: {{ matrix_backup_borg_postgresql_databases_hostname }}
+    username: {{ matrix_backup_borg_postgresql_databases_username }}
+    password: {{ matrix_backup_borg_postgresql_databases_password }}
```
spantaleev, line 35: `|to_json` for `username` and `password` is probably important. Some people manually define very strong passwords, which contain special characters that may break the YAML syntax here. From what I remember, we've had similar problems before.

Using `|to_json` everywhere just to be on the same side is preferable.

spantaleev (changes requested):
I've submitted some feedback with potential improvements.

roles/matrix-backup-borg/defaults/main.yml
```diff
@@ -1,17 +1,17 @@
 ---
 matrix_backup_borg_enabled: true
 
+matrix_backup_borg_base_path: "{{ matrix_base_data_path }}/backup-borg"
+matrix_backup_borg_config_path: "{{ matrix_backup_borg_base_path }}/config"
+
 matrix_backup_borg_container_image_self_build: false
 matrix_backup_borg_docker_repo: "https://github.com/borgmatic-collective/docker-borgmatic"
```
spantaleev, line 8: Is this correct, by the way?

If self-building, wouldn't we want to self-build https://gitlab.com/etke.cc/borgmatic instead?

roles/matrix-backup-borg/tasks/setup_install.yml
etkecc: I changed a bit your suggestion, because it doesn't matter if postgres is self-hosted or not for backups

spantaleev:
Looks great! Thank you for adding Borg support, which also includes Postgres backups! :+1:

PR #1759, opened 2022-04-17
optional media cache with range requests support

borisrunakov:
Not thoroughly tested but looks good for now!
you can verify cache access with `inotifywait -m -r /matrix/nginx-proxy/data/`


borisrunakov:
3rd day testing today without issues. I think you can merge :)

etkecc:
why not use tmpfs for that?
With your approach cache will be saved on disk (persistent), but if you check the nginx systemd unit template, you will find that you can bind tmpfs volume (ephemeral) to the container.

Could you use it instead? I doubt saving such ephemeral data persistently has any advantages against tmpfs

borisrunakov:
> why not use tmpfs for that? With your approach cache will be saved on disk (persistent), but if you check the nginx systemd unit template, you will find that you can bind tmpfs volume (ephemeral) to the container.
> 
> Could you use it instead? I doubt saving such ephemeral data persistently has any advantages against tmpfs

so `matrix_nginx_proxy_synapse_cache_path: "/tmp/nginx-proxy/cache"` for example ?

etkecc:
you don't need to add a new var for it, just mount a tmpfs somewhere inside the container and use that internal path in the nginx config

borisrunakov:
I think you need a var for it in case `matrix_nginx_proxy_enabled: false` like me, right ?

etkecc:
I didn't think you use the playbook's nginx config with disabled playbook's nginx.
In that case - you're right

borisrunakov:
But in general yes, this solution can be of course further optimized. There can a be an additional `{% if matrix_nginx_proxy_enabled %}` for example. Or a more general solution that applies irrelevant to `matrix_nginx_proxy_enabled`

spantaleev:
Using `tmpfs` sounds good!

For when `matrix-nginx-proxy` is disabled, you can probably just drop support for this feature.. Unless you want to think of some other workaround.

borisrunakov:
Please check now.

roles/matrix-nginx-proxy/defaults/main.yml
```diff
@@ -557,6 +557,16 @@ matrix_nginx_proxy_synapse_media_repository_locations: []
 matrix_nginx_proxy_synapse_user_dir_locations: []
 matrix_nginx_proxy_synapse_frontend_proxy_locations: []
 
+# synapse content caching
+matrix_nginx_proxy_synapse_cache_enabled: false
+matrix_nginx_proxy_synapse_cache_path: "{{ matrix_nginx_proxy_data_path+'/cache' if matrix_nginx_proxy_synapse_cache_enabled else '/tmp/cache' }}"
```
spantaleev, line 562: Did you mean `if matrix_nginx_proxy_enabled` instead of `if matrix_nginx_proxy_synapse_cache_enabled`?

roles/matrix-nginx-proxy/tasks/setup_nginx_proxy.yml
```diff
@@ -21,6 +21,7 @@
     - "{{ matrix_nginx_proxy_base_path }}"
     - "{{ matrix_nginx_proxy_data_path }}"
     - "{{ matrix_nginx_proxy_confd_path }}"
+    - "{{ matrix_nginx_proxy_synapse_cache_path }}"
```
spantaleev, line 24: This has the potential to attempt to create `/tmp/cache` on the host for certain setups (when `matrix_nginx_proxy_enabled: false`). It should probably be made conditional somehow.

roles/matrix-nginx-proxy/templates/systemd/matrix-nginx-proxy.service.j2
```diff
@@ -22,6 +22,9 @@ ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-nginx-proxy \
 			--cap-drop=ALL \
 			--read-only \
 			--tmpfs=/tmp:rw,noexec,nosuid,size={{ matrix_nginx_proxy_tmp_directory_size_mb }}m \
+			{% if matrix_nginx_proxy_enabled and matrix_nginx_proxy_synapse_cache_enabled %}
```
spantaleev, line 25: I believe that the `matrix-nginx-proxy.service` file is only created if `matrix-nginx-proxy` is enabled (`matrix_nginx_proxy_enabled`), so adding this in the `if` statement is superfluous.

We should leave `if matrix_nginx_proxy_synapse_cache_enabled` only.

Which makes me think.. if this cache is just for Synapse, and we're calling the variable and cache zone name "synapse", we should probably make the directory `/tmp/synapse-cache` instead?

spantaleev (changes requested):

roles/matrix-nginx-proxy/defaults/main.yml
borisrunakov: yeap you re right !

borisrunakov: and the conditional is the other way around, i.e.: `matrix_nginx_proxy_synapse_cache_path: "{{ '/tmp/cache' if matrix_nginx_proxy_enabled else matrix_nginx_proxy_data_path+'/cache' }}"`

roles/matrix-nginx-proxy/tasks/setup_nginx_proxy.yml
borisrunakov: I try to make it conditional like so : 
```
- name: Ensure Matrix nginx-proxy paths exist
  file:
    path: "{{ item.directory }}"
    state: directory
    mode: 0750
  with_items:
    - { directory: "{{ matrix_nginx_proxy_base_path }}" }
    - { directory: "{{ matrix_nginx_proxy_data_path }}" }
    - { directory: "{{ matrix_nginx_proxy_confd_path }}" }
    - { directory: "{{ matrix_nginx_proxy_synapse_cache_path }}", when: matrix_nginx_proxy_synapse_cache_enabled and not matrix_nginx_proxy_enabled }
```
however yamllint check complains for "too many spaces inside spaces". 
Any hint ?

borisrunakov: changed it to :
```
- name: Ensure Matrix nginx-proxy paths exist
  file:
    path: "{{ item.directory }}"
    state: directory
    mode: 0750
  with_items:
    - directory: "{{ matrix_nginx_proxy_base_path }}"
    - directory: "{{ matrix_nginx_proxy_data_path }}"
    - directory: "{{ matrix_nginx_proxy_confd_path }}"
    - directory: "{{ matrix_nginx_proxy_synapse_cache_path }}"
      when: "matrix_nginx_proxy_synapse_cache_enabled and not matrix_nginx_proxy_enabled"
  ```
  and the checks passed

roles/matrix-nginx-proxy/templates/systemd/matrix-nginx-proxy.service.j2
borisrunakov: You are right. Changed it accordingly

spantaleev:
Thank you for making media work better (faster for subsequent access and seekable)! :+1: 

This is currently:

- not enabled by default. We should probably enable it at some point. Perhaps we should test it a little more first
- not doing anything when Synapse is running in non-worker mode (or even when in worker mode, but without media workers?). It'd be great if we can bring this to work even for people that use Synapse without workers.

PR #1766, opened 2022-04-19
update synapse 1.56.0 -> 1.57.0

etkecc:
no config changes

### DIFF
--- a/roles/matrix-synapse/defaults/main.yml
+++ b/roles/matrix-synapse/defaults/main.yml
@@ -9,7 +9,7 @@ matrix_synapse_container_image_self_build_repo: "https://github.com/matrix-org/s
 
 matrix_synapse_docker_image: "{{ matrix_synapse_docker_image_name_prefix }}matrixdotorg/synapse:{{ matrix_synapse_docker_image_tag }}"
 matrix_synapse_docker_image_name_prefix: "{{ 'localhost/' if matrix_synapse_container_image_self_build else matrix_container_global_registry_prefix }}"
-matrix_synapse_version: v1.56.0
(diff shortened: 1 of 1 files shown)


spantaleev:
Thanks! :+1: 

------

[This changelog entry for Synapse v1.57](https://github.com/matrix-org/synapse/blob/v1.57.0rc1/docs/upgrade.md#changes-to-database-schema-for-application-services) may be important to some users:

> # Changes to database schema for application services
>
> Synapse v1.57.0 includes a https://github.com/matrix-org/synapse/pull/12209 to the way transaction IDs are managed for application services. If your deployment uses a dedicated worker for application service traffic, it must be stopped when the database is upgraded (which normally happens when the main process is upgraded), to ensure the change is made safely without any risk of reusing transaction IDs.
> 
> Deployments which do not use separate worker processes can be upgraded as normal. Similarly, deployments where no applciation services are in use can be upgraded as normal.

etkecc:
But the `start` tag does the full stop

spantaleev:
It does bring all services to a full stop, and then starts them again.

Various `matrix-synapse-worker-*.service` services depend on `matrix-synapse.service` (and `sleep` 5 seconds), so the main one should be starting before worker processes. Although.. those systemd dependencies and delays are not very reliable, so I suppose some worker could potentially start while `matrix-synapse.service` is still starting up or is slowly running some migrations.

From what I understand, it's safest if some application service worker only gets to start **after** `matrix-synapse.service` has managed to fully complete the migration.

---

Well, let's ignore it and see what happens. :crossed_fingers:

etkecc:
Could you add a notice in the changelog about it?

PR #1768, opened 2022-04-19
matrix-backup-borg - cleanup the .service from the systemd list

etkecc:
ref: #1750


spantaleev:
Thanks for the fixup! :+1:

Issue #176, opened 2019-05-11
Synapse S3 Storage Provider instead of Goofys?


tompinzler:
Just getting started with that playbook (thanks for creating it!) and want to use S3 as a storage back-end. I realized that you are integrating S3 using Goofys instead of the official [synapse-s3-storage-provider](https://github.com/matrix-org/synapse-s3-storage-provider).

I would think that Goofys is more like a workaround for applications that don't natively support S3.

Wouldn't synapse-s3-storage-provider be the natural choice to connect Matrix to S3?

spantaleev:
You're right! Using Goofys is a workaround (and quite hacky implementation-wise).
It'd be much better if we support something else (also see #67).

A few things that are worrisome about [synapse-s3-storage-provider](https://github.com/matrix-org/synapse-s3-storage-provider):
- requires [boto3](https://github.com/boto/boto3), so likely just adding the module to the Synapse Docker container is not enough to enable it. Unless the Docker image already includes boto3 (doubtful).
- it doesn't support other (S3-like) object stores (work happening/stuck here: https://github.com/matrix-org/synapse-s3-storage-provider/pull/13). Even our Goofys usage didn't until recently though (see #132)
- seems unmaintained (some seemingly important pull requests being stuck; some issues saying "not working"). This may just be a wrong first impression though.
- seems like Synapse would still store files locally [according to this comment](https://github.com/matrix-org/synapse-s3-storage-provider/issues/6#issuecomment-419601139) as well as remotely on S3. What happens when the local cache fills up? Does Synapse trim it before we run out of disk space or?

We should try to clear some of these up, but.. for now, it seems like relying on the Goofys workaround is easier and safer.

Also, adding support for synapse-s3-storage-provider doesn't mean that we need to get rid of Goofys or support all of the features we have there. It can be a completely separate option that we support, with its own pros and cons. But before we do that, we should figure out what we gain and whether anyone gains enough to spend the time to work on it.

tompinzler:
Thanks for the detailed answer. Totally understand your decision now.

ptman:
synapse-s3-storage-provider seems to support other S3-compatible stores now

dotZoki:
It would be nice if support for synapse-s3-storage-provider is added.

shalzz:
Hi, time to maybe revisit this?

[...]

PR #1769, opened 2022-04-19
Added official S3 Storage Module for Synapse

BackInBash:
Hey 🖖,

this is a Pull to implement the [S3 storage backend](https://github.com/matrix-org/synapse-s3-storage-provider) as described in #176.

In order for the Plugin to work I modified the build process of the synapse container and injected the module installation.
As of now its doing its thing, but uploads both to the local and the remote store. In any possible configuration.

https://github.com/BackInBash/matrix-docker-ansible-deploy/blob/542e8ec119cb7d093ddeafb02ccdbbc980fe32ab/roles/matrix-synapse/defaults/main.yml#L583-L588


roles/matrix-synapse/tasks/synapse/setup_install.yml
```diff
@@ -47,7 +56,7 @@
       environment:
         DOCKER_BUILDKIT: 1
       when: "matrix_synapse_git_pull_results.changed|bool or matrix_synapse_docker_image_check_result.stdout == ''"
-  when: "matrix_synapse_container_image_self_build|bool"
+  when: "matrix_synapse_container_image_self_build|bool or matrix_synapse_s3_media_store_enabled|bool"
```
spantaleev, line 59: I think that instead of adding ` or matrix_synapse_s3_media_store_enabled|bool` here, we should instead influence the `matrix_synapse_container_image_self_build` variable.

In `roles/matrix-synapse/defaults/main.yml`:

```diff
-matrix_synapse_container_image_self_build: false
+matrix_synapse_container_image_self_build: "{{ matrix_synapse_s3_media_store_enabled }}"
```

And in `group_vars/matrix_servers`:

```diff
-matrix_synapse_container_image_self_build: "{{ matrix_architecture not in ['arm64', 'amd64'] }}"
+matrix_synapse_container_image_self_build: "{{ matrix_architecture not in ['arm64', 'amd64'] or matrix_synapse_s3_media_store_enabled }}"
```

------

For a while, I thought that redefining `matrix_synapse_container_image_self_build` is dirtier, but looking a few lines below (for the "Ensure Synapse Docker image is pulled"), it seems like your change introduces a bug.

The "Ensure Synapse Docker image is pulled" task's `when` either needs to be changed:

```diff
-when: "not matrix_synapse_container_image_self_build"
+when: "not matrix_synapse_container_image_self_build or not matrix_synapse_s3_media_store_enabled"
```

.. or, we need to influence the `matrix_synapse_container_image_self_build` variable as described above and then it will all work.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -1039,6 +1039,27 @@ media_store_path: "/matrix-media-store-parent/{{ matrix_synapse_media_store_dire
 #    config:
 #       directory: /mnt/some/other/directory
 
+# Matrix S3 Storage Provider
+{% if matrix_synapse_s3_media_store_enabled %}
+media_storage_providers:
+- module: s3_storage_provider.S3StorageProviderBackend
+  # In the context of this module the local store means the S3 Bucket
+  # and the remote store the local volume ¯\_(ツ)_/¯ (i swaped it for convenience)
```
spantaleev, line 1047: While swapping does make some sense (`store_local` being the S3 bucket is weird indeed..), I think it may introduce some confusion sometimes.

We normally name configuration keys as they appear upstream (`matrix_SERVICE_SETTING`), so someone looking at `roles/matrix-synapse/defaults/main.yml` and configuring `matrix_synapse_s3_media_store_remote` may think they're influencing the `store_remote` configuration property. In fact, they're influencing `store_local`.

Since both things are confusing, we should probably do something else -- using completely different variables (`matrix_synapse_s3_media_store_on_s3` and `matrix_synapse_s3_media_store_on_local_filesystem`, for example).

Keeping your comment about the meaning of `store_local` and `store_remote` will still be helpful, so I think we should leave it there.

The comments around `matrix_synapse_s3_media_store_on_s3` and `matrix_synapse_s3_media_store_on_local_filesystem` in `roles/matrix-synapse/defaults/main.yml` may point out which configuration key they influence.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -1039,6 +1039,27 @@ media_store_path: "/matrix-media-store-parent/{{ matrix_synapse_media_store_dire
...
+  store_local: {{ matrix_synapse_s3_media_store_remote }}
+  store_remote: {{ matrix_synapse_s3_media_store_local }}
+  store_synchronous: {{ matrix_synapse_s3_media_store_sync }}
```
spantaleev, line 1050: This does not follow our naming conventions (`matrix_SERVICE_SETTING`).

It should be called `matrix_synapse_media_storage_provider_s3_store_synchronous` instead.

The same applies to various other options as well (`region` -> `region_name`, `access` -> `access_key_id`, etc.).

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -1039,6 +1039,27 @@ media_store_path: "/matrix-media-store-parent/{{ matrix_synapse_media_store_dire
...
+  config:
+    bucket: {{ matrix_synapse_s3_media_store_bucket }}
+    # All of the below options are optional, for use with non-AWS S3-like
+    # services, or to specify access tokens here instead of some external method.
+    region_name: {{ matrix_synapse_s3_media_store_region }}
+    endpoint_url: {{ matrix_synapse_s3_media_store_endpoint }}
+    access_key_id: {{ matrix_synapse_s3_media_store_access }}
+    secret_access_key: {{ matrix_synapse_s3_media_store_key }}
```
spantaleev, line 1058: Using [`|to_json`](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#formatting-data-yaml-and-json) for all of these is safer.

```diff
-secret_access_key: {{ matrix_synapse_s3_media_store_key }}
+secret_access_key: {{ matrix_synapse_media_storage_provider_s3_secret_access_key|to_json }}
```

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -1039,6 +1039,27 @@ media_store_path: "/matrix-media-store-parent/{{ matrix_synapse_media_store_dire
 #    config:
 #       directory: /mnt/some/other/directory
 
+# Matrix S3 Storage Provider
+{% if matrix_synapse_s3_media_store_enabled %}
+media_storage_providers:
```
spantaleev, line 1044: A better way to inject this configuration is to do it how we do for Synapse modules (see `modules:` in `roles/matrix-synapse/templates/synapse/homeserver.yaml.j2` and the `matrix_synapse_modules` list variable: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/dd83d88cb04791a19f343fd3750f974a3a19c64b/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2#L24

Certain tasks in the playbook may inject modules into the `matrix_synapse_modules` at runtime, and then the whole `matrix_synapse_modules` list is just spewed out into the configuration.

The benefit of this is that users can manipulate the `matrix_synapse_modules` (and in our case `matrix_synapse_media_storage_providers`) list manually as well. If they enable some ready-made media storage provider (`matrix_synapse_media_storage_provider_s3_enabled`) it will get injected into their list without overwriting it.

To see how some module injects itself into the list, see how shared-secret-auth does it: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/dd83d88cb04791a19f343fd3750f974a3a19c64b/roles/matrix-synapse/tasks/ext/shared-secret-auth/setup_install.yml#L26-L37

spantaleev (changes requested):
Hey! Thanks for contributing this! I'm sending some feedback for how to make it more consistent with the rest of the playbook.

BackInBash:
Thanks for the review. I'll looking into it ;)

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
BackInBash: Added the `|toJson` attribute to all String based defined input variables.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
BackInBash: All vars renamed.

docs/configuring-playbook-synapse.md
```diff
@@ -74,3 +74,30 @@ In case you encounter errors regarding the parsing of the variables, you can try
...
+## Synapse S3 Storage Provider
+
+If you'd like to use Synapse with a S3 storage backend.
+This method is implementing the S3 module for Synapse from [Matrix-Org](https://github.com/matrix-org/synapse-s3-storage-provider)
+
+You need to enable it in the config as followed:
+
+```yaml
+# To enable the Module
+matrix_synapse_s3_media_store_enabled: true
```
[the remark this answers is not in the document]
BackInBash: Renamed.

spantaleev:
Thanks for working on improving this! Let me know when it's ready for review.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
BackInBash: Renamed the media options.

roles/matrix-synapse/tasks/synapse/setup_install.yml
BackInBash: I like the idea to influence the `self_build` var. I already added a disclaimer in the docs to not disabling or overwriting the Synapse `self-build` var.

roles/matrix-synapse/defaults/main.yml
```diff
@@ -595,6 +595,18 @@ matrix_synapse_media_storage_provider_s3_secret_access_key: ""
...
+  "bucket": {{ matrix_synapse_media_storage_provider_s3_bucket|to_json }},
+  # All of the below options are optional, for use with non-AWS S3-like
+  # services, or to specify access tokens here instead of some external method.
+  "region_name": {{ matrix_synapse_media_storage_provider_s3_region_name|to_json }},
+  "endpoint_url": {{ matrix_synapse_media_storage_provider_s3_endpoint_url|to_json }},
+  "access_key_id": {{ matrix_synapse_media_storage_provider_s3_access_key_id|to_json }},
+  "secret_access_key": {{ matrix_synapse_media_storage_provider_s3_secret_access_key|to_json }},
+  "storage_class": {{ matrix_synapse_s3_media_store_storage_class|to_json }},
+  "threadpool_size": {{ matrix_synapse_media_storage_provider_s3_threadpool_size }}
+
```
BackInBash, line 609: Here is the Bucket Config

roles/matrix-synapse/tasks/synapse/setup_install.yml
```diff
@@ -119,3 +128,21 @@
...
+            "module": "s3_storage_provider.S3StorageProviderBackend",
+            "config":{
+              "store_local": {{ matrix_synapse_s3_media_store_on_s3 }},
+              "store_remote": {{ matrix_synapse_s3_media_store_on_local_filesystem }},
+              "store_synchronous": {{ matrix_synapse_media_storage_provider_s3_store_synchronous }},
+              "config": {{ matrix_synapse_media_storage_provider_s3_config }}
+            }
+          }
+        ]
+      }}
```
BackInBash, line 148: Here I injected the whole module config. But it does not get in the homeserver.yaml

BackInBash (commented):
The Config Injection is not working as intended.

I've looked at the antispam and shared secret implementaion but not getting any further.

roles/matrix-synapse/tasks/synapse/setup_install.yml
spantaleev: You should inject into a new variable (`matrix_synapse_media_storage_providers: []` in `defaults/main.yml`).

Then you should have this in `homeserver.yaml.j2`:

```yaml
media_storage_providers: {{ matrix_synapse_media_storage_providers|to_json }}
```

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -1039,7 +1039,9 @@ media_store_path: "/matrix-media-store-parent/{{ matrix_synapse_media_store_dire
 #    config:
 #       directory: /mnt/some/other/directory
 
+{% if matrix_synapse_media_storage_provider_s3_enabled %}
```
spantaleev, line 1042: This `if` is unnecessary. It's also undersireable even.

With our new setup, people should be able to inject media storage providers manually into `matrix_synapse_media_storage_providers`.

Enabling `matrix_synapse_media_storage_provider_s3_enabled` makes the playbook automatically inject a storage provider there, but manual management is also possible.

roles/matrix-synapse/tasks/synapse/setup_install.yml
```diff
@@ -130,7 +130,7 @@
     mode: 0755
 
 - set_fact:
-    matrix_synapse_media_storage_providers: |
+    media_storage_providers: |
```
spantaleev, line 133: Why not `matrix_synapse_media_storage_providers` here?

spantaleev (changes requested):

roles/matrix-synapse/tasks/synapse/setup_install.yml
BackInBash: I thought it needs to match `media_storage_providers:` in the `homeserver.yaml`

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
BackInBash: yeah got it removed.

roles/matrix-synapse/tasks/synapse/setup_install.yml
spantaleev: The variable is part of the `matrix-synapse` role, so we prefix it with `matrix_synapse_`.

And then it's `_media_storage_providers`, because that's the configuration key that it controls.

The `matrix_synapse_media_storage_providers` variable should be defined in `defaults/main.yml` as an empty list: `matrix_synapse_media_storage_providers: []`.

People can override it manually, as well as enable various media storage providers modules (like the S3 module you're adding here). These modules will inject their configuration into the `matrix_synapse_media_storage_providers` variable.

BackInBash: regardless the naming of the options.
the `media_storage_providers` stays empty

homeserver.yaml
```yaml
max_spider_size: 10M
max_upload_size: 4096M
media_storage_providers: []
media_store_path: /matrix-media-store-parent/media-store
metrics_flags: null
```

xangelix:
what's missing here?

BackInBash:
Config Nesting from `setup_install.yml` into `homeserver.yaml.j2` is not working as intended.
Example:
https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/1769/files/6d305e432cae8e6d68463bcb3e13ab620e2a2aee#diff-3f01d105d3b729eb1a3cedbdc4ec41036d66dab6c4983e5464e2709a3d9ae637

Implementation:
https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/1769/files/589a821f1b83bb1221425e6192e7bd71a5d37b13..33bd2f3faba7d838583aa14eea8fcf0670b9995c#diff-b0da8a56987a4474d3b70fb97aa67db6ed7874b0a94e62575e17e32045c02892

xangelix:
@spantaleev able to close after https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/ac7cb3619dfe522af280875566e859007f7a0374

PR #1770, opened 2022-04-19
add retires to all get_url actions

etkecc:
This PR adds automatic retries to all `get_url` actions to avoid errors like this one:

```
Request failed: <urlopen error [Errno -3] Temporary failure in name resolution>
```

The same was done to docker pulls before


spantaleev:
Nice! Thanks! :+1:

PR #1771, opened 2022-04-20
Add matrix-registration-bot

moan0s:
This adds an install and uninstall task plus helpers. The bot is disabled by default.
This commit does not include documentation, yet. In short, the bot can be enabled by adding
```
matrix_bot_matrix_registration_bot_enabled: true
matrix_bot_matrix_registration_bot_matrix_user_password: "verysecret"
matrix_bot_matrix_registration_bot_matrix_admin_token: "supersecret"
```
to the host_vars

I look forward to comments. If this seems to be something that will be added, I will provide documentation


moan0s:
The task is heavily inspired by the matrix-reminder-bot task of this repository :)

group_vars/matrix_servers
```diff
@@ -987,6 +987,15 @@ matrix_bot_matrix_reminder_bot_container_image_self_build: "{{ matrix_architectu
...
 
+
+######################################################################
+#
+# matrix-bot-matrix-registration-bot
+#
+######################################################################
+# We don't enable bots by default.
+matrix_bot_matrix_registration_bot_enabled: false
+
```
spantaleev, line 998: We should probably add this here:

```yaml
matrix_bot_matrix_registration_bot_container_image_self_build: "{{ matrix_architecture not in ['amd64'] }}"

matrix_bot_matrix_registration_bot_systemd_required_services_list: |
  {{
    ['docker.service']
    +
    ['matrix-' + matrix_homeserver_implementation + '.service']
    +
  }}
```

roles/matrix-bot-matrix-registration-bot/defaults/main.yml
```diff
@@ -0,0 +1,62 @@
+---
+# matrix-registration-bot creates and manages registration tokens for a matrix server
+# See: https://github.com/moan0s/matrix-registration-bot
+
+matrix_bot_matrix_registration_bot_enabled: true
+matrix_bot_matrix_registration_bot_container_image_self_build: false
+matrix_bot_matrix_registration_bot_docker_repo: "https://github.com/moan0s/matrix-registration-bot.git"
+matrix_bot_matrix_registration_bot_docker_src_files_path: "{{ matrix_base_data_path }}/matrix-registration-bot/"
```
spantaleev, line 8: We normally name this directory `docker-src` and put it under the component's base path (`matrix_bot_matrix_registration_bot_base_path`), not under the main base path (`matrix_base_data_path` = `/matrix`).

```diff
-matrix_bot_matrix_registration_bot_docker_src_files_path: "{{ matrix_base_data_path }}/matrix-registration-bot/"
+matrix_bot_matrix_registration_bot_docker_src_files_path: "{{ matrix_bot_matrix_registration_bot_base_path }}/docker-src"
```

roles/matrix-bot-matrix-registration-bot/tasks/setup_install.yml
```diff
@@ -0,0 +1,69 @@
+---
+
+- set_fact:
+    matrix_bot_matrix_registration_bot_requires_restart: false
```
spantaleev, line 4: This can be removed. Together with the reference to `matrix_bot_matrix_registration_bot_requires_restart` below.

In some other roles, we've used a variable like this for doing data migrations and restarting the service, but we don't really use this for anything here.

roles/matrix-bot-matrix-registration-bot/tasks/setup_install.yml
```diff
@@ -0,0 +1,69 @@
...
+  when: "matrix_bot_matrix_registration_bot_container_image_self_build|bool"
+
+- name: Ensure matrix-registration-bot image is built
+  docker_image:
+    name: "{{ matrix_bot_matrix_registration_bot_docker_image }}"
+    source: build
+    force_source: "{{ matrix_bot_matrix_registration_bot_git_pull_results.changed if ansible_version.major > 2 or ansible_version.minor >= 8 else omit }}"
+    force: "{{ omit if ansible_version.major > 2 or ansible_version.minor >= 8 else matrix_mailer_git_pull_results.changed }}"
+    build:
+      dockerfile: docker/Dockerfile
```
spantaleev, line 48: Looking at https://github.com/moan0s/matrix-registration-bot, the `Dockerfile` is at the root path, not in `docker/Dockerfile`, so I believe this won't work.

You can test the self-building functionality even on an `amd64` architecture by explicitly setting `matrix_bot_matrix_registration_bot_container_image_self_build: true` in your `vars.yml` file.

roles/matrix-bot-matrix-registration-bot/templates/systemd/matrix-bot-matrix-registration-bot.service.j2
```diff
@@ -0,0 +1,38 @@
...
+ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-bot-matrix-registration-bot \
+			--log-driver=none \
+			--cap-drop=ALL \
+			--network={{ matrix_docker_network }} \
+			-e 'BOT_SERVER=https://{{ matrix_server_fqn_matrix }}' \
+			-e 'BOT_USERNAME={{ matrix_bot_matrix_registration_bot_matrix_user_id_localpart }}' \
+			-e 'BOT_PASSWORD={{ matrix_bot_matrix_registration_bot_matrix_user_password }}' \
+			-e 'API_BASE_URL=https://{{ matrix_server_fqn_matrix }}' \
+			-e 'API_TOKEN={{ matrix_bot_matrix_registration_bot_matrix_admin_token }}' \
+			-e 'LOGGING_LEVEL=info' \
```
spantaleev, line 28: Looks like you've decided to configure the both with environment variables, not with a `config.yml` file.

This is fine, but it's better (and safer) to not use `-e` here, but to pass an environment variables file using `--env-file`. Doing it via a file helps keep the password hidden from system users doing `ps`, etc. It's also a cleaner way to manage a bunch of environment variables.

To see how to use an `--env-file`, see how the `matrix-bot-honoroit` role does it.

Also note its usage of the `matrix_bot_honoroit_environment_variables_extension` variable, which allows users to inject additional variables, besides the one we define explicitly.

I think you should also have a dedicated variable for each of this environemnt variables.

Instead of defining `BOT_SERVER` as `https://{{ matrix_server_fqn_matrix }}`, it'd be better to have a `matrix_bot_matrix_registration_bot_bot_server` variable that is defined like this in `defaults/main.yml`:

```yaml
matrix_bot_matrix_registration_bot_bot_server: "https://{{ matrix_server_fqn_matrix }}"
```

This allows advanced users to override the whole `matrix_bot_matrix_registration_bot_bot_server` variable, should they need to.

The same goes for the other environment variables.

------

`LOGGING_LEVEL` seems completely hardcoded. You may wish to add a variable (`matrix_bot_matrix_registration_bot_logging_level: info`) for this one as well.

roles/matrix-bot-matrix-registration-bot/defaults/main.yml
```diff
@@ -0,0 +1,62 @@
...
+  # Example configuration extension follows:
+  #
+  # matrix:
+  #   device_name: My-Registration-Bot
+
+matrix_bot_matrix_registration_bot_configuration_extension: "{{ matrix_bot_matrix_registration_bot_configuration_extension_yaml|from_yaml if matrix_bot_matrix_registration_bot_configuration_extension_yaml|from_yaml is mapping else {} }}"
+
+# Holds the final configuration (a combination of the default and its extension).
+# You most likely don't need to touch this variable. Instead, see `matrix_bot_matrix_registration_bot_configuration_yaml`.
+matrix_bot_matrix_registration_bot_configuration: "{{ matrix_bot_matrix_registration_bot_configuration_yaml|from_yaml|combine(matrix_bot_matrix_registration_bot_configuration_extension, recursive=True) }}"
```
spantaleev, line 62: All of this appears to be unused. You're not defining a `config.yml` file, but instead passing configuration using environment variables. All of this can be removed.

The `matrix_bot_matrix_registration_bot_config_path` path should probably be kept, because we'll need it for defining the `env` file (passed as `--env-file` to `docker run`; discussed in another comment in this review).

spantaleev (changes requested):
Hey!

Thank you for working on matrix-registration-bot request and for doing a good job as a first time contributor!

I've submitted some feedback for making this new role even better and more consistent with the rest of the playbook.

moan0s:
I hope the changes address all the issues, thank you for the helpful feedback!

roles/matrix-bot-matrix-registration-bot/templates/env.j2
```diff
@@ -0,0 +1,6 @@
+BOT_SERVER={{ matrix_bot_matrix_registration_bot_bot_server }}
+BOT_USERNAME={{ matrix_bot_matrix_registration_bot_matrix_user_id_localpart }}
+BOT_PASSWORD={{ matrix_bot_matrix_registration_bot_matrix_user_password }}
+API_BASE_URL={{ matrix_bot_matrix_registration_bot_api_base_url }}
+API_TOKEN={{ matrix_bot_matrix_registration_bot_matrix_admin_token }}
+LOGGING_LEVEL={{ matrix_bot_matrix_registration_bot_logging_level }}
```
spantaleev, line 6: Let's add an "additional variables" variable, like `matrix_bot_honoroit_environment_variables_extension`, so that users can put more variables here easily.

moan0s: Will do

moan0s:
Yes, will do :)

docs/configuring-playbook-bot-matrix-registration-bot.md
```diff
@@ -0,0 +1,71 @@
...
+matrix_bot_matrix_registration_bot_matrix_user_password: "PASSWORD_FOR_THE_BOT"
+# Token obtained via an administrator account
+matrix_bot_matrix_registration_bot_matrix_admin_token: "syt_bW9hbm9z_XXXXXXXXXXXXXr_2kuzbE"
+
+# Enables registration
+matrix_synapse_enable_registration: true
+
+# Restrict registration to users with a token
+matrix_synapse_configuration_extension_yaml: |
+  registration_requires_token: true # For token based registration
```
spantaleev, line 50: Let's add a dedicated variable (`matrix_synapse_registration_requires_token: false` by default) for this. We can then simplify by not having to use `matrix_synapse_configuration_extension_yaml`.

roles/matrix-bot-matrix-registration-bot/templates/env.j2
```diff
@@ -0,0 +1,8 @@
+BOT_SERVER={{ matrix_bot_matrix_registration_bot_bot_server }}
+BOT_USERNAME={{ matrix_bot_matrix_registration_bot_matrix_user_id_localpart }}
+BOT_PASSWORD={{ matrix_bot_matrix_registration_bot_matrix_user_password }}
+API_BASE_URL={{ matrix_bot_matrix_registration_bot_api_base_url }}
+API_TOKEN={{ matrix_bot_matrix_registration_bot_matrix_admin_token }}
```
spantaleev, line 5: We should probably rename `matrix_bot_matrix_registration_bot_matrix_admin_token` to `matrix_bot_matrix_registration_bot_api_token`.

Similarly, renaming `matrix_bot_matrix_registration_bot_matrix_user_password` to `matrix_bot_matrix_registration_bot_bot_password` probably makes sense.

roles/matrix-bot-matrix-registration-bot/templates/systemd/matrix-bot-matrix-registration-bot.service.j2
```diff
@@ -0,0 +1,33 @@
...
+
+[Service]
+Type=simple
+Environment="HOME={{ matrix_systemd_unit_home_path }}"
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} kill matrix-bot-matrix-registration-bot 2>/dev/null || true'
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} rm matrix-bot-matrix-registration-bot 2>/dev/null || true'
+
+ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-bot-matrix-registration-bot \
+			--log-driver=none \
+			--cap-drop=ALL \
```
spantaleev, line 21: Can we add `--read-only` for additional security here or does the container choke with that?

docs/configuring-playbook-bot-matrix-registration-bot.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+Choose a strong password for the bot. You can generate a good password with a command like this: `pwgen -s 64 1`.
+
+## Obtaining an admin API token
+
+In order to use the bot you need to add an admin API token to the configuration. Log into Element/Schildichat with an
+administrator account and navigate to `Settings->Help&About` and scroll to the bottom. You can expand "Access token" to copy it.
+
+![Obatining an admin access token with Element](assets/obtain_admin_access_token_element.png)
```
spantaleev, line 32: We don't have such an image, do we?

docs/configuring-playbook-bot-matrix-registration-bot.md
```diff
@@ -0,0 +1,71 @@
...
+## Adjusting the playbook configuration
+
+Add the following configuration to your `inventory/host_vars/matrix.DOMAIN/vars.yml` file:
+
+```yaml
+matrix_bot_matrix_registration_bot_enabled: true
+# Adjust this to whatever password you chose when registering the bot user
+matrix_bot_matrix_registration_bot_matrix_user_password: "PASSWORD_FOR_THE_BOT"
+# Token obtained via an administrator account
+matrix_bot_matrix_registration_bot_matrix_admin_token: "syt_bW9hbm9z_XXXXXXXXXXXXXr_2kuzbE"
```
spantaleev, line 43: Wouldn't it be most simple if we:

- register the bot user as an admin user (`admin=yes`)
- then obtain a token for that user
- then use that for both `api_token` and for `bot_access_token` (instead of a password)

Given that we need an admin's token anyway, we might as well make the bot an administrator? Given that we need to obtain a token, we might as well use a token (not a password). The same token could be used for both configuration options.

What do you think? It sounds like this simplifies the process.

Later on, when tokens are created with Synapse, it's obvious which service (user) is creating them. It's actually the bot, not your other admin account for which you've revelead the token.

docs/configuring-playbook-bot-matrix-registration-bot.md
moan0s: I forgot to commit it 🙈

docs/configuring-playbook-bot-matrix-registration-bot.md
moan0s: That is a very good idea!

roles/matrix-bot-matrix-registration-bot/templates/systemd/matrix-bot-matrix-registration-bot.service.j2
moan0s: A simple `--read-only`does not work (`matrix-bot-matrix-registration-bot[1870772]: OSError: [Errno 30] Read-only file system: 'config.toml'` but I think I will rewrite configuration bit of the role to mount the config directory. This has the advantage that the `config.toml` stays accessible to the user.

spantaleev: We're configuring `matrix-registration-bot` using environment variables though, not via `config.yml`. If the bot can be fully-configured using environment varialbes only, we'd better refrain from using `config.yml` at all.

Seems like your `matrix-registration-bot` code tries to write some configuration on startup? Perhaps that could be reworked?

moan0s: The bot can only partially be configured via `config.yml` as allowed users etc... live in `config.toml` which is provided by the underlying `simplematrixbotlib`. As this file has to be persistent and users may want to modify is manually I would add `config.yml` and `config.toml` to the config folder (and get rid of the environment variables). The bot furthermore writes a `session.txt` (which can be ephemeral, but it is better if it's not).

spantaleev: Should we switch away from environment variables to `config.yml` then?

As for the `session.txt` file (or other persistent data managed by the bot itself), it should probably go into a `/data` directory that is mounted separately.

The `config/` directory (or `config/config.yml`) can be mounted into the container as read-only, while the `data/` directory is mounted writable (no read-only modifier).

moan0s: Yes, I think that would be good for now. I will have a look at the reminder bot that uses a similar structure.

docs/configuring-playbook-bot-matrix-registration-bot.md
moan0s: This was already done: https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/f6cb59116b93bc826fdb2a7b8999046eaa08e0d6
I will be using it now (didn't exist when I looked for it last time)

roles/matrix-bot-matrix-registration-bot/templates/env.j2
```diff
@@ -0,0 +1,8 @@
+BOT_SERVER={{ matrix_bot_matrix_registration_bot_bot_server }}
+BOT_USERNAME={{ matrix_bot_matrix_registration_bot_matrix_user_id_localpart }}
+BOT_ACCESS_TOKEN={{ matrix_bot_matrix_registration_bot_api_token }}
```
spantaleev, line 3: Why is it called `bot_api_token` when it influences the `bot_access_token` variable? It should probably be called `matrix_bot_matrix_registration_bot_bot_access_token`. That `bot_bot_` prefix might look weird, but it's because the role name is `matrix-registration-bot` (which gives us the `matrix_bot_matrix_registration_bot_` part), and then we have the actual property name (`bot_access_token`).

Why use environment variables? I thought you're moving all this to `config.yml` only? Or you abandoned that idea?

roles/matrix-bot-matrix-registration-bot/templates/env.j2
```diff
@@ -0,0 +1,8 @@
...
+API_BASE_URL={{ matrix_bot_matrix_registration_bot_api_base_url }}
+API_TOKEN={{ matrix_bot_matrix_registration_bot_matrix_admin_token }}
```
spantaleev, line 5: This should be called `matrix_bot_matrix_registration_bot_api_token`.

roles/matrix-bot-matrix-registration-bot/tasks/validate_config.yml
```diff
@@ -0,0 +1,9 @@
+---
+
+- name: Fail if required settings not defined
+  fail:
+    msg: >-
+      You need to define a required configuration setting (`{{ item }}`).
+  when: "vars[item] == ''"
+  with_items:
+    - "matrix_bot_matrix_registration_bot_api_token"
```
spantaleev, line 9: It's probably best to validate both `matrix_bot_matrix_registration_bot_bot_access_token` and `matrix_bot_matrix_registration_bot_api_token` here.

roles/matrix-bot-matrix-registration-bot/defaults/main.yml
```diff
@@ -0,0 +1,41 @@
...
+matrix_bot_matrix_registration_bot_systemd_wanted_services_list: []
+
+# The bot's username. This user needs to be created manually beforehand.
+# Also see `matrix_bot_matrix_registration_bot_user_password`.
+matrix_bot_matrix_registration_bot_matrix_user_id_localpart: "bot.matrix-registration-bot"
+
+matrix_bot_matrix_registration_bot_matrix_user_id: '@{{ matrix_bot_matrix_registration_bot_matrix_user_id_localpart }}:{{ matrix_domain }}'
+
+# The password that the bot uses to authenticate.
+matrix_bot_matrix_registration_bot_matrix_user_password: ''
```
spantaleev, line 39: This property is useless now?

We should define the other variables with proper defaults and comments that describe the situation and wiring:

```
# The access token that the bot uses to communicate in Matrix chats
# This does not necessarily need to be a privileged (admin) access token.
matrix_bot_matrix_registration_bot_bot_access_token: ''

# The access token that the bot uses to call the Matrix API for creating registration tokens.
# This needs to be a privileged (admin) access token.
# By default, we assume `matrix_bot_matrix_registration_bot_bot_access_token` is such a privileged token and we use it as is.
# If necessary, you can define your own other access token here, which might even be for a different Matrix user.
matrix_bot_matrix_registration_bot_api_token: "{{ matrix_bot_matrix_registration_bot_bot_access_token }}"

spantaleev (changes requested):

roles/matrix-bot-matrix-registration-bot/templates/env.j2
moan0s: Sorry I forgot to delete that file it is no longer needed as I indeed changed to a config file

roles/matrix-bot-matrix-registration-bot/defaults/main.yml
moan0s: I now removed the password and added your explanation

roles/matrix-bot-matrix-registration-bot/tasks/validate_config.yml
moan0s: Done

roles/matrix-bot-matrix-registration-bot/templates/env.j2
moan0s: Removed the file but changed the variable name accordingly

roles/matrix-bot-matrix-registration-bot/tasks/validate_config.yml
moan0s: Removed the file but changed the variable name accordingly

moan0s:
I tried to implement your suggestions. Sorry, it's my first time using ansible (and I am not experienced in docker too), so thank you for your patience

spantaleev:
Thanks a lot for developing this bot and adding support for it to the playbook, so that others can install it easily now! :+1: 

I've announced it in the changelog via e435c55458571ee

PR #1774, opened 2022-04-20
Slack appservice: Enable RTM by default

OdyX:
It is very confusing to debug why messages only go from Matrix to Slack
but not from Slack to Matrix. RTM should be enabled by default, as
that's the recommended way to make this work.

### DIFF
--- a/roles/matrix-bridge-appservice-slack/templates/config.yaml.j2
+++ b/roles/matrix-bridge-appservice-slack/templates/config.yaml.j2
@@ -9,8 +9,20 @@ homeserver:
     url: "{{ matrix_appservice_slack_homeserver_url }}"
     media_url: "{{ matrix_appservice_slack_homeserver_media_url }}"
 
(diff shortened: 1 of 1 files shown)


spantaleev:
Thanks! :+1:

spantaleev:
As I said in b2105f35ecd6f4ad0dec612d927, it would probably be great if we reconciled our `config.yaml` with upstream to:

- bring in things we've missed (like this `rtm` thing)
- bring in comments for what we're using already
- bring in new options that may have appeared since we gained appservice-slack support

If you've got time and would like to help maintain the appservice-slack bridge a bit, it'd be great!

OdyX:
Hello @spantaleev ,

thank you for merging, and adding much-needed comments and documentation!

We (@liip) 're still in the evaluation / investigation phase, so I only have that much of energy to put towards this. But if we go further with Matrix, I'll certainly get more involved!

PR #1775, opened 2022-04-21
Updated: mautrix-signal to v0.3.0 & signald to 0.18.0

GoliathLabs:
https://github.com/mautrix/signal/releases/tag/v0.3.0

### DIFF
--- a/roles/matrix-bridge-mautrix-signal/defaults/main.yml
+++ b/roles/matrix-bridge-mautrix-signal/defaults/main.yml
@@ -8,8 +8,8 @@ matrix_mautrix_signal_container_image_self_build: false
 matrix_mautrix_signal_docker_repo: "https://mau.dev/mautrix/signal.git"
 matrix_mautrix_signal_docker_src_files_path: "{{ matrix_base_data_path }}/mautrix-signal/docker-src"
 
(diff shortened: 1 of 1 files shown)


spantaleev:
Nice! Thanks! :+1:

GoliathLabs:
@borisrunakov normally the 0.18.0 image should correct the permissions and then drop privileges. Yet I had the same error. I did a manually chown and now everything seems to be working

What's your opinion regarding this @spantaleev

borisrunakov:
Chown to `matrix:matrix`?

spantaleev:
We use `--user` when starting it to start it with the `matrix` user (and group), so it probably can't do its thing (start as `root` and drop). We prefer forcing containers into a non-privileged user instead of giving them `root` and hoping/trusting that they'll drop their privileges later.

Does that mean that we need to use the `-non-root` variation of the Signald daemon now? Is this change (root vs non-root) documented somewhere upstream?

borisrunakov:
There is this: https://signald.org/articles/install/docker/#migrating-from-versions-before-0180

tulir:
> the maunium one does not appear up-to-date anymore.

(it's always up to date, but it has never published tags)

PR #1776, opened 2022-04-21
HOTFIX: matrix-registration-bot, fixed systemd template that caused setup error

wibbene:
Sorry to request to pull into master, but a hotfix is needed. I have removed a line in [matrix-bot-matrix-registration-bot.service.j2](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/roles/matrix-bot-matrix-registration-bot/templates/systemd/matrix-bot-matrix-registration-bot.service.j2). 

If this line is left open, docker will crash when it tries to open a file that doesn't exist. Looks like this may have been left over from something. I have tested this and I was able to install it on my own server. Once this line was removed, I had no issues.

I also noticed a documentation error that referenced the wrong bot. I corrected this too.

### DIFF
--- a/docs/configuring-playbook-bot-matrix-registration-bot.md
+++ b/docs/configuring-playbook-bot-matrix-registration-bot.md
@@ -63,7 +63,7 @@ ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start
 
 ## Usage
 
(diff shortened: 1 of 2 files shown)


spantaleev:
Indeed, https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/1771 first used `--env-file`, but later switched to configuring matrix-registration-bot using a configuration file (`config.yaml`), instead of using environment variables.

Thanks for removing these leftovers! :+1:

PR #1777, opened 2022-04-22
Change list of public servers, old as not functional

moan0s:
The old link returned a 404 so I thought I throw in joinmatrix.org :)


spantaleev:
Thanks! :+1:

PR #1779, opened 2022-04-22
Bump signald to 0.18.1

mattcen:
According to https://signald.org/articles/install/docker/#migrating-from-versions-before-0180,
This release only chowns files if the container is running as root. See also this upstream commit:
https://gitlab.com/signald/signald/-/commit/3bb7e8d2c128681473e324f811cff25e0883b88d


spantaleev:
Thanks! 👍

PR #1780, opened 2022-04-23
Make synapse htpasswd file path configurable

Shaps:
When setting `matrix_nginx_proxy_enabled: false` and enabling authentication on the metrics endpoint, the htpasswd file is hardcoded to the nginx-proxy container dir.
This changes the hardcoded value to a variable so the path can be updated.

Added a new variable in the nginx-proxy defaults `matrix_nginx_proxy_proxy_synapse_metrics_basic_auth_path` pointing to the currently hardcoded value


spantaleev:
Thank you for making this configurable! :+1: 

I've even made it automatically do the right thing in [4a0b839](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/1780/commits/4a0b8397680119432287c175dad68f479324cb82).

Shaps:
Oh, I didn't see there was a var to configure the path! That makes it even better :). Thanks for merging

PR #1781, opened 2022-04-23
Give a hint to the default variables

JokerGermany:
I think it is something with is not obvious to find for everyone.

### DIFF
--- a/docs/configuring-playbook.md
+++ b/docs/configuring-playbook.md
@@ -20,7 +20,8 @@ You can then follow these steps inside the playbook directory:
 
 
 For a basic Matrix installation, that's all you need.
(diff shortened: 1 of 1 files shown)


docs/configuring-playbook.md
```diff
@@ -20,7 +20,8 @@ You can then follow these steps inside the playbook directory:
 
 
 For a basic Matrix installation, that's all you need.
-For a more custom setup, see the [Other configuration options](#other-configuration-options) below.
+For a more custom setup, see the [Other configuration options](#other-configuration-options) below.
```
spantaleev, line 23: Linking to `matrix-base`'s `defaults/main.yml` is of limited use. There are only a handful of things there, while the tens of other roles contain hundreds of other variables.

PR #1783, opened 2022-04-25
Update path to signald Dockerfile

Sekki21956:
Fixes permissions bug for arm64.


mattcen:
For a bit more clarity on this, unlike amd64, arm64 self-builds the signald image (even though upstream provides an arm64 image on Docker Hub; fixed by #1784), but it was being built using the wrong Dockerfile path.

spantaleev:
Thanks! :+1:

PR #1787, opened 2022-04-26
Support ansible vault strings for homeserver secret key

dooleydevin:
Running the setup playbook on a fresh Ubuntu 21 VM, I was getting the same error described in #1420 because I was using `ansible-vault` to encrypt the `matrix_homeserver_generic_secret_key` variable.

I implemented the same solution I saw in #1420 everywhere I found this variable and was able to get the playbook to go through successfully :smile:


spantaleev:
Thanks! :+1:

PR #1790, opened 2022-04-26
matrix-change-user-admin-status: do not allocate tty

aine-etke:


spantaleev:
Thanks! 👍 

I wonder if we need the `-i` part even.

etkecc:
The `-i` part is needed to get the output (eg it's `UPDATE 1` in successful case), otherwise it will be omitted

Issue #1801, opened 2022-05-07
Use upstream mx-puppet-discord


dsonck92:
**Is your feature request related to a problem? Please describe.**
Some weeks ago, a decision was made to switch from the official mx-puppet-bridge to the beeper maintained fork. This was done as the build of the Dockerfile was failing due to NPM + Github shenanigans. Meanwhile, I've migrated the official project over to gitlab (https://gitlab.com/mx-puppet/discord/mx-puppet-discord) and set-up proper CI pipelines to automatically build new releases, alleviating the need for manual building through the Dockerfile (even though this also still works).

Therefore I propose that it's reverted back, also because in the support channel for mx-puppet-bridge, some beeper specific issues have been cropping up (which, obviously, the upstream project does not want to deal with).

**Describe the solution you'd like**
Revert back to the original mx-puppet-bridge, which can now be found at `registry.gitlab.com/mx-puppet/discord/mx-puppet-discord`. Tags exist for versions (`vX.Y.Z`), specific branches (`main`) and of course `latest`.

NOTE: I cannot say whether shifting between the beeper and original version can be done without ill effects. Potentially the database needs to be fixed manually if any migrations happened on beepers fork.

dsonck92:
Needless to say, I would really like this to go through, since I've had to mention "we don't support the ansible deployment, it's not **the** mx-puppet-discord we develop" a few times.

dsonck92:
Also, the tracking issue there is: https://gitlab.com/mx-puppet/skype/mx-puppet-skype/-/issues/17

spantaleev:
Good to hear that you've made so many improvements! We're happy to get back on the upstream train!

I've migrated my installation from the Beeper fork to yours (via #1802) and (so far) it appears to work well.

Thank you! :+1:

PR #1802, opened 2022-05-07
Change back to original mx-puppet-discord

dsonck92:
This MR will revert the change to use beepers fork back to the upstream project.

Closes: #1801


spantaleev:
Thanks! :+1: 

It's good to be back on upstream where active development is happening!
Nice to see that you provide tagged releases now.

Any plans for non-amd64 builds? It's not a hard necessity, as self-building also seems to work well, but it will make setup faster and more stable for users on these other architectures.

dsonck92:
Currently I don't have the infrastructure for that. I opted to utilize my own hardware to run the pipelines (not to be limited by running time of gitlab shared runners) but that's a pure amd64 cluster. I would need to investigate how I can easily cross compile (because there are some dependencies with native code). So I might consider it at some point, but for now the focus lies on doing the same things for slack and other bridges so they also get an easy to install container.

PR #1803, opened 2022-05-07
Upgrade Hookshot (1.5.0 -> 1.6.1)

HarHarLinks:
https://github.com/matrix-org/matrix-hookshot/releases/tag/1.6.1
https://github.com/matrix-org/matrix-hookshot/releases/tag/1.6.0

tl;dr:
- gitlab can be configured with commands and a widget
- rss/atom feed support
- docker image is now only 300MB instead of 1.3GB! :tada:

no api changes = easy upgrade

PR #1804, opened 2022-05-07
Add hookshot feeds config support

HarHarLinks:
I actually missed adding this config in #1803 so here we go


spantaleev:
Thanks! :+1: 

I've merged it as it is, although it could use the following improvements:

- `|to_json` when spewing things into the configuration - for extra safety
- variabla naming that matches the thing being configured (`matrix_hookshot_feeds_interval` -> `matrix_hookshot_feeds_pollIntervalSeconds`) - this makes naming consistent with the rest of the playbook

PR #1806, opened 2022-05-08
Clarify reverse-proxying of well-known files with Caddy2

Arkonos:
The suggestions for how to serve/reverse proxy .well-known files from base domain using Caddy 2 seems to be outdated. https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-well-known.md#option-3-setting-up-reverse-proxying-of-the-well-known-files-from-the-base-domains-server-to-the-matrix-server

Instead what is suggested in this issue (https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/1442) should be used.

### DIFF
--- a/docs/configuring-well-known.md
+++ b/docs/configuring-well-known.md
@@ -116,8 +116,22 @@ server {
 **For Caddy 2**, it would be something like this:
 
 ```caddy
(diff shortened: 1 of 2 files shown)


spantaleev:
Thanks! 👍

PR #1818, opened 2022-05-12
Bridge mautrix signal and mautrix whatsapp encryption

### LINKED ISSUE #877: E2EE from Element to Signal Bridge
I'm trying to set up E2E encryption from my Android Element client, when I'm inside the Settings of Signal room there is next message `You don't have permission to enable encryption in this room`, is there ability to activate it?

paulRbr:
Hello,

This PR adds two new variables, one in each Signal & WhatsApp bridges, to be able to enable end-to-bridge encryption.

Closes #877

### DIFF
--- a/roles/matrix-bridge-mautrix-signal/defaults/main.yml
+++ b/roles/matrix-bridge-mautrix-signal/defaults/main.yml
@@ -127,3 +127,7 @@ matrix_mautrix_signal_registration_yaml: "{{ lookup('template', 'templates/regis
 matrix_mautrix_signal_registration: "{{ matrix_mautrix_signal_registration_yaml|from_yaml }}"
 
 matrix_mautrix_signal_log_level: 'DEBUG'
+
+matrix_mautrix_signal_bridge_encryption_allow: false
+matrix_mautrix_signal_bridge_encryption_default: "{{ matrix_mautrix_signal_bridge_encryption_allow }}"
+matrix_mautrix_signal_bridge_encryption_key_sharing_allow: "{{ matrix_mautrix_signal_bridge_encryption_allow }}"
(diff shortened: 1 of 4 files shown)


roles/matrix-bridge-mautrix-signal/defaults/main.yml
```diff
@@ -127,3 +127,5 @@ matrix_mautrix_signal_registration_yaml: "{{ lookup('template', 'templates/regis
 matrix_mautrix_signal_registration: "{{ matrix_mautrix_signal_registration_yaml|from_yaml }}"
 
 matrix_mautrix_signal_log_level: 'DEBUG'
+
+matrix_mautrix_signal_enable_encryption: false
```
spantaleev, line 131: This should be called `matrix_mautrix_signal_bridge_encryption_allow` to match the configuration key.

roles/matrix-bridge-mautrix-signal/templates/config.yaml.j2
```diff
@@ -152,15 +152,15 @@ bridge:
     # this to work. See https://github.com/tulir/mautrix-telegram/wiki/End‐to‐bridge-encryption
     encryption:
         # Allow encryption, work in group chat rooms with e2ee enabled
-        allow: false
+        allow: {{ matrix_mautrix_signal_enable_encryption }}
         # Default to encryption, force-enable encryption in all portals the bridge creates
         # This will cause the bridge bot to be in private chats for the encryption to work properly.
-        default: false
+        default: {{ matrix_mautrix_signal_enable_encryption }}
```
spantaleev, line 158: This should use a separate variable (`matrix_mautrix_signal_bridge_encryption_default`).

For convenience, you can define it in terms of `matrix_mautrix_signal_bridge_encryption_allow`:

```yaml
matrix_mautrix_signal_bridge_encryption_default: "{{ matrix_mautrix_signal_bridge_encryption_allow }}"
```

spantaleev (changes requested):

roles/matrix-bridge-mautrix-signal/defaults/main.yml
paulRbr: :+1: fixed

spantaleev:
I've added `|to_json` for extra safety and consistency with how we do things. These templates lack `|to_json` in various other places. Those would be nice to fix as well.

Thank you for working on this! :+1:

Issue #1815, opened 2022-05-11
Reverse proxy example code for well-known for Apache2 is incorrect


brush:
**Describe the bug**

Installation appears to work correctly. Services including synapse are working. However, self check fails with this error:

-----------
TASK [matrix-synapse : Check Matrix Client API] ********************************************************************************************

fatal: [matrix.example.com]: FAILED! => changed=false 
  connection: close
  content_length: '153'
  content_type: text/html
  date: Wed, 11 May 2022 01:58:26 GMT
  elapsed: 0
  msg: 'Status code was 404 and not [200]: HTTP Error 404: Not Found'
  permissions_policy: interest-cohort=()
  redirected: false
  server: nginx
  status: 404
  strict_transport_security: max-age=31536000; includeSubDomains
  url: https://matrix.example.com/_matrix/client/versions
...ignoring

TASK [matrix-synapse : Fail if Matrix Client API not working] 
fatal: [matrix.example.com]: FAILED! => changed=false 
  msg: 'Failed checking Matrix Client API is up at `matrix.example.com` (checked endpoint: `https://matrix.example.com/_matrix/client/versions`). Is Synapse running? Is port 443 open in your firewall? Full error: {''redirected'': False, ''url'': ''https://matrix.example.com/_matrix/client/versions'', ''status'': 404, ''date'': ''Wed, 11 May 2022 01:58:26 GMT'', ''server'': ''nginx'', ''content_type'': ''text/html'', ''content_length'': ''153'', ''strict_transport_security'': ''max-age=31536000; includeSubDomains'', ''permissions_policy'': ''interest-cohort=()'', ''connection'': ''close'', ''elapsed'': 0, ''changed'': False, ''failed'': True, ''msg'': ''Status code was 404 and not [200]: HTTP Error 404: Not Found''}'
-----------

Element runs, but results in this error:

`
There was a problem communicating with the homeserver, please try again later. (HTTP 404)
Cannot reach homeserver
Ensure you have a stable internet connection, or get in touch with the server admin
`

In addition, debug logging from synapse includes this:

`
matrix-synapse[192282]: 2022-05-11 01:18:29,846 - synapse.util.httpresourcetree - 47 - INFO - sentinel - Attaching <synapse.rest.ClientRestResource object at 0x7fa917fdad60> to path b'/_matrix/client'
`

... and many like it, mapping various files to urls I believe -- and I'm wondering if that "b" in front of all of them is spurious.

[...]

PR #1820, opened 2022-05-13
Suggested fix to docs for reverse proxy for well-known file

brush:
Link to issue https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/1815


spantaleev:
Thanks! :+1:

PR #1825, opened 2022-05-14
Change back to working original mx-puppet-slack

dsonck92:
Now that v0.1.1 is out, which depends on the functional matrix-slack-parser it works again.


spantaleev:
Thank you! :+1:

PR #1828, opened 2022-05-16
Update Element self build repo URL

aaronraimist:
It forwards to the correct place but might as well just update it to the current URL.


spantaleev:
Thanks! :+1:

PR #1837, opened 2022-05-20
Fix synapse LDAP simple bind config

### LINKED ISSUE #1496: LDAP auth issue with matrix-synapse-ldap3
Hi,

I found an issue with LDAP authentication in anonymous bind mode. When following the example LDAP configuration, with empty strings as `bind_dn` and `bind_password`, it generates a synapse LDAP configuration also with empty strings for `bind_dn` and `bind_password` which does not seem to work anymore. I needed to remove those config keys in order to make LDAP authentication work.

Playbook conf:
```yaml
matrix_synapse_ext_password_provider_ldap_bind_dn: ""
matrix_synapse_ext_password_provider_ldap_bind_password: ""
```

Generated synapse conf (authentication does not work):
```yaml
password_providers:
...
    bind_dn: ''
    bind_password: ''
...
```
Those two keys must be removed or the authentication will fail with those logs:
```
ldap_auth_provider - 499 - WARNING - sentinel - Error during LDAP authentication: user name is mandatory in simple bind
ldap_auth_provider - 602 - WARNING - sentinel - Error during LDAP authentication: user name is mandatory in simple bind
ldap_auth_provider - 207 - WARNING - sentinel - Error during ldap authentication: user name is mandatory in simple bind
```

It did work before, and I do not really know when it started to fail as already connected clients seemed to be able to login. I could not find in this playbook, the `matrix-synapse-ldap3` project or even the python ldap3 module which recent change could have triggered this issue.

Lyokovic:
When using LDAP simple bind, the `bind_dn` and `bind_password`
configuration values must not be present.

Fixes #1496

### DIFF
--- a/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
+++ b/roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
@@ -2521,8 +2521,10 @@ password_providers:
         uid: {{ matrix_synapse_ext_password_provider_ldap_attributes_uid|string|to_json }}
         mail: {{ matrix_synapse_ext_password_provider_ldap_attributes_mail|string|to_json }}
         name: {{ matrix_synapse_ext_password_provider_ldap_attributes_name|string|to_json }}
+{% if matrix_synapse_ext_password_provider_ldap_bind_dn %}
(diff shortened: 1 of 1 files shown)


spantaleev:
Thanks! :+1:

PR #1870, opened 2022-06-05
Add configuration example for Jitsi behind a NAT with no audio or video for more than 2 call participants.

EtienneHenger:


EtienneHenger:
It seems that this was already implemented with `matrix_jitsi_jvb_stun_servers` with `meet-jit-si-turnrelay.jitsi.net:443` as the default. For some reason this STUN server does not seem to work for my setup behind a NAT, however adding a Google STUN server seems to work.

PR #1882, opened 2022-06-15
Added option to add env variables to mautrix signal daemon container

IUCCA:
from signald doku: when a remote key changes, set trust level to TRUSTED_UNVERIFIED instead of UNTRUSTED 
I find it much more convenient when new identities are automatically recognized as trusted, as the process to do that manually is cumbersome.
Should this the default behavior, or should i add an option to configure this behavior?


spantaleev:
What is the recommended default by Signald folks? Having ours set to that sounds better.

Making this configurable, so more or less paranoid people can choose their own setting also sounds like a good idea.

IUCCA:
The answer from Finn to another person from the signal bridge matrix room who had trouble with verifying a new number:
> turn on trust all safety numbers mode
> i should make that default
> i've been hesitant to do it but honestly it trips up so many people
> and first party clients have fully eliminated hard stops around safety number changes

If I make it configurable, what should be default?

aaronraimist:
Comparing safety numbers is Signal's method of key verification.

aaronraimist:
> I find it much more convenient when new identities are automatically recognized as trusted, as the process to do that manually is cumbersome.

@IUCCA Sure that's more convenient because if you automatically trust everyone then you don't have to do anything. **Of course it also defeats the entire purpose of verifying since you haven't actually verified anything...**

I haven't used the bridge but surely you can just to talk without verifying, is that correct? Verification is completely optional when using the actual Signal app. If that is the case, then I don't think we should turn on some option to automatically verify all safety numbers by default.

IUCCA:
At the first start of an chat every security number is recorded as TRUSTED_UNVERIFIED, if it changes afterwards it is marked as untrusted, after that the bridge doesn't bridge anymore to this person. Sometimes you get an error message and sometimes not.

There is an third state of TRUSTED_VERIFIED, which is used if you want to check the numbers.

IUCCA:
This number is like a public key and changes if a person install Signal on a new phone

aaronraimist:
Sounds like bad design on the bridge's part. They should really allow verification to be optional rather than forcing it which encourages "verifying" without actually verifying.

roles/matrix-bridge-mautrix-signal/defaults/main.yml
```diff
@@ -130,6 +130,10 @@ matrix_mautrix_signal_configuration_extension: "{{ matrix_mautrix_signal_configu
 # You most likely don't need to touch this variable. Instead, see `matrix_mautrix_signal_configuration_yaml`.
 matrix_mautrix_signal_configuration: "{{ matrix_mautrix_signal_configuration_yaml|from_yaml|combine(matrix_mautrix_signal_configuration_extension, recursive=True) }}"
 
+# Prevents the puppet from breaking when the signal security nuber changes.
```
spantaleev, line 133: nuber? Did you mean number or something?

roles/matrix-bridge-mautrix-signal/defaults/main.yml
```diff
@@ -130,6 +130,10 @@ matrix_mautrix_signal_configuration_extension: "{{ matrix_mautrix_signal_configu
...
+# The new security nuber will marked as trusted_unverified if this is set to true
+matrix_mautrix_signal_deamon_trust_new_security_nuber: false
```
spantaleev, line 135: The name of the variable does not correspond to the configuration key that it influences (`SIGNALD_TRUST_NEW_KEYS`). This makes things harder for playbook users and is inconsistent with the rest of the playbook.

roles/matrix-bridge-mautrix-signal/templates/systemd/matrix-mautrix-signal-daemon.service.j2
```diff
@@ -34,7 +34,9 @@ ExecStartPre=-{{ matrix_host_command_docker }} run --rm --name matrix-mautrix-si
 # We can't use `--read-only` for this bridge.
 ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-mautrix-signal-daemon \
 			--log-driver=none \
+            {% if matrix_mautrix_signal_deamon_trust_new_security_nuber %}
 			--env SIGNALD_TRUST_NEW_KEYS=true \
+            {% endif %}
```
spantaleev, line 39: If Signald exposes other things to be configured via environment variables, it may be better (and cleaner) to do this using an environment variables file -- one file, where we template all of them, which is later referenced here using `--env-file ...`.

See how the Jitsi role does it for its various components.

Additional benefits of using `--env-file` instead of `--env` (besides the cleaner `.service` file) is:

- changes to environment files do not need to reload systemd (`systemctl daemon-reload`) for the changes to take effect. Simplify restarting the service will make it pick up the new environment variable values
- we can make it extensible in a clean way. See this: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/9abcb10f9fc2341bcc02da72710f53b7db255c52/roles/matrix-jitsi/defaults/main.yml#L162-L172

spantaleev (changes requested):

roles/matrix-bridge-mautrix-signal/defaults/main.yml
IUCCA: The Bridge Bot called the Key 'new identity' when it displays a failure Message (only some times):
`⚠ Your message was not bridged: Identity failure occurred while sending message to +000. New identity: xxx`
The signal-daemon called it key and users of the original Signal App called the key 'safety number'. [Signal support article](https://support.signal.org/hc/en-us/articles/360007060632-What-is-a-safety-number-and-why-do-I-see-that-it-changed-)

Security Number was a translating mistake by me.
I would prefer to call it safety number, because that is the name which was known by signal users.

Or should i name it in the comment safety number and the variable `matrix_mautrix_signal_deamon_trust_new_keys`?

spantaleev: If the configuration setting that we ultimately control is named `SIGNALD_TRUST_NEW_KEYS`, we should probably use a variable that has the same name (e.g. `matrix_mautrix_signal_deamon_trust_new_keys`).

The comment describing the variable can say "Security Number" and link to the Signal support article even.

roles/matrix-bridge-mautrix-signal/defaults/main.yml
```diff
@@ -130,6 +130,10 @@ matrix_mautrix_signal_configuration_extension: "{{ matrix_mautrix_signal_configu
 # You most likely don't need to touch this variable. Instead, see `matrix_mautrix_signal_configuration_yaml`.
 matrix_mautrix_signal_configuration: "{{ matrix_mautrix_signal_configuration_yaml|from_yaml|combine(matrix_mautrix_signal_configuration_extension, recursive=True) }}"
 
+# Prevents the puppet from breaking when the signal safety number changes.
+# The new safety number will be marked as trusted_unverified if this is set to true
+matrix_mautrix_signal_daemon_trust_new_keys: false
```
spantaleev, line 135: This PR is named "Auto trust new signal identities", but the `matrix_mautrix_signal_daemon_trust_new_keys` variable is defaulted to `false`.

Should we rename the PR? Or perhaps switch the default to `true`? I think our conversation was hinting that this shouldn't be set to `true` by default, but.. I'm not really sure what we settled on.

IUCCA: The default behavior of the bridge is to have this option disabled. The official Signal programs behave as if this is set to true.
When this option is enabled, it significantly increases the comfort.

If paranoid people want to verify the safety numbers, they can still do that even if the option is on. 
When the option is off, the bridge will no longer send messages to a person when the security number of that person changes, until the safety number is marked manually as verified.

I would prefer the option to be on by default, but I am also happy with off as well.

We could rename the PR to Added option to accept changed signal keys without verifying them manually

Unfortunately, the synonymous use of key, safety number and identity is confusing and is caused by the signal folks.
This thing is something like an individual public key for each conversation and changes every time when a person installs Signal again or linked the Signal Account to a second device (signal desktop, this bridge).

IUCCA:
Now this PR adds only the option to add your own env variables to the signal daemon

spantaleev:
Looks good! Thanks! :+1: 

Since `SIGNALD_TRUST_NEW_KEYS` is an important setting, it doesn't hurt having a dedicated variable for it, but.. it's okay as it is too, I suppose.

If you'd like to add a variable for it in another PR, feel free to do so!

PR #1884, opened 2022-06-17
add synapse media_retention

etkecc:
Ref: https://matrix-org.github.io/synapse/develop/usage/configuration/config_documentation.html#media_retention

PS: Not tested yet


spantaleev:
Thanks! :+1:

PR #1886, opened 2022-06-18
Synapse workers should respect X-Forwarded headers

davidmehren:
Currently, Synapse workers ignore the X-Forwarded headers, which leads to internal Docker IP addresses randomly appearing in the device list of users.

This adds the `x_forwarded: true` option to the worker config, fixing the issue.


spantaleev:
Thanks for fixing it! :+1:

PR #1888, opened 2022-06-20
Update Buscarron 1.1.0 -> 1.2.0

etkecc:
Pipeline: https://gitlab.com/etke.cc/buscarron/-/jobs/2615582403


roles/matrix-bot-buscarron/defaults/main.yml
```diff
@@ -76,6 +76,9 @@ matrix_bot_buscarron_homeserver: "{{ matrix_homeserver_container_url }}"
 # forms configuration
 matrix_bot_buscarron_forms: []
 
+# Disable encryption
+matrix_bot_buscarron_noencryption:
```
spantaleev, line 80: Setting an explicit value here (e.g. `false`, `0`) will make this more clear what this variable needs to contain, without people having to go and dig into the Buscarron documentation.

Alternatively, we can leave it as `null`, but have 1 more comment line that tells you how to make use of this variable.

spantaleev:
Thanks! :+1:

roles/matrix-bot-buscarron/defaults/main.yml
etkecc: Will do with the next release

PR #1894, opened 2022-06-24
Maubot moanos

moan0s:
this PR is mainly based on #622 and for me it was easier copy the work that @Cadair did into my branch.
I tested this role and can successfully install maubot with reverse proxy and configuration options.

The role could be more configurable but I think it is a good start.

If this is something with perspective to be merged I can write up some documentation


roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
```diff
@@ -0,0 +1,38 @@
...
+ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-maubot \
+			--log-driver=none \
+            -e UID={{ matrix_user_uid }} \
+            -e GID={{ matrix_user_gid }} \
+            -v {{ matrix_bot_maubot_data_path }}:{{ matrix_bot_maubot_container_data_dir }}:z \
+			{% for arg in matrix_bot_maubot_container_extra_arguments %}
+			{{ arg }} \
+			{% endfor %}
+			--network={{ matrix_docker_network }} \
+            -p {{ matrix_bot_maubot_port }}:29316 \
```
Cadair, line 28: Exposing this to the host should be a config option

roles/matrix-bot-maubot/tasks/setup_install.yml
```diff
@@ -0,0 +1,111 @@
...
+      {% endif %}
+      }
+
+- name: Register Maubot's proxying configuration with matrix-nginx-proxy
+  set_fact:
+    matrix_nginx_proxy_proxy_matrix_additional_server_configuration_blocks: |
+      {{
+        matrix_nginx_proxy_proxy_matrix_additional_server_configuration_blocks|default([])
+        +
+        [matrix_bot_maubot_matrix_nginx_proxy_configuration]
```
Cadair, line 49: I don't know what the usual pattern in this repo is for this, but people might not want to expose the maubot management interface through the nginx proxy service if that service is exposed to the internet. I would think this should be optional and maybe there's some other thing to deal with that.

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
moan0s: You mean the line   `-p {{ matrix_bot_maubot_port }}:29316 \` ?

roles/matrix-bot-maubot/tasks/setup_install.yml
moan0s: Yeah true!

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
Cadair: yeah

roles/matrix-bot-maubot/tasks/setup_install.yml
moan0s: Done

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
moan0s: Done

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
```diff
@@ -0,0 +1,40 @@
...
+DefaultDependencies=no
+
+[Service]
+Type=simple
+Environment="HOME={{ matrix_systemd_unit_home_path }}"
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} kill matrix-maubot 2>/dev/null || true'
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} rm matrix-maubot 2>/dev/null || true'
+
+ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-maubot \
+			--log-driver=none \
```
spantaleev, line 20: Could you test starting the container directly with the matrix user and group, like we do for our other services?

`--user={{ matrix_user_uid }}:{{ matrix_user_gid }}`, instead of starting as root and expecting that the container will drop to whatever specify in the `UID` and `GID` environment variables.

This may mean that we won't use its own entrypoint.. And that likely we won't need to be passing `UID`/`GID` anymore.

Along the same lines, would this work if we add `--cap-drop=ALL`, so that we can further reduce the container process' priviledges?

Can we also use `--read-only`? For this, you may need to first adjust the bot's logging configuration, so that it won't try to write to `/var/log`.

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
```diff
@@ -0,0 +1,40 @@
...
+            -e UID={{ matrix_user_uid }} \
+            -e GID={{ matrix_user_gid }} \
+            -v {{ matrix_bot_maubot_data_path }}:{{ matrix_bot_maubot_container_data_dir }}:z \
+			{% for arg in matrix_bot_maubot_container_extra_arguments %}
+			{{ arg }} \
+			{% endfor %}
+			--network={{ matrix_docker_network }} \
```
spantaleev, line 27: Indentation seems a bit off compared to the other lines.

roles/matrix-bot-maubot/tasks/setup_install.yml
```diff
@@ -0,0 +1,113 @@
...
+    mode: "u=rwx"
+
+- name: Generate Maubot proxying configuration for matrix-nginx-proxy
+  set_fact:
+    matrix_bot_maubot_matrix_nginx_proxy_configuration: |
+      location ~ ^/(_matrix/maubot/.*) {
+      {% if matrix_nginx_proxy_enabled|default(False) %}
+      	{# Use the embedded DNS resolver in Docker containers to discover the service #}
+      	resolver 127.0.0.11 valid=5s;
+      	set $backend "matrix-maubot:{{ matrix_bot_maubot_port }}/$1";
```
spantaleev, line 31: The port inside the container is always `29316`, isn't it? This probably doesn't need to use this variable.

roles/matrix-bot-maubot/templates/config/config.yaml.j2
```diff
@@ -0,0 +1,129 @@
...
+        colored:
+            (): maubot.lib.color_log.ColorFormatter
+            format: '[%(asctime)s] [%(levelname)s@%(name)s] %(message)s'
+        normal:
+            format: '[%(asctime)s] [%(levelname)s@%(name)s] %(message)s'
+    handlers:
+        file:
+            class: logging.handlers.RotatingFileHandler
+            formatter: normal
+            filename: /var/log/maubot.log
```
spantaleev, line 114: Can we disable logging to files and just use stdout, so that logs are collected by journald, like we're doing for all other services?

I suppose that `/var/log` is just an in-container path, which is not mounted outside of the container, so this file will probably not persist anyway.

Still, we'd better avoid writing to files. And if you go for `--read-only` for the container filesystem as I suggest in my other review comment, writing to `/var/log` may even fail.

roles/matrix-bot-maubot/tasks/setup_install.yml
```diff
@@ -0,0 +1,113 @@
...
+  file:
+    path: "{{ item.path }}"
+    state: directory
+    mode: 0755
+    owner: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_groupname }}"
+  with_items:
+    - {path: "{{ matrix_bot_maubot_base_path }}", when: true}
+    - {path: "{{ matrix_bot_maubot_data_path }}", when: true}
+    - {path: "{{ matrix_bot_maubot_docker_src_files_path }}", when: true}
```
spantaleev, line 13: `matrix_bot_maubot_docker_src_files_path` only needs to be created when `matrix_bot_maubot_container_image_self_build` is `true`.

I imagine you need 

```suggestion
    - {path: "{{ matrix_bot_maubot_docker_src_files_path }}", when: "{{ matrix_bot_maubot_container_image_self_build }}"}
```

roles/matrix-bot-maubot/defaults/main.yml
```diff
@@ -0,0 +1,36 @@
...
+matrix_bot_maubot_docker_src_files_path: "{{ matrix_bot_maubot_base_path }}/docker-src"
+
+matrix_bot_maubot_version: latest
+matrix_bot_maubot_docker_image: "dock.mau.dev/maubot/maubot:{{ matrix_bot_maubot_version }}"
+matrix_bot_maubot_docker_image_force_pull: "{{ matrix_bot_maubot_docker_image.endswith(':latest') }}"
+
+matrix_bot_maubot_base_path: "{{ matrix_base_data_path }}/maubot"
+matrix_bot_maubot_data_path: "{{ matrix_bot_maubot_base_path }}/data"
+matrix_bot_maubot_config_path: "{{ matrix_bot_maubot_base_path }}/config"
+matrix_bot_maubot_container_data_dir: "/data"
```
spantaleev, line 15: `matrix_bot_maubot_container_data_dir` doesn't seem very useful.

If the bot always expects its configuration in `/data`, what's the point of making it configurable?

```suggestion
```

roles/matrix-bot-maubot/defaults/main.yml
```diff
@@ -0,0 +1,36 @@
+---
+
+matrix_bot_maubot_enabled: true
+matrix_bot_maubot_container_image_self_build: false
+matrix_bot_maubot_docker_repo: "https://mau.dev/maubot/maubot.git"
+matrix_bot_maubot_docker_src_files_path: "{{ matrix_bot_maubot_base_path }}/docker-src"
+
+matrix_bot_maubot_version: latest
```
spantaleev, line 8: `v0.3.1` seems to be available. We'd better use a tagged release, not `latest`.
```suggestion
matrix_bot_maubot_version: v0.3.1
```

group_vars/matrix_servers
```diff
@@ -1058,6 +1058,33 @@ matrix_bot_matrix_registration_bot_systemd_required_services_list: |
 #
 ######################################################################
 
+######################################################################
+#
+# matrix-maubot
```
spantaleev, line 1063:
```suggestion
# matrix-bot-maubot
```

group_vars/matrix_servers
```diff
@@ -1058,6 +1058,33 @@ matrix_bot_matrix_registration_bot_systemd_required_services_list: |
...
+    +
+    ['matrix-' + matrix_homeserver_implementation + '.service']
+    +
+    (['matrix-nginx-proxy.service'] if matrix_nginx_proxy_enabled else [])
+  }}
+
+
+######################################################################
+#
+# /matrix-maubot
```
spantaleev, line 1084:
```suggestion
# /matrix-bot-maubot
```

roles/matrix-bot-maubot/defaults/main.yml
```diff
@@ -0,0 +1,36 @@
...
+matrix_bot_maubot_docker_image: "dock.mau.dev/maubot/maubot:{{ matrix_bot_maubot_version }}"
+matrix_bot_maubot_docker_image_force_pull: "{{ matrix_bot_maubot_docker_image.endswith(':latest') }}"
+
+matrix_bot_maubot_base_path: "{{ matrix_base_data_path }}/maubot"
+matrix_bot_maubot_data_path: "{{ matrix_bot_maubot_base_path }}/data"
+matrix_bot_maubot_config_path: "{{ matrix_bot_maubot_base_path }}/config"
+matrix_bot_maubot_container_data_dir: "/data"
+matrix_bot_maubot_container_config_dir: "/root/.config/"
+
+matrix_bot_maubot_bot_server: "https://{{ matrix_server_fqn_matrix }}"
```
spantaleev, line 18: Looks like this variable is used in multiple places, both as `public_url` in the configuration and in other places.

I suppose that not all instances need to use this "public URL".

We can define one more variable here:

```suggestion
matrix_bot_maubot_bot_server_url: "{{ matrix_homeserver_container_url }}"
matrix_bot_maubot_bot_server_public: "https://{{ matrix_server_fqn_matrix }}"
```

.. and then use these accordingly.

Not sure if it makes sense though.. Maybe maubot expects all server URLs defined in its configuration to indeed be public? Feel free to disregard this if so.

roles/matrix-bot-maubot/defaults/main.yml
```diff
@@ -0,0 +1,36 @@
...
+matrix_bot_maubot_proxy_management_interface: false
+matrix_bot_maubot_expose_management_interface: true
+
+
+matrix_bot_maubot_logging_level: info
+matrix_bot_maubot_secret: ''
+matrix_bot_maubot_admin_user: ''
+matrix_bot_maubot_admin_password: ''
+matrix_mau_environment_variables_extension: ''
```
spantaleev, line 27: These variables seem to be unused.

```suggestion
```

roles/matrix-bot-maubot/defaults/main.yml
```diff
@@ -0,0 +1,36 @@
...
+matrix_bot_maubot_config_path: "{{ matrix_bot_maubot_base_path }}/config"
+matrix_bot_maubot_container_data_dir: "/data"
+matrix_bot_maubot_container_config_dir: "/root/.config/"
+
+matrix_bot_maubot_bot_server: "https://{{ matrix_server_fqn_matrix }}"
+matrix_bot_maubot_proxy_management_interface: false
+matrix_bot_maubot_expose_management_interface: true
+
+
+matrix_bot_maubot_logging_level: info
```
spantaleev, line 23: This variable seems to be unused.

```suggestion
```

Feel free to add variables for the different loggers instead though.

roles/matrix-bot-maubot/templates/config/config.yaml.j2
```diff
@@ -0,0 +1,129 @@
...
+    # The base path for plugin endpoints. The instance ID will be appended directly.
+    plugin_base_path: /_matrix/maubot/plugin/
+    # Override path from where to load UI resources.
+    # Set to false to using pkg_resources to find the path.
+    override_resource_path: /opt/maubot/frontend
+    # The base appservice API path. Use / for legacy appservice API and /_matrix/app/v1 for v1.
+    appservice_base_path: /_matrix/app/v1
+    # The shared secret to sign API access tokens.
+    # Set to "generate" to generate and save a new token at startup.
+    unshared_secret: {{ matrix_bot_maubot_secret }}
```
spantaleev, line 67: It's safer to use `|to_json`, especially around secrets, which people tend to populate with all sorts of weird characters.

```suggestion
    unshared_secret: {{ matrix_bot_maubot_secret|to_json }}
```

roles/matrix-bot-maubot/templates/config/config.yaml.j2
```diff
@@ -0,0 +1,129 @@
...
+# clients in the management UI, since you can also just type the homeserver URL
+# into the box there.
+homeservers:
+  {{  matrix_domain }}:
+    # Client-server API URL
+        url: {{ matrix_bot_maubot_bot_server }}
+    # registration_shared_secret from synapse config
+    # You can leave this empty if you don't have access to the homeserver.
+    # When this is empty, `mbc auth --register` won't work, but `mbc auth` (login) will.
+        secret: {{ matrix_registration_shared_secret }}
```
spantaleev, line 80:
```suggestion
        secret: {{ matrix_bot_maubot_registration_shared_secret|to_json }}
```

This does a couple of things:
- it adds `|to_json`
- **and more importantly**, to switches away from using the `matrix_registration_shared_secret` variable which is **part of the `matrix-registration` role**.

You probably intended to use `matrix_synapse_registration_shared_secret` (in the case of Synapse, at least).

In any case, the shared secret (coming from the Synapse role.. or Dendrite role) should be passed via another variable (let's call it `matrix_bot_maubot_registration_shared_secret`), which is defined as empty in `roles/matrix-bot-maubot/defaults/main.yml` and is later redefined in terms of `matrix_synapse_registration_shared_secret` (or the Dendrite equivalent) via `group_vars/matrix_servers`.

`group_vars/matrix_servers`:

```yaml
matrix_bot_maubot_registration_shared_secret: |-
  {{
    {
      'synapse': matrix_synapse_registration_shared_secret,
      'dendrite': matrix_dendrite_registration_shared_secret,
    }[matrix_homeserver_implementation]
  }}
```

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
```diff
@@ -0,0 +1,40 @@
...
+            {% if matrix_bot_maubot_expose_management_interface|bool %}
+            -p {{ matrix_bot_maubot_port }}:29316 \
+            {% endif %}
+			{{ matrix_bot_maubot_docker_image }}
+
+ExecStop=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} kill matrix-maubot 2>/dev/null || true'
+ExecStop=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} rm matrix-maubot 2>/dev/null || true'
+Restart=always
+RestartSec=30
+SyslogIdentifier=matrix-maubot
```
spantaleev, line 37:
```suggestion
SyslogIdentifier=matrix-bot-maubot
```

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
```diff
@@ -0,0 +1,40 @@
+#jinja2: lstrip_blocks: "True"
```
spantaleev, line 1: Renaming this file to `matrix-bot-maubot.service` would be preferable.

Not just the template, but the file on the Matrix server as well.

Don't forget to update `init.yml` too, so that it points to `matrix-bot-maubot.service`.

spantaleev (changes requested):
Thanks for picking this up and getting it to work! :bow: 

I have some feedback that aims to make it more polished and more consistent with our naming conventions, etc.

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
moan0s: --user is not possible as it fails with 
```
Jun 24 15:05:08 X-Wing matrix-maubot[1016487]: mkdir: can't create directory '/var/log/maubot': Permission denied
Jun 24 15:05:09 X-Wing matrix-maubot[1016487]: FAILED: No config file 'alembic.ini' found, or file has no '[alembic]' section
Jun 24 15:05:09 X-Wing matrix-maubot[1016487]: chown: /var/log: Operation not permitted
Jun 24 15:05:09 X-Wing matrix-maubot[1016487]: chown: /var/log: Operation not permitted
Jun 24 15:05:09 X-Wing matrix-maubot[1016487]: su-exec: setgroups(1000): Operation not permitted

```

moan0s: Also `--cap-drop=ALL`
```

Jun 24 15:06:42 X-Wing matrix-maubot[1017218]: Error: open /data/config.yaml: permission denied
Jun 24 15:06:42 X-Wing matrix-maubot[1017218]: Error: open /data/config.yaml: permission denied
Jun 24 15:06:42 X-Wing matrix-maubot[1017218]: su-exec: setgroups(1000): Operation not permitted

```

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
moan0s: Done

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
moan0s: Done

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
```diff
@@ -0,0 +1,40 @@
...
+{% endfor %}
+{% for service in matrix_bot_maubot_systemd_wanted_services_list %}
+Wants={{ service }}
+{% endfor %}
+DefaultDependencies=no
+
+[Service]
+Type=simple
+Environment="HOME={{ matrix_systemd_unit_home_path }}"
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} kill matrix-maubot 2>/dev/null || true'
```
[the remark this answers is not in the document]
moan0s: Done

group_vars/matrix_servers
moan0s: Done

roles/matrix-bot-maubot/templates/config/config.yaml.j2
moan0s: Done

roles/matrix-bot-maubot/defaults/main.yml
moan0s: done

roles/matrix-bot-maubot/tasks/setup_install.yml
moan0s: Yeah it is. I thought this port refers to the exposed port which could be changed -> therefore the variable

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
moan0s: Read-only is done

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
spantaleev: image

Here's what I meant. Some of these lines are indented with spaces, others with tabs. In some editors, this may look the same, but on the diff, you can clearly see that it's not.

We generally prefer tabs.

moan0s: Ah did not see that in my editor 😅

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
spantaleev: Looking at the entrypoint command [here](https://github.com/maubot/maubot/blob/47d7e2b483aa598a601231b88bbd4af78f71e4c6/docker/run.sh), it seems like this script does a initialization stuff like this:
- creating some directories under `/data` (`/data/plugins`, `/data/trash`, `/data/dbs`) - we may be able to pre-create these via the playbook (just add them to `setup_install.yml`)
- creates the `/var/log` directory in the container. We'd ideally stop logging to files, as discussed elsewhere. It's probably done the same way it's done for other mautrix bridges. See [our configuration for mautrix-facebook](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/bd05e0c1b46bfb0acc0e540535a5efafb12c41dc/roles/matrix-bridge-mautrix-facebook/templates/config.yaml.j2#L235-L260). If we can't stop it, we can probably mount a `tmpfs` volume at `/var/log`.
- updates the configuration (relative to absolute paths) - we can skip this by defining absolute paths in our configuration from the beginning, instead of relying on the script to fix it up. This is probably already the case, so maybe there is nothing to do about it.
- moves some databases that were previously in `/data/plugins` to `/data/dbs` - this seems like a data migration from some olden times where databases were stored in `/data/plugins`. We don't need to care about it

With these changes, we may be able to start with `--user` and `--cap-drop=ALL`, and to invoke `python3 -m maubot -c /data/config.yaml` directly, instead of using [the default command specified in the Dockerfile](https://github.com/maubot/maubot/blob/47d7e2b483aa598a601231b88bbd4af78f71e4c6/Dockerfile#L63).

To start the continer with a custom command, simply append it at the end of the `docker run` invocation in this `.service` file - [we do a similar thing for Synapse](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/bd05e0c1b46bfb0acc0e540535a5efafb12c41dc/roles/matrix-synapse/templates/synapse/systemd/matrix-synapse.service.j2#L64).

roles/matrix-bot-maubot/tasks/setup_install.yml
spantaleev: It seems useful in the `{% else %}` clause. In there, it refers to the exposed port (on `127.0.0.1`).

The main `{% if %}` part (before `{% else %}`) handles proxying over the container network. There, `matrix-bot-maubot:29316` is always true.

By the way, there's a typo here. Now that the container is named `matrix-bot-maubot` (not `matrix-maubot`), this line needs to be fixed up:

```suggestion
      	set $backend "matrix-bot-maubot:29316/$1";
```

spantaleev: By the way, it seems like this `set $backend` line (and likely the surrounding lines too) mix different ways indenting - tabs and spaces.

We should switch to **spaces-only** here, because YAML doesn't like tabs.

roles/matrix-bot-maubot/defaults/main.yml
moan0s: We do need the public url everywhere but for different reasons. I updated accordingly

roles/matrix-bot-maubot/defaults/main.yml
moan0s: Done (removing)

roles/matrix-bot-maubot/templates/config/config.yaml.j2
moan0s: Done

roles/matrix-bot-maubot/templates/config/config.yaml.j2
```diff
@@ -64,7 +64,7 @@ server:
     appservice_base_path: /_matrix/app/v1
     # The shared secret to sign API access tokens.
     # Set to "generate" to generate and save a new token at startup.
-    unshared_secret: {{ matrix_bot_maubot_secret }}
+    unshared_secret: {{ matrix_bot_maubot_secret|to_json }}
```
spantaleev, line 67: Calling this variable `matrix_bot_maubot_server_unshared_secret` will be even better.

Where appropriate, we try to name variables after the configuration key they control.

Similarly, it's probably better if `matrix_bot_maubot_bot_server_public` is called `matrix_bot_maubot_bot_server_public_url`.

Cadair:
I have a couple of commits here: https://github.com/moan0s/matrix-docker-ansible-deploy/pull/1

roles/matrix-bot-maubot/tasks/setup_install.yml
Cadair: I fixed this in my PR to moan0s' branch.

roles/matrix-bot-maubot/templates/config/config.yaml.j2
```diff
@@ -0,0 +1,129 @@
+# The full URI to the database. SQLite and Postgres are fully supported.
+# Other DBMSes supported by SQLAlchemy may or may not work.
+# Format examples:
+#   SQLite:   sqlite:///filename.db
+#   Postgres: postgresql://username:password@hostname/dbname
+database: sqlite:////data/maubot.db
```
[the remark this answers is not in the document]
Cadair: In my PR I have made this customisable again, but I don't use the playbooks managed postgres so I haven't done the hard work of integrating it with that.

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
moan0s: done

roles/matrix-bot-maubot/templates/config/config.yaml.j2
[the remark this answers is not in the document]
moan0s: I now added postgres support. It currently doesn't work with sqlite anymore as I didn't think of a clean way to make this work. Would it be preferable to switch between the two or should I remove sqlite support completly?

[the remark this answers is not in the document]
spantaleev: To make both work, you'd need to make `matrix_bot_maubot_database_uri` dynamic, based on the value of `matrix_bot_maubot_database_engine`.

See how the mautrix-facebook roles does it here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/569b52f0c1818d988aa0ebf0eb93a7ff9e37b503/roles/matrix-bridge-mautrix-facebook/defaults/main.yml#L73-L80

roles/matrix-bot-maubot/templates/systemd/matrix-bot-maubot.service.j2
```diff
@@ -0,0 +1,42 @@
...
+Type=simple
+Environment="HOME={{ matrix_systemd_unit_home_path }}"
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} kill matrix-bot-maubot 2>/dev/null || true'
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} rm matrix-bot-maubot 2>/dev/null || true'
+
+ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-bot-maubot \
+            --log-driver=none \
+			--user={{ matrix_user_uid }}:{{ matrix_user_gid }} \
+            --read-only \
+            --cap-drop=ALL \
```
spantaleev, line 23: Looks like indentation is mixed here. Some lines are using spaces, some are using tabs.

Better switch everything to tabs in this file.

roles/matrix-bot-maubot/templates/config/config.yaml.j2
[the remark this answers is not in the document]
moan0s: tested and pushed (I for some reason thought this was not possible in the `main.yml`

roles/matrix-bot-maubot/tasks/setup_install.yml
```diff
@@ -0,0 +1,113 @@
...
+    src: "{{ role_path }}/templates/config/config.yaml.j2"
+    dest: "{{ matrix_bot_maubot_data_path }}/config.yaml"
+    owner: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_groupname }}"
+    mode: "u=rwx"
+
+- name: Generate Maubot proxying configuration for matrix-nginx-proxy
+  set_fact:
+    matrix_bot_maubot_matrix_nginx_proxy_configuration: |
+      location ~ ^/(_matrix/maubot/.*) {
```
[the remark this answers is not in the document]
moan0s: Mhhh true.
> What's the magic incantation to make the trailing slash optional?
I don't know

roles/matrix-bot-maubot/templates/config/config.yaml.j2
moan0s: done

roles/matrix-bot-maubot/templates/systemd/matrix-bot-maubot.service.j2
moan0s: done

roles/matrix-bot-maubot/templates/systemd/matrix-maubot.service.j2
```diff
@@ -0,0 +1,40 @@
...
+ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-maubot \
+			--log-driver=none \
+            -e UID={{ matrix_user_uid }} \
+            -e GID={{ matrix_user_gid }} \
+            -v {{ matrix_bot_maubot_data_path }}:{{ matrix_bot_maubot_container_data_dir }}:z \
+			{% for arg in matrix_bot_maubot_container_extra_arguments %}
+			{{ arg }} \
+			{% endfor %}
+			--network={{ matrix_docker_network }} \
+            {% if matrix_bot_maubot_expose_management_interface|bool %}
```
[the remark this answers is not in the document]
moan0s: done

roles/matrix-bot-maubot/tasks/setup_install.yml
moan0s: done

roles/matrix-bot-maubot/tasks/setup_install.yml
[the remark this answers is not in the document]
spantaleev: How about:

```
location /_matrix/maubot/ {
.....
.....
# Just hostname + port here, no paths
set $backend "matrix-maubot:{{ matrix_bot_maubot_port }}";
# Note the trailing slash here
proxy_pass http://$backend/;
```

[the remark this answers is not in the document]
moan0s: So I tried the following which was inspired by the official docs (did not work, just wanted to leave it here). Will try 
```
- name: Generate Maubot proxying configuration for matrix-nginx-proxy
  set_fact:
    matrix_bot_maubot_matrix_nginx_proxy_configuration: |
      location /_matrix/maubot {
      {% if matrix_nginx_proxy_enabled|default(False) %}
         resolver 127.0.0.11 valid=5s;
         proxy_pass http://matrix-bot-maubot:29316";
      {% else %}
        proxy_pass http://localhost:{{ matrix_bot_maubot_management_interface_port }};
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
      {% endif %}
      }
      location /_matrix/maubot/v1/logs {
      {% if matrix_nginx_proxy_enabled|default(False) %}
        resolver 127.0.0.11 valid=5s;
        proxy_pass http://matrix-bot-maubot:29316";
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "Upgrade";
      {% else %}
        proxy_pass http://localhost:29316;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "Upgrade";
      {% endif %}
      }
```
@spantaleev Your approach redirects `https://synapse.hyteck.de/_matrix/maubot` to `https://synapse.hyteck.de:8443/_matrix/maubot/` and `https://synapse.hyteck.de/_matrix/maubot/` results in a 404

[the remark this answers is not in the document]
spantaleev: I think hardcoding `proxy_pass http://matrix-bot-maubot:29316";` makes nginx try to resolve `matrix-bot-maubot` just once, on startup.

This can fail if the bot service is not running yet. If it works, it's still not good, because if the bot service restarts for whatever reason and gets a new IP address, nginx would still use the old one.

With the `resolver 127.0.0.11 valid=5s;` and `set $backend` dance, we force nginx to resolve the IP address once every 5 seconds.

-----

Why would my approach redirect `https://synapse.hyteck.de/_matrix/maubot` to `https://synapse.hyteck.de:8443/_matrix/maubot/`? Did you mean it forwards it to `https://synapse.hyteck.de:29316/_matrix/maubot/`?

I would expect that it would forward `https://synapse.hyteck.de/_matrix/maubot/SOMETHING` to ``https://synapse.hyteck.de:29316/SOMETHING`. I figured that that is your intention -- removing the `/_matrix/maubot` prefix from the URL when forwarding to the bot. If you'd like **to preserve prefix** instead, then, you'd need to remove the trailing slash from the `proxy_pass` line, like this: `proxy_pass http://$backend;`

You can also try my approach without the trailing slash in the `location` statement, like this: `location /_matrix/maubot`.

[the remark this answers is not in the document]
moan0s: I honestly don't know why your approach behaves that way. I am puzzled myself.
I also tried `rewrite ^/_matrix_maubot$ ^/_matrix/maubot/ permanent;` and it resulted in the same behavior (rewriting `https://synapse.hyteck.de/_matrix/maubo`t to `https://synapse.hyteck.de:8443/_matrix/maubot/`)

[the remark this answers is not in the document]
moan0s: Here is a snippet of my nginx conf


```

rewrite ^/_matrix/maubot$ $scheme://$server_name/_matrix/maubot/ permanent;
location ~ ^/(_matrix/maubot/.*) {
     resolver 127.0.0.11 valid=5s;
   set $backend "matrix-bot-maubot:29316/$1";
   proxy_pass http://$backend;
   proxy_set_header Upgrade $http_upgrade;
   proxy_set_header Connection "upgrade";
}

rewrite ^/synapse-admin$ $scheme://$server_name/synapse-admin/ permanent;

```



I included the synapse admin bit because this woks with rewriting to the correct location.


Maybe we can merge this without solving it? I know it is not as satisfying, on the other hand I don't see it as strictly necessary.

PR #1895, opened 2022-06-24
Added a note on managing the basic_auth password on external prometheus

nono-lqdn:
Hello,

Following a bug I had on setting up my instance, I wrote down the issue I faced so that people don't suffer it. 

Cheers,


spantaleev:
Thanks for clarifying this! :+1:

PR #1898, opened 2022-06-26
expose prometheus process args

etkecc:
the predefined list copied from prometheus docker image's `CMD`

### DIFF
--- a/roles/matrix-prometheus/defaults/main.yml
+++ b/roles/matrix-prometheus/defaults/main.yml
@@ -26,6 +26,13 @@ matrix_prometheus_systemd_wanted_services_list: []
 # Takes an "<ip>:<port>" or "<port>" value (e.g. "127.0.0.1:9090"), or empty string to not expose.
 matrix_prometheus_container_http_host_bind_port: ''
 
+# A list of extra arguments to pass to the prometheus process
+matrix_prometheus_process_extra_arguments:
+  - "--config.file=/etc/prometheus/prometheus.yml"
+  - "--storage.tsdb.path=/prometheus"
+  - "--web.console.libraries=/usr/share/prometheus/console_libraries"
+  - "--web.console.templates=/usr/share/prometheus/consoles"
+
 # Tells whether the "synapse" scraper configuration is enabled.
 matrix_prometheus_scraper_synapse_enabled: false
 
--- a/roles/matrix-prometheus/defaults/main.yml
+++ b/roles/matrix-prometheus/defaults/main.yml
@@ -26,12 +26,18 @@ matrix_prometheus_systemd_wanted_services_list: []
 # Takes an "<ip>:<port>" or "<port>" value (e.g. "127.0.0.1:9090"), or empty string to not expose.
 matrix_prometheus_container_http_host_bind_port: ''
 
-# A list of extra arguments to pass to the prometheus process
-matrix_prometheus_process_extra_arguments:
+# A list of default arguments to pass to the prometheus process
+matrix_prometheus_process_default_arguments:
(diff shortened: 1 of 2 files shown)


roles/matrix-prometheus/defaults/main.yml
```diff
@@ -26,6 +26,13 @@ matrix_prometheus_systemd_wanted_services_list: []
 # Takes an "<ip>:<port>" or "<port>" value (e.g. "127.0.0.1:9090"), or empty string to not expose.
 matrix_prometheus_container_http_host_bind_port: ''
 
+# A list of extra arguments to pass to the prometheus process
+matrix_prometheus_process_extra_arguments:
+  - "--config.file=/etc/prometheus/prometheus.yml"
+  - "--storage.tsdb.path=/prometheus"
+  - "--web.console.libraries=/usr/share/prometheus/console_libraries"
+  - "--web.console.templates=/usr/share/prometheus/consoles"
```
spantaleev, line 34: Are these some defaults use by the Prometheus image already?

Shouldn't we use `matrix_prometheus_process_extra_arguments: []` as a default.

roles/matrix-prometheus/templates/systemd/matrix-prometheus.service.j2
```diff
@@ -31,7 +31,8 @@ ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-prometheus \
 			{% for arg in matrix_prometheus_container_extra_arguments %}
 			{{ arg }} \
 			{% endfor %}
-			{{ matrix_prometheus_docker_image }}
+			{{ matrix_prometheus_docker_image }} \
+			{{ matrix_prometheus_process_extra_arguments|join(' ') }}
```
spantaleev, line 35: (Untested) If `matrix_prometheus_process_extra_arguments` is `[]`, then this may result in an empty line.. Which I'm not sure if it will cause a problem or not.. but it will certainly look weird.

It may be better to put these on the same line, just in case.

```suggestion
			{{ matrix_prometheus_docker_image }} {{ matrix_prometheus_process_extra_arguments|join(' ') }}
```

roles/matrix-prometheus/defaults/main.yml
etkecc: Yes, that is is the default [CMD of prometheus' dockerfile](https://github.com/prometheus/prometheus/blob/main/Dockerfile#L25).

As it's `CMD`, I'm not sure just appending custom args is a good idea, because (IIRC) `CMD` will be re-defined with any args you pass after the image in the `docker run` command

roles/matrix-prometheus/templates/systemd/matrix-prometheus.service.j2
etkecc: the idea was to always have defaults in the extra_arguments var, but yes, that's a valid point!

roles/matrix-prometheus/defaults/main.yml
spantaleev: Maybe we should have a new variable `matrix_prometheus_process_arguments: "{{ matrix_prometheus_process_default_arguments + matrix_prometheus_process_extra_arguments }}`.

And we can put the defaults in `matrix_prometheus_process_default_arguments` and allow people to add additions to `matrix_prometheus_process_extra_arguments`. If someone needs to override the defaults as well, they can manipulate `matrix_prometheus_process_default_arguments` or even `matrix_prometheus_process_arguments`

etkecc: ok, added

spantaleev:
Great! Thank you! 👍

PR #1899, opened 2022-06-26
mx-puppet-bridges - set log level = warn, set presence interval = 5s

etkecc:
upstreaming etke.cc changes


spantaleev:
Thanks! :+1:

PR #1900, opened 2022-06-27
fix typo

etkecc:


spantaleev:
Ohhh.. good catch! :man_facepalming:

PR #1901, opened 2022-06-28
add missing retry to the apache docker image pull

aine-etke:


spantaleev:
Thanks! :+1: 

I've copy/pasted this from some other role, so it means that other role also doesn't have retry logic. It probably should. I don't know which one.

PR #1902, opened 2022-06-28
Update configuring-playbook-bridge-mautrix-facebook.md

ThellraAK:
Logging in from the IP address is no longer always enough, but 2FA seems to work eventually


spantaleev:
Thanks! :+1:

PR #1877, opened 2022-06-12
Add Go Skype Bridge

CyberShadow:
Based on mautrix-whatsapp, as that's what the bridge software is based on.

This one logs me in successfully, unlike mx-puppet-skype, which seems to be dead right now.

I'm not 100% sure that 100% of the added lines are correct - haven't tested things like database migration, just the basic functionality.

PR #1903, opened 2022-06-28
Update README.md to include Go Skype Bridge

ChrisKar96:
Update README.md to include Go Skype Bridge, added in #1877


spantaleev:
Thanks! :+1:

PR #1906, opened 2022-06-30
Update configuring-playbook-bridge-mautrix-instagram.md

ThellraAK:
Copy/Pasting from docs/configuring-playbook-bridge-mautrix-facebook.md but with the relevant variable names changed to add turning on encryption and a puppet admin.

### DIFF
--- a/docs/configuring-playbook-bridge-mautrix-instagram.md
+++ b/docs/configuring-playbook-bridge-mautrix-instagram.md
@@ -20,13 +20,18 @@ matrix_mautrix_instagram_configuration_extension_yaml: |
 
 If you would like to be able to administrate the bridge from your account it can be configured like this:
 ```yaml
+# The easy way. The specified Matrix user ID will be made an admin of all bridges
+matrix_admin: "@YOUR_USERNAME:{{ matrix_domain }}"
+
+# OR:
+# The more verbose way. Applies to this bridge only. You may define multiple Matrix users as admins.
 matrix_mautrix_instagram_configuration_extension_yaml: |
   bridge:
     permissions:
       '@YOUR_USERNAME:YOUR_DOMAIN': admin
 ```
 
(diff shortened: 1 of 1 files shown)


spantaleev:
I've fixed it up a little. Thanks! :+1:

spantaleev:
The format for the `_extension_yaml` variable is.. **2 spaces more-indented** compared to what you see in `defaults/main.yml`.

```yaml
matrix_something_configuration_extension_yaml: |
  # stuff here (<- indented with 2 additional spaces)
```

---

When it comes to logging, it might be more helpful to introduce variables for configuring the various logging configuration keys. Here's [an example from the mautrix-instagram bridge's configuration](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/95ca182ab50bc3cf04e56d7ae1cc57cb49e59265/roles/matrix-bridge-mautrix-instagram/templates/config.yaml.j2#L204-L231).

We can redo it by introducing a variable like this:

```yaml
# Controls the logging level for all bridge loggers.
# If you'd like to adjust each logger separately, use matrix_mautrix_instagram_configuration_extension_yaml.
matrix_mautrix_instagram_logging_level: WARNING
```

And then replace `level: WARNING` with `level: {{ matrix_mautrix_instagram_logging_level|to_json }}` everywhere in `roles/matrix-bridge-mautrix-instagram/templates/config.yaml.j2`.

With this, changing the log-level (e.g. `WARNING` -> `DEBUG`) becomes a simple variable change in `vars.yml`. Example: `matrix_mautrix_instagram_logging_level: DEBUG`, instead of:

```yaml
matrix_mautrix_instagram_configuration_extension_yaml: |
  logging:
    loggers:
      mau:
        level: DEBUG
      mauigpapi:
        level: DEBUG
      paho:
        level: DEBUG
      aiohttp:
        level: DEBUG
    root:
      level: DEBUG
```

It's both easier and the documentation becomes cleaner.

PR #1908, opened 2022-07-01
Update configuring-playbook-bridge-mautrix-facebook.md

ThellraAK:
Adding the defaults in addition to template, switching YOUR_DOMAIN to {{ matrix_domain }}, and giving example of the two combined, as the playbook gives a warning about things being defined twice, so only using the last one in the vars.yml

### DIFF
--- a/docs/configuring-playbook-bridge-mautrix-facebook.md
+++ b/docs/configuring-playbook-bridge-mautrix-facebook.md
@@ -24,10 +24,22 @@ If you would like to be able to administrate the bridge from your account it can
 matrix_mautrix_facebook_configuration_extension_yaml: |
   bridge:
     permissions:
-      '@YOUR_USERNAME:YOUR_DOMAIN': admin
+      '@YOUR_USERNAME:{{ matrix_domain }}': admin
 ```
 
-You may wish to look at `roles/matrix-bridge-mautrix-facebook/templates/config.yaml.j2` to find other things you would like to configure.
+Using both would look like
+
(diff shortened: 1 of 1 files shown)


spantaleev:
Thanks! :+1:  It could also mention the `matrix_admin` variable, which makes permission configuration much cleaner.. And it works across all bridges, which is even better.

ThellraAK:
Awesome, will change it to that, also I finally found the ` key vs the ' key.

Do you think it'd be helpful to still have the per-bridge admin stuff there, or should it be kept simple with showing how to do it for all of them in one spot?

spantaleev:
We can have a generic bridge setup guide. It could mentions that each bridge component (role) has a `matrix_SERVICE_configuration_extension_yaml` variable, which can be used to override configuration. It may give some examples of how it's used. It should point people to the default configuration for a bridge (`roles/matrix-bridge-*/templates/config*`), to the default variables being used (roles/matrix-bridge-*/defaults/main.yml`) and to the playbook overrides (`group_vars/matrix_servers`).

This guide can point out some other useful variables, such as `matrix_admin`.

ThellraAK:
I tried uploading it here, and it made a 'commit' instead of a pull request, then gave me a warning it wasn't a part of anything, and then I lost it...

https://github.com/ThellraAK/matrix-docker-ansible-deploy/pull/1

Is what I came up with, I'm not convinced it's any good and don't see how to improve it.

PR #1909, opened 2022-07-01
Adding logging variable for facebook and setting it's default

ThellraAK:
Added a variable and set it's default for logging in the facebook bridge.

### DIFF
--- a/roles/matrix-bridge-mautrix-facebook/defaults/main.yml
+++ b/roles/matrix-bridge-mautrix-facebook/defaults/main.yml
@@ -128,3 +128,4 @@ matrix_mautrix_facebook_registration_yaml: |
   de.sorunome.msc2409.push_ephemeral: true
 
 matrix_mautrix_facebook_registration: "{{ matrix_mautrix_facebook_registration_yaml|from_yaml }}"
+matrix_mautrix_facebook_logging_level: DEBUG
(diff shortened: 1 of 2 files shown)


roles/matrix-bridge-mautrix-facebook/defaults/main.yml
```diff
@@ -128,3 +128,5 @@ matrix_mautrix_facebook_registration_yaml: |
   de.sorunome.msc2409.push_ephemeral: true
 
 matrix_mautrix_facebook_registration: "{{ matrix_mautrix_facebook_registration_yaml|from_yaml }}"
+#Sets the default log level for the bridge
```
spantaleev, line 131: As yamllint says:

> missing starting space in comment

```suggestion
# Sets the default log level for the bridge
```

ThellraAK:
Awesome, is there anything else I should do?

And should I do a different one of these for each one, or do one bigger one for the rest of them?

spantaleev:
It looks good! Thank you! :+1: 

You can do 1 single PR for updating all the other bridges.

etkecc:
Why DEBUG? Please, use the WARNING level as default, because debug generates **a lot** of logs

spantaleev:
Oh, yes! This was overlooked. Fixed in 71006393

PR #1911, opened 2022-07-01
Adding logging configuration and default to the rest of the mautrixes…

ThellraAK:
… that don't have them

Put the default config below the botname, it was already defined in whatsapp with a different naming convention, so I just left it.


spantaleev:
Thank you! :+1:

Issue #1889, opened 2022-06-21
[Jitsi-meet] Support for arm64 is now available in master (form 20 May)


n00bsteam:
**Describe the bug**

**To Reproduce**
My `vars.yml` file looks like this:

```yaml
matrix_docker_installation_enabled: true

matrix_domain: *******

# The Matrix homeserver software to install.
# See `roles/matrix-base/defaults/main.yml` for valid options.
matrix_homeserver_implementation: synapse

# A secret used as a base, for generating various other secrets.
# You can put any string here, but generating a strong one is preferred (e.g. `pwgen -s 64 1`).
matrix_homeserver_generic_secret_key: '*******'

# This is something which is provided to Let's Encrypt when retrieving SSL certificates for domains.
#
# In case SSL renewal fails at some point, you'll also get an email notification there.
#
# If you decide to use another method for managing SSL certificates (different than the default Let's Encrypt),
# you won't be required to define this variable (see `docs/configuring-playbook-ssl-certificates.md`).
#
# Example value: someone@example.com
matrix_ssl_lets_encrypt_support_email: '*******'

# A Postgres password to use for the superuser Postgres user (called `matrix` by default).
#
# The playbook creates additional Postgres users and databases (one for each enabled service)
# using this superuser account.
matrix_postgres_connection_password: '*******'

matrix_jitsi_enabled: true
matrix_jitsi_jicofo_auth_password: *******
matrix_jitsi_jvb_auth_password: *******
matrix_jitsi_jibri_recorder_password: *******
matrix_jitsi_jibri_xmpp_password: *******
matrix_jitsi_enable_auth: true
matrix_jitsi_enable_guests: true
matrix_jitsi_prosody_auth_internal_accounts:
  - username: "*******"
    password: "*******"
  - username: "*******"
    password: "*******"

matrix_prometheus_enabled: true
matrix_prometheus_node_exporter_enabled: true
matrix_grafana_enabled: true
matrix_grafana_anonymous_access: false
# This has no relation to your Matrix user id. It can be any username you'd like.
# Changing the username subsequently won't work.
matrix_grafana_default_admin_user: "*******"
# Changing the password subsequently won't work.
matrix_grafana_default_admin_password: "*******"


matrix_client_hydrogen_enabled: true
matrix_synapse_admin_enabled: true


matrix_architecture: "arm64"
```

**Expected behavior**
[...]

PR #1913, opened 2022-07-01
feat(jitsi+arm64): Enable Jitsi on arm64

Kab1r:
Support for arm64 images tracked in jitsi/docker-jitsi-meet#1214 and added in jitsi/docker-jitsi-meet#1269

fixes spantaleev/matrix-docker-ansible-deploy#1889


spantaleev:
Thanks! :+1:

PR #1915, opened 2022-07-04
bots: make command_prefix configurable

3hhh:


spantaleev:
Using `|to_json` in the config files might have been better, but it's okay nevertheless.

Looks like the only breaking change is the googlechat bridge now using the `!gc` prefix, instead of `!HO`. Not sure how important this is to necessitate a `CHANGELOG.md` entry though.

Thank you for improving the playbook! :+1:

PR #1920, opened 2022-07-06
Changed whatsapp logging variable name to match

ThellraAK:
I also moved it just below botname as is the case on the other mautrix bridges, and added  the needed lines to `roles/matrix-bridge-mautrix-whatsapp/tasks/validate_config.yml`


spantaleev:
Thanks! :+1:

PR #1921, opened 2022-07-06
signald: 0.18.5 -> 0.20.0

teutat3s:
https://gitlab.com/signald/signald/-/blob/main/releases/0.19.0.md
https://gitlab.com/signald/signald/-/blob/main/releases/0.19.1.md
https://gitlab.com/signald/signald/-/blob/main/releases/0.20.0.md

Notable changes in version `0.19.0`
- All critical data now stored in database
  - Avatars, group images, attachments, etc are still stored on the disk for now
  - None of these need to be kept around for signald to continue working
- New command line flag `--migrate-data` tells signald to preform all required data migrations and exit

I'm not sure if we'd like to automate this or add a hint to CHANGELOG.md about it?

### DIFF
--- a/roles/matrix-bridge-mautrix-signal/defaults/main.yml
+++ b/roles/matrix-bridge-mautrix-signal/defaults/main.yml
@@ -9,7 +9,7 @@ matrix_mautrix_signal_docker_repo: "https://mau.dev/mautrix/signal.git"
 matrix_mautrix_signal_docker_src_files_path: "{{ matrix_base_data_path }}/mautrix-signal/docker-src"
 
 matrix_mautrix_signal_version: v0.3.0
-matrix_mautrix_signal_daemon_version: 0.18.5
+matrix_mautrix_signal_daemon_version: 0.20.0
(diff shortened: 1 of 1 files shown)


spantaleev:
What happens if you try to start the new version with the old data (without running the migration command)?

akdk7:
Hi!

I've just updated the signal bridge and had some trouble with the connection to my signal account. It was disconnected and I received the following message while sending text or pictures:
`Your message was not bridged: org.whispersystems.signalservice.api.push.exceptions.AuthorizationFailedException: [401] Authorization failed!`

The last message I've sent was on March 22nd. So I don't know if this error appears because of an automatic device logout due to inactivity. So I re-registered my matrix device and were able to send messages and pictures again.
There are two other users on the server who are using the bidge and they seem to be DISCTONNECTED from the signal network too. I will watch the log file for the next days.

I can also find some migration information in the log file without any error messages.

Greetings
AkDk7

teutat3s:
Thank you @spantaleev for merging and improving on this in your recent commits! :+1:

PR #1932, opened 2022-07-15
Update recommended Skype bridge in docs

alexhartley:
The skype link recommended is the same as the deprecated.


spantaleev:
Thanks! :+1:

PR #1939, opened 2022-07-17
add ansible-lint; cleanup yamllint

etkecc:


spantaleev:
Thanks! :+1: 

Looks like the ansible-lint job already reports quite a few problems and will report CI as failing until we get those solved. No need to postpone though. Merged as-is.

PR #1947, opened 2022-07-18
Make ansible-lint happy

teutat3s:
- fqcn-builtins: Use FQCN for builtin actions


teutat3s:
> Work on this is already progressing in the `ansible-lint-fixes` branch.

Alright in that case, I'll stop tackling this for now.

PR #1954, opened 2022-07-19
Update telemetry documentation to more closely match upstream

aaronraimist:
Synapse no longer describes the stats as anonymized since the `server_name` is included.
https://github.com/matrix-org/synapse/pull/13321


spantaleev:
Thanks! :+1: 

I wonder if we should mention it more explicitly that the statistics are not anonymized and that they'd be tied to your server's domain name.

PR #1931, opened 2022-07-13
Implement MSC 1929 Homeserver Admin Contact

teutat3s:
Fixes #1612

For details to the proposed (not accepted yet) MSC, see:
https://github.com/matrix-org/matrix-spec-proposals/blob/hs/proposal-admin-contact-1/proposals/1929-admin-contact.md

PR #1957, opened 2022-07-19
Update configuring-well-known.md

mholt:
Make use of `example.com` more consistent (replace `DOMAIN`) and simplify Caddy 2 config. Remove Caddy 1 config since Caddy 1 is long past EOL.

**I would also like to add an example of serving the .well-known "files" without needing actual files.** IMO that's actually the best way: you don't need a file server or a reverse proxy, so it's a lot simpler, and a lot less can go wrong:

```
example.com {
	header /.well-known/matrix/* Content-Type application/json
	header /.well-known/matrix/* Access-Control-Allow-Origin *
	respond /.well-known/matrix/server `{"m.homeserver": {"base_url": "https://matrix.example.com"}}`
	respond /.well-known/matrix/client `{"m.server": "matrix.example.com:443"}`
}
```

Would this addition be acceptable?

### DIFF
--- a/docs/configuring-well-known.md
+++ b/docs/configuring-well-known.md
@@ -79,7 +79,7 @@ If you're managing the base domain by yourself somehow, you'll need to set up se
 
 To make things easy for you to set up, this playbook generates and hosts 2 well-known files on the Matrix domain's server. The files are generated at `/matrix/static-files/.well-known/matrix/` and hosted at `https://matrix.example.com/.well-known/matrix/server` and `https://matrix.example.com/.well-known/matrix/client`, even though this is the wrong place to host them.
 
-You have 4 options when it comes to installing the files on the base domain's server:
+You have 3 options when it comes to installing the files on the base domain's server:
 
 
 ### (Option 1): **Copying the files manually** to your base domain's server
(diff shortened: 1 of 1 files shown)


spantaleev:
Thank you for improving the configuration!

As for the inlined-configuration, it might work for the simplest of cases, but:

- the `/.well-known/matrix/client` file is very simplistic. The playbook may generate a much more complete `client` file in certain situations (various components being installed; various options being set). See [roles/matrix-base/templates/static-files/well-known/matrix-client.j2](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/7d75af338c1824c57e9e6d7134c86d81d6642125/roles/matrix-base/templates/static-files/well-known/matrix-client.j2)
- other `/.well-known/matrix` files may be generated in certain situations. One example is `/.well-known/matrix/support` (still an unaccepted MSC; as such, it's not enabled/generated by default, for now) - see https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/1931

So even though the inlining method **may** work, just like the **copying the files** method, it's not ideal.

If you'd like you can add it as a 4th option though (with these caveats noted), we could do it.

spantaleev:
Looks good! Thank you! :+1:

PR #1959, opened 2022-07-19
Fix self-build error image names

xangelix:
Fixes self-build errors from an unsupported Ansible version that references only the Element image.


spantaleev:
Thanks! :+1:

PR #1969, opened 2022-07-21
Borg backup provide ssh key example format

kleo:
After copy pasting from current documentation it had a tab. I had to attempt a couple of times to get the correct formatting on yml. Private key is base64'd lorem ipsum.

Conversation https://matrix.to/#/!cNSQwPqhHKkIZdBnvt:devture.com/$otTtC0eojvNEKI5PV3p3rqCfy9BHXfHdj8cWKY_qbwk?via=devture.com&via=matrix.org&via=virtualenigma.net


spantaleev:
Thanks for improving our examples! :+1:

PR #1970, opened 2022-07-21
Fix link to maubot docs in README

anoadragon453:
This PR fixes a 404'd link in the README.


spantaleev:
Thanks! :+1: 

I'm not sure why we were calling it `matrix-maubot` either, so I shortened it to just `maubot`.

PR #1971, opened 2022-07-21
dendrite_recaptcha_siteverify_api

Maxdeso:


spantaleev:
Thanks! :+1: 

I wonder why the sample configuration leaves it empty as well: https://github.com/matrix-org/dendrite/blob/35ce551c8ffaef4d4bbfc5a4a8fd41a8b8ef2b27/dendrite-sample.monolith.yaml#L168

PR #1977, opened 2022-07-25
Add KakaoTalk bridge

hnarjis:
Adding KakaoTalk bridge, it works with the fix in the bridge PR: https://src.miscworks.net/fair/matrix-appservice-kakaotalk/pulls/3


spantaleev:
I've also submitted a PR that adds Kakaotalk support - https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/1978

It includes various other improvements (documented there). I've also taken some inspiration from yours. I believe that that PR is a better candidate for merging.

Would you review/test the other PR?

-----

The bot and its puppets are named the same way as in your PR, so you shouldn't need custom configuration to override naming.

The other PR uses `/matrix/appservice-kakaotalk` for the base path, while this one uses `/matrix/matrix-appservice-kakaotalk` (which is more consistent with the rest of the playbook), so there should be no file conflicts. The database is named `matrix_appservice_kakaotalk` in both, so it should pick up the same data. The systemd service names are the same, so it should be fine.

If all is fine, you can probably remove `/matrix/matrix-appservice-kakaotalk` completely.

hnarjis:
Closing in favor of #1978.

PR #1979, opened 2022-07-26
Double quote to prevent globbing and word splitting.

mcnesium:
[SC2086](https://www.shellcheck.net/wiki/SC2086)


spantaleev:
Thanks! :+1:

PR #1980, opened 2022-07-26
run the playbook on multiple hosts with different credentials

mcnesium:
This is a bash script to run the playbook on multiple hosts with different credentials.

It first iterates over any `inventory/<my host>.yml` to get all the sudo passwords, and then runs the playbook on each host afterwards. The PR includes an example host file and documentation. The host file comes with yaml syntax, so that the for loop in the bash script can use the asterisk wildcard syntax: `inventory/*.yml`.

I am aware that running playbooks on multiple hosts can be done using [default ansible methods](https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#using-multiple-inventory-sources). Though, this playbook ought to be usable as a standalone configuration imho.

Feel free to rephrase, extend, alter or reject anything of this contribution. Just wanted to share the work I did for my own use case.

### DIFF
--- a/docs/configuring-playbook.md
+++ b/docs/configuring-playbook.md
@@ -18,7 +18,7 @@ You can then follow these steps inside the playbook directory:
 
 1. edit the inventory hosts file (`inventory/hosts`) to your liking
 
-1. if you want to run multiple servers with different credentials, you can copy the sample inventory hosts yaml file for each of your hosts: (`cp examples/host.yml inventory/my_host1.yml` …) and use the [`ansible-all-hosts.sh`](inventory/scripts/ansible-all-hosts.sh) script [in then next step](installing.md).
+1. (optional, advaned) to run Ansible against multiple servers with different `sudo` credentials, you can copy the sample inventory hosts yaml file for each of your hosts: (`cp examples/host.yml inventory/my_host1.yml` …) and use the [`ansible-all-hosts.sh`](../inventory/scripts/ansible-all-hosts.sh) script [in then installation step](installing.md).
 
 For a basic Matrix installation, that's all you need.
 For a more custom setup, see the [Other configuration options](#other-configuration-options) below.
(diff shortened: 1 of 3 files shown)


docs/configuring-playbook.md
```diff
@@ -18,6 +18,7 @@ You can then follow these steps inside the playbook directory:
 
 1. edit the inventory hosts file (`inventory/hosts`) to your liking
 
+1. if you want to run multiple servers with different credentials, you can copy the sample inventory hosts yaml file for each of your hosts: (`cp examples/host.yml inventory/my_host1.yml` …) and use the [`ansible-all-hosts.sh`](inventory/scripts/ansible-all-hosts.sh) script [in then next step](installing.md).
```
spantaleev, line 21:
```suggestion
1. (optional, advaned) to run Ansible against multiple servers with different `sudo` credentials, you can copy the sample inventory hosts yaml file for each of your hosts: (`cp examples/host.yml inventory/my_host1.yml` …) and use the [`ansible-all-hosts.sh`](../inventory/scripts/ansible-all-hosts.sh) script [in then installation step](installing.md).
```

examples/host.yml
```diff
@@ -0,0 +1,8 @@
+---
```
spantaleev, line 1:
```suggestion
---

# This is a host file for usage with the `ansible-all-hosts.sh` script,
# which runs Ansible against a bunch of hosts, each with its own `sudo` password.
```

spantaleev:
Thanks! :+1:

etkecc:
IIRC, you can encrypt password with ansible-vault and save it to the vars.yml of your host. In such way you can run ansible with all hosts at once

mcnesium:
@etkecc that is true. We talked about this problem it in [the chat](https://matrix.to/#/#matrix-docker-ansible-deploy:devture.com) last week and there were many suggestions on how to deal with multiple home servers. Most of them were presented with a similar uncertainty regarding whether this playbook is actually meant to be used this way. Using a vault being one of them, most suggestions required further knowledge about how Ansible works, which probably exceeds the scope of the documentation of this playbook. The easiest suggestion was to use multiple host files. That in turn required to run the `ansible-playbook …` command multiple times with lots of time to wait in between, which was annoying and lead me to writing this script.

PR #1984, opened 2022-07-28
Make http_host_bind_port vars more useful

ssnailed:
With a minimally invasive change, the generated "matrix-domain.conf" nginx file could automatically support custom host port binds for the exporter containers. 
This is useful if you want to use the builtin exporters but are supplying your own nginx. 
The functionality already exists when generating the systemd service files but it seems that there was an oversight when generating this configuration file.


roles/matrix-prometheus-node-exporter/tasks/init.yml
```diff
@@ -23,10 +23,10 @@
               resolver 127.0.0.11 valid=5s;
               set $backend "matrix-prometheus-node-exporter:9100";
               proxy_pass http://$backend/metrics;
+            {% elif matrix_prometheus_postgres_exporter_container_http_host_bind_port != '' %}
+              proxy_pass http://{{ matrix_prometheus_postgres_exporter_container_http_host_bind_port }}/metrics;
```
spantaleev, line 27: `matrix_prometheus_postgres_exporter_container_http_host_bind_port` (and other `_host_bind_port` variables we have) can contain different kinds of values. For example:

- `1234` - bare port number
- `127.0.0.1:1234` - loopback interface + port number
- `0.0.0.0:1234` - all IPv4 interfaces + port number

Your quick fix is likely to work only for the `127.0.0.1:1234` case and generate invalid configuration for the others.

----

This can be solved by adding a new variable, which is defined in terms of `matrix_prometheus_postgres_exporter_container_http_host_bind_port` somehow.

Example `roles/matrix-prometheus-postgres-exporter/vars/main.yml`:

```yaml
# `matrix_prometheus_postgres_exporter_container_http_host_bind_port_number_raw` contains the raw port number extracted from `matrix_prometheus_postgres_exporter_container_http_host_bind_port`,
# which can contain values like this: ('1234', '127.0.0.1:1234', '0.0.0.0:1234')
matrix_prometheus_postgres_exporter_container_http_host_bind_port_number_raw: "{{ '' if matrix_prometheus_postgres_exporter_container_http_host_bind_port == '' else (matrix_prometheus_postgres_exporter_container_http_host_bind_port.split(':')[1] if ':' in matrix_prometheus_postgres_exporter_container_http_host_bind_port else matrix_prometheus_postgres_exporter_container_http_host_bind_port) }}
```

Example `roles/matrix-prometheus-postgres-exporter/defaults/main.yml`:

```yaml
# proxy_pass host to use when matrix-nginx-proxy is NOT enabled.
matrix_prometheus_postgres_exporter_matrix_nginx_proxy_not_enabled_proxy_pass_host: "{{ '' if matrix_prometheus_postgres_exporter_container_http_host_bind_port_number_raw == '' else ('127.0.0.1' + matrix_prometheus_postgres_exporter_container_http_host_bind_port_number_raw) }}"
```

You can then use `matrix_prometheus_postgres_exporter_matrix_nginx_proxy_not_enabled_proxy_pass_host` in `tasks/init.yml`, after checking that it's not empty.

ssnailed:
I did fail to consider that 0.0.0.0 is a possibility, but in the defaults/main.yml file it is stated that an <ip>:<port> value is expected.
I assume that allowing just a port number might be preferable so I clarified that comment. (I didn't look in depth to see if it is assumed that this value contains an IP in other parts of the playbook, so this might also be a bad change.)
I also added some checks to ensure that 127.0.0.1:1234 isn't used in init.yml when 127.0.0.55:1234 is specified. (Previously it would just take take the port value and append it to 127.0.0.1 regardless of what IP was set.)

P.S. I have been using these playbooks for a while and they're really useful! This is also my first contribution on Github so I hope I'm not making any big mistakes.
Thank you so much for all the effort put into this repo!

roles/matrix-prometheus-node-exporter/defaults/main.yml
```diff
@@ -54,3 +54,12 @@ matrix_prometheus_node_exporter_metrics_proxying_enabled: false
...
+# If matrix_prometheus_node_exporter_container_http_host_bind_port is set to just a port number, this will default to "127.0.0.1:<port>"
+# If matrix_prometheus_node_exporter_container_http_host_bind_port is set to an IP that is not  0.0.0.0 and a port, that "<ip>:<port>" value will be used
+# Otherwise this value will be empty and you will have to manually configure your NGINX config file. (If you are using the config files generated by this playbook, you will have to edit matrix-domain.conf)
+matrix_prometheus_node_exporter_matrix_nginx_proxy_not_enabled_proxy_pass_host: "{{ '127.0.0.1' + matrix_prometheus_node_exporter_container_http_host_bind_port_number_raw if not ':' in matrix_prometheus_node_exporter_container_http_host_bind_port else (matrix_prometheus_node_exporter_container_http_host_bind_port if matrix_prometheus_node_exporter_container_http_host_bind_port.split(':')[0] != '0.0.0.0' else '') }}"
+
+matrix_prometheus_node_exporter_dashboard_urls:
+  - "https://grafana.com/api/dashboards/9628/revisions/7/download"
```
spantaleev, line 65: This seems unused. Looks like it got copied over from `roles/matrix-prometheus-postgres-exporter/defaults/main.yml`.

roles/matrix-prometheus-node-exporter/tasks/init.yml
```diff
@@ -23,10 +23,10 @@
               resolver 127.0.0.11 valid=5s;
               set $backend "matrix-prometheus-node-exporter:9100";
               proxy_pass http://$backend/metrics;
+            {% elif matrix_prometheus_postgres_exporter_container_http_host_bind_port != '' %}
```
spantaleev, line 26:
```suggestion
            {% elif matrix_prometheus_node_exporter_matrix_nginx_proxy_not_enabled_proxy_pass_host != '' %}
```

roles/matrix-prometheus-postgres-exporter/tasks/init.yml
```diff
@@ -23,10 +23,10 @@
               resolver 127.0.0.11 valid=5s;
               set $backend "matrix-prometheus-postgres-exporter:9187";
               proxy_pass http://$backend/metrics;
+            {% elif matrix_prometheus_node_exporter_container_http_host_bind_port != '' %}
```
spantaleev, line 26:
```suggestion
            {% elif matrix_prometheus_postgres_exporter_matrix_nginx_proxy_not_enabled_proxy_pass_host != '' %}
```

spantaleev:
I've submitted a few more comments!

It looks good otherwise! Thank you for improving the playbook! :bow:

spantaleev:
Please note the failing CI checks. Also, my recommendations above (for these new variables, etc.) are untested, but I assume you've tested it all out.

roles/matrix-prometheus-node-exporter/defaults/main.yml
ssnailed: Yep, I fatfingered when pasting and then replaced every instance of the word postgres with node

ssnailed:
I unfortunately don't have time for a lot of in depth testing, but the playbook runs without any errors and I have fixed the linting errors.

spantaleev:
Thanks! :+1:

PR #1985, opened 2022-07-28
Add example configuration for Caddy v2

vaivars:
Add a basic example how to get synapse-admin running behind Caddy v2 proxy.


spantaleev:
Thanks! :+1:

PR #1989, opened 2022-07-30
Fix mautrix-telegram permissions

etkecc:
Because for telegram bridge there is a special `full` level to allow double-puppeting for users


spantaleev:
Oh, thanks for finding/fixing this! 👍

PR #1992, opened 2022-08-02
[Go-Skype] Change Docker Hub repository

RoiArthurB:
According to [this commit](https://github.com/kelaresg/go-skype-bridge/commit/59e27903cd6888399706ec5b7b1757411834990c), the docker hub account did change. This PR follow this update.

Also, this new image is automatically build (latest & release) for AMD64 and ARM64 :) Don't know how to apply those informations, but it can be handy :smile:


spantaleev:
Nice! Thank you! :+1: 

I've made it use the prebuilt image on `amd64` and `arm64` in cf6e38a.

PR #1997, opened 2022-08-03
Update Element 1.11.1 -> 1.11.2

etkecc:


spantaleev:
Still not pushed to Docker Hub, so this will have to wait.

PR #1998, opened 2022-08-03
Use prebuilt ARM images for Element

krassle:
* element-web arm64 builds available since 2022-08-03 v.1.11.2 [vectorim/element-web:v1.11.2](https://hub.docker.com/layers/element-web/vectorim/element-web/v1.11.2/images/sha256-776f82281936226d91cc1b3b587f4aa28fd46934b8045427ced7c72668eda223?context=explore)


spantaleev:
Nice! Thank you! :+1:

PR #1999, opened 2022-08-04
Fix Dendrite extra arguments getting lost

cvwright:
This PR moves the `matrix_dendrite_process_extra_arguments` line up so that it doesn't get lost, and adds a trailing backslash.

## Motivation
Without this PR, anything below the `..._bind_address` lines gets lost.  I ran into this bug because I needed to enable the "Really enable open registration" option on a server in a test environment.  But even after I added it to the `_extra_arguments` variable, Dendrite still wasn't seeing it.

## Alternatives
One possible approach would be to do figure out whether we have any extra args for the process or not.  And if we do, then add the trailing \ after the `_bind_address` lines.

Instead, here I just move the `_extra_arguments` line up so that it's no longer the last one.  Then everything is easy -- we always have the \ after the `_extra_arguments` line, and we never need one after `_bind_address`.


spantaleev:
Oh, good catch! Thank you for fixing it! :+1:

madprogramer:
Kudos @cvwright, I spent two hours yesterday trying to figure out _why_  `matrix_dendrite_process_extra_arguments` was not always updating. Nicely done :)

PR #2000, opened 2022-08-04
Add back sqlite plugin database config option to maubot

Cadair:
I found that all my plugins (other than rss which supports postgres) broke because this config option was missing.

@moan0s


spantaleev:
Thanks! :+1:

PR #2003, opened 2022-08-05
Fix bug that prevented user with external nginx from launch

moan0s:
The backslash was missing and prevented users from starting the bot

### DIFF
--- a/docs/configuring-playbook-bot-maubot.md
+++ b/docs/configuring-playbook-bot-maubot.md
@@ -61,23 +61,3 @@ You can expand "Access token" to copy it.
 ![Obatining an admin access token with Element](assets/obtain_admin_access_token_element.png)
 
 **IMPORTANT**: once you copy the token, just close the Matrix client window/tab. Do not "log out", as that would invalidate the token.
-
-## External nginx
-
-To use the management interface with an [external nginx webserver](configuring-playbook-own-webserver.md) the configuration will default to
-a http bind port at `127.0.0.1:29316`. If you need to use a different interface add the following to your `vars.yml` (e.g. with `0.0.0.0`).
-
-
-```yaml
-# Controls whether the maubot container exposes its HTTP management interface port (tcp/29316 in the container).
-#
-# Takes an "<ip>:<port>" or "<port>" value (e.g. "127.0.0.1:29316"), or empty string to not expose.
-# If you'll be setting this at all, it should be defined in terms of `matrix_bot_maubot_management_interface_port`.
-# Example:
-# matrix_bot_maubot_management_interface_http_bind_port: "127.0.0.1:{{ matrix_bot_maubot_management_interface_port }}"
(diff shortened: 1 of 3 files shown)


docs/configuring-playbook-bot-maubot.md
```diff
@@ -62,9 +62,21 @@ You can expand "Access token" to copy it.
 
 **IMPORTANT**: once you copy the token, just close the Matrix client window/tab. Do not "log out", as that would invalidate the token.
 
+## External nginx
```
spantaleev, line 65: Instead of having this documentation thing, why don't we adjust `matrix_bot_maubot_management_interface_http_bind_port` dynamically based on `matrix_nginx_proxy_enabled`.

You can do it via `group_vars/matrix_servers`:

```yaml
matrix_bot_maubot_management_interface_http_bind_port: "{{ '' if matrix_nginx_proxy_enabled else ('127.0.0.1:' + matrix_bot_maubot_management_interface_port | string) }}"
``

moan0s: I thought this would mess with people that did not use `127.0.0.1` but I simply added this to the configuration

spantaleev: I believe we do the same for all playbook components (when `matrix-nginx-proxy` is disabled, we expose on `127.0.0.1` by default), so it's no surprise.

spantaleev: The heading says "External nginx", but I think external-nginx people (if they run nginx on the same machine and consume the generated configs from `/matrix/nginx-proxy/conf.d`, like we advise) do not need to worry about it. We expose the port on `127.0.0.1` and inject nginx configuration which will proxy to it on `127.0.0.1`. See here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/ecfcbd6f1f9f9c44db111759c0626157680bf458/roles/matrix-bot-maubot/tasks/init.yml#L21-L26

People who may need to worry about `matrix_bot_maubot_management_interface_http_bind_port` are:

- people who use an external nginx which is on another host - they may wish to see how they do reverse-proxying for maubot, as well as all other services
- people who use a non-nginx reverse proxy server - they will need to figure out how to do reverse-proxying for maubot, as well as all other services

So, I think this documentation section can be deleted. It's just for very special cases and the same thing is documented in `roles/matrix-bot-maubot/defaults/main.yml` already. People hosting nginx on another machine or those using a non-nginx server, already have a lot of other troubles proxying all other services. They'll learn about these `_bind_port` variables sooner or later (probably sooner, as reverse-proxying Synapse and other services is a higher priority for them). Educating them about `_bind_port` doesn't have to be on this bot page.

Unless I'm missing something..?

moan0s: No you are right - thanks for the feedback and the explanation!

spantaleev:
Thanks! :+1:

PR #2007, opened 2022-08-08
Use arm64 images for Cinny

ganyuke:
Cinny now builds arm64 docker images since [v2.0.4](https://hub.docker.com/layers/cinny/ajbura/cinny/v2.0.4/images/sha256-a7202136f8568eb0397a3d644725a8fb7dca230e08bcfc42040238bda0382057?context=explore).


spantaleev:
Great! Thanks! :+1:

PR #2031, opened 2022-08-17
added rc_invites configuration for synapse

benjictln:
Hello, I had to be able to change the rate limit for the invitation event on my server.
I needed to be able to add ~50+ members quickly.

Thus, I added the config for `matrix_synapse_rc_invites` the same way it was added for `matrix_synapse_rc_joins ` ( not sure why it wasn't added before ?) 

for the default values, I took them from here `https://github.com/matrix-org/synapse/blob/f4ab6a4a96ceb02e260a3d025ff6c1e6cfefe4ed/docs/usage/configuration/config_documentation.md`

### DIFF
--- a/roles/matrix-synapse/defaults/main.yml
+++ b/roles/matrix-synapse/defaults/main.yml
@@ -142,6 +142,19 @@ matrix_synapse_rc_joins:
     per_second: 0.01
     burst_count: 3
 
+
+matrix_synapse_rc_invites:
+  per_room:
+    per_second: 0.5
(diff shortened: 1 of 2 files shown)


spantaleev:
Not sure why it wasn't added. Perhaps it got introduced later? People who needed to customize this part of the configuration could always do it via the `matrix_synapse_configuration_extension_yaml` variable anyway. Having a dedicated variable probably makes sense, given that we have the same kind of rate limit variables for other stuff.

PR #2035, opened 2022-08-19
Update beeper-linkedin v0.5.2 -> v0.5.3 (repo change)

xangelix:
beeper-linkedin has moved to GitHub at https://github.com/beeper/linkedin and a new version has been published, v0.5.3.

PR #2036, opened 2022-08-20
Remove arch from beeper-linkedin tag

xangelix:
Fixes issue with beeper-linkedin tag introduced in https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2035 (sorry!). 
The new tag naming structure does not include the architecture.


spantaleev:
Thanks for fixing it up!

I was wondering if the new tags are offering multi-architecture images. It seems like Github's container registry doesn't make it easy to say.

I've tested it though and it seems like the `v0.5.3` tag is just an amd64 image. That's what we expect right now, so it's all good. We use self-building for all other architectures.

PR #2037, opened 2022-08-20
Create dedicated doc page on obtaining access tokens

DennisCiba:
Right now, the documentation pages on how to set up bots and bridges each describe how to obtain the required access token. This means a lot of redundancy. Also the documentation is not always complete, often lacking one of the two options to generate a token. 

I propose in this PR to create a dedicated page in the documentation on how to obtain an access token, and then just link to that page on other documentation pages requiring the setup of a token, similar to how there is already a dedicated page on how to register a new user. 

Please let me know what you think.


spantaleev:
That's a great idea! I've often thought about this, but never got around to doing it. Thank you! :+1:

Issue #1144, opened 2021-06-27
Include matrix_ldap_registration_proxy to support registration on LDAP setups


TheOneWithTheBraid:
I would love to see [matrix_ldap_registration_proxy](https://gitlab.com/activism.international/matrix_ldap_registration_proxy) included into this role.

matrix_ldap_registration_proxy is a simple drop-in proxy enabling registration on LDAP setups by catching the registration requests and forwarding them to an LDAP server.
It's dockerized by default.

PR #2038, opened 2022-08-21
Draft: feat: include matrix_ldap_registration_proxy

TheOneWithTheBraid:
Fixes: #1144

Signed-off-by: TheOneWithTheBraid <the-one@with-the-braid.cf>


roles/matrix-ldap-registration-proxy/defaults/main.yml
```diff
@@ -0,0 +1,47 @@
...
+
+# List of systemd services that matrix_ldap_registration_proxy.service wants
+matrix_ldap_registration_proxy_systemd_wanted_services_list: []
+
+# Default ma1sd configuration template which covers the generic use case.
+# You can customize it by controlling the various variables inside it.
+matrix_ldap_registration_proxy_configuration_env: "{{ lookup('template', 'templates/ldap-registration-proxy.env.j2') }}"
+
+# Holds the final ma1sd configuration (a combination of the default and its extension).
+matrix_ldap_registration_proxy_configuration: "{{ matrix_ldap_registration_proxy_configuration_env }}"
```
spantaleev, line 47: There are some `ma1sd` leftovers here.

Also, it may be unnecessary to go through such a complicated process (`template`, etc.) for some `env` file.

In other roles where we have env files like this, we usually offer some `_extension` variable which lets people append additional environment variables to the file. All existing environment variables (part of the default `env` file) can be redefined using playbook variables (like `matrix_ldap_registration_proxy_ldap_base_dn`, for example).

You can take inspiration from the `matrix-jitsi` role.

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5fafbab7ae4b6fd090c152144932da6e57f883cd/roles/matrix-jitsi/defaults/main.yml#L162-L172

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5fafbab7ae4b6fd090c152144932da6e57f883cd/roles/matrix-jitsi/templates/web/env.j2#L110

roles/matrix-ldap-registration-proxy/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
+---
+
+- ansible.builtin.import_tasks: "{{ role_path }}/tasks/init.yml"
+  tags:
+    - always
+
+- ansible.builtin.import_tasks: "{{ role_path }}/tasks/validate_config.yml"
```
spantaleev, line 7: `validate_config.yml` being empty (as it is now) will cause some Ansible error when importing.

In another review comment, I'm advising that you start making use of this file though (adding stuff to `validate_config.yml`), so it will no longer be a problem.

roles/matrix-ldap-registration-proxy/tasks/self_check_matrix_ldap_registration_proxy.yml
```diff
@@ -0,0 +1,22 @@
+---
+
+- ansible.builtin.set_fact:
+    matrix_ldap_registration_proxy_url_endpoint_public: "https://{{ matrix_server_fqn_matrix }}/_matrix/client/r0/register"
```
spantaleev, line 4: This is exercising the homeserver more than the LDAP proxy though, right?

This API endpoint will always return stuff if the homeserver is up, regardless of whether it's wired to use LDAP proxy correctly or not.

Maybe removing this whole self-check thing is a good idea.

spantaleev (changes requested):
Thanks for working on adding this! I'm submitting some feedback

TheOneWithTheBraid:
@spantaleev Seems like a fellow added the requested changes.

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
```diff
@@ -129,6 +129,20 @@
...
+                {# Use the embedded DNS resolver in Docker containers to discover the service #}
+                resolver 127.0.0.11 valid=5s;
+                set $backend "{{ matrix_nginx_proxy_proxy_matrix_3pid_registration_addr_with_container }}";
+                proxy_pass http://$backend/register;
+            {% else %}
+                {# Generic configuration for use outside of our container setup #}
+                proxy_pass http://{{ matrix_nginx_proxy_proxy_matrix_3pid_registration_addr_sans_container }}/register;
+            {% endif %}
+        }
+    {% endif %}
```
spantaleev, line 144: There are 2 ways to do this in a more independent way (one that doesn't cross-reference variables between `matrix-nginx-proxy` and the `matrix-ldap-registration-proxy` role:

One is by introducing new variables in the `matrix-nginx-proxy` role:
- for enabling this behavior (disabled by default)
- for the actual addresses that it will proxy to

`matrix_nginx_proxy_proxy_matrix_3pid_registration_addr_with_container` and `matrix_nginx_proxy_proxy_matrix_3pid_registration_addr_sans_container` that you're using here are used for other purposes (forwarding email 3pid stuff to ma1sd).. Not sure if it's a good idea to reuse them for your own purposes - for proxying to `ldap-registration-proxy`. I also don't see any logic that points these variables to `ldap-registration-proxy`, so I'm not sure how it's expected to work.. Or am I missing something?

-----

The other (**better**) way to solve this is **without adding new variables** to the `matrix-nginx-proxy` role.

Your role can inject additional configuration into the `matrix-nginx-proxy` role by injecting blocks into `matrix_nginx_proxy_proxy_matrix_additional_server_configuration_blocks` at runtime.

This is typically done from the role's `tasks/init.yml` file.

See an example from the `matrix-bridge-mautrix-telegram` role here https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/427806674980325d308987661df305d7ba694b02/roles/matrix-bridge-mautrix-telegram/tasks/init.yml#L30-L75

For this to work, your role needs to be defined **before** the `matrix-nginx-proxy` role in the main [`setup.yml` file](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/setup.yml). Speaking of which, I don't see you adjusting the `setup.yml` file in this PR, so this role will not be taken into account at all when running the playbook.

group_vars/matrix_servers
```diff
@@ -1538,6 +1538,26 @@ matrix_jitsi_etherpad_base: "{{ matrix_etherpad_base_url if matrix_etherpad_enab
...
+######################################################################
+
+# This is only for users with a specific LDAP setup
+matrix_ldap_registration_proxy_enabled: false
+
+# Use the LDAP values specified for the synapse role to setup LDAP proxy
+matrix_ldap_registration_proxy_ldap_uri: "{{ matrix_synapse_ext_password_provider_ldap_uri }}"
+matrix_ldap_registration_proxy_ldap_base_dn: "{{ matrix_synapse_ext_password_provider_ldap_base }}"
+matrix_ldap_registration_proxy_ldap_user: "{{ matrix_synapse_ext_password_provider_ldap_bind_dn }}"
+matrix_ldap_registration_proxy_ldap_password: "{{ matrix_synapse_ext_password_provider_ldap_bind_password }}"
```
spantaleev, line 1554: Defining these settings in terms of `matrix_synapse_ext_password_provider_ldap_` is convenient if you're switching from the LDAP password provider for Synapse.. However it makes me think that `ldap-registration-proxy` **must** be used together with the LDAP password provider for Synapse. Is that true?

If it can be used on its own, you'd better separate these (by removing all these variable re-definitions here).

If these 2 components really need to be used together, you'd better:

- update [the upstream repository](https://gitlab.com/activism.international/matrix_ldap_registration_proxy), as it's not giving any such indication
- add documentation in `roles/matrix-ldap-registration-proxy/defaults/main.yml`, etc.

----

Speaking of which, adding a `docs/` page about this component and how to install/use it within the context of the playbook would be beneficial.

Does one simply need to set `matrix_ldap_registration_proxy_enabled` to `true` and run the playbook (`--tags=setup-all,start`)? Or does one need to configure the LDAP password provider for Synapse first?

Any other things that need to be done?

TheOneWithTheBraid:
CC: @moan0s

roles/matrix-ldap-registration-proxy/defaults/main.yml
```diff
@@ -0,0 +1,57 @@
...
+matrix_ldap_registration_proxy_self_check_validate_certificates: true
+
+matrix_ldap_registration_proxy_container_port: 8080
+# Controls whether the matrix_ldap_registration_proxy container exposes its HTTP port (tcp/{{ matrix_ldap_registration_proxy_container_port }} in the container).
+#
+# Takes an "<ip>:<port>" or "<port>" value (e.g. "127.0.0.1:8080"), or empty string to not expose.
+matrix_ldap_registration_proxy_container_http_host_bind_port: ''
+
+matrix_ldap_registration_proxy_registration_addr_with_container: "matrix-ldap_-egistration-proxy:{{ matrix_ldap_registration_proxy_container_port }}"
+matrix_ldap_registration_proxy_registration_addr_sans_container: "127.0.0.1:{{ matrix_ldap_registration_proxy_container_port }}"
```
spantaleev, line 39: This looks incorrect.

It should probably use the port value from `matrix_ldap_registration_proxy_container_http_host_bind_port` instead of `matrix_ldap_registration_proxy_container_port`.

You can't use `matrix_ldap_registration_proxy_container_http_host_bind_port` directly, because it may contain a port value (e.g. `1234`), but it may also contain an IP address + port value (e.g. `127.0.0.1:1234`). We probably need an additional variable `matrix_ldap_registration_proxy_container_http_host_bind_port_number_raw`. See `matrix_prometheus_node_exporter_container_http_host_bind_port_number_raw` (defined in `roles/matrix-prometheus-node-exporter/vars/main.yml`) for an example: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5f9f891322effa21d81c468cdcde701b6ba8c72c/roles/matrix-prometheus-node-exporter/vars/main.yml#L1-L5

roles/matrix-ldap-registration-proxy/templates/systemd/matrix-ldap-registration-proxy.service.j2
```diff
@@ -0,0 +1,43 @@
...
+{% for service in matrix_ldap_registration_proxy_systemd_wanted_services_list %}
+Wants={{ service }}
+{% endfor %}
+DefaultDependencies=no
+
+[Service]
+Type=simple
+Environment="HOME={{ matrix_systemd_unit_home_path }}"
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} kill matrix_ldap_registration_proxy 2>/dev/null || true'
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} rm matrix_ldap_registration_proxy 2>/dev/null || true'
```
spantaleev, line 17:
```suggestion
ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} kill matrix-ldap-registration-proxy 2>/dev/null || true'
ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} rm matrix-ldap-registration-proxy 2>/dev/null || true'
```

spantaleev:
Looks like there are only a few things left to do and this can be merged

group_vars/matrix_servers
moan0s: I don't know. If it needs to be used together we could indicate that in the docs, else I would move this block to group vars and activate it when the LDAP password provider for Synapse is used.
@TheOneWithTheBraid has to answer this, I don't have much knowledge of LDAP setups

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
moan0s: Done

roles/matrix-ldap-registration-proxy/templates/systemd/matrix-ldap-registration-proxy.service.j2
moan0s: Done

docs/configuring-playbook-matrix-ldap-registration-proxy.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Quickstart
+
+Add the following configuration to your `inventory/host_vars/matrix.DOMAIN/vars.yml` file:
+
+```yaml
+matrix_ldap_registration_proxy_enabled: true
+```
+
+That is enough if you use the synapse external password provider via LDAP.
```
spantaleev, line 18:
```suggestion
This is enough **if** you use the [synapse external password provider via LDAP](docs/configuring-playbook-ldap-auth.md) (that is, you have `matrix_synapse_ext_password_provider_ldap_enabled: true` and other options in your configuration).
```

Still, as discussed [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2038#discussion_r963057592) I wonder if this is a good idea.

Are these 2 things used together often? How does it help if you use both at the same time? If there is a benefit, we should probably document it as well. I bet lots of people will wonder "which one do I use? how is this one better? is it better still if I enable both at the same time?".

My understanding is that:

- the Synapse LDAP auth provider just handles login requests and tries to authenticate them against an LDAP server
- ldap-registration-proxy handles registration requests and creates users on the LDAP server (open registration?)

.. so they do complement each other and it's useful to have both at the same time (if you'd like both login and open registration functionality).

moan0s: I have the same understanding of the functionality. So I'd say there is no benefit to use them together, it is just possible. If you do, it could be an option (but not a good one if I think about it opinion) to reuse the credentials of Synapse LDAP auth provider. I think it would be better to have both roles specify which credentials to use and I'll add that to the docs and check in the `validate_config.yml ` if the values are set. An advanced user could then set these to 
```
matrix_ldap_registration_proxy_ldap_uri: "{{ matrix_synapse_ext_password_provider_ldap_uri }}"
matrix_ldap_registration_proxy_ldap_base_dn: "{{ matrix_synapse_ext_password_provider_ldap_base }}"
matrix_ldap_registration_proxy_ldap_user: "{{ matrix_synapse_ext_password_provider_ldap_bind_dn }}"
matrix_ldap_registration_proxy_ldap_password: "{{ matrix_synapse_ext_password_provider_ldap_bind_password }}"
```

spantaleev: Isn't the benefit that one handles logins and the other handles new registrations (for an open registration server)?

TheOneWithTheBraid: Exactly, that's the idea.

group_vars/matrix_servers
TheOneWithTheBraid: They are *technically* completely independent from each other even though I's state the ldap_registration_proxy does not make any sense without the LDAP password provider in most cases.

But of course, theoretically, one could use the ldap_registration_proxy in order to register users while any other unknown implementation handles the channel back from LDAP to synapse or maybe any other homeserver implementation.

TheOneWithTheBraid: I'd anyway appreciate naming the ldap_registration_proxy on the docs' LDAP page along with the Synapse ext password LDAP and ma1dn LDAP.

docs/configuring-playbook-matrix-ldap-registration-proxy.md
moan0s: But as far as I know they work independently from each other so I would not tie them together in `group_vars`

PR #2043, opened 2022-08-22
add postmoogle

etkecc:


group_vars/matrix_servers
```diff
@@ -1198,6 +1198,35 @@ matrix_bot_buscarron_container_image_self_build: "{{ matrix_architecture not in
...
+    +
+    (['matrix-postgres.service'] if matrix_postgres_enabled else [])
+    +
+    (['matrix-synapse.service'] if matrix_synapse_enabled else [])
+  }}
+
+# Postgres is the default, except if not using `matrix_postgres` (internal postgres)
+matrix_bot_postmoogle_database_engine: "{{ 'postgres' if matrix_postgres_enabled else 'sqlite' }}"
+matrix_bot_postmoogle_database_password: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'postmoogle.db') | to_uuid }}"
+matrix_bot_postmoogle_container_image_self_build: "{{ matrix_architecture not in ['amd64', 'arm32', 'arm64'] }}"
```
spantaleev, line 1222:
```suggestion

matrix_bot_postmoogle_container_image_self_build: "{{ matrix_architecture not in ['amd64', 'arm32', 'arm64'] }}"
```

Separating this a bit from the `_database_` stuff makes it easier to see. We normally put this more toward the top of the configuration too.

roles/matrix-bot-postmoogle/defaults/main.yml
```diff
@@ -0,0 +1,105 @@
...
+matrix_bot_postmoogle_noowner: false
+
+# Allow Postmoogle use by users over federation
+matrix_bot_postmoogle_federation: false
+
+# Sentry DSN
+matrix_bot_postmoogle_sentry: ''
+
+# Log level
+matrix_bot_postmoogle_loglevel: ''
```
spantaleev, line 91: No default log level? What are the possible options here?

roles/matrix-bot-postmoogle/tasks/setup_install.yml
```diff
@@ -0,0 +1,99 @@
...
+  ansible.builtin.file:
+    path: "{{ item.path }}"
+    state: directory
+    mode: 0750
+    owner: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_groupname }}"
+  with_items:
+    - {path: "{{ matrix_bot_postmoogle_config_path }}", when: true}
+    - {path: "{{ matrix_bot_postmoogle_data_path }}", when: true}
+    - {path: "{{ matrix_bot_postmoogle_docker_src_files_path }}", when: true}
```
spantaleev, line 37:
```suggestion
    - {path: "{{ matrix_bot_postmoogle_docker_src_files_path }}", when: matrix_bot_postmoogle_container_image_self_build}
```

roles/matrix-bot-postmoogle/templates/systemd/matrix-bot-postmoogle.service.j2
```diff
@@ -0,0 +1,40 @@
...
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} rm matrix-bot-postmoogle 2>/dev/null || true'
+
+ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-bot-postmoogle \
+			--log-driver=none \
+			--user={{ matrix_user_uid }}:{{ matrix_user_gid }} \
+			--cap-drop=ALL \
+			--read-only \
+			--network={{ matrix_docker_network }} \
+			--env-file={{ matrix_bot_postmoogle_config_path }}/env \
+			-p {{ matrix_bot_postmoogle_port }}:{{ matrix_bot_postmoogle_port }} \
```
spantaleev, line 26: Isn't this (with the default configuration) trying to bind to port `25` inside the container?

Since it's a [privileged port](https://www.w3.org/Daemon/User/Installation/PrivilegedPorts.html) (` < 1024`), this is supposed to require special permissions. We either need to run the container as `root` (not great) or add `--cap-add=NET_ADMIN` (also not great).

Alternatively, we can make the ports different. We can make Postmoogle (in the container) bind to a non-privileged port (e.g. `2525`) and then have Docker itself do the forwarding (`-p 25:2525`).

----

The `email2matrix` bridge hardcodes port `2525` for the container program and only ever changes the exposed port, controlled by a `matrix_email2matrix_smtp_host_bind_port` setting.

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5fafbab7ae4b6fd090c152144932da6e57f883cd/roles/matrix-email2matrix/templates/config.json.j2#L4

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5fafbab7ae4b6fd090c152144932da6e57f883cd/roles/matrix-email2matrix/templates/systemd/matrix-email2matrix.service.j2#L20

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5fafbab7ae4b6fd090c152144932da6e57f883cd/roles/matrix-email2matrix/defaults/main.yml#L25-L30

roles/matrix-bot-postmoogle/defaults/main.yml
```diff
@@ -0,0 +1,105 @@
...
+
+# Disable encryption
+matrix_bot_postmoogle_noencryption: false
+
+matrix_bot_postmoogle_domain: "{{ matrix_server_fqn_matrix }}"
+
+matrix_bot_postmoogle_port: "25"
```
spantaleev, line 98: I was wondering "why not a number here instead of a string", but I guess this is similar to the `_host_bind` varialbes we have in other roles, where it could either be a port number or an IP address + port (e.g. `127.0.0.1:25`, `0.0.0.0:25`, etc.).

---

Also, see my other comment (for `matrix-bot-postmoogle.service.j2`) about potentially redoing this in a way that doesn't use privileged ports.

Renaming it for consistency with the other _`host_bind_port` variables and adding comments like see here may also be beneficial, so that users know how to use it:

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5fafbab7ae4b6fd090c152144932da6e57f883cd/roles/matrix-email2matrix/defaults/main.yml#L25-L30

Of course, as of right now, `matrix_bot_postmoogle_port` is really meant to be a port number, because it's passed directly to Postmoogle as well, but that may change. See the other review comment about `matrix-bot-postmoogle.service.j2`.

docs/configuring-playbook-bot-postmoogle.md
```diff
@@ -0,0 +1,54 @@
...
+After configuring the playbook, run the [installation](installing.md) command again:
+
+```
+ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start
+```
+
+
+## Usage
+
+To use the bot, invite the `@postmoogle:DOMAIN` into a room you want to use as a mailbox, after that send `!pm mailbox NAME` to enable email address `NAME@matrix.domain` and start receiving emails
```
spantaleev, line 50:
```suggestion
To use the bot, invite the `@postmoogle:DOMAIN` into a room you want to use as a mailbox.

Then send `!pm mailbox NAME` to expose this Matrix room as an inbox with the email address `NAME@matrix.domain`. Emails sent to that email address will be forwarded to the room.
```

Splitting this long sentence into multiple simpler and more descriptive sentences.

spantaleev (changes requested):
Submitted some feedback

roles/matrix-bot-postmoogle/defaults/main.yml
etkecc: Default log level set in [separate logger library and it's `INFO` by default](https://gitlab.com/etke.cc/go/logger/-/blob/main/logger.go#L65). Choices: https://gitlab.com/etke.cc/go/logger/-/blob/main/logger.go#L65 (case doesn't matter)

roles/matrix-bot-postmoogle/templates/systemd/matrix-bot-postmoogle.service.j2
etkecc: > Isn't this (with the default configuration) trying to bind to port 25 inside the container?

Yes, but you can do that inside the container, [even with custom user `postmoogle`](https://gitlab.com/etke.cc/postmoogle/-/blob/main/Dockerfile#L17). It works with the 25 port without additional capability

roles/matrix-bot-postmoogle/defaults/main.yml
etkecc: Sure. In that case I'll add additional variable, because `POSTMOOGLE_PORT` is actual env var of the postmoogle config

spantaleev (approved):

PR #2045, opened 2022-08-23
appservice: add and use homeserver_container_* vars

shalzz:


group_vars/matrix_servers
```diff
@@ -2337,6 +2337,9 @@ matrix_synapse_redis_password: "{{ matrix_redis_connection_password if matrix_re
 #
 ######################################################################
 
+matrix_synapse_container_extra_arguments: "{{ matrix_homeserver_container_extra_arguments }}"
+matrix_synapse_app_service_config_files: "{{ matrix_homeserver_container_app_service_config_files }}"
```
spantaleev, line 2341: These should be a few lines above, so that they are within the `matrix-synapse` section.

group_vars/matrix_servers
```diff
@@ -2602,6 +2605,9 @@ matrix_dendrite_systemd_wanted_services_list: |
 #
 ######################################################################
 
+matrix_dendrite_container_extra_arguments: "{{ matrix_homeserver_container_extra_arguments }}"
+matrix_dendrite_app_service_config_files: "{{ matrix_homeserver_container_app_service_config_files }}"
+
```
spantaleev, line 2610: These should be a few lines above, so that they are within the `matrix-dendrite` section.

spantaleev (changes requested):

spantaleev:
Something else that worries me is that people may use `matrix_synapse_container_extra_arguments` and `matrix_synapse_app_service_config_files` in their own `vars.yml` file already.

Right now, you can populate these variables manually and then have the playbook inject additional stuff into them during runtime. Your own manually-added configuration will still be there.

After this PR's changes, we don't allow people to change `matrix_synapse_container_extra_arguments` and `matrix_synapse_app_service_config_files` anymore. People who do change these variables will find that:

- their changes are indeed still applied
- they would now be overriding the default wiring (`matrix_synapse_container_extra_arguments: "{{ matrix_homeserver_container_extra_arguments }}"` and  `matrix_synapse_app_service_config_files: "{{ matrix_homeserver_container_app_service_config_files }}"`). Given that they're throwing it away, they'll notice that bridges will no longer be working

So I think we need to think more about how we handle this.

group_vars/matrix_servers
shalzz: Oh, I see. FIxed.

group_vars/matrix_servers
shalzz: Fixed

shalzz:
> Something else that worries me is that people may use `matrix_synapse_container_extra_arguments` and `matrix_synapse_app_service_config_files` in their own `vars.yml` file already.
> 
> Right now, you can populate these variables manually and then have the playbook inject additional stuff into them during runtime. Your own manually-added configuration will still be there.
> 
> After this PR's changes, we don't allow people to change `matrix_synapse_container_extra_arguments` and `matrix_synapse_app_service_config_files` anymore. People who do change these variables will find that:
> 
>     * their changes are indeed still applied
> 
>     * they would now be overriding the default wiring (`matrix_synapse_container_extra_arguments: "{{ matrix_homeserver_container_extra_arguments }}"` and  `matrix_synapse_app_service_config_files: "{{ matrix_homeserver_container_app_service_config_files }}"`). Given that they're throwing it away, they'll notice that bridges will no longer be working
> 
> 
> So I think we need to think more about how we handle this.

How about, we append `matrix_homeserver_container_extra_arguments` to the `matrix_synapse_container_extra_arguments` variable instead of overwriting it?

Something like this (I'm not sure ansible syntax will allow this):
`matrix_synapse_container_extra_arguments: "{{ matrix_synapse_container_extra_arguments + matrix_homeserver_container_extra_arguments }}"`

spantaleev:
Ansible doesn't allow such a thing:

```yaml
matrix_synapse_container_extra_arguments: "{{ matrix_synapse_container_extra_arguments + matrix_homeserver_container_extra_arguments }}"
```

----

Here's a potential solution.

We can define 2 new variables for each pair. Example:

```yaml
# .. Same as before ..
matrix_synapse_container_extra_arguments: []

# matrix_synapse_container_runtime_injected_arguments is a list of extra arguments to pass to the container.
# This list is built during runtime. You're not meant to override this variable.
# If you'd like to inject your own arguments, see `matrix_synapse_container_extra_arguments`.
matrix_synapse_container_runtime_injected_arguments: []

# matrix_synapse_container_arguments holds the final list of extra arguments to pass to the container.
# You're not meant to override this variable.
# If you'd like to inject your own arguments, see `matrix_synapse_container_extra_arguments`.
matrix_synapse_container_arguments: "{{ matrix_synapse_container_extra_arguments + matrix_synapse_container_runtime_injected_arguments }}"
```

I believe that it may be possible to define `matrix_synapse_container_runtime_injected_arguments` and `matrix_synapse_container_arguments` in `roles/matrix-synapse/vars/main.yml` even, but I haven't tested it.

`group_vars/matrix_servers` will then inject `matrix_homeserver_container_extra_arguments` into `matrix_synapse_container_runtime_injected_arguments`.

We may even rename `matrix_homeserver_container_extra_arguments` to `matrix_homeserver_container_runtime_injected_arguments`, as that makes more sense.

Roles will then inject themselves into `matrix_homeserver_container_runtime_injected_arguments`.

---

What do you think?

shalzz:
Okay, makes sense. I've made the changes.

roles/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -1647,7 +1647,7 @@ room_prejoin_state:
 
 # A list of application service config files to use
 #
-app_service_config_files: {{ matrix_synapse_app_service_config_files|to_json }}
+app_service_config_files: {{ matrix_synapse_config_files|to_json }}
```
spantaleev, line 1650: `matrix_synapse_config_files` seems like an odd and inconsistent name.

We try to make variable names match the setting that they control. In this case, we can't use `matrix_synapse_app_service_config_files` as discussed before, because that would break backward compatibility.

We can probably think of a different name though. One which is more consistent.

How about `matrix_synapse_app_service_config_files_final` or `matrix_synapse_app_service_config_files_combined`?

roles/matrix-synapse/defaults/main.yml
```diff
@@ -63,6 +63,16 @@ matrix_synapse_container_manhole_api_host_bind_port: ''
 # A list of extra arguments to pass to the container
 matrix_synapse_container_extra_arguments: []
```
spantaleev, line 64:
```suggestion
# Also see `matrix_synapse_container_arguments`
matrix_synapse_container_extra_arguments: []
```

roles/matrix-synapse/defaults/main.yml
```diff
@@ -326,6 +336,16 @@ matrix_synapse_additional_loggers: []
 # You may wish to use this together with `matrix_synapse_container_additional_volumes` or `matrix_synapse_container_extra_arguments`.
 matrix_synapse_app_service_config_files: []
```
spantaleev, line 337:
```suggestion
# Also see `matrix_synapse_app_service_config_files_final`
matrix_synapse_app_service_config_files: []
```

(well, or whatever name you decide to go with when renaming the `matrix_synapse_config_files` varialbe, as proposed in another review comment)

roles/matrix-dendrite/defaults/main.yml
```diff
@@ -106,6 +116,16 @@ matrix_dendrite_container_additional_volumes: []
 # You may wish to use this together with `matrix_dendrite_container_additional_volumes` or `matrix_dendrite_container_extra_arguments`.
 matrix_dendrite_app_service_config_files: []
```
spantaleev, line 117:
```suggestion
# Also see `matrix_dendrite_app_service_config_files_final`
matrix_dendrite_app_service_config_files: []
```

roles/matrix-dendrite/defaults/main.yml
```diff
@@ -106,6 +116,16 @@ matrix_dendrite_container_additional_volumes: []
...
 
+# matrix_dendrite_app_service_runtime_injected_config_files is a list of appservice config files.
+# This list is built during runtime. You're not meant to override this variable.
+# If you'd like to inject your own arguments, see `matrix_dendrite_app_service_config_files`.
+matrix_dendrite_app_service_runtime_injected_config_files: []
+
+# matrix_dendrite_config_files holds the final list of config files to pass to the container.
+# You're not meant to override this variable.
+# If you'd like to inject your own arguments, see `matrix_dendrite_app_service_config_files`.
+matrix_dendrite_config_files: "{{ matrix_dendrite_app_service_config_files + matrix_dendrite_app_service_runtime_injected_config_files }}"
```
spantaleev, line 127:
```suggestion
# matrix_dendrite_app_service_config_files_final holds the final list of config files to pass to the container.
# You're not meant to override this variable.
# If you'd like to inject your own arguments, see `matrix_dendrite_app_service_config_files`.
matrix_dendrite_app_service_config_files_final: "{{ matrix_dendrite_app_service_config_files + matrix_dendrite_app_service_runtime_injected_config_files }}"
```

roles/matrix-dendrite/defaults/main.yml
```diff
@@ -45,6 +45,16 @@ matrix_dendrite_container_https_host_bind_address: ""
 # A list of extra arguments to pass to the container (`docker run` command)
 matrix_dendrite_container_extra_arguments: []
```
spantaleev, line 46:
```suggestion
# Also see `matrix_dendrite_container_arguments`
matrix_dendrite_container_extra_arguments: []
```

roles/matrix-dendrite/templates/dendrite/dendrite.yaml.j2
```diff
@@ -145,7 +145,7 @@ app_service_api:
   disable_tls_validation: {{ matrix_dendrite_disable_tls_validation|to_json }}
 
   # Appservice configuration files to load into this homeserver.
-  config_files: {{ matrix_dendrite_app_service_config_files|to_json }}
+  config_files: {{ matrix_dendrite_config_files|to_json }}
```
spantaleev, line 148:
```suggestion
  config_files: {{ matrix_dendrite_app_service_config_files_final|to_json }}
```

spantaleev (changes requested):
Looks like we're almost done! Thank you for your patience!

shalzz:
Okay, applied your suggestions @spantaleev. The CI failure looks unrelated.
(Feel free to squash the commits if you want before merging)

spantaleev:
Looks good! Thank you! :+1:

PR #2048, opened 2022-08-23
dendrite: fix user-registration command

shalzz:
It currently fails with:
```
FATA[0003] Failed to create the account: unable to get nonce: Get "https://localhost:8448/_synapse/admin/v1/register": dial tcp 127.0.0.1:8448: connect: connection refused
```

Also go ahead and disable open registration by default to match the synapse defaults.

### DIFF
--- a/roles/matrix-dendrite/defaults/main.yml
+++ b/roles/matrix-dendrite/defaults/main.yml
@@ -83,7 +83,7 @@ matrix_dendrite_rate_limiting_threshold: 5
 matrix_dendrite_rate_limiting_cooloff_ms: 500
 
 # Controls whether people with access to the homeserver can register by themselves.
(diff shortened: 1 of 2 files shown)


roles/matrix-dendrite/defaults/main.yml
```diff
@@ -83,7 +83,7 @@ matrix_dendrite_rate_limiting_threshold: 5
 matrix_dendrite_rate_limiting_cooloff_ms: 500
 
 # Controls whether people with access to the homeserver can register by themselves.
-matrix_dendrite_registration_disabled: false
+matrix_dendrite_registration_disabled: true
```
spantaleev, line 86: Huh! I wonder why this wasn't set to `true` to begin with. Looks like a huge omission

PR #2052, opened 2022-08-24
dendrite: enable MSC2946 (Spaced Summary) by default

shalzz:


spantaleev:
Is this an upstream default? If not, why should we enable experimental things?

Note that you can always enable it just for yourself by making use of `matrix_dendrite_configuration_extension_yaml`: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/f674afe5e8d7d4df50ae9f92439f0818c8086a0d/roles/matrix-dendrite/defaults/main.yml#L176-L191

shalzz:
Okay, was enabling it since it was enabled by default in synapse, but closing it now.

PR #2086, opened 2022-09-02
Make Conduit's rocksdb_max_open_files parameter configurable

cvwright:
This PR exposes the Conduit configuration variable `rocksdb_max_open_files` to the playbook and sets it to a slightly higher default value.

The parameter exists to prevent RocksDB from opening too many files and exceeding the system's maximum number of open files.

By default, Conduit sets the parameter to 20 if no value is given in the config file.  Here I'm setting it to 64 instead, so that we avoid creating a bottleneck on larger systems.  Some RocksDB documentation suggests allowing for at least one open file per CPU.

### DIFF
--- a/roles/matrix-conduit/defaults/main.yml
+++ b/roles/matrix-conduit/defaults/main.yml
@@ -36,6 +36,11 @@ matrix_conduit_template_conduit_config: "{{ role_path }}/templates/conduit/condu
 # Max size for uploads, in bytes
 matrix_conduit_max_request_size: 20_000_000
 
+# Maximum number of open files for Conduit's embedded RocksDB database
+# See https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide#tuning-other-options
+# If not specified, Conduit defaults to a relatively low value of 20
+matrix_conduit_rocksdb_max_open_files: 64
+
 # Enables registration. If set to false, no users can register on this server.
 matrix_conduit_allow_registration: false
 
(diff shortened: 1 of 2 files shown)


spantaleev:
Thanks!

Not sure if this is helpful though:

> If not specified, Conduit defaults to a relatively low value of 20

It makes it sound like not specifying that variable (or unsetting/clearing) will somehow make it default to 20.

Will it though? Or will `rocksdb_max_open_files = ` cause a configuration syntax error?

cvwright:
If `rocksdb_max_open_files` is not in `conduit.toml`, then Conduit will set it to 20.

If the file says `rocksdb_max_open_files = ` without giving a value, then I expect that's a syntax error.

spantaleev:
I suspect so too, which is why I was thinking that telling people this:

> If not specified, Conduit defaults to a relatively low value of 20

right next to the variable definition, could potentially make them want to unset it (in an effort to get the actual default value hardcoded in Conduit), and thus get a syntax error.

This comment being there doesn't seem of much value to anyone.

cvwright:
> right next to the variable definition, could potentially make them want to unset it (in an effort to get the actual default value hardcoded in Conduit), and thus get a syntax error.
> 
> This comment being there doesn't seem of much value to anyone.

Ah right, now I see what you mean.  Yes, I agree.  Please feel free to delete that comment.

Or I can submit another PR to clean up the wording, e.g. "By default, Conduit uses a relatively low value of 20".

Issue #2024, opened 2022-08-14
Adding Cactus Comments as a service to this playbook


N07070:
Hello everyone,

I've been pondering on using Cactus comments for my blog for a while, and as I've found time to reinstall this playbook on my servers, I was working on installing Cactus Comments too. 

Now, it seems that a role to install Cactus Comments via Ansible doesn't exist yet. As such, I wanted to create one. Would anyone be interested in working on one together ? 

I would also very much like for the role to be included in this playbook, as it would make it much much easier for me to manage all of it. Would there be interest in it ? Or at least, does anyone oppose it ? 

Cheers,

spantaleev:
I haven't heard of anyone working on a Cactus Comments role. It sounds like a nice addition to the playbook though! If you'd like to work on a PR, that'd be great! If you need any guidance, feel free to ask here or in [our support room](https://github.com/spantaleev/matrix-docker-ansible-deploy#support).

N07070:
Thank for your answer ! I'm already in the support room :) 

I'll let you know there ( and here ) when I make progress on the role.

moan0s:
I also use Cactus Comments and would love to bring this to life. Feel free to drop a DM to @moanos:hyteck.de if you need help

moan0s:
I started working on it here: https://github.com/moan0s/matrix-docker-ansible-deploy/tree/cactus-comments
Help or advice welcome (and don't expect fast progress, I am on vacation)

PR #2089, opened 2022-09-05
Cactus comments

moan0s:
As mentioned in #2024 I worked on adding the cactus comments server.
This role is tested but not extensively.

### DIFF
--- a/docs/configuring-playbook-cactus-comments.md
+++ b/docs/configuring-playbook-cactus-comments.md
@@ -1,0 +1,43 @@
+# Setting up Cactus Comments (optional)
+
+The playbook can install and configure [Cactus Comments](https://cactus.chat) for you.
+
+Cactus Comments is a **federated comment system** built on Matrix. The role allows you to self-host the system.
+It respects your privacy, and puts you in control.
+
+See the project's [documentation](https://cactus.chat/docs/getting-started/introduction/) to learn what it
+does and why it might be useful to you.
+
+
+## Configuration
+
+Add the following block to your `vars.yaml` and make sure to exchange the tokens to randomly generated values.
+
+```ỳaml
+#################
+## Cactus Chat ##
+#################
+
+matrix_cactus_comments_enabled: true
+matrix_cactus_comments_hs_token: wXchqGY94uhlk4OQ2ObPf0jsOMsnqXE9xuS2szB5
+matrix_cactus_comments_as_token: cvv8Y6OhAJqDbd1KCpRhc4mecZE9xuS2skBopaU5
+
+# To allow guest comments without users needing to log in, you need to have guest registration enabled.
+# To do this you need to uncomment one of the following lines (depending if you are using synapse or dentrite as a homeserver)
+# If you don't know which one you use: The default is synapse ;)
+# matrix_synapse_allow_guest_access: true
+# matrix_dentrite_allow_guest_access
+```
+
+## Installing
+
+After configuring the playbook, run the [installation](installing.md) command again:
+
+```
+ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start
+```
+
+
+## Usage
+
+To learn how to use cactus comments message @cactusbot:your-homeserver.com and type `help` or have a look in the [documentation](https://cactus.chat/docs/getting-started/quick-start/).
--- a/docs/configuring-playbook-cactus-comments.md
+++ b/docs/configuring-playbook-cactus-comments.md
@@ -44,7 +44,7 @@ Now you are good to go and can include the comment section on your website!
 
 **Careful:** To really make use of self-hosting you need change a few things in comparison to the official docs!
 
-Use the following snippet and make sure to replace `example.com` with your base domain!
+Insert the following snippet into you page and make sure to replace `example.com` with your base domain!
 
 
 ```html
--- a/docs/configuring-playbook-cactus-comments.md
+++ b/docs/configuring-playbook-cactus-comments.md
@@ -38,4 +38,4 @@ ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start
 
 ## Usage
 
-To learn how to use cactus comments message @cactusbot:your-homeserver.com and type `help` or have a look in the [documentation](https://cactus.chat/docs/getting-started/quick-start/).
+To learn how to use cactus comments message @bot.cactusbot:your-homeserver.com and type `help` or have a look in the [documentation](https://cactus.chat/docs/getting-started/quick-start/).
--- a/docs/configuring-playbook-cactus-comments.md
+++ b/docs/configuring-playbook-cactus-comments.md
@@ -19,8 +19,6 @@ Add the following block to your `vars.yaml` and make sure to exchange the tokens
 #################
 
 matrix_cactus_comments_enabled: true
-matrix_cactus_comments_hs_token: wXchqGY94uhlk4OQ2ObPf0jsOMsnqXE9xuS2szB5
-matrix_cactus_comments_as_token: cvv8Y6OhAJqDbd1KCpRhc4mecZE9xuS2skBopaU5
 
 # To allow guest comments without users needing to log in, you need to have guest registration enabled.
 # To do this you need to uncomment one of the following lines (depending if you are using synapse or dentrite as a homeserver)
--- a/docs/configuring-playbook-cactus-comments.md
+++ b/docs/configuring-playbook-cactus-comments.md
@@ -38,4 +38,26 @@ ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start
 
 ## Usage
 
-To learn how to use cactus comments message @bot.cactusbot:your-homeserver.com and type `help` or have a look in the [documentation](https://cactus.chat/docs/getting-started/quick-start/).
+To get started wit cactus comments message @bot.cactusbot:your-homeserver.com and type `help` to make sure it works.
+Then register a site by typing: `register <sitename>`. You will then be invited into a moderation room.
+Now you are good to go and can include the comment section on your website!
+
+**Careful:** To really make use of self-hosting you need change a few things in comparison to the official docs!
+
+Use the following snippet and make sure to replace `example.com` with your base domain!
+
+
+```html
+<script type="text/javascript" src="https://matrix.example.com/cactus-comments/cactus.js"></script>
+<link rel="stylesheet" href="https://matrix.example.com/cactus-comments/style.css" type="text/css">
+<div id="comment-section"></div>
+<script>
+initComments({
+  node: document.getElementById("comment-section"),
+  defaultHomeserverUrl: "https://matrix.example.com:8448",
+  serverName: "example.com",
+  siteName: "YourSiteName",
+  commentSectionId: "1"
+})
+</script>
+```
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1342,12 +1342,6 @@ matrix_cactus_comments_as_token: "{{ '%s' | format(matrix_homeserver_generic_sec
 matrix_cactus_comments_hs_token: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'cactus-comments.hs.token') | to_uuid }}"
 
 matrix_cactus_comments_container_image_self_build: "{{ matrix_architecture not in ['amd64', 'arm32', 'arm64'] }}"
-
-# Normally, matrix-nginx-proxy is enabled and nginx can reach cactus-comments over the container network.
-# If matrix-nginx-proxy is not enabled, or you otherwise have a need for it, you can expose
-# cactus-comments web-server port.
-#matrix_cactus_comments_container_http_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:' + matrix_cactus_comments_container_port | string }}"
-
 matrix_cactus_comments_systemd_required_services_list: |
   {{
     (['docker.service'])
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1329,6 +1329,29 @@ matrix_backup_borg_systemd_required_services_list: |
 # /matrix-backup-borg
 #
 ######################################################################
+######################################################################
+#
+# matrix-cactus-comments
+#
+######################################################################
+
+matrix_cactus_comments_enabled: false
+
+matrix_cactus_comments_container_image_self_build: "{{ matrix_architecture not in ['amd64', 'arm32', 'arm64'] }}"
+
+matrix_corporal_systemd_required_services_list: |
+  {{
+    (['docker.service'])
+    +
+    (['matrix-' + matrix_homeserver_implementation + '.service'])
+  }}
+matrix_cactus_comments_homeserver_url: {{ matrix_homeserver_url }}
+
+######################################################################
+#
+# /matrix-cactus-comments
+#
+######################################################################
 
 ######################################################################
 #
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1337,6 +1337,10 @@ matrix_backup_borg_systemd_required_services_list: |
 
 matrix_cactus_comments_enabled: false
 
+# Derive secret values from homeserver secret
+matrix_cactus_comments_as_token: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'cactus-comments.as.token') | to_uuid }}"
+matrix_cactus_comments_hs_token: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'cactus-comments.hs.token') | to_uuid }}"
+
 matrix_cactus_comments_container_image_self_build: "{{ matrix_architecture not in ['amd64', 'arm32', 'arm64'] }}"
 
 # Normally, matrix-nginx-proxy is enabled and nginx can reach cactus-comments over the container network.
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1349,7 +1349,7 @@ matrix_synapse_app_service_config_files: |
 # cactus-comments web-server port.
 #matrix_cactus_comments_container_http_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:' + matrix_cactus_comments_container_port | string }}"
 
-matrix_corporal_systemd_required_services_list: |
+matrix_cactus_comments_systemd_required_services_list: |
   {{
     (['docker.service'])
     +
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1339,12 +1339,6 @@ matrix_cactus_comments_enabled: false
 
 matrix_cactus_comments_container_image_self_build: "{{ matrix_architecture not in ['amd64', 'arm32', 'arm64'] }}"
 
-# Doesn't work as it must be in matrix container
-#matrix_synapse_app_service_config_files: |
-#  {{
-#    ([ matrix_cactus_comments_app_service_config_file ] if matrix_cactus_comments_enabled else [])
-#  }}
-
 # Normally, matrix-nginx-proxy is enabled and nginx can reach cactus-comments over the container network.
 # If matrix-nginx-proxy is not enabled, or you otherwise have a need for it, you can expose
 # cactus-comments web-server port.
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1351,6 +1351,8 @@ matrix_cactus_comments_systemd_required_services_list: |
     (['matrix-' + matrix_homeserver_implementation + '.service'])
   }}
 
+matrix_cactus_comments_client_nginx_path: {{ '/cactus-comments/' if matrix_nginx_proxy_enabled else matrix_cactus_comments_client_path + '/' }}
+
 ######################################################################
 #
 # /matrix-cactus-comments
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -1346,6 +1346,8 @@ matrix_cactus_comments_systemd_required_services_list: |
   {{
     (['docker.service'])
     +
+    (['matrix-nginx-proxy.service'] if matrix_nginx_proxy_enabled else [])
+    +
     (['matrix-' + matrix_homeserver_implementation + '.service'])
   }}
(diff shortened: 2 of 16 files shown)


roles/matrix-cactus-comments/defaults/main.yml
```diff
@@ -0,0 +1,38 @@
+---
+# Cactus Comments is a federated comment system built on Matrix
+
+matrix_cactus_comments_enabled: true
+matrix_cactus_comments_container_image_self_build: false
+matrix_cactus_comments_docker_repo: "https://gitlab.com/cactus-comments/cactus-appservice.git"
```
carlbordum, line 6: Just FYI, this is the git repo. The Docker Hub is here: https://hub.docker.com/r/cactuscomments/cactus-appservice

Not sure what's intended.

carlbordum (commented):
Awesome! I'm not familiar with the way `matrix-docker-ansible-deploy` does stuff, but this LGTM

roles/matrix-cactus-comments/defaults/main.yml
moan0s: This is intended (I think) as this is the repo where the Dockerfile for self-building can be found AFAIK

docs/configuring-playbook-cactus-comments.md
```diff
@@ -0,0 +1,43 @@
...
+Add the following block to your `vars.yaml` and make sure to exchange the tokens to randomly generated values.
+
+```ỳaml
+#################
+## Cactus Chat ##
+#################
+
+matrix_cactus_comments_enabled: true
+matrix_cactus_comments_hs_token: wXchqGY94uhlk4OQ2ObPf0jsOMsnqXE9xuS2szB5
+matrix_cactus_comments_as_token: cvv8Y6OhAJqDbd1KCpRhc4mecZE9xuS2skBopaU5
```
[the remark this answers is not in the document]
spantaleev: We recommend something like `pwgen -s 64 1` in various places.

----

You can also consider not asking the user to specify anything and derive these tokens from `matrix_homeserver_generic_secret_key`. Here's an example:

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5f9f891322effa21d81c468cdcde701b6ba8c72c/group_vars/matrix_servers#L78-L80

Doing this, enabling Cactus Comments will be as simple as adding `matrix_cactus_comments_enabled: true` to the configuration.

roles/matrix-cactus-comments/defaults/main.yml
```diff
@@ -0,0 +1,38 @@
...
+matrix_cactus_comments_docker_repo_version: "{{ matrix_cactus_comments_version if matrix_cactus_comments_version != 'latest' else 'main' }}"
```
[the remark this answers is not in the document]
spantaleev: When self-building is enabled (it may be manually enabled or it may automatically get enabled if you're on an unsupported CPU architecture), then the image will be built from source from that git repository. When the Docker version is set to `latest`, we'll automatically build the `main` git repository branch. If it's pinned to some version (e.g. `v1.2.3`), we'll automatically pull the branch that goes by the same name (`v1.2.3`).

It seems fine.

roles/matrix-cactus-comments/defaults/main.yml
```diff
@@ -0,0 +1,38 @@
...
+matrix_cactus_comments_base_path: "{{ matrix_base_data_path }}/cactus-comments"
+matrix_cactus_comments_container_tmp_path: "{{ matrix_cactus_comments_base_path }}/tmp"
+
+matrix_cactus_comments_app_service_config_file: "{{ matrix_cactus_comments_base_path }}/cactus_appservice.yaml"
+matrix_cactus_comments_app_service_env_file: "{{ matrix_cactus_comments_base_path }}/cactus.env"
+
+matrix_cactus_comments_as_token: ''
+matrix_cactus_comments_hs_token: ''
+matrix_cactus_comments_homeserver_url: ''
+matrix_cactus_comments_user_id: "cactusbot"
```
spantaleev, line 20: We usually name bots `bot.something`. It will be more consisent if we go for that.

The documentation will also need to be adjusted.

group_vars/matrix_servers
```diff
@@ -1329,6 +1329,34 @@ matrix_backup_borg_systemd_required_services_list: |
...
+# Normally, matrix-nginx-proxy is enabled and nginx can reach cactus-comments over the container network.
+# If matrix-nginx-proxy is not enabled, or you otherwise have a need for it, you can expose
+# cactus-comments web-server port.
+#matrix_cactus_comments_container_http_host_bind_port: "{{ '' if matrix_nginx_proxy_enabled else '127.0.0.1:' + matrix_cactus_comments_container_port | string }}"
+
+matrix_cactus_comments_systemd_required_services_list: |
+  {{
+    (['docker.service'])
+    +
+    (['matrix-' + matrix_homeserver_implementation + '.service'])
```
spantaleev, line 1351:
```suggestion
    +
    (['matrix-' + matrix_homeserver_implementation + '.service'])
    +
    (['matrix-nginx-proxy.service'] if matrix_nginx_proxy_enabled else [])
```

Services usually acccess the homeserver through the nginx proxy, and not directly. Adding it as a dependency makes sense.

In this case, `matrix_cactus_comments_homeserver_url` is set to `matrix_homeserver_url` which is the public URL (`https://matrix.DOMAIN`). That may need to be redone as well using `matrix_homeserver_container_url`.

group_vars/matrix_servers
```diff
@@ -1329,6 +1329,34 @@ matrix_backup_borg_systemd_required_services_list: |
...
+  }}
+matrix_cactus_comments_homeserver_url: "{{ matrix_homeserver_url }}"
```
spantaleev, line 1353: Can this be changed to `matrix_homeserver_container_url` (which is usually someting lke `http://matrix-nginx-proxy:12080` if nginx-proxy is enabled and something like `http://matrix-synapse:8008` otherwise) or does Cactus Comments expect a public URL (e.g. `https://matrix.DOMAIN`)?

https://cactus.chat/docs/server/self-host/ seems to indicate that container URLs also work.

I don't see a separate configuration option for configuring the public homeserver URL though.. I wonder how it auto-detects it, etc., for generating the URLs for the client embed files.

roles/matrix-cactus-comments/tasks/setup_install.yml
```diff
@@ -0,0 +1,82 @@
...
+  ansible.builtin.file:
+    path: "{{ item.path }}"
+    state: directory
+    mode: 0750
+    owner: "{{ matrix_user_username }}"
+    group: "{{ matrix_user_groupname }}"
+  with_items:
+    - {path: "{{ matrix_cactus_comments_base_path }}", when: true}
+    - {path: "{{ matrix_cactus_comments_container_tmp_path }}", when: true}
+    - {path: "{{ matrix_cactus_comments_docker_src_files_path }}", when: true}
```
spantaleev, line 13:
```suggestion
    - {path: "{{ matrix_cactus_comments_docker_src_files_path }}", when: matrix_cactus_comments_container_image_self_build}
```

roles/matrix-cactus-comments/templates/cactus_appservice.yaml.j2
```diff
@@ -0,0 +1,19 @@
...
+id: "Cactus Comments"
+
+# Where the cactus-appservice is hosted:
+url: "http://matrix-cactus-comments:{{ matrix_cactus_comments_container_port }}"
+
+# Unique tokens used to authenticate requests between our service and the
+# homeserver (and the other way). Use the sha256 hashes of something random.
+# CHANGE THESE VALUES.
+as_token: "{{ matrix_cactus_comments_as_token }}"
+hs_token: "{{ matrix_cactus_comments_hs_token }}"
```
spantaleev, line 11:
```suggestion
as_token: {{ matrix_cactus_comments_as_token | to_json }}
hs_token: {{ matrix_cactus_comments_hs_token | to_json }}
```

It's safer this way

roles/matrix-cactus-comments/templates/cactus_appservice.yaml.j2
```diff
@@ -0,0 +1,19 @@
...
+
+# User associated with our service. In this case "@cactusbot:example.com"
+sender_localpart: "{{ matrix_cactus_comments_user_id }}"
```
spantaleev, line 14: The comment here is confusing. It makes it look like we full MXID is expected, even though it actually says `_localpart` and we're using a user id (localpart) as well.

Maybe it's how it is in the upstream configuration though? If so, we may leave it.

roles/matrix-cactus-comments/templates/systemd/matrix-cactus-comments.service.j2
```diff
@@ -0,0 +1,36 @@
...
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} kill matrix-cactus-comments 2>/dev/null || true'
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} rm matrix-cactus-comments 2>/dev/null || true'
+
+ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-cactus-comments \
+			--log-driver=none \
+			--cap-drop=ALL \
+			--user={{ matrix_user_uid }}:{{ matrix_user_gid }} \
+			--read-only \
+			--env-file {{ matrix_cactus_comments_app_service_env_file }} \
+            --mount type=bind,src={{ matrix_cactus_comments_container_tmp_path }},dst=/tmp \
```
spantaleev, line 25: What is this used for? Do these temporary files really need to be persisted to the filesystem or can we use a tmpfs here?

Example: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5f9f891322effa21d81c468cdcde701b6ba8c72c/roles/matrix-dendrite/templates/dendrite/systemd/matrix-dendrite.service.j2#L30

docs/configuring-playbook-cactus-comments.md
```diff
@@ -0,0 +1,43 @@
...
+After configuring the playbook, run the [installation](installing.md) command again:
+
+```
+ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start
+```
+
+
+## Usage
+
+To learn how to use cactus comments message @cactusbot:your-homeserver.com and type `help` or have a look in the [documentation](https://cactus.chat/docs/getting-started/quick-start/).
```
spantaleev, line 43: Maybe this needs to be expanded.

When self-hosting is one still supposed to be embedding the files like this:

```html
<script type="text/javascript" src="https://latest.cactus.chat/cactus.js"></script>
<link rel="stylesheet" href="https://latest.cactus.chat/style.css" type="text/css">
```

or can one serve these files from the homeserver and be completely independent from `https://latest.cactus.chat`?

Besides the dependency, serving JS/CSS files from `latest.cactus.chat` while using an older version via the playbook will likely lead to various incompatibilities.

I believe we should serve everything from our own self-hosted instance of Cactus Comments.

Perhaps you need to inject some additional nginx configuration into `matrix-nginx-proxy` to proxy these files to the Cactus Comments container (which hopefully includes and can serve these files).

You can inject additional nginx configuration into any nginx vhost. Perhaps the `matrix.DOMAIN` vhost will be best, as it's always guaranteed to be available. See an example of how this can be done here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5f9f891322effa21d81c468cdcde701b6ba8c72c/roles/matrix-bridge-mautrix-telegram/tasks/init.yml#L30-L75

spantaleev (changes requested):

roles/matrix-cactus-comments/defaults/main.yml
moan0s: True. Thought that provoke problems later on but that was a misunderstanding of mine

group_vars/matrix_servers
```diff
@@ -1329,6 +1329,38 @@ matrix_backup_borg_systemd_required_services_list: |
...
+#
+# matrix-cactus-comments
+#
+######################################################################
+
+matrix_cactus_comments_enabled: false
+
+# Derive secret values from homeserver secret
+matrix_cactus_comments_as_token: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'cactus-comments.as.token') | to_uuid }}"
+matrix_cactus_comments_hs_token: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'cactus-comments.hs.token') | to_uuid }}"
```
spantaleev, line 1342:
```suggestion
matrix_cactus_comments_as_token: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'cactus.as.token') | to_uuid }}"
matrix_cactus_comments_hs_token: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'cactus.hs.token') | to_uuid }}"
```

`cactus-comments.as.token` and `cactus-comments.hs.token` are very long.

Some versions of Ansible on some platforms have trouble if these salt-strings are longer than 16 characters. Consider shortening them.

docs/configuring-playbook-cactus-comments.md
moan0s: Very good catch. My first idea (as the the container does not serve these files) is to download the files and let nginx serve them on the homeserver domain. An alternative would be to implement serving these files in upstream. The problem with that is, that the files are located in the client repo https://gitlab.com/cactus-comments/cactus-client/-/tree/main/src whereas we would serve it from https://gitlab.com/cactus-comments/cactus-appservice
I will not do this right away but think a bit more abut it...

group_vars/matrix_servers
moan0s: Oh thanks, did not know that

roles/matrix-cactus-comments/templates/systemd/matrix-cactus-comments.service.j2
moan0s: No they do not I just didn't know better ;) Thanks!

docs/configuring-playbook-cactus-comments.md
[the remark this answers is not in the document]
moan0s: Done with d965685

group_vars/matrix_servers
moan0s: done

group_vars/matrix_servers
moan0s: Yep, tested it and works 🥳

roles/matrix-cactus-comments/templates/cactus_appservice.yaml.j2
moan0s: Comment is corrected

docs/configuring-playbook-cactus-comments.md
carlbordum: Just to clarify some stuff, so everyone's on the same page:
The Cactus appservice and client are basically completely decoupled - they only talk to Synapse, send spec'ed events and don't really know of each others existence. You can think of the client as a special-purpose Matrix client that renders a public Matrix room as a comment section, and the appservice is just there to make some stuff easy, like auto-creating rooms and replicating bans.

Downloading the files is fine to me. Alternatively, we can build a very boring nginx image in the client repo / ci pipeline upstream and use that. Whatever you prefer.

group_vars/matrix_servers
```diff
@@ -1329,6 +1329,34 @@ matrix_backup_borg_systemd_required_services_list: |
...
+matrix_cactus_comments_container_image_self_build: "{{ matrix_architecture not in ['amd64', 'arm32', 'arm64'] }}"
+matrix_cactus_comments_systemd_required_services_list: |
+  {{
+    (['docker.service'])
+    +
+    (['matrix-nginx-proxy.service'] if matrix_nginx_proxy_enabled else [])
+    +
+    (['matrix-' + matrix_homeserver_implementation + '.service'])
+  }}
+matrix_cactus_comments_homeserver_url: "{{ matrix_homeserver_container_url }}"
```
spantaleev, line 1353: This can probably be moved to `defaults/main.yml`. I see we've been using `matrix_homeserver_container_url` from `defaults/main.yml` for many many other roles already.

docs/configuring-playbook-cactus-comments.md
spantaleev: Downloading the files is okay too. However, downloading from `https://latest.cactus.chat` means we've got a moving target and each time you run the playbook you're potentially getting new files.

I was wondering if we could download via `git` from some branch/tag.

Having a separate Docker image that serves the client files could also be nice. We can have pinned versions there, etc., and have the nginx configuration just proxy there. However, it's probably more difficult for you to create and publish such an image. Will it be multi-arch is another question.

carlbordum: Ah, right. Should probably download from [the releases page](https://gitlab.com/cactus-comments/cactus-client/-/releases) then :)

moan0s: I now use the release or a local file if provided

roles/matrix-cactus-comments/defaults/main.yml
```diff
@@ -0,0 +1,47 @@
...
+matrix_cactus_comments_tmp_directory_size_mb: 1
+
+matrix_cactus_comments_container_port: 5000
+
+matrix_cactus_comments_version: latest
+matrix_cactus_comments_docker_image: "{{ matrix_container_global_registry_prefix }}cactuscomments/cactus-appservice:{{ matrix_cactus_comments_version }}"
+matrix_cactus_comments_docker_image_force_pull: "{{ matrix_cactus_comments_docker_image.endswith(':latest') }}"
+
+matrix_cactus_comments_client_version: "0.13.0"
+matrix_cactus_comments_client_local_dir: ''
```
spantaleev, line 33:
```suggestion
# matrix_cactus_comments_client_version specifies the version of the cactus-client release to use.
# For available versions, see: https://gitlab.com/cactus-comments/cactus-client/-/releases
# Also see: `matrix_cactus_comments_client_local_dir`
matrix_cactus_comments_client_version: "0.13.0"

# matrix_cactus_comments_client_local_dir specifies a local directory (on the Ansible controller, not on the remote server) with cactus-client files to use.
# This is an alternative to `matrix_cactus_comments_client_version`, to be used when you'd like to
# provide the files locally / manually.
matrix_cactus_comments_client_local_dir: ''
```

roles/matrix-cactus-comments/tasks/init.yml
```diff
@@ -0,0 +1,67 @@
...
+          {{
+                    matrix_nginx_proxy_container_additional_volumes | default([])
+            +
+            [{"src": "{{ matrix_cactus_comments_client_path }}", "dst": "/cactus-comments/cactus-comments", "options": "ro"}]
+          }}
+    - name: Generate Cactus Comment proxying configuration for matrix-nginx-proxy
+      ansible.builtin.set_fact:
+        matrix_cactus_comments_nginx_proxy_configuration: |
+          location {{ matrix_cactus_comments_client_endpoint }} {
+              root /cactus-comments/;
```
spantaleev, line 46: I wonder if this path should be made configurable, especially for people who disable `matrix-nginx-proxy`.

You can put it in a variable:

```yaml
# matrix_cactus_comments_client_nginx_path specifies the path where nginx can access the client files.
# The default value assumes a container setup. If you're running nginx without a container, consider adjusting this path
matrix_cactus_comments_client_nginx_path: /cactus-comments/
```

And then in `group_vars/matrix_servers`:

```yaml
matrix_cactus_comments_client_nginx_path: {{ '/cactus-comments/' if matrix_nginx_proxy_enabled else matrix_cactus_comments_client_path + '/' }}
```

roles/matrix-cactus-comments/templates/systemd/matrix-cactus-comments.service.j2
```diff
@@ -0,0 +1,36 @@
...
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} kill matrix-cactus-comments 2>/dev/null || true'
+ExecStartPre=-{{ matrix_host_command_sh }} -c '{{ matrix_host_command_docker }} rm matrix-cactus-comments 2>/dev/null || true'
+
+ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-cactus-comments \
+			--log-driver=none \
+			--cap-drop=ALL \
+			--user={{ matrix_user_uid }}:{{ matrix_user_gid }} \
+			--read-only \
+			--env-file {{ matrix_cactus_comments_app_service_env_file }} \
+             --tmpfs=/tmp:rw,noexec,nosuid,size={{ matrix_cactus_comments_tmp_directory_size_mb }}m \
```
spantaleev, line 25:
```suggestion
			--env-file {{ matrix_cactus_comments_app_service_env_file }} \
			--tmpfs=/tmp:rw,noexec,nosuid,size={{ matrix_cactus_comments_tmp_directory_size_mb }}m \
```

roles/matrix-cactus-comments/tasks/init.yml
moan0s: True. Done

spantaleev (approved):

spantaleev:
Would you fix the CI errors? I suppose we can get this merged after that :)

moan0s:
Thank you again for all your helpful feedback and the work you do 🤗

spantaleev:
Thank you too, for your patience and for adding new and awesome services to the playbook! :bow:

spantaleev:
Does one need to create the `bot.cactusbot` user manually or will the appservice create it automatically when it starts?

Whatever the answer, perhaps we'd better indicate that on the docs page. Most services require you to create the bot manually, so people may be surprised if this one does it automatically.

spantaleev:
I've pinned the Cactus Comments version to `0.9.0`. We try to not use `latest` unless we have to.

carlbordum:
> I've pinned the Cactus Comments version to 0.9.0. We try to not use latest unless we have to.

Note that the client and appservice are versioned independently.

> Does one need to create the bot.cactusbot user manually or will the appservice create it automatically when it starts?
> 
> Whatever the answer, perhaps we'd better indicate that on the docs page. Most services require you to create the bot manually, so people may be surprised if this one does it automatically.

The appservice creates the user automatically.

spantaleev:
> Note that the client and appservice are versioned independently.

Sure! We pin the client to `0.13.0` via a separate variable. 

> The appservice creates the user automatically.

That's great! I've mentioned it in b510848c

PR #2092, opened 2022-09-09
Update Postmoogle 0.9.0 -> 0.9.1

etkecc:


docs/configuring-dns.md
```diff
@@ -77,3 +81,8 @@ This is an optional feature for the optionally-installed [ma1sd service](configu
 Note: This `_matrix-identity._tcp` SRV record for the identity server is different from the `_matrix._tcp` that can be used for Synapse delegation. See [howto-server-delegation.md](howto-server-delegation.md) for more information about delegation.
 
 When you're done with the DNS configuration and ready to proceed, continue with [Getting the playbook](getting-the-playbook.md).
+
+## `_dmarc`, `postmoogle._domainkey` TXT and `matrix` MX records setup
+
+To make the [postmoogle](https://gitlab.com/etke.cc/postmoogle) email bridge enable its email sending features, you need to configure
```
spantaleev, line 87: We should probably link to our own docs (`docs/configuring-playbook-bot-postmoogle.md`) as well here.

docs/configuring-dns.md
```diff
@@ -28,18 +28,22 @@ If you are using Cloudflare DNS, make sure to disable the proxy and set all reco
...
+| CNAME | `goneb`                        | -        | -      | -    | `matrix.<your-domain>`      |
+| CNAME | `sygnal`                       | -        | -      | -    | `matrix.<your-domain>`      |
+| CNAME | `ntfy`                         | -        | -      | -    | `matrix.<your-domain>`      |
+| CNAME | `hydrogen`                     | -        | -      | -    | `matrix.<your-domain>`      |
+| CNAME | `cinny`                        | -        | -      | -    | `matrix.<your-domain>`      |
+| CNAME | `buscarron`                    | -        | -      | -    | `matrix.<your-domain>`      |
+| MX    | `matrix`                       | 10       | 0      | -    | `matrix.<your-domain>`      |
+| TXT   | `matrix`                       | -        | -      | -    | `v=spf1 ip4:<your-ip> -all` |
+| TXT   | `_dmarc.matrix`                | -        | -      | -    | `v=DMARC1; p=quarantine;`   |
+| TXT   | `postmoogle._domainkey.matrix` | -        | -      | -    | get it from `!pm dkim`      |
```
spantaleev, line 46: I wonder if we should add some `Matrix Component` column at the end, which specifies which component requires this service.

On each row, we can link to specific component's `docs/` page.

group_vars/matrix_servers
```diff
@@ -1791,6 +1791,8 @@ matrix_ssl_domains_to_obtain_certificates_for: |
     +
     ([matrix_server_fqn_ntfy] if matrix_ntfy_enabled else [])
     +
+    ([matrix_bot_postmoogle_domain] if matrix_bot_postmoogle_enabled else [])
```
spantaleev, line 1794:
```suggestion
    ([matrix_bot_postmoogle_domain] if matrix_bot_postmoogle_enabled and matrix_bot_postmoogle_domain != matrix_server_fqn_matrix else [])
```

---

It might be a good idea to add ` | uniq` (or is it ` | unique` or something else?) somewhere as well. It could be in the actual `with_items` iteration in `roles/matrix-nginx-proxy`. This way, even if someone unintentionally duplicates domains in `matrix_ssl_domains_to_obtain_certificates_for` (or `matrix_ssl_additional_domains_to_obtain_certificates_for` used within it), we won't be doing double work.

roles/matrix-bot-postmoogle/defaults/main.yml
```diff
@@ -110,11 +111,20 @@ matrix_bot_postmoogle_noencryption: false
...
+matrix_bot_postmoogle_tls_required: false
+
+# in-container ports
 matrix_bot_postmoogle_port: '2525'
+matrix_bot_postmoogle_tls_port: '25587'
 
-# on-host port
+# on-host ports
 matrix_bot_postmoogle_smtp_host_bind_port: '25'
+matrix_bot_postmoogle_smtps_host_bind_port: '587'
```
spantaleev, line 123:
```suggestion
matrix_bot_postmoogle_submission_host_bind_port: '587'
```

- https://en.wikipedia.org/wiki/SMTPS#Difference_between_SMTPS_and_smtps
- https://en.wikipedia.org/wiki/Message_submission_agent

`smtps` kind of hints towards port `465`.

roles/matrix-bot-postmoogle/defaults/main.yml
```diff
@@ -9,14 +9,15 @@ matrix_bot_postmoogle_docker_repo: "https://gitlab.com/etke.cc/postmoogle.git"
...
-matrix_bot_postmoogle_version: v0.9.0
+matrix_bot_postmoogle_version: v0.9.1
 matrix_bot_postmoogle_docker_image: "{{ matrix_bot_postmoogle_docker_image_name_prefix }}postmoogle:{{ matrix_bot_postmoogle_version }}"
 matrix_bot_postmoogle_docker_image_name_prefix: "{{ 'localhost/' if matrix_bot_postmoogle_container_image_self_build else 'registry.gitlab.com/etke.cc/' }}"
 matrix_bot_postmoogle_docker_image_force_pull: "{{ matrix_bot_postmoogle_docker_image.endswith(':latest') }}"
 
 matrix_bot_postmoogle_base_path: "{{ matrix_base_data_path }}/postmoogle"
 matrix_bot_postmoogle_config_path: "{{ matrix_bot_postmoogle_base_path }}/config"
 matrix_bot_postmoogle_data_path: "{{ matrix_bot_postmoogle_base_path }}/data"
+matrix_bot_postmoogle_ssl_path: "{{ matrix_ssl_config_dir_path }}"
```
spantaleev, line 20: `matrix_ssl_config_dir_path` is part of `matrix-nginx-proxy` and so far it has only been used in that role.

It would probably be better if we set this to empty, and override it from `group_vars/matrix_servers`:

```yaml
matrix_bot_postmoogle_ssl_path: "{{ matrix_ssl_config_dir_path }}"
matrix_bot_postmoogle_tls_cert: "/ssl/live/{{ matrix_bot_postmoogle_domain }}/fullchain.pem"
matrix_bot_postmoogle_tls_key: "/ssl/live/{{ matrix_bot_postmoogle_domain }}/privkey.pem"
```

roles/matrix-bot-postmoogle/defaults/main.yml
```diff
@@ -110,11 +111,20 @@ matrix_bot_postmoogle_noencryption: false
...
+matrix_bot_postmoogle_tls_port: '25587'
 
-# on-host port
+# on-host ports
 matrix_bot_postmoogle_smtp_host_bind_port: '25'
+matrix_bot_postmoogle_smtps_host_bind_port: '587'
+
+# in-container SSL paths
+matrix_bot_postmoogle_tls_cert: "/ssl/live/{{ matrix_bot_postmoogle_domain }}/fullchain.pem"
+matrix_bot_postmoogle_tls_key: "/ssl/live/{{ matrix_bot_postmoogle_domain }}/privkey.pem"
```
spantaleev, line 127: These can probably be left empty as well. If `matrix_bot_postmoogle_ssl_path` is not set (the default in `defaults/main.yml`, as suggested by my other comment), then it won't be mounted and certificate files won't be available.

Adding some comments with example values might be helpful.

```suggestion
# matrix_bot_postmoogle_tls_cert is the SSL certificate's certificate.
# This is likely set via group_vars/matrix_servers, so you don't need to set it.
# If you do need to set it manually, note that this is an in-container path. To mount a certificates volumes into the container, use matrix_bot_postmoogle_ssl_path
# Example value: /ssl/live/{{ matrix_bot_postmoogle_domain }}/fullchain.pem
matrix_bot_postmoogle_tls_cert: "/ssl/live/{{ matrix_bot_postmoogle_domain }}/fullchain.pem"

# matrix_bot_postmoogle_tls_key is the SSL certificate's key.
# This is likely set via group_vars/matrix_servers, so you don't need to set it.
# If you do need to set it manually, note that this is an in-container path. To mount a certificates volumes into the container, use matrix_bot_postmoogle_ssl_path
# Example value: /ssl/live/{{ matrix_bot_postmoogle_domain }}/privkey.pem
matrix_bot_postmoogle_tls_key: "/ssl/live/{{ matrix_bot_postmoogle_domain }}/privkey.pem"
```

roles/matrix-bot-postmoogle/templates/systemd/matrix-bot-postmoogle.service.j2
```diff
@@ -24,7 +24,9 @@ ExecStart={{ matrix_host_command_docker }} run --rm --name matrix-bot-postmoogle
 			--network={{ matrix_docker_network }} \
 			--env-file={{ matrix_bot_postmoogle_config_path }}/env \
 			-p {{ matrix_bot_postmoogle_smtp_host_bind_port }}:{{ matrix_bot_postmoogle_port }} \
+			-p {{ matrix_bot_postmoogle_smtps_host_bind_port }}:{{ matrix_bot_postmoogle_tls_port }} \
 			--mount type=bind,src={{ matrix_bot_postmoogle_data_path }},dst=/data \
+			--mount type=bind,src={{ matrix_bot_postmoogle_ssl_path }},dst=/ssl \
```
spantaleev, line 29:
```suggestion
			{% if matrix_bot_postmoogle_ssl_path %}
			--mount type=bind,src={{ matrix_bot_postmoogle_ssl_path }},dst=/ssl \
			{% endif %}
```

If `matrix_bot_postmoogle_ssl_path` defaults to an empty value (as suggested by my other review comment), we need to ensure we don't try to mount it.

docs/configuring-dns.md
etkecc: Tried to add it, looks **very** bad

group_vars/matrix_servers
etkecc: Good point about `unique` filter, but I doubt the `and` condition needed in the group_vars - it's already hard to read, so let's just leave one `if` condition and apply `unique` filter when needed

spantaleev: Each thing is on a line of its own, so it's not like it's becoming very complicated. Given that we have ` | unique` now, I suppose we're good and can keep this simple.

docs/configuring-dns.md
spantaleev: Let's leave it as it is then

PR #2109, opened 2022-09-17
change stream writer worker metrics range start

borisrunakov:
`one-of-each` setting for workers,  recently broke with error messages from `matrix-synapse-worker-stream-writer-0-events.service` with  `port 19111 already in use` . 
Changing `matrix_synapse_workers_stream_writer_workers_metrics_range_start` from  `19111` to `19211` seem to fixed for me the issue


spantaleev:
Thanks! What did it conflict with though?

(on mobile, so can't easily grep myself)

borisrunakov:
Also on mobile but I think the commit description, mentions the conflict.

spantaleev:
Stream writers' metrics ports were conflicting with generic workers' metrics ports (both starting from `19111` until now).

Issue #2119, opened 2022-09-20
Use `matrix_coturn_turn_static_auth_secret` for external turn server


Samonitari:
**Describe the bug**
The supplied turn shared secret is not used for synapse (or denrite) config, if external coturn server is used instead of the playbook's one.

**To Reproduce**
My `vars.yml` file looks like this:

```yaml
matrix_coturn_turn_static_auth_secret: "my_shared_secret"
matrix_coturn_enabled: false
matrix_synapse_turn_uris:
- turn:my.turn.server?transport=tcp
- turn:my.turn.server?transport=udp
```

**Expected behavior**
The `matrix_coturn_turn_static_auth_secret` variable gets inserted into homeserver.yml as `turn_shared_secret`

spantaleev:
This is by design.

If you use an external Coturn server, we don't automatically wire stuff. It's on you to set the appropriate Synapse (or Dendrite) variables so that your homeserver points to where you'd like.

Setting any `matrix_coturn_*` variables when `matrix_coturn_enabled: false` is set has no effect.

Samonitari:
Hmm, would you still look at #2121?
Seems relatively harmless, backward compatible to me

PR #2121, opened 2022-09-20
Insert turn shared secret to server if turn URIs is not empty

Samonitari:
Possible fix for #2119

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -2340,7 +2340,7 @@ matrix_synapse_turn_uris: |
     ] if matrix_coturn_enabled else []
   }}
 
-matrix_synapse_turn_shared_secret: "{{ matrix_coturn_turn_static_auth_secret if matrix_coturn_enabled else '' }}"
+matrix_synapse_turn_shared_secret: "{{ matrix_coturn_turn_static_auth_secret if (matrix_synapse_turn_uris | length > 0) else '' }}"
 
 matrix_synapse_self_check_validate_certificates: "{{ false if matrix_ssl_retrieval_method == 'self-signed' else true }}"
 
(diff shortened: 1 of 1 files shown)


spantaleev:
While this may work for you.. it doesn't make much sense.

What you're saying is "if TURN URIs are specified in the Synapse role, use the secret from the Coturn role".

What it used to say is "if the Coturn role is enabled, use the secrets from that role".

----

If you're providing the TURN URIs for Synapse manually in https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/2119, why not provide it with a secret too, instead of passing the secret to the *disabled* Coturn role and expecting that it will be picked up from there. You can use `matrix_synapse_turn_shared_secret` and have your configuration be like this:

```yaml
# We use our own Coturn, so we disable the Coturn role and want nothing to do with it.
matrix_coturn_enabled: false

# Below is our custom TURN server configuration

matrix_synapse_turn_uris:
- turn:my.turn.server?transport=tcp
- turn:my.turn.server?transport=udp

matrix_synapse_turn_shared_secret: "my_shared_secret"
```

In the rest of the playbook, when a service is *disabled*, we strive to completely eliminate it from influencing the other services. If you do `matrix_mailer_enabled: false`, none of the other `matrix_mailer_*` variables matter anymore. The mailer is simply not used.

`matrix-nginx-proxy` may be the only exception. When disabled (`matrix_proxy_enabled: false`), we still generate nginx configuration in `/matrix/nginx-proxy/conf.d`, just because a lot of people want to use their own nginx and it's convenient to generate nginx configuration for them.

Samonitari:
Thanks for the comment, I missed the bigger picture a bit!

PR #2130, opened 2022-09-24
whatsapp: do not turn on synapse only options

shalzz:
Dendrite does not support MSC2409 as of now, so read receipts fail with `ephemeral_events` turned on by default on dendrite.


spantaleev:
Looks like I had made this regress by syncing our default configuration with upstream. Thanks for fixing it on Dendrite!

PR #2133, opened 2022-09-25
prometheus: don't just default to synapse

shalzz:
I don't think dendrite is completely hookedup to export data just yet but at least now the playbook shouldn't just listen to synapse by default.

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -2310,8 +2310,7 @@ matrix_synapse_tls_private_key_path: ~
 
 matrix_synapse_federation_port_openid_resource_required: "{{ not matrix_synapse_federation_enabled and (matrix_dimension_enabled or matrix_ma1sd_enabled) }}"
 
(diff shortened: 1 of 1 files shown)


spantaleev:
I think the reason you'd like to change `matrix_synapse_metrics_enabled: "{{ matrix_prometheus_enabled }}"` to `"{{ matrix_synapse_enabled and matrix_prometheus_enabled }}"` is because of how `matrix_synapse_metrics_enabled` is used.

One would say that "if Synapse is not in use, it doesn't matter what `matrix_synapse_metrics_enabled` is set to, since it won't be used either".

However, `matrix_synapse_metrics_enabled` is used outside of the Synapse role too.. which is the problematic part. I believe it should be addressed where it's used, and not by redefining `matrix_synapse_metrics_enabled`.

I believe I've improved the situation in 6c928d87ca3bd2af4 and f623cf355065f3, so this PR should no longer be necessary.

shalzz:
Okay, sounds good. Closing this PR.

PR #2135, opened 2022-09-27
Exempt Matrix server from ntfy rate limit

karmanyaahm:
Add the Matrix server to ntfy's exemption list. Also allow all ntfy rate limits to be configured through Ansible variables.

Reasoning: we (UnifiedPush) have seen some support requests for people not receiving notifications due to rate limits. At some point, someone using this playbook for a large-ish ntfy deployment will also run into that. Adding this as a precaution should mitigate that.

I will test this on my server soon, and then make this ready for review.

### DIFF
--- a/roles/custom/matrix-ntfy/defaults/main.yml
+++ b/roles/custom/matrix-ntfy/defaults/main.yml
@@ -15,11 +15,13 @@ matrix_ntfy_docker_image_force_pull: "{{ matrix_ntfy_docker_image.endswith(':lat
 matrix_ntfy_base_url: "https://{{ matrix_server_fqn_ntfy }}"
 
 # Rate limits
-matrix_ntfy_limit_burst_rate: 60 # default
-matrix_ntfy_limit_replenish_rate: "5s" # default
-matrix_ntfy_limit_rate_exempt: "{{matrix_server_fqn_matrix}},localhost" # exempt our matrix server from rate limits, this may not work when the homeserver's outgoing IP is different from the incoming IP, but most small deployments should be fine.
-matrix_ntfy_limit_global_topic: 15000 # default
-matrix_ntfy_limit_visitor_subscriptions: 30 # default
+
+matrix_ntfy_global_topic_limit: 15000 # default
+matrix_ntfy_visitor_subscription_limit: 30 # default
+matrix_ntfy_visitor_request_limit_burst: 60 # default
+matrix_ntfy_visitor_request_limit_replenish: "5s" # default
+matrix_ntfy_visitor_request_limit_exempt_hosts: "{{matrix_server_fqn_matrix}},localhost" # exempt our matrix server from rate limits, this may not work when the homeserver's outgoing IP is different from the incoming IP, but most small deployments should be fine.
+
 
 # Controls whether the container exposes its HTTP port (tcp/80 in the container).
 #
--- a/roles/custom/matrix-ntfy/defaults/main.yml
+++ b/roles/custom/matrix-ntfy/defaults/main.yml
@@ -20,7 +20,6 @@ matrix_ntfy_global_topic_limit: 15000 # default
 matrix_ntfy_visitor_subscription_limit: 30 # default
 matrix_ntfy_visitor_request_limit_burst: 60 # default
 matrix_ntfy_visitor_request_limit_replenish: "5s" # default
-matrix_ntfy_visitor_request_limit_exempt_hosts: "{{matrix_server_fqn_matrix}},localhost" # exempt our matrix server from rate limits, this may not work when the homeserver's outgoing IP is different from the incoming IP, but most small deployments should be fine.
 
 
 # Controls whether the container exposes its HTTP port (tcp/80 in the container).
--- a/roles/custom/matrix-ntfy/defaults/main.yml
+++ b/roles/custom/matrix-ntfy/defaults/main.yml
@@ -14,6 +14,13 @@ matrix_ntfy_docker_image_force_pull: "{{ matrix_ntfy_docker_image.endswith(':lat
 # Public facing base URL of the ntfy service
 matrix_ntfy_base_url: "https://{{ matrix_server_fqn_ntfy }}"
 
+# Rate limits
+matrix_ntfy_limit_burst_rate: 60 # default
+matrix_ntfy_limit_replenish_rate: "5s" # default
+matrix_ntfy_limit_rate_exempt: "{{matrix_server_fqn_matrix}},localhost" # exempt our matrix server from rate limits, this may not work when the homeserver's outgoing IP is different from the incoming IP, but most small deployments should be fine.
+matrix_ntfy_limit_global_topic: 15000 # default
+matrix_ntfy_limit_visitor_subscriptions: 30 # default
+
 # Controls whether the container exposes its HTTP port (tcp/80 in the container).
 #
 # Takes an "<ip>:<port>" or "<port>" value (e.g. "127.0.0.1:2586"), or empty string to not expose.
--- a/roles/custom/matrix-ntfy/defaults/main.yml
+++ b/roles/custom/matrix-ntfy/defaults/main.yml
@@ -16,10 +16,10 @@ matrix_ntfy_base_url: "https://{{ matrix_server_fqn_ntfy }}"
 
 # Rate limits
 
-matrix_ntfy_global_topic_limit: 15000 # default
-matrix_ntfy_visitor_subscription_limit: 30 # default
-matrix_ntfy_visitor_request_limit_burst: 60 # default
-matrix_ntfy_visitor_request_limit_replenish: "5s" # default
+matrix_ntfy_global_topic_limit: 15000  # default
+matrix_ntfy_visitor_subscription_limit: 30  # default
+matrix_ntfy_visitor_request_limit_burst: 60  # default
+matrix_ntfy_visitor_request_limit_replenish: "5s"  # default
 
 
 # Controls whether the container exposes its HTTP port (tcp/80 in the container).
--- a/roles/custom/matrix-ntfy/templates/ntfy/server.yml.j2
+++ b/roles/custom/matrix-ntfy/templates/ntfy/server.yml.j2
@@ -8,5 +8,4 @@ global-topic-limit: {{ matrix_ntfy_global_topic_limit | to_json }}
 visitor-subscription-limit: {{ matrix_ntfy_visitor_subscription_limit | to_json }}
 
 visitor-request-limit-burst: {{ matrix_ntfy_visitor_request_limit_burst | to_json }}
-visitor-request-limit-replenish: "{{ matrix_ntfy_visitor_request_limit_replenish | to_json }}"
-visitor-request-limit-exempt-hosts: "{{ matrix_ntfy_visitor_request_limit_exempt_hosts | to_json }}"
+visitor-request-limit-replenish: "{{ matrix_ntfy_visitor_request_limit_replenish }}"
--- a/roles/custom/matrix-ntfy/templates/ntfy/server.yml.j2
+++ b/roles/custom/matrix-ntfy/templates/ntfy/server.yml.j2
@@ -4,9 +4,9 @@ cache_file: /data/cache.db
 listen-http: :8080
 
 # Rate Limits
-global-topic-limit: {{ matrix_ntfy_limit_global_topic }}
-visitor-subscription-limit: {{ matrix_ntfy_limit_visitor_subscriptions }}
+global-topic-limit: {{ matrix_ntfy_global_topic_limit | to_json }}
+visitor-subscription-limit: {{ matrix_ntfy_visitor_subscription_limit | to_json }}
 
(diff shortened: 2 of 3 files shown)


roles/matrix-ntfy/defaults/main.yml
```diff
@@ -14,6 +14,13 @@ matrix_ntfy_docker_image_force_pull: "{{ matrix_ntfy_docker_image.endswith(':lat
 # Public facing base URL of the ntfy service
 matrix_ntfy_base_url: "https://{{ matrix_server_fqn_ntfy }}"
 
+# Rate limits
+matrix_ntfy_limit_burst_rate: 60 # default
+matrix_ntfy_limit_replenish_rate: "5s" # default
+matrix_ntfy_limit_rate_exempt: "{{matrix_server_fqn_matrix}},localhost" # exempt our matrix server from rate limits, this may not work when the homeserver's outgoing IP is different from the incoming IP, but most small deployments should be fine.
```
spantaleev, line 20: How does the homeserver currently hit Ntfy? Via `matrix_ntfy_base_url`? Via container URLs like `http://matrix-ntfy:PORT`?

As per our `examples/hosts`, we ask people to use an external IP address for `ansible_host`: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/d159408feb0888dce1110d65955e1e4916c44739/examples/hosts#L1-L2

We pass this to `matrix_coturn_turn_external_ip_address` here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/d159408feb0888dce1110d65955e1e4916c44739/group_vars/matrix_servers#L1413

You can do something similar:

- introduce a new `matrix_ntfy_external_ip_address` variable in this `defaults/main.yml` file, set to empty
- add a `matrix_ntfy_external_ip_address: "{{ ansible_host }}"` definition to the `matrix-ntfy` section in `group_vars/matrix_servers`
- use `matrix_ntfy_external_ip_address` (if not empty) in `matrix_ntfy_limit_rate_exempt`

Not sure if this is necessary or helpful though. `matrix_ntfy_limit_rate_exempt` seems to use hostnames, not IP addresses. It likely resolves these domain names correctly, so `matrix_server_fqn_matrix` would end up reliably pointing to the external IP address of the deployment. There's probably no need to improve on that.

roles/matrix-ntfy/templates/ntfy/server.yml.j2
```diff
@@ -1,3 +1,11 @@
...
 behind_proxy: true
 cache_file: /data/cache.db
+
+# Rate Limits
+global-topic-limit: {{ matrix_ntfy_limit_global_topic }}
+visitor-subscription-limit: {{ matrix_ntfy_limit_visitor_subscriptions }}
+
+visitor-request-limit-burst: {{ matrix_ntfy_limit_burst_rate }}
+visitor-request-limit-replenish: "{{ matrix_ntfy_limit_replenish_rate }}"
+visitor-request-limit-exempt-hosts: "{{ matrix_ntfy_limit_rate_exempt }}"
```
spantaleev, line 11:
```suggestion
global-topic-limit: {{ matrix_ntfy_limit_global_topic | to_json }}
visitor-subscription-limit: {{ matrix_ntfy_limit_visitor_subscriptions | to_json }}

visitor-request-limit-burst: {{ matrix_ntfy_limit_burst_rate | to_json }}
visitor-request-limit-replenish: "{{ matrix_ntfy_limit_replenish_rate | to_json }}"
visitor-request-limit-exempt-hosts: "{{ matrix_ntfy_limit_rate_exempt | to_json }}"
```

It's always more reliable to use ` | to_json` in YAML/JSON config files.

-----

However, another thing that stands out is: variables are named differently than what they control. We usually strive for a match here, as it's more straightforward. A setting called `global-topic-limit` should be controlled via a variable called `matrix_ntfy_global_topic_limit`. Renaming all these variables will make this more consistent and straightforward.

karmanyaahm: will do

roles/matrix-ntfy/defaults/main.yml
karmanyaahm: > How does the homeserver currently hit Ntfy? Via `matrix_ntfy_base_url`? Via container URLs like `http://matrix-ntfy:PORT`?

The homeserver reaches ntfy the same way that a phone reaches ntfy, via the public host ntfy.BASE_DOMAIN. It can't tell the difference between a ntfy instance hosted by this playbook or one on another server.

karmanyaahm: > Not sure if this is necessary or helpful though. `matrix_ntfy_limit_rate_exempt` seems to use hostnames, not IP addresses. It likely resolves these domain names correctly, so `matrix_server_fqn_matrix` would end up reliably pointing to the external IP address of the deployment. There's probably no need to improve on that.

It can use both hostnames and IP addresses. I am setting up a temporary server for testing rn, and will see if it properly detects it.

karmanyaahm: Also, clarification: I added localhost here just in case some weird thing happens with IP detection since both services are running on the same machine. It is unlikely, since nginx would get in the way, but there's no harm in adding it.

spantaleev: Sure! No harm in keeping it.

roles/matrix-ntfy/templates/ntfy/server.yml.j2
karmanyaahm: For some reason, when I `to_json` the last two, they break with
```
fatal: [matrix.matrix753.malhotra.cc]: FAILED! => 
  msg: |-
    An unhandled exception occurred while templating '{{ matrix_ntfy_configuration_yaml | from_yaml | combine(matrix_ntfy_configuration_extension, recursive=True) }}'. Error was a <class 'yaml.parser.ParserError'>, original message: while parsing a block mapping
      in "<unicode string>", line 1, column 1
    did not find expected key
      in "<unicode string>", line 11, column 39
```

karmanyaahm:
Is there any way to get the docker network bridge (?) IP address `172.18.0.1` as a variable? For whatever weird reason, when I send a request from the synapse container to https://ntfy.mydomain, that's the source IP in the x-forwarded header.

roles/matrix-ntfy/templates/ntfy/server.yml.j2
spantaleev: Oh, sorry.. When you use `| to_json`, you shouldn't add double-quotes around it:

```diff
-"{{ something | to _json }}"
+{{ something | to_json }}
```

My suggested code above is wrong.

spantaleev:
You can get the whole `matrix` network's CIDR by doing `docker network inspect matrix -f '{{ (index .IPAM.Config 0).Subnet }}'`.

However, that's probably not of too much help. We could put this in an environment variable from an `ExecStart` call in the `matrix-ntfy.service` systemd service.. And then reference that environment variable in the config, although.. I doubt ntfy supports that.

Another alternative is to use Ansible's [`command` module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/command_module.html) in `roles/matrix-ntfy/tasks/setup_install.yml`, to run the above command and to capture its output into some `matrix_ntfy_matrix_docker_network_cidr` variable and then use that in `roles/matrix-ntfy/templates/ntfy/server.yml.j2`.

We may also modify `roles/matrix-base/tasks/setup_matrix_base.yml` to `register` the result of creating the network (that's where creating the `matrix` network happens) and use that.. However tasks in `roles/matrix-base/tasks/setup_matrix_base.yml` are only executed for `--tags=setup-all` and not for other playbook invocations (`--tags=setup-ntfy` will not run those tasks and we'll be left without this result).

There may even be a hybrid approach - use `register` in `roles/matrix-base/tasks/setup_matrix_base.yml` (e.g. `register: matrix_docker_network_result`) and later (in `roles/matrix-ntfy/tasks/setup_install.yml`) use `matrix_docker_network_result` if available or do the `command` thing otherwise. This potentially saves us one `command` call in some cases (`--tags=setup-all`), but complicates things - we now have 2 separate code paths and variables that may or may not be used across roles. We'd better not go with a hybrid approach.

Using the whole network's CIDR is better than using the IP of the `matrix-synapse` container, because:

- if Synapse uses workers, it may not be the main process that hits Ntfy, but some `pusher` worker process
- if some other homeserver is used (Dendrite, Conduit), it may be the one hitting us, not Synapse

Before you dive into all of this though... **Will a network CIDR work in `roles/matrix-ntfy/templates/ntfy/server.yml.j2`**? You can try manually adding it to confirm if it's helpful.

karmanyaahm:
> However, that's probably not of too much help. We could put this in an environment variable from an ExecStart call in the matrix-ntfy.service systemd service.. And then reference that environment variable in the config, although.. I doubt ntfy supports that.

The `NTFY_VISITOR_REQUEST_LIMIT_EXEMPT_HOSTS` env var overrides the config file option.

> Will a network CIDR work in roles/matrix-ntfy/templates/ntfy/server.yml.j2?

It does not, but it should be simple, I can patch ntfy to make it support it. I can ask binwiederhier if he will accept a pull request for that.

Then, it should be possible to set `NTFY_VISITOR_REQUEST_LIMIT_EXEMPT_HOSTS=docker network inspect matrix -f '{{ (index .IPAM.Config 0).Subnet }}'` in systemd somewhere.

spantaleev:
> The `NTFY_VISITOR_REQUEST_LIMIT_EXEMPT_HOSTS` env var overrides the config file option.

Ah.. then we'd probably need to build that whole env var during runtime (in `ExecStart` or something) -- so that it contains both the `matrix` network's CIDR and possibly a few other things like `matrix.DOMAIN` and `localhost`, etc.

Making ntfy accept whole networks in the exempt hosts would be nice, I suppose

karmanyaahm:
Since this is merged into ntfy, I'll now work on setting it as an environment variable through systemd.

karmanyaahm:
I fixed the Systemd service, and this is working on my end. Just waiting on a ntfy release.

karmanyaahm:
Tested with ntfy v1.29.0 on my server. Everything is working properly.

Script: `docker exec -it matrix-synapse bash -c "for i in {1..1000}; do curl -X POST https://ntfy.mypersonaldomain.com/up234 -d hi; done"`

karmanyaahm:
> The `docker network inspect` line being added inline there is quite ugly too - I woner if we can make it better by using some `Environment=..` line that initializes the value by running `docker inspect ...`, and then us passing this environment variable to `--env`.

Environment doesn't support shell substitutions, the next best thing would be an ExecStartPre setting an environment variable, but that's global which is more of a mess [1].

I'll fix the lint errors.

Sorry, it took me a while to get to this.
[1] https://superuser.com/a/1379686

spantaleev:
> Environment doesn't support shell substitutions, the next best thing would be an ExecStartPre setting an environment variable, but that's global which is more of a mess [1].

Why do you need shell substitutions though?


I believe something like this could work:

```jinja
Environment="MATRIX_CONTAINER_NETWORK_SUBNET=$({{ devture_systemd_docker_base_host_command_docker }} inspect --format='{% raw %}{{ .(index .IPAM.Config 0).Subnet }}{% endraw %}' {{ matrix_docker_network | quote }})"
```

.. and then you can pass it like this:

```
--env NTFY_VISITOR_REQUEST_LIMIT_EXEMPT_HOSTS={{matrix_server_fqn_matrix}},localhost,${MATRIX_CONTAINER_NETWORK_SUBNET} \
```

karmanyaahm:
From `man systemd.exec` under `Environment=`:

> Variable expansion is not performed inside the strings, however, specifier expansion is possible. The "$" character has no special meaning. 

i.e. `$()` doesn't do anything

karmanyaahm:
I manually tested the following:
```ini
Environment="MATRIX_CONTAINER_NETWORK_SUBNET=$(/usr/bin/env docker inspect --format='{{ .(index .IPAM.Config 0).Subnet }}' 'matrix')"
# ...
                        --env NTFY_VISITOR_REQUEST_LIMIT_EXEMPT_HOSTS=matrix.mydomain.cc,localhost,${MATRIX_CONTAINER_NETWORK_SUBNET} \
```
and it just passes the literal `$(/usr/bin/env docker inspect --format=`... to ntfy

spantaleev:
Hmm.. Strange.. I believe I've used `$()` in `Environment` and had it work in the past, but.. maybe it doesn't anymore or something.

We can probably still make use of `{% raw %}... {% endraw %}` to make the current code prettier?

karmanyaahm:
> We can probably still make use of `{% raw %}... {% endraw %}` to make the current code prettier?

:+1: done

spantaleev:
Anything else to do before we merge this? It looks good, I think. Except for the extra whitespace just before `{% endraw %}`.

karmanyaahm:
Removed the whitespace, everything is working well on my server, so I don't think anything else is needed. Can you "squash and merge" this?

spantaleev:
Thank you for the great work and patience in bringing this important feature to our ntfy setup! :bow:

PR #2141, opened 2022-09-29
Update Mjolnir Config from its #962 state to todays template.

FSG-Cat:
This PR is not fully ready as far as i am concerned. Someone should take a second pass and give all options not commented out a way to be modified. I dont know if overides is that way or not. If its deemed that no changes are needed because you can change all the options then this PR should be in a merge ready state. 

I edited the data path since we use the data/ path instead of data/storage/ in our current config and this is supposed to be able to merge clean.

### DIFF
--- a/roles/matrix-bot-mjolnir/templates/production.yaml.j2
+++ b/roles/matrix-bot-mjolnir/templates/production.yaml.j2
@@ -1,136 +1,184 @@
-# Where the homeserver is located (client-server URL). This should point at
-# pantalaimon if you're using that.
+# Endpoint URL that Mjolnir uses to interact with the matrix homeserver (client-server API),
+# set this to the pantalaimon URL if you're using that.
 homeserverUrl: "{{ matrix_homeserver_url }}"
 
-# The access token for the bot to use. Do not populate if using Pantalaimon.
+# Endpoint URL that Mjolnir could use to fetch events related to reports (client-server API and /_synapse/),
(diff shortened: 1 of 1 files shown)


aaronraimist:
> This PR is not fully ready as far as i am concerned. Someone should take a second pass and give all options not commented out a way to be modified. I dont know if overides is that way or not. If its deemed that no changes are needed because you can change all the options then this PR should be in a merge ready state.

All the options are configurable through `matrix_bot_mjolnir_configuration_extension_yaml` so this can be merged without changes

FSG-Cat:
> > This PR is not fully ready as far as i am concerned. Someone should take a second pass and give all options not commented out a way to be modified. I dont know if overides is that way or not. If its deemed that no changes are needed because you can change all the options then this PR should be in a merge ready state.
> 
> All the options are configurable through `matrix_bot_mjolnir_configuration_extension_yaml` so this can be merged without changes

I susspected this but i wasnt sure so i wanted to be cautious.

spantaleev:
If the new config works well (which I assume it does), it's fine. As Aaron says, we don't need special variables for controlling the other options. In fact, we don't want to add hundreds of variables for controlling the various esoteric options of each component.

Some important features (I see there's a new web UI feature for Mjolnir?) may be worth adding support for though. For the web UI, we'd need a new variable and likely some tasks in `tasks/init.yml` which will inject stuff into the `matrix-nginx-proxy` configuration, etc. Feel free to contribute this in another PR!

PR #2143, opened 2022-09-30
Update proxied nginx for maubot

Warrows:
I've got a situation where I'm using maubot with gitlab bot. 
My matrix server is setup behind a nginx configured as a proxy.
The bot is giving me error 400 : room not found on webhooks.
The proposed change is working fixes the problem in my situation. However I have not tested it on other configurations nor ran any test.

### DIFF
--- a/roles/matrix-bot-maubot/tasks/init.yml
+++ b/roles/matrix-bot-maubot/tasks/init.yml
@@ -14,14 +14,13 @@
           {% if matrix_nginx_proxy_enabled | default(False) %}
             {# Use the embedded DNS resolver in Docker containers to discover the service #}
              resolver 127.0.0.11 valid=5s;
-             set $backend "matrix-bot-maubot:{{ matrix_bot_maubot_management_interface_port }}$request_uri";
-             proxy_pass http://$backend;
-             proxy_set_header Host $host;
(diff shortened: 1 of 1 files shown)


spantaleev:
A few questions:

1. Does it really not work without this extra `Host` header?
2. Why not `proxy_pass http://$backend/$request_uri`, instead of messing with `set $backend "matrix-bot-maubot:{{ matrix_bot_maubot_management_interface_port }}";`?
3. Why not apply the same fix to the `{% else %}` clause below as well (for when `matrix_nginx_proxy_enabled: false`)?

Warrows:
1. You're right, it seems to work without the `Host` header.
2. As long as `/$1`is here the fix doesn't work in my situation and I get 404. I need to remove it for it to work. The `$request_uri` (not `/$request_uri`) then can be put at the end of `set $backend` or `proxy_pass`, it's all the same. I just put it where the `/$1` was.
3. I just did not test it but I figure the same could be applied.

spantaleev:
Looks good! Thanks!

mjarr:
@spantaleev
This commit has quote issues. It broke the management interface for me.

spantaleev:
Thanks for reporting this! It should be fixed in https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2149

PR #2146, opened 2022-10-02
Fix error introduced in #2141

FSG-Cat:
#2141 introduced a change to the config that makes mjolnir poll for reports. This change is reverted in this PR as Mjolnir is not happy to try to poll for Reports but get denied due to lack of being a homeserver admin that is required. 

If your mjolnir is homeserver admin the changes where safe but i am not going to expect all mjolnirs to be homeserver admins. There are other config flags that require it being HS admin that are not default enabled in #2141 for that very reason.


spantaleev:
Thanks for fixing it up!

PR #2155, opened 2022-10-07
dendrite: update config

shalzz:


roles/matrix-dendrite/templates/dendrite/dendrite.yaml.j2
```diff
@@ -349,10 +349,16 @@ sync_api:
...
   real_ip_header: {{ matrix_dendrite_sync_api_real_ip_header|to_json }}
-  fulltext:
+  # Configuration for the full-text search engine.
+  search:
+    # Whether or not search is enabled.
     enabled: false
-    index_path: "./fulltextindex"
-    language: "en" # more possible languages can be found at https://github.com/blevesearch/bleve/tree/master/analysis/lang
+    # The path where the search index will be created in.
+    index_path: "/matrix-media-store-parent/searchindex"
```
spantaleev, line 357: I was about to say:

> Why `/matrix-media-store-parent/searchindex`? This is not media.

But it seems like we have `matrix_dendrite_storage_path: "{{ matrix_dendrite_base_path }}/storage"` and mount it to `/matrix-media-store-parent`.

It'd be nice to fix this up in the future. For now, I'll merge this as it is.

PR #2161, opened 2022-10-09
fix(mautrix-discord): allow configuring the restricted_rooms option

meenzen:
This is a workaround for a conduit issue. Trying to join some spaces created by the mautrix-discord bridge results in M_FORBIDDEN errors. Setting the restricted_rooms option to false fixes the issue.


roles/matrix-bridge-mautrix-discord/templates/config.yaml.j2
```diff
@@ -101,7 +101,7 @@ bridge:
     message_error_notices: true
     # Should the bridge use space-restricted join rules instead of invite-only for guild rooms?
     # This can avoid unnecessary invite events in guild rooms when members are synced in.
-    restricted_rooms: true
+    restricted_rooms: {{ matrix_mautrix_discord_restricted_rooms }}
```
spantaleev, line 104: This `bridge.restricted_rooms` setting should be controlled by a variabled named: `matrix_mautrix_discord_bridge_restricted_rooms`, not `matrix_mautrix_discord_restricted_rooms`.

roles/matrix-bridge-mautrix-discord/defaults/main.yml
```diff
@@ -140,3 +140,7 @@ matrix_mautrix_discord_registration: "{{ matrix_mautrix_discord_registration_yam
 matrix_mautrix_discord_bridge_encryption_allow: false
 matrix_mautrix_discord_bridge_encryption_default: "{{ matrix_mautrix_discord_bridge_encryption_allow }}"
 matrix_mautrix_discord_bridge_encryption_key_sharing_allow: "{{ matrix_mautrix_discord_bridge_encryption_allow }}"
+
+# On conduit this option may prevent you from joining spaces created by the bridge.
+# Setting this to false fixes the issue.
+matrix_mautrix_discord_restricted_rooms: true
```
spantaleev, line 146: Even now, one doesn't need extra variables for controlling all the esoteric settings of a bridge (or other component).

You can use something like this to control this setting:

```yaml
matrix_mautrix_discord_configuration_extension_yaml: |
  bridge:
    restricted_rooms: true
```

----

Still, it could be worth merging this PR, because once we have a dedicated variable, it would be easier to automatically influence it via `group_vars/matrix_servers` so that we can do the right thing depending on the server implementation.

We could add something like this in the `mautrix-discord` section of `group_vars/matrix_servers`:

```yaml
# Enabling bridge.restricted_rooms for this bridge does not work well with Conduit, so we disable it by default.
# See: URL to upstream Conduit issue or mautrix-discord issue, which explains more about the problem
matrix_mautrix_discord_bridge_restricted_rooms: "{{ false if matrix_homeserver_implementation == 'conduit' else true }}"
```

Feel free to add this to your PR!

spantaleev (changes requested):

meenzen:
I've marked this as a draft for now because we don't have an upstream issue yet.

meenzen:
Currently it is not clear if the issue is with Conduit or the bridge. Once that's known I'll create an issue and update this pr.

spantaleev:
Merged! Thanks! I hope we'll remember to remove the `group_vars/matrix_servers` part when Conduit 0.5.0 actually gets released.

PR #2162, opened 2022-10-10
Update Caddy examples and outdated Jitsi docker env variable/documentation link

fdarveau:
I encountered a few issues when setting up my instance, some of them caused by lacking or outdated documentation.

1. Updated the `DOCKER_HOST_ADDRESS` for Jitsi to its new name (`JVB_ADVERTISE_IPS`)
2. Moved a part of the external webserver documentation that seemed to be specific to nginx because it was in the nginx section to the generic section since the configuration needed for all external webservers on another docker network or host. Also includes an example for a default setup (Synapse + Element)
3. Updated Caddyfile to :
    - Support CORS
    - Update some click-jacking protection directives to use a different approach that allows different subdomains to use the website as a frame, not just the exact same origin (ex : `element.example.com` trying to use `jitsi.example.com` in an iframe)
    - Add an example to handle `.well-knowns` in cases where the base domain already has `.well-knowns`

### DIFF
--- a/docs/configuring-playbook-own-webserver.md
+++ b/docs/configuring-playbook-own-webserver.md
@@ -35,7 +35,7 @@ No matter which external webserver you decide to go with, you'll need to:
         ```
     - if using an external server on another host, add the `<service>_http_host_bind_port` or `<service>_http_bind_port` variables for the services that will be exposed by the external server on the other host. The actual name of the variable is listed in the `roles/<service>/defaults/vars.yml` file for each service. Most variables follow the `<service>_http_host_bind_port` format.
        
-      These variables will make Docker expose the ports to the local network instead of localhost only.
+      These variables will make Docker expose the ports on all network interfaces instead of localhost only.
       [Keep in mind that there are some security concerns if you simply proxy everything.](https://github.com/matrix-org/synapse/blob/master/docs/reverse_proxy.md#synapse-administration-endpoints)
 
       Here are the variables required for the default configuration (Synapse and Element)
--- a/docs/configuring-playbook-own-webserver.md
+++ b/docs/configuring-playbook-own-webserver.md
@@ -33,16 +33,16 @@ No matter which external webserver you decide to go with, you'll need to:
         ```yaml
         matrix_nginx_proxy_enabled: false
         ```
-    - if using an external server on another docker network or host, add the `<service>_http_host_bind_port` or `<service>_http_bind_port` variables for the services that will be exposed by the external server on the other docker network or host. The actual name of the variable is listed in the `roles/<service>/defaults/vars.yml` file for each service. Most variables follow the `<service>_http_host_bind_port` format.
+    - if using an external server on another host, add the `<service>_http_host_bind_port` or `<service>_http_bind_port` variables for the services that will be exposed by the external server on the other host. The actual name of the variable is listed in the `roles/<service>/defaults/vars.yml` file for each service. Most variables follow the `<service>_http_host_bind_port` format.
        
       These variables will make Docker expose the ports to the local network instead of localhost only.
       [Keep in mind that there are some security concerns if you simply proxy everything.](https://github.com/matrix-org/synapse/blob/master/docs/reverse_proxy.md#synapse-administration-endpoints)
 
       Here are the variables required for the default configuration (Synapse and Element)
        ```
-        matrix_synapse_container_client_api_host_bind_port: '8008'
-        matrix_synapse_container_federation_api_plain_host_bind_port: '8048'
-        matrix_client_element_container_http_host_bind_port: "8765"
+        matrix_synapse_container_client_api_host_bind_port: '0.0.0.0:8008'
+        matrix_synapse_container_federation_api_plain_host_bind_port: '0.0.0.0:8048'
+        matrix_client_element_container_http_host_bind_port: "0.0.0.0:8765"
        ```
 
 3) **If you'll manage SSL certificates by yourself**, edit your configuration file (`inventory/host_vars/matrix.<your-domain>/vars.yml`) to disable SSL certificate retrieval:
--- a/docs/configuring-playbook-own-webserver.md
+++ b/docs/configuring-playbook-own-webserver.md
@@ -27,11 +27,23 @@ No matter which external webserver you decide to go with, you'll need to:
 
 1) Make sure your web server user (something like `http`, `apache`, `www-data`, `nginx`) is part of the `matrix` group. You should run something like this: `usermod -a -G matrix nginx`. This allows your webserver user to access files owned by the `matrix` group. When using an external nginx webserver, this allows it to read configuration files from `/matrix/nginx-proxy/conf.d`. When using another server, it would make other files, such as `/matrix/static-files/.well-known`, accessible to it.
 
-2) Edit your configuration file (`inventory/host_vars/matrix.<your-domain>/vars.yml`) to disable the integrated nginx server:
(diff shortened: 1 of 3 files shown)


docs/configuring-playbook-own-webserver.md
```diff
@@ -27,11 +27,23 @@ No matter which external webserver you decide to go with, you'll need to:
...
+    - if using an external server on another docker network or host, add the `<service>_http_host_bind_port` or `<service>_http_bind_port` variables for the services that will be exposed by the external server on the other docker network or host. The actual name of the variable is listed in the `roles/<service>/defaults/vars.yml` file for each service. Most variables follow the `<service>_http_host_bind_port` format.
+       
+      These variables will make Docker expose the ports to the local network instead of localhost only.
+      [Keep in mind that there are some security concerns if you simply proxy everything.](https://github.com/matrix-org/synapse/blob/master/docs/reverse_proxy.md#synapse-administration-endpoints)
+
+      Here are the variables required for the default configuration (Synapse and Element)
+       ```
+        matrix_synapse_container_client_api_host_bind_port: '8008'
+        matrix_synapse_container_federation_api_plain_host_bind_port: '8048'
+        matrix_client_element_container_http_host_bind_port: "8765"
```
spantaleev, line 45: Why do you need to tell people to do this explicitly?

We already do it automatically when `matrix_nginx_proxy_enabled: false`.

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/06161073468027361585b9c850f8975ff71139d2/group_vars/matrix_servers#L2293-L2298

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/06161073468027361585b9c850f8975ff71139d2/group_vars/matrix_servers#L2300-L2301

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/06161073468027361585b9c850f8975ff71139d2/group_vars/matrix_servers#L2197-L2200

---

The above also bind to `127.0.0.01:PORT`, instead of just `PORT` (which exposes the port on all interfaces).

---

When someone is using a reverse-proxy on another network, it's probably easier to reconcile that difference by making them share the network (see the `matrix_docker_network` variable).

For reverse-proxying using yet another server (fronting matrix-nginx-proxy), we have these docs too: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/06161073468027361585b9c850f8975ff71139d2/docs/configuring-playbook-own-webserver.md#method-2-fronting-the-integrated-nginx-reverse-proxy-webserver-with-another-reverse-proxy

examples/caddy2/Caddyfile
```diff
@@ -123,6 +144,8 @@ element.DOMAIN.tld {
                 X-Content-Type-Options "nosniff"
                 # Disallow the site to be rendered within a frame (clickjacking protection)
                 X-Frame-Options "DENY"
```
spantaleev, line 146: You're not removing `X-Frame-Options` here, but you are removing it from all other places?

examples/caddy2/Caddyfile
```diff
@@ -185,9 +212,9 @@ element.DOMAIN.tld {
 #
 #        # Prevent some browsers from MIME-sniffing a response away from the declared Content-Type
 #        X-Content-Type-Options "nosniff"
-#
-#        # Disallow the site to be rendered within a frame (clickjacking protection)
-#        X-Frame-Options "SAMEORIGIN"
+
+#        # Only allow same base domain to render this website in a frame; Can be removed if the client (Element for example) is hosted on another domain
+#        # Content-Security-Policy frame-ancestors https://*.DOMAIN.tld
```
spantaleev, line 217: The comment says "Only allow the base domain", but.. `*.DOMAIN.tld` seems to say "all sibling domains".

This seems to change the behavior compared to before - and to a more insecure way of doing things.

Previously, by default, **no framing was allowed at all**. Now, this `Content-Security-Policy` is there, but it's commented out by default, so **framing is allowed from anywhere**.

docs/configuring-playbook-own-webserver.md
fdarveau: You are right, it is not needed when the other webproxy is on a different docker network (same host). I will update my changes to remove that part.

`127.0.0.01:PORT` binds to the loopback interface only. This works only when the other webserver is on the same host (both on `192.158.0.1` for example). There is a good explanation of the difference between 127.0.0.1 and 0.0.0.0 (default for Docker if we don't specifiy an IP address) in this stack overflow : https://stackoverflow.com/a/20778887

If the other webserver is on `192.168.0.2`, it won't be able to reach the service exposed with `127.0.0.1:PORT` on `192.168.0.1`.

Binding to `127.0.0.1` can be a good default, but documenting how to bind the ports to allow another device on the same network to reach the service is necessary. The default in the snippets you linked to could also be changed to bind to `0.0.0.0:PORT` or just `PORT`.

examples/caddy2/Caddyfile
fdarveau: The reasoning is that Jitsi and Dimension can be framed into Element, but Element isn't framed in anything by default since it is the "root" client app.

docs/configuring-playbook-own-webserver.md
spantaleev: The default try to be safe.

Binding to `0.0.0.0` by default for various services is not great.

Very few people use a reverse-proxy on another machine and need to bind on all interfaces. Most others only need to disable matrix-nginx-proxy and will use another local proxy - fo them, binding to `127.0.0.1` is ideal, and safe.

examples/caddy2/Caddyfile
fdarveau: Maybe "base domain" could be rephrased. The goal here is to allow the Jitsi integration in Element to work.

It is less secure, but should still be secure enough since you usually own the whole domain. It could be changed to specifically allow `https://element.DOMAIN.tld` only for a more secure default/example or subdomain renting services.

fdarveau: To clarify, `X-Frame-Options "SAMEORIGIN"` specifically only allows `https://jitsi.DOMAIN.tld` to frame `jitsi.DOMAIN.tld`,

spantaleev: It's still commented out by default though and I suspect most peope copy-pasting the file may leave it as it is (commented out), thus being exposed to framing from anywhere.

It's not the end of the world, but.. it is very different compared to what we had as a default before (no framing allowed from anywhere).

fdarveau: That's a mistake on my end. I will fix it for dimension

docs/configuring-playbook-own-webserver.md
```diff
@@ -27,11 +27,23 @@ No matter which external webserver you decide to go with, you'll need to:
...
-```
+2) Edit your configuration file (`inventory/host_vars/matrix.<your-domain>/vars.yml`)
+   - to disable the integrated nginx server:
+
+        ```yaml
+        matrix_nginx_proxy_enabled: false
+        ```
+    - if using an external server on another host, add the `<service>_http_host_bind_port` or `<service>_http_bind_port` variables for the services that will be exposed by the external server on the other host. The actual name of the variable is listed in the `roles/<service>/defaults/vars.yml` file for each service. Most variables follow the `<service>_http_host_bind_port` format.
+       
+      These variables will make Docker expose the ports to the local network instead of localhost only.
```
spantaleev, line 38:
```suggestion
      These variables will make Docker expose the ports on all network interfaces instead of localhost only.
```

Issue #1740, opened 2022-04-10
playbook do not create container goofys


Seele-Vollerei32:
**Playbook Configuration**:

My `vars.yml` file looks like this:

```yaml
---
# The bare domain name which represents your Matrix identity.
# Matrix user ids for your server will be of the form (`@user:<matrix-domain>`).
#
# Note: this playbook does not touch the server referenced here.
# Installation happens on another server ("matrix.<matrix-domain>").
#
# If you've deployed using the wrong domain, you'll have to run the Uninstalling step,
# because you can't change the Domain after deployment.
#
# Example value: example.com
matrix_domain: atunemic.cn

# The Matrix homeserver software to install.
# See `roles/matrix-base/defaults/main.yml` for valid options.
matrix_homeserver_implementation: synapse

# A secret used as a base, for generating various other secrets.
# You can put any string here, but generating a strong one is preferred (e.g. `pwgen -s 64 1`).
matrix_homeserver_generic_secret_key: 'SECRET_KEY'

# This is something which is provided to Let's Encrypt when retrieving SSL certificates for domains.
#
# In case SSL renewal fails at some point, you'll also get an email notification there.
#
# If you decide to use another method for managing SSL certificates (different than the default Let's Encrypt),
# you won't be required to define this variable (see `docs/configuring-playbook-ssl-certificates.md`).
#
# Example value: someone@example.com
matrix_ssl_lets_encrypt_support_email: 'we123445@outlook.com'

# A Postgres password to use for the superuser Postgres user (called `matrix` by default).
#
# The playbook creates additional Postgres users and databases (one for each enabled service)
# using this superuser account.
matrix_postgres_connection_password: 'PASSWORD'

matrix_synapse_enable_registration: true
matrix_synapse_registrations_require_3pid: 'email'

matrix_prometheus_enabled: true

matrix_prometheus_node_exporter_enabled: true

matrix_grafana_enabled: true

matrix_grafana_anonymous_access: false

# This has no relation to your Matrix user id. It can be any username you'd like.
# Changing the username subsequently won't work.
matrix_grafana_default_admin_user: "kevin"

# Changing the password subsequently won't work.
matrix_grafana_default_admin_password: "PASSWORD"

matrix_synapse_admin_enabled: true

matrix_synapse_ext_password_provider_shared_secret_auth_enabled: true
matrix_synapse_ext_password_provider_shared_secret_auth_shared_secret: TOKEN

[...]

Issue #1107, opened 2021-06-05
matrix-goofys: error related to fusermount


luixxiul:
The service to enable and run [goofys](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-s3.md#migrating-from-local-filesystem-storage-to-s3) for filesystem fails due to the error related to `fusermount` when you restart `matrix-goofys`:

````
Jun 05 23:20:05 ik1-432-48064 systemd[1]: Starting Matrix Goofys media store...
Jun 05 23:20:05 ik1-432-48064 matrix-goofys[13842]: Error response from daemon: Cannot kill container: matrix-goofys.service: No such container: matrix-goofys.service
Jun 05 23:20:05 ik1-432-48064 matrix-goofys[13852]: Error: No such container: matrix-goofys.service
Jun 05 23:20:05 ik1-432-48064 systemd[1]: Started Matrix Goofys media store.
Jun 05 23:20:05 ik1-432-48064 matrix-goofys[13861]: 2021/06/05 14:20:05.982870 main.FATAL Mounting file system: Mount: mount: running fusermount: exit status 1
Jun 05 23:20:05 ik1-432-48064 matrix-goofys[13861]: stderr:
Jun 05 23:20:05 ik1-432-48064 matrix-goofys[13861]: fusermount: mountpoint is not empty
Jun 05 23:20:05 ik1-432-48064 matrix-goofys[13861]: fusermount: if you are sure this is safe, use the 'nonempty' mount option
Jun 05 23:20:06 ik1-432-48064 systemd[1]: matrix-goofys.service: Main process exited, code=exited, status=1/FAILURE
Jun 05 23:20:06 ik1-432-48064 systemd[1]: matrix-goofys.service: Failed with result 'exit-code'.
Jun 05 23:20:11 ik1-432-48064 systemd[1]: matrix-goofys.service: Service RestartSec=5s expired, scheduling restart.
Jun 05 23:20:11 ik1-432-48064 systemd[1]: matrix-goofys.service: Scheduled restart job, restart counter is at 338.
````

Restarting the server did not fix the error, I am wondering how it should be possible to troubleshoot and fix it to make matrix-goofys run successfully. Thanks!

````
$ uname -v
#1 SMP Debian 4.19.181-1 (2021-03-19)
````

The setup is based on 8cf3bdb7720d7b61fa32bc7e1290439c533fbe3a.

[...]

PR #2163, opened 2022-10-11
fix(goofys): fix synapse systemd unit file to correctly require goofys

jimeh:
On Debian 10 (buster) at least, while the Synapse systemd service unit was referring to Goofys as `matrix-goofys` without a `.service` suffix, systemd was ignoring the goofys dependency, starting Synapse before Goofys. All other dependant units which work are using the ".service" suffix.

This generally leads to the mount path goofys using having been populated by Synapse before Goofys starts, causing it to fail due to the mount target not being empty.

The fix seems to be to ensure that the Synapse service unit refers to Goofys as `matrix-goofys.service`.

This change causes the following two lines in `/etc/systemd/system/matrix-synapse.service`:

    Requires=matrix-goofys
    After=matrix-goofys

To be modified to:

    Requires=matrix-goofys.service
    After=matrix-goofys.service

---

I believe this may fix both #1107 and #1740.

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -2364,7 +2364,7 @@ matrix_synapse_systemd_required_services_list: |
     +
     (['matrix-postgres.service'] if matrix_postgres_enabled else [])
     +
-    (['matrix-goofys'] if matrix_s3_media_store_enabled else [])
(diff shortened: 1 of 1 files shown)


spantaleev:
Good catch! Thank you for fixing it!

PR #2164, opened 2022-10-11
Enable hookshot feeds by default

etkecc:
as per documentation: "Services that need no further configuration are enabled by default" and feeds service doesn't require it


spantaleev:
The default config does not enable feeds: https://github.com/matrix-org/matrix-hookshot/blob/107b6c0c2f33410b99a9e7f84d81caccc03311e6/config.sample.yml#L88

On the other hand, it doesn't enable webhooks either: https://github.com/matrix-org/matrix-hookshot/blob/107b6c0c2f33410b99a9e7f84d81caccc03311e6/config.sample.yml#L79

It doesn't seem like it enables anything by default, so it doesn't seem worth following it.

PR #2166, opened 2022-10-11
Provision extra Jitsi JVB services on additional hosts

warrenbailey:
Adding tasks to provision one or more JVN servers to additional hosts. Opening the prosody port 5222 so that these JVB can then register with prosody and be used by jicofo to manage conferences.

This task is not part of `setup-all` but rather it's own task `setup-additional-jvb` and will only act on servers list in the ansible host file under the `[jvb_servers]` section. As such should not impact the playback of the setup-all playbook.

In order to achieve this a minor refactor of matrix-base was required, splitting the sanity check into two parts.


roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-jitsi.conf.j2
```diff
@@ -54,6 +54,20 @@
 		tcp_nodelay on;
 	}
 
+	# colibri (JVB) websockets for additional JVBs
+	location ~ ^/colibri-ws/([a-zA-Z0-9-\.]+)/(.*) {
+		proxy_pass http://$1:9090/colibri-ws/$1/$2$is_args$args;
```
spantaleev, line 59: It sounds dangerous to just get something from the URL path and to use it as a host.

It might expose various hosts that we try to keep private and don't reverse-proxy to.

I wonder if we can tighten this up somehow (regex whitelist with known allowed hosts; some prefix; etc.)?

setup.yml
```diff
@@ -73,3 +73,12 @@
     - matrix-postgres-backup
     - matrix-backup-borg
     - matrix-common-after
+
+
+- name: "Set up additional Jitsi JVB servers"
+  hosts: "jvb_servers"
```
spantaleev, line 79: Maybe a group named `jitsi_jvb_servers` will be clearer to people reading the `inventory` file in the future

roles/matrix-base/tasks/main.yml
```diff
@@ -1,18 +1,24 @@
...
 
+- ansible.builtin.import_tasks: "{{ role_path }}/tasks/sanity_check.yml"
+  tags:
+    - setup-all
+
 - ansible.builtin.import_tasks: "{{ role_path }}/tasks/clean_up_old_files.yml"
   when: run_setup | bool
   tags:
     - setup-all
+    - setup-additional-jvb
```
spantaleev, line 15: Perhaps `setup-additional-jitsi-jvb` is a better name for this.

I wonder if we should introduce a new tag (called `common` or something of this sort) and use this tag in `matrix-base/tasks/main.yml`, so that we don't need to mention Jitsi stuff.

Later, we can ask people to invoke the "setup additional Jitsi JVB command like this: `ansible-playbook -i inventory/hosts --limit jitsi_jvb_servers setup.yml --tags=common,setup-additional-jitsi-jvb`

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-jitsi.conf.j2
warrenbailey: Yes you're right, in my scenario I'm using a private IP address here which is only accessible to nginx but in hindsight you are correct this is probably a bit of a security hole. I'll have a think.

setup.yml
warrenbailey: Agreed, will change

roles/matrix-base/tasks/main.yml
warrenbailey: Yeah I think that's a good idea. 👍

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-jitsi.conf.j2
spantaleev: The problem is that if you Matrix server is public, one could hit:

- `https://jitsi.DOMAIN/colibri-ws/google.com/search` and hit `http://google.com:9090/colibri-ws/search`, etc.
- `https://jitsi.DOMAIN/colibri-ws/matrix-synapse/search` and hit `http://matrix-synapse:9090/colibri-ws/search`, etc.

Thankfully, none of these services are exposed on the `9090` port, and none offer services under the`/colibri-ws` path prefix, so it can't do too much damage.

Because of this, we may consider merging this as is.

It would have been better if it only handled whitelsited hosts though. It may be with some mapping (host id to hostname). By passing it to a `matrix_jitsi_*` variable, and then wiring that to some `matrix_nginx_proxy_..jittsi..*` variable (via `group_vars/matrix_servers`), we could pass the mapping to this file and it could generate a new block for each service.

By default, the mapping could include `jvb-1`, pointing to `matrix-jitsi-jvb:9090`.

One could add more things to it:

- `jvb-2` pointing to `https://jvb.DOMAIN`
- `jvb-3` pointing to `http://matrix-jvb-3:9090`
- etc.

The mapping keys (`jvb-1`, `jvb-2`, `jvb-3`, etc.) are probably useful for passing to Jitsi somehow? How does it know to invoke these other JVB servers otherwise? Is one meant to redefine `JVB_WS_SERVER_ID={{ matrix_jitsi_jvb_server_id }}` or change `matrix_jitsi_jvb_server_id` to a comma-separated list of JVB ids (`jvb-1,`jvb-2`)?

Would it be hard to change to something like this?

roles/matrix-nginx-proxy/defaults/main.yml
```diff
@@ -646,3 +646,9 @@ matrix_nginx_proxy_synapse_cache_proxy_cache_valid_time: "24h"
 # http://nginx.org/en/docs/ngx_core_module.html#worker_connections
 matrix_nginx_proxy_worker_processes: auto
 matrix_nginx_proxy_worker_connections: 1024
+
+# A mapping of JVB ids to hostname for additional jvb blocks to add to Jitsi's server configuration (matrix-jitsi.conf)
+matrix_nginx_proxy_proxy_jitsi_additional_jvbs_enabled: false
+matrix_nginx_proxy_proxy_jitsi_additional_jvbs:
+  - jvb-2: 192.168.0.1
+  - jvb-3: 192.168.0.2
```
spantaleev, line 654: Using a default value which doesn't contain stuff would be better.

```suggestion
# Example:
# matrix_nginx_proxy_proxy_jitsi_additional_jvbs:
#   jvb-2: 192.168.0.1
#   jvb-3: 192.168.0.2
matrix_nginx_proxy_proxy_jitsi_additional_jvbs: {}
```

Doing this, you don't even need the `_enabled` variable.

You can just iterate over it in `roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-jitsi.conf.j2`. If `matrix_nginx_proxy_proxy_jitsi_additional_jvbs` is empty, nothing happens. If it's non-empty (JVBs added to the list), it will inject a `location` block for each one.

My `matrix_nginx_proxy_proxy_jitsi_additional_jvbs` above uses a dictionary, which can be iterated over directly like this: `{% for id, ip_address in matrix_nginx_proxy_proxy_jitsi_additional_jvbs %}`, which removes the need for 2 nested `for` loops.

----

I'm also a bit confused about how `matrix_nginx_proxy_proxy_jitsi_additional_jvbs` is meant to be used. Maybe we need more comments there and in `docs/configuring-playbook-jitsi.md`.

Can `matrix_nginx_proxy_proxy_jitsi_additional_jvbs` be populated automatically based on your inventory? Even if it can, should it?

Do we still need `matrix_jitsi_jvb_server_id: jvb-1`?

It seems like using `jvb-1` inside `matrix_nginx_proxy_proxy_jitsi_additional_jvbs` will cause trouble. Perhaps this should be pointed out. Maybe pointing to `matrix_jitsi_jvb_server_id` might be a good idea.

`matrix_jitsi_jvb_server_id`'s comment (if this variable will remain in use) could also point to `matrix_nginx_proxy_proxy_jitsi_additional_jvbs`.

I'm still not sure how it all works. How do you tell Jitsi to use these other JVBs? The reverse proxy config is there, but what tells Jitsi that there's `jvb-2` as well? Is this part still missing from the PR or am I misunderstanding things?

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-jitsi.conf.j2
```diff
@@ -53,20 +53,25 @@
 
 		tcp_nodelay on;
 	}
-
+	{% if matrix_nginx_proxy_proxy_jitsi_additional_jvbs_enabled %}
+	{% for jvb in matrix_nginx_proxy_proxy_jitsi_additional_jvbs %}
```
spantaleev, line 57: There should be a way to iterate over this without 2 nested `for` loops.

With my other review suggestion (which redoes `matrix_nginx_proxy_proxy_jitsi_additional_jvbs` to be a dictionary, instead of a list of a dictionaries), the need for 2 `for` loops is removed.

You can use `{% for id, ip_address in matrix_nginx_proxy_proxy_jitsi_additional_jvbs %}` here.

You also don't need the `matrix_nginx_proxy_proxy_jitsi_additional_jvbs_enabled` part.

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-jitsi.conf.j2
```diff
@@ -53,20 +53,25 @@
...
+	{% for id in jvb %}
 	# colibri (JVB) websockets for additional JVBs
-	location ~ ^/colibri-ws/([a-zA-Z0-9-\.]+)/(.*) {
-		proxy_pass http://$1:9090/colibri-ws/$1/$2$is_args$args;
+	location ~ ^/colibri-ws/{{ id }}/(.*) {
+		proxy_pass http://{{ jvb[id] }}:9090/colibri-ws/{{ id }}/$2$is_args$args;
```
spantaleev, line 61: Why `$2`? I don't see 2 regex groups in the location statement (`location ~ ^/colibri-ws/{{ id }}/(.*) {`).

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-jitsi.conf.j2
```diff
@@ -53,20 +53,25 @@
...
 		tcp_nodelay on;
 	}
-
+	{% if matrix_nginx_proxy_proxy_jitsi_additional_jvbs_enabled %}
+	{% for jvb in matrix_nginx_proxy_proxy_jitsi_additional_jvbs %}
+	{% for id in jvb %}
 	# colibri (JVB) websockets for additional JVBs
-	location ~ ^/colibri-ws/([a-zA-Z0-9-\.]+)/(.*) {
-		proxy_pass http://$1:9090/colibri-ws/$1/$2$is_args$args;
+	location ~ ^/colibri-ws/{{ id }}/(.*) {
```
spantaleev, line 60:
```suggestion
	location ~ ^/colibri-ws/{{ id | regex_escape }}/(.*) {
```

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-jitsi.conf.j2
warrenbailey: Yeah that's a mistake on my behalf forgot to change it to $1

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-jitsi.conf.j2
warrenbailey: I added in the enabled flag but after testing with it I agree its not enabled. I can remove the double loop and use a dictionary instead. 👍

roles/matrix-nginx-proxy/defaults/main.yml
warrenbailey: Yeah sorry, I wanted to test that it all worked before adding some more documentation about how to use. 

We will need the matrix_jitsi_jvb_server_id because the JVB themselves need to know their server id. This field ends up in JVB_WS_SERVER_ID environment variable of the additional JVB once it has been provisioned.

warrenbailey:
Thanks for the review comments. I'll make the changes you've recommended and add some more documentation once I know if all works 👍

roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-jitsi.conf.j2
warrenbailey: The id is no longer a regex though

warrenbailey: fair point, will add

docs/configuring-playbook-jitsi.md
```diff
@@ -134,19 +134,44 @@ By default a single JVB is deployed on the same hosts as the matrix server. To t
...
 ```
 
-This will provision a single JVB instance to that host which will then register itself with the prosody service and make itself available for jicofo to route conferences too.
+``` INI
+[jitsi_jvb_servers]
+jvb-2.example.com ansible_host=192.168.0.1 matrix_jitsi_jvb_server_id=jvb-2
+jvb-3.example.com ansible_host=192.168.0.1 matrix_jitsi_jvb_server_id=jvb-2
+```
+
+Note that the server id jvb-1 is reserved for JVB instance running on the main host and therefore should not be used as the id of an additional jvb host.
```
spantaleev, line 157:
```suggestion
Note that the server id `jvb-1` is reserved for the JVB instance running on the Matrix host and therefore should not be used as the id of an additional jvb host.
```

docs/configuring-playbook-jitsi.md
```diff
@@ -134,19 +134,44 @@ By default a single JVB is deployed on the same hosts as the matrix server. To t
...
+
+The nginx configuration will also need to be updated in order to deal with the additional jvbs server. This is achieved via its own configuration variable
```
spantaleev, line 159:
```suggestion
The nginx configuration will also need to be updated in order to deal with the additional JVB servers. This is achieved via its own configuration variable
```

docs/configuring-playbook-jitsi.md
```diff
@@ -134,19 +134,44 @@ By default a single JVB is deployed on the same hosts as the matrix server. To t
...
+
+``` yaml
+matrix_nginx_proxy_proxy_jitsi_additional_jvbs:
+   jvb-2: 192.168.0.1
+   jvb-3: 192.168.0.2
+```
+matrix_nginx_proxy_proxy_jitsi_additional_jvbs: {}
+
+
+Applied together this will allow you to provision a extra JVB instances which will register themselves with the prosody service and be available for jicofo
```
spantaleev, line 172:
```suggestion
Applied together this will allow you to provision extra JVB instances which will register themselves with the prosody service and be available for jicofo 
```

docs/configuring-playbook-jitsi.md
```diff
@@ -134,19 +134,44 @@ By default a single JVB is deployed on the same hosts as the matrix server. To t
...
 
-For this role to work you will need an additional section in the ansible hosts file with the details of the JVB hosts, for example 
+For this role to work you will need an additional section in the ansible hosts file with the details of the JVB hosts, for example:
 ```
 [jitsi_jvb_servers]
 <your jvb hosts> ansible_host=<ip address of the jvb host>
 ```
 
-The followings variables will need to be set in the vars file for each JVB:
+Each JVB will require a server id to be set so that it can be uniquely identify and this allows jitsi to keep track of which conferences are on which JVB.
```
spantaleev, line 143:
```suggestion
Each JVB will require a server id to be set so that it can be uniquely identified and this allows Jitsi to keep track of which conferences are on which JVB.  
```

docs/configuring-playbook-jitsi.md
```diff
@@ -134,19 +134,44 @@ By default a single JVB is deployed on the same hosts as the matrix server. To t
...
+For this role to work you will need an additional section in the ansible hosts file with the details of the JVB hosts, for example:
 ```
 [jitsi_jvb_servers]
 <your jvb hosts> ansible_host=<ip address of the jvb host>
 ```
 
-The followings variables will need to be set in the vars file for each JVB:
+Each JVB will require a server id to be set so that it can be uniquely identify and this allows jitsi to keep track of which conferences are on which JVB.  
+The server id is set with the variable `matrix_jitsi_jvb_server_id` which ends up as the JVB_WS_SERVER_ID environment variables in the JVB docker container. 
+This variable can be set via the host file, a parameter to the ansible command or in the vars.yaml for the host which will have the additional JVB. For example:
```
spantaleev, line 145:
```suggestion
This variable can be set via the host file, a parameter to the ansible command or in the `vars.yaml` for the host which will have the additional JVB. For example:
```

docs/configuring-playbook-jitsi.md
```diff
@@ -134,19 +134,44 @@ By default a single JVB is deployed on the same hosts as the matrix server. To t
...
+`matrix_nginx_proxy_proxy_jitsi_additional_jvbs`, which contains a dictionary of server ids to ip addresses.
+
+For example,
+
+``` yaml
+matrix_nginx_proxy_proxy_jitsi_additional_jvbs:
+   jvb-2: 192.168.0.1
+   jvb-3: 192.168.0.2
+```
+matrix_nginx_proxy_proxy_jitsi_additional_jvbs: {}
```
spantaleev, line 169:
```suggestion
```

This was probably added here by mistake.

docs/configuring-playbook-jitsi.md
```diff
@@ -127,6 +127,52 @@ Read how it works [here](https://github.com/jitsi/jitsi-videobridge/blob/master/
 
 You may want to **limit the maximum video resolution**, to save up resources on both server and clients.
 
+## (Optional) Additional JVBs
+
+By default a single JVB is deployed on the same hosts as the matrix server. To the scale the JVBs and allow more conferences you may wish to provision additional JVB services on other hosts.
```
spantaleev, line 132:
```suggestion
By default a single JVB ([Jitsi VideoBridge](https://github.com/jitsi/jitsi-videobridge)) is deployed on the same host as the Matrix server. To allow more video-conferences to happen at the same time, you may need to provision additional JVB services on other hosts.
```

warrenbailey:
I also have the matrix_jitsi_xmpp_server variable set on the additional JVB. This is needed to so that the JVB knows where the prosody server is. I'll update the documentation but it needs to point to the Matrix server.

spantaleev:
Let me know when you think this is ready for another review!

roles/matrix-jitsi/defaults/main.yml
```diff
@@ -266,4 +270,4 @@ matrix_jitsi_jvb_container_rtp_tcp_host_bind_port: "{{ matrix_jitsi_jvb_rtp_tcp_
 # Controls whether the matrix-jitsi-jvb container exposes its Colibri WebSocket port (tcp/9090 in the container).
 #
 # Takes an "<ip>:<port>" or "<port>" value (e.g. "127.0.0.1:12090"), or empty string to not expose.
-matrix_jitsi_jvb_container_colibri_ws_host_bind_port: ''
+matrix_jitsi_jvb_container_colibri_ws_host_bind_port: 9090
```
spantaleev, line 273: Oh.. why are we exposing these by default now?

warrenbailey: Ah it's needed on the additional jvb as it's outside the docker network but you're right it shouldn't be exposed on the matrix host and hence shouldn't be a default. Good catch, i'll change.

warrenbailey:
merged in master and resolved conflicts

warrenbailey:
@spantaleev anything else you'd like me to change?

setup.yml
```diff
@@ -93,3 +93,15 @@
...
+  hosts: "jitsi_jvb_servers"
+  become: true
+
+  roles:
+    - role: galaxy/com.devture.ansible.role.playbook_help
+    - role: galaxy/com.devture.ansible.role.systemd_docker_base
+
+    - custom/matrix-base
+    - custom/matrix-jitsi
+    - custom/matrix-common-after
```
spantaleev, line 107: Why not make this a completely different playbook file, other than `setup.yml`?

I suppose we could have:

- `playbook/matrix.yml` be what `setup.yml` is now
- `setup.yml` be a symlink to `playbook/matrix.yml`
- a new `playbook/jitsi_jvb.yml` setup file for `jitsi_jvb_servers`

warrenbailey: Sounds good to me

warrenbailey: I tried moving the contents of setup.yml into playbook/matrix.yml by adding an import playbook statement. e.g. in setup.yml

```
- name: Include the matrix playbook
  ansible.builtin.import_playbook: playbooks/matrix.yml
```

However this requires a lot of changes throughout the playbook to support change in directory structure as its using relative paths in a lot of places. For example, everywhere it refers to `custom/<role>` I've had to change to `roles/custom/<role>` Also I needed to move the `groups_vars` directory under playbooks so that it's alongside the playbook.

I've pushed another branch to my fork if you want to see what I mean "https://github.com/warrenbailey/matrix-docker-ansible-deploy/tree/multi-jvb-playbooks/playbooks"

@spantaleev  I just wanted to check my understand of what you're suggesting is right before I push the change to this PR?

spantaleev: I did what I posted in my previous comment above and it worked:

- `mkdir playbook`
- `mv setup.yml playbook/matrix.yml`
- `ln -s playbook/matirx.yml setup.yml`
- `ansible-playbook -i inventory/hosts setup.yml --tags=setup-all`

----

Maybe you're `cd`-ing into the `playbook` directory and trying to use relative paths to your inventory, etc.? You probably shouldn't do that -- trying staying at the project root.

We could use `import_playbook`, but.. I don't see why we should do that when a symlink also works.

warrenbailey: Ah ok, I did miss understand then. I assumed you'd have to do an import playbook and wasn't sure symlinks would work like that with ansible.  

I'll get it sorted asap and pushed up to this PR. 👍

warrenbailey:
I'll update the documentation too.

warrenbailey:
@spantaleev is this what you were thinking?

spantaleev:
Thank you for your patience while bringing this to the playbook!

warrenbailey:
No worries, thank you for merging. It's useful functionality for a great playbook :)

PR #2175, opened 2022-10-15
Add ability to sync only direct Chats in TG-Bridge

smargold476:
Hi all,

as a bloody beginner I tried to add the ability to sync only direct-chats with the bridge, because for me TG-Groups and Channels are way to noisy ... 

I would love to have this reviewed regarding:
1. Is that the right way to extend the project?
2. Can somebody confirm the filter-mode works like this?

Regards & thanks for that really nice project!


spantaleev:
Looks good! Thank you for contributing!

PR #2180, opened 2022-10-19
Bump postgres version to newly released 15

dnapier:
Kid tested, mother approved.


spantaleev:
There was a missign part that I've added in a32eea41fe56b.

Thanks!

PR #2182, opened 2022-10-21
grammar correction

array-in-a-matrix:
the word "both" would refer to only 2 urls in this context


spantaleev:
Thanks! They used to be 2, but not anymore ;)

PR #2183, opened 2022-10-21
Add postgresql 15 as supported by Borg

alemairebe:
As the upgrade to postgres 15 is proposed when running this script. It would be good to add it to the list of supported versions by borg


alemairebe:
It needs alpine to publish postgresql15-client package first; then ekte.cc to update its gitlab pipeline.

PR #2201, opened 2022-10-26
MSC2499 integration: make self-check of /.well-known/matrix/client accept redirects

aine-etke:
Rationale: the /.well-known/matrix/server should return only 200 as well, but it has hardcoded "safe" value, regardless the spec: https://spec.matrix.org/v1.4/client-server-api/#well-known-uri

So, why not do the same for the /.well-known/matrix/client?


spantaleev:
For ` /.well-known/matrix/server` we don't have a variable and hardcode `safe` here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/21eabf45c256073738760592132c6fa4a9ebf9ce/roles/matrix-nginx-proxy/tasks/self_check_well_known.yml#L19

It's because the [Server-Server spec about well-known](https://spec.matrix.org/v1.4/server-server-api/#well-known-uri) says:

> 30x redirects should be followed, however redirection loops should be avoided.

So, the playbook is following the spec for both `/.well-known/matrix/server` (redirects allowed, so we allow them) and `/.well-known/matrix/client` (redirects are not allowed according to the spec, so we don't allow them).

---

The question is.. if everyone is allowing redirects in practice, should we disregard the spec and just allow it too?

Or maybe bring this into discussion with the Matrix folks who would:

- either bring it into the spec (because it makes sense and because it's somewhat common usage already)
- or fix the Element bug which allows this against-spec behavior

----

I think your PR title and PR description are incorrect, mentioning `server` where it should say `client` and vice-versa.

aaronraimist:
> either bring it into the spec (because it makes sense and because it's somewhat common usage already)

The relevant MSC is https://github.com/matrix-org/matrix-spec-proposals/pull/2499

PR #2203, opened 2022-10-26
Update Synapse 1.69.0 -> 1.70.0

adam-kress:


spantaleev:
1.70.0 has some known regressions, wihch don't sound great:

> The regressions we're aware of are:
> - Application Services that can use an Authorization header will likely not work because Synapse is not sending them correctly. (https://github.com/matrix-org/synapse/pull/14301)
> - In some (perhaps unusual) cases, users may be unable to /sync (i.e. receive new messages and generally use their account). (https://github.com/matrix-org/synapse/issues/14300)

I wonder if we should upgrade or wait for 1.70.1..

joecool1029:
For what it's worth it's a vote from me to stay until the point release in a day or two. That first issue will likely break things for anyone using mautrix bridges.

adam-kress:
Agree, I will close this PR. Thank you.

PR #2206, opened 2022-10-27
Change signald version fom 0.22.2 to 0.23.0

cyber-simon:
As mentioned in https://github.com/mautrix/signal/issues/314 signald version is not directly managed by mautrix-bridge-signal. But mautrix-signal 0.4.0 seems to work with signald 0.23.0.

First test seems good. Connection to Signal is working again.


spantaleev:
Merged in https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2207

PR #2210, opened 2022-10-28
use fqcns for some task

rndmh3ro:
Signed-off-by: Sebastian Gumprich <github@gumpri.ch>


rndmh3ro:
This was somehow way harder than expected.

spantaleev:
Thanks!

I think `community.docker.docker_image` was failing on some older Ansible versions from what I remember, but maybe I am misremembering.. Also the Ansible version that we support has increased anyway, so we're probably fine.

rndmh3ro:
> Also the Ansible version that we support has increased anyway, so we're probably fine.

I did not find any documentation on that..

PR #2214, opened 2022-10-28
Update mautrix-signal 0.4.0 -> 0.4.1

etkecc:
ref: https://github.com/mautrix/signal/releases/tag/v0.4.1

Note that part:
> The docker image now has an option to bypass the startup script by setting the MAUTRIX_DIRECT_STARTUP environment variable. Additionally, it will refuse to run as a non-root user if that variable is not set (and print an error message suggesting to either set the variable or use a custom command).


spantaleev:
I think another mautrix image had the same warning and it wasn't causing trouble for us there (likely because we're sending a custom start command).

In any case, I've confirmed that this bridge starts successfully.

PR #2216, opened 2022-10-29
fix: copy vars.yml only

ikkemaniac:


spantaleev:
`matrix_vars_yml_snapshotting_src` defaults to: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/f108f87c8f2ce4ecbac08ff9745ae6921bed35fd/roles/matrix-base/defaults/main.yml#L262

It's already a path to `vars.yml`. There's no need to adjust it.

ikkemaniac:
Sorry, my bad. I somehow had changed the original var.

PR #2217, opened 2022-10-29
Save the GIT hash when running the playbook to allow for easy recovery

ikkemaniac:
To recover from a failure or allow to quickly reset to known state we need to know with what playbook-commit the sever is installed. This commit saves the GIT hash/status when we run the playbook in a file called git_hash.yml. It also backs up that file by copying it to the target machine.


spantaleev:
This breaks backward compatibility for people who download the playbook from an archive and don't have `git` on their system. Not sure how many of them there are. They can always toggle the variable to disable this behavior though.

---

I was wondering about naming:

- All variables we define (except for these `run_*` variables, which we can probably get rid of..) are prefixed with `matrix_`. This one (`git_save_hash`) should probably get the same treatment.
- Moreover, this certain enable/disalbe functionality within the playbook is often named `something_something_enabled`, so it makes sense to use such naming.
- Moreover, this is similar to the `vars.yml`-snapshotting feature that we have, the variable for which is named `matrix_vars_yml_snapshotting_enabled`. It would make sense to use a similar name.

So.. it seems like a variable name such as `matrix_playbook_commit_hash_preservation_enabled` might be a better variable name.

roles/matrix-base/tasks/setup_matrix_base.yml
```diff
@@ -19,6 +19,59 @@
...
+    - name: Git hash
+      debug:
+        msg: "Git hash: {{ git_hash }}"
+    
+    - name: Save git hash in git_hash.yml
+      become: false
+      local_action: 
+        copy
+        content="git_hash_last_run{{ ":" }} {{ git_hash }}\n"
+        dest="{{ matrix_vars_yml_snapshotting_src }}/git_hash.yml"
```
spantaleev, line 48: `matrix_vars_yml_snapshotting_src` defaults to:

```yaml
matrix_vars_yml_snapshotting_src: "{{ inventory_dir }}/host_vars/{{ inventory_hostname }}/vars.yml"
```

So.. using a path like `{{ inventory_dir }}/host_vars/{{ inventory_hostname }}/vars.yml/git_hash.yml` doesn't seem to make much sense.

Also.. There's probably no need to save the git hash locally - you can just save it remotely.

roles/matrix-base/tasks/setup_matrix_base.yml
```diff
@@ -19,6 +19,59 @@
...
+
+    - name: Copy git_hash.yml file to target
+      copy:
+        src: "{{ matrix_vars_yml_snapshotting_src }}/git_hash.yml"
```
spantaleev, line 52:
```suggestion
        content: "{{ git_hash }}"
```

roles/matrix-base/tasks/setup_matrix_base.yml
```diff
@@ -19,6 +19,59 @@
...
+      when: "git_describe.stderr.find('not found') != -1"
+
+    - name: GIT hash error
+      ansible.builtin.fail:
+        msg: >-
+          Error when trying to get the GIT hash.
+          You can disable saving the GIT hash by setting 'git_save_hash: false' in vars.yml
+      when: "git_describe.stderr.find('not found') == -1"
+
+  when: "matrix_vars_yml_snapshotting_enabled|bool and git_save_hash|bool"
```
spantaleev, line 73: The `matrix_vars_yml_snapshotting_enabled|bool` part can be removed.

Saving the git hash can be completel separate from saving `vars.yml` on the server.

ikkemaniac:
> This breaks backward compatibility for people who download the playbook from an archive and don't have `git` on their system. Not sure how many of them there are. They can always toggle the variable to disable this behavior though.
> 

I don't believe it breaks backward compatibility. The first rescue check catches if it failed with a "git not found". If it did, it just prints a msg and continues - no biggie. If it does have git and fails in other ways it raise an error with `fail` + the msg.

> I was wondering about naming:
> 
>     * All variables we define (except for these `run_*` variables, which we can probably get rid of..) are prefixed with `matrix_`. This one (`git_save_hash`) should probably get the same treatment.
> 
>     * Moreover, this certain enable/disalbe functionality within the playbook is often named `something_something_enabled`, so it makes sense to use such naming.
> 
>     * Moreover, this is similar to the `vars.yml`-snapshotting feature that we have, the variable for which is named `matrix_vars_yml_snapshotting_enabled`. It would make sense to use a similar name.
> 
> 
> So.. it seems like a variable name such as `matrix_playbook_commit_hash_preservation_enabled` might be a better variable name.

Understood, didn't even consider to think about that.
Will fix.

roles/matrix-base/tasks/setup_matrix_base.yml
ikkemaniac: understood, i saw your comment on the other PR as well. I guess somewhere during my mods i modified the original `matrix_vars_yml_snapshotting_src` (removed the vars.yml) as it nicely give me the root where we need to store the git hash.
Any suggestion on how to get the correct snapshot directory without having to cut the `vars.yml` off of `matrix_vars_yml_snapshotting_src`? Or are you ok with changing `matrix_vars_yml_snapshotting_src` to only have the dir and than accept my other PR as well? that seems to be the only location where `matrix_vars_yml_snapshotting_src` is used currently. I prefer the latter.

I'll change it to store the `git_hash.yml` only on the target.

roles/matrix-base/tasks/setup_matrix_base.yml
ikkemaniac: it could but i figured it would make sense to keep them linked as i didn't see any value in it if you don't save `vars.yml` as well.
Will fix.

roles/matrix-base/tasks/setup_matrix_base.yml
ikkemaniac: ok, scratch that directory comment.... we don't need the directory as we won't save it locally.
I need some coffee

spantaleev:
Would you work on fixing the CI errors?

They're mostly about whitespace issues and using short name for Ansible modules as opposed to fully-qualified ones (`shell` -> `ansible.builtin.shell`).

I think that instead of `shell`, you can use `ansible.builtin.command` too.

ikkemaniac:
Everything should be fixed.

`shell` is preferred in this case as `command` returns an error msg that's to generic to distinguish between 'git not installed' or other stuff.

ikkemaniac:
now it's fixed, for real

spantaleev:
The `command` module's `result` also contains `stdout` and `stderr`, so I don't see why we need the `shell` module and these `noqa` things. We can probably just use the `command` module as is

ikkemaniac:
The output is quite different AND `stderr` is empty using `command`.

`command`:

```
TASK [matrix-base : Get local git hash] ********************************
fatal: [matrix -> 127.0.0.1]: FAILED! => changed=false 
  cmd: git describe --always --tags --dirty --long --all
  msg: '[Errno 2] No such file or directory: b''git'''
  rc: 2
  stderr: ''
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
```


`shell`:
```
TASK [matrix-base : Get local git hash] ********************************
fatal: [matrix -> 127.0.0.1]: FAILED! => changed=false 
  cmd: git describe --always --tags --dirty --long --all
  delta: '0:00:00.004084'
  end: '2022-11-01 16:04:53.868744'
  msg: non-zero return code
  rc: 127
  start: '2022-11-01 16:04:53.864660'
  stderr: '/bin/sh: 1: git: not found'
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
```

spantaleev:
OK, looks good! Merged! Thank you for your patience and effort!

ikkemaniac:
I was testing some more: i botched up a running install to see if I could get it back up.

The problem i ran into is that the `vars.yml` and `git_hash.yml` are overwritten and THAN the playbook fails for whatever reason. This means that if we want to recover, grab the files and use the hash and vars to start over, we will get the hash and vars of the latest -failed- attempt!

Wouldn't it be a lot better if we add the `vars.yml` and `git_has.yml` file copy all the way to the end as part of `matrix-common-after`?

What do you think?

spantaleev:
`vars.yml` snapshotting was never meant to be a "last known working configuration".

It was meant to help people who lose their Ansible playbook's `vars.yml` file. Such people often have their Matrix server still running (or some backup of the Matrix server) and they can at least get an up-to-date `vars.yml` file that they can set up a new playbook from.

What you're saying sounds like a good idea though. Let's do it!

If we move `vars.yml` and `git_hash.yml` snapshotting to the end of the playbook though and have it serve both purposes:

- the original purpose of providing some backup in case you lose your `vars.yml` file
- having it actually serve as a "last known good configuration" backup

----

I'm actually in the process of reorganizing the whole playbook into roles (distributed as part of the playbook or pulled using `ansible-galaxy`). The `vars.yml` snapshotting feature has already been extracted out into a [`com.devture.ansible.role.vars_preserver`](https://github.com/devture/com.devture.ansible.role.vars_preserver) role and is already used in [gitea-docker-ansible-deploy](https://github.com/spantaleev/gitea-docker-ansible-deploy) and [https://github.com/spantaleev/nextcloud-docker-ansible-deploy](https://github.com/spantaleev/nextcloud-docker-ansible-deploy). Here's how it's hooked into the Gitea playbook:

- https://github.com/spantaleev/gitea-docker-ansible-deploy/blob/6934cb105670232d6ababb29c82d7c2f914831b8/requirements.yml#L15-L16
- https://github.com/spantaleev/gitea-docker-ansible-deploy/blob/6934cb105670232d6ababb29c82d7c2f914831b8/setup.yml#L28-L31
- https://github.com/spantaleev/gitea-docker-ansible-deploy/blob/6934cb105670232d6ababb29c82d7c2f914831b8/group_vars/gitea_servers#L93-L101

So.. moving `vars.yml` snapshotting into a new role that runs toward the end of the playbook is already easy. We just need to emulate what the Gitea playbook does. For Gitea and Nextcloud, I've now also moved this role more towards the end of the playbook:

- https://github.com/spantaleev/gitea-docker-ansible-deploy/commit/29559d9bdf1d44d01fd95bad4ef751f156c455a4
- https://github.com/spantaleev/nextcloud-docker-ansible-deploy/commit/a930e087e4584b6436be262bba00a781506e49f8

I'll start redoing the Matrix playbook (at least the `vars.yml` snapshotting part) the same way.

ikkemaniac:
sounds like a plan. Let me know if can do anything for the `git describe` task

spantaleev:
The [com.devture.ansible.role.vars_preserver](https://github.com/devture/com.devture.ansible.role.vars_preserver) has been superseded by a new role - [com.devture.ansible.role.playbook_state_preserver](https://github.com/devture/com.devture.ansible.role.playbook_state_preserver). This new role includes `vars.yml` preservation support as well as commit hash preservation support (your contribution from this PR).

Since c3a7237de702b4f4ec26, the playbook now includes this new `com.devture.ansible.role.playbook_state_preserver` role and removes the functionality from `matrix-base`.

The new role is included toward the bottom of playbook execution (only "playbook runtime messages") is below it, so it effectively saves the last-known-good-configuration.

NewOcean11:
galaxy what？u edited setup.yml so make setup to be failed at the beginning......

spantaleev:
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/CHANGELOG.md#2022-11-04

ikkemaniac:
can we have the playbook run `make roles`?

spantaleev:
I don't think so. The playbook will fail very early if it can't fine all roles that are referenced in the playbook file.

BryceChrispy:
I've seen "make roles" referenced a few places but absolutely zero guidance on how to run it. Someone fill me in here?

MirrorCY:
make roles
rm -rf roles/galaxy
ansible-galaxy install -r requirements.yml -p roles/galaxy/ --force
make: ansible-galaxy: No such file or directory
make: *** [Makefile:8: roles] Error 127

spantaleev:
Sounds like you don't have [ansible-galaxy](https://docs.ansible.com/ansible/latest/cli/ansible-galaxy.html) tool installed. Check some guide on the internet about how to install it.

It should come included with Ansible though. So maybe you are on an old version of Ansible which doesn't have it. Consider upgrading Ansible or running it in a container (see `docs/ansible.md`)

PR #2219, opened 2022-10-29
Always use resolver variable in nginx conf.d files

swarnat:
For some tests with this playbook I currently do, I need to add the _**matrix_nginx_proxy_http_level_resolver**_ variable within all nginx configurations.

The variable already exists, is defined [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/roles/matrix-nginx-proxy/defaults/main.yml#L506), but is not used in any files..

During a first test, it runs without any difference for result, because default value is same then current value and the empty "else" definition is not used, because all usages are also surrounded by **matrix_nginx_proxy_enabled** check.

### DIFF
--- a/roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-base-domain.conf.j2
+++ b/roles/matrix-nginx-proxy/templates/nginx/conf.d/matrix-base-domain.conf.j2
@@ -45,7 +45,7 @@ server {
 		location /.well-known/acme-challenge {
 			{% if matrix_nginx_proxy_enabled %}
 				{# Use the embedded DNS resolver in Docker containers to discover the service #}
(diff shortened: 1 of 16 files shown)


spantaleev:
Makes sense!

It should be noted that `matrix_nginx_proxy_http_level_resolver` is currently defined as:

```yaml
matrix_nginx_proxy_http_level_resolver: "{{ '127.0.0.11' if matrix_nginx_proxy_enabled else '' }}"
```

So it will have an awkward empty value if `matrix_nginx_proxy_enabled` is `false`.

This is not a problem though, because all `resolver {{ matrix_nginx_proxy_http_level_resolver }} valid=5s;` instances occur within a `{% if matrix_nginx_proxy_enabled %}` block.

PR #2222, opened 2022-10-31
Document login to Discord using token

array-in-a-matrix:


spantaleev:
Maybe these instructions belong to the **Usage** section below, although they sound fine in **Prerequisites**.

They're not very actionable though. You tell the user that they'll need to scan a QR code, but they don't tell them when and how it's going to happen (answer: `login-qr` command later, during **Usage**). About the token, it's the same thing - you tell the user how to retrieve it, but you don't tell them when and how to make use of it (answer: `login-token` command later, during **Usage**)

----

I suppose that you could mention these 2 login methods in Prerequisites and then describe each one in detail in **Usage**.

**Usage** could be split into **Logging in** and **Bridging** or something, and the **Logging in** section could have 2 subsections (**Via a QR code (recommended)** and **Via a token (not recommended - against the Discord ToS)**).

array-in-a-matrix:
Is using a token against TOS as using it for bridging might be considered selfbotting?

spantaleev:
I'm actually not sure.. Maybe both login methods are kind of the same in the end -- against the ToS.

array-in-a-matrix:
Yeah... 
It is still better to recommend using a QR code as leaking a token isn't ideal.

spantaleev:
I think a QR token is easier for people who already use the Discord app and can scan QR codes.

Logging in via a browser and retrieving the token using Developer Tools could be a little unpleasant to regular people.

array-in-a-matrix:
@spantaleev Is there anything else I should change?

spantaleev:
Now that I think about it, the `(not recommended)` wording may sound a bit too strong, but.. it's okay.

Merged! Thank you!

array-in-a-matrix:
I think it is still better to discourage getting the token because Discord can (and have) change their API to obscure or change where the token can be copied from. Also the user might accidentally paste the token in chat or somewhere public and cause them chaos.

PR #2227, opened 2022-11-02
Add matrix-bridge-mautrix-slack role

xangelix:
Add support for [mautrix-slack](https://github.com/mautrix/slack). Inherits greatly from the mautrix-discord role.


docs/configuring-playbook-bridge-mautrix-slack.md
```diff
@@ -0,0 +1,78 @@
...
+
+
+### Additional configuration
+
+There are some additional things you may wish to configure about the bridge.
+
+Take a look at:
+
+- `roles/matrix-bridge-mautrix-slack/defaults/main.yml` for some variables that you can customize via your `vars.yml` file
+- `roles/matrix-bridge-mautrix-slack/templates/config.yaml.j2` for the bridge's default configuration. You can override settings (even those that don't have dedicated playbook variables) using the `matrix_mautrix_slack_configuration_extension_yaml` variable
```
spantaleev, line 46: Maybe it's somewhere (and on `docs/configuring-playbook-bridge-appservice-slack.md`) here that we should mention that the `appservice-slack` bridge also uses the same bot username (`slackbot`) by default..

And that if one intends to enable both bridges at the same time, they probably need to change either `matrix_mautrix_slack_appservice_bot_username` (this role) or `matrix_appservice_slack_bot_name` (the `appservice-slack` role).

---

Instead of a notice (which adds to the wall-of-text and may be easily missed), a `fail` task in each of these roles may be **better**. It's a little bit against our rule of not referencing variables from other roles, but ..

**Untested** example (`roles/matrix-bridge-mautrix-slack/tasks/validate_config.yml`):

```yaml
- name: Fail if appservice-slack and mautrix-slack have conflicting bot usernames
  when: matrix_appservice_slack_enabled | default(False) | bool and matrix_mautrix_slack_appservice_bot_username == matrix_appservice_slack_bot_name | default ('')
  ansible.builtin.fail:
    msg: |
      The appservice-slack and mautrix-slack components are both enabled and use the same bot username ({{ matrix_mautrix_slack_appservice_bot_username }}), as per their default configuration, which causes a conflcit.
      To resolve the conflict, make one of these components use a different username.
      Consider either changing `matrix_mautrix_slack_appservice_bot_username` (the bot username for the mautrix-slack component) or `matrix_appservice_slack_bot_name` (the bot username for the appservice-slack component).
      We recommend that you change the username for the newly-added (and yet unused) component.
```

In `roles/matrix-bridge-appservice-slack/tasks/validate_config.yml` we'd need the same thing but with a different `when`:

```yaml
when: matrix_mautrix_slack_enabled | default(False) | bool and matrix_mautrix_slack_appservice_bot_username | default ('') == matrix_appservice_slack_bot_name
```

--

^^ We use ` | default(..)` above to allow the roles to not have a hard dependency on one another.

spantaleev: Alternatively, we could make the new role (`mautrix-slack`) use s a different username by default.

Zocker1999NET:
Also probably the role should be moved to `roles/custom` where the other roles live now after the last change

Zocker1999NET (commented):
Except of merging `setup.yml` manually & moving the role to `roles/custom`, these should be all changes required to make this role compatible to the current main branch.
At least I got the bridge up & running.

@xangelix Thanks for your work of migrating mautrix/slack into this playbook!

docs/configuring-playbook-bridge-mautrix-slack.md
xangelix: Good catch, added the checks. Do we want to additionally change the default username?

xangelix:
Thanks a ton @Zocker1999NET for the checklist, makes the migration a lot easier.

Unrelated to that change, were the slack puppets able to be created properly in your instance? I may have gotten configs messed up, it seems the bridge is only able to puppet my own account.

docs/configuring-playbook-bridge-mautrix-slack.md
spantaleev: Sorry for somehow missing your reply here. It's just now that I see it.

While I'd love for the playbook to be as conflict-free as possible by default (that is, you can enable and disable stuff as you wish at any time and not worry about conflicts), I'd also like to stick to upstream defaults as much as possible. Otherwise, we might get people reading the the [upstream mautrix-slack Authentication page](https://docs.mau.fi/bridges/go/slack/authentication.html) and trying to message `@slackbot`. If we've renamed it to `@mautrixslackbot`, we might get contacted by people saying "it's not working".

I'm not sure how many people will follow these upstream docs.. And how many will also disregard our own docs and warnings. Probably only a few, if at all?

Maybe deviating from upstream defaults is better in this case. If upstream wanted consistency, they should have avoided reusing names from other (older) projects.

If we go for this new default username, we could probably get rid of these conflict checks in both `validate_config.ym` files.

Zocker1999NET:
> Unrelated to that change, were the slack puppets able to be created properly in your instance? I may have gotten configs messed up, it seems the bridge is only able to puppet my own account.

@xangelix sadly I have a very inactive Slack, so I couldn't test if messages from the puppets are bridged ... But I have problems with DMs not being bridged.

docs/configuring-playbook-bridge-mautrix-slack.md
xangelix: I think in-line with upstream sounds like the better priority than user config as it is now then. Keeping the checks in and username as default.

spantaleev:
Do you think this works well enough to be merged now?

As for various problems with bridging.. I wonder if the bridge is fully-featured or stable enough for us to be recommending it over appservice-slack, etc.

shawnp0wers:
> As for various problems with bridging.. I wonder if the bridge is fully-featured 
> or stable enough for us to be recommending it over appservice-slack, etc.

For what it's worth, this bridge appears to function without the need for a slack-side app/api. Several of the Slack workspaces I need to bridge with don't allow app/api access. That might make this bridge worth any warts it might have.

(Full disclosure, I'm very new to Matrix, but I've been watching this pull request because of that nuance regarding this Slack bridge vs the others)

RoiArthurB:
I tried this on my personal instance, and so far, so good with the several slacks I bridged.

However, since this project moved to galaxy roles, some works have to be redone (I'd be happy to commit the modifications I did on this branch if you give me the access). 

Also, there's a small "bug" of opening plenty of "Slack Bridge Bot" rooms... Not sure what it is about so far (bug or just threads? 🤷🤷🤷)

xangelix:
@RoiArthurB were the above changes mentioned by Zocker after moving to galaxy roles not all of the needed ones?
also, did you test this role specifically, or the bridge?

RoiArthurB:
> @RoiArthurB were the above changes mentioned by Zocker after moving to galaxy roles not all of the needed ones?

Was enough, moving files and replace galaxy's project new variable (like the link to postgresql).

> also, did you test this role specifically, or the bridge?

This role precisely in my `matrix-docker-ansible-deploy` instance :)

xangelix:
Only seems to bridge the logged-in user both ways, other users not working.
Not sure why getting these errors / what's wrong with the localpart:
`Failed to send media message 2023-01-03 01:53:02.000469079 +0000 UTC to matrix: failed to ensure joined: failed to ensure registered: failed to POST /_matrix/client/v3/register: M_EXCLUSIVE (HTTP 400): Invalid user localpart for this application service.`

xangelix:
Scratch that, all is now in order--appears to be fully functional except for displaynames.

spantaleev:
What's wrong with display names? Is it an upstream problem or some misconfiguration? Do upstream report them as working?

xangelix:
Config properly sets these vars to:
```yml
  bot_displayname_template: '{{.Name}} (bot)'
  channel_name_template: '#{{.Name}}'
  displayname_template: '{{.RealName}} (S)'
```
However all the bridged users have their full Slack ID as the display names.
image

I posted in the mautrix slack matrix channel for now, but will post an issue upstream if I don't hear back.

xangelix:
I believe all the vars here "should" work.

https://github.com/slack-go/slack/blob/master/users.go

xangelix:
Submitted issue here: https://github.com/mautrix/slack/issues/9

xangelix:
Naming issue appears fixed in https://github.com/mautrix/slack/commit/f933a14343ad3bef3059bf64eec02f8a34d81461.
This merge request should be ready to go!

PR #2230, opened 2022-11-04
Keep Synapse worker definitions in vars.yml to avoid an extra file include on each playbook task

### LINKED ISSUE #895: ERROR! vars file roles/matrix-synapse/vars/workers.yml was not found
Hello,
if you run `setup.yml` playbook from different place (or symlink), it will fail, because vars file `roles/matrix-synapse/vars/workers.yml` has been imported directly to playbook, full error message:

```
ERROR! vars file roles/matrix-synapse/vars/workers.yml was not found
Could not find file on the Ansible Controller.
If you are using a module and expect the file to exist on the remote, see the remote_src option
```

In my case, i use different ansible project structure with symlink to `setup.yml`.

Could you move that file's contents to `defaults/main.yml` of matrix-synapse role, please?
Or is it some special case and that file must be included directly in playbook?

etkecc:
The diff between actual (manually adjusted) and freshly generated content:

```diff
-# These appear to be conditional and should not be enabled by default.
-# We need to fix up our workers-doc-to-yaml.awk parsing script to exclude them.
-# For now, they've been commented out manually.
-#  # Account data requests
-#  - ^/_matrix/client/(r0|v3|unstable)/.*/tags
-#  - ^/_matrix/client/(r0|v3|unstable)/.*/account_data
-#
-#  # Receipts requests
-#  - ^/_matrix/client/(r0|v3|unstable)/rooms/.*/receipt
-#  - ^/_matrix/client/(r0|v3|unstable)/rooms/.*/read_markers
-#
-#  # Presence requests
-#  - ^/_matrix/client/(api/v1|r0|v3|unstable)/presence/
+  # Account data requests
+  - ^/_matrix/client/(r0|v3|unstable)/.*/tags
+  - ^/_matrix/client/(r0|v3|unstable)/.*/account_data
+
+  # Receipts requests
+  - ^/_matrix/client/(r0|v3|unstable)/rooms/.*/receipt
+  - ^/_matrix/client/(r0|v3|unstable)/rooms/.*/read_markers
+
+  # Presence requests
+  - ^/_matrix/client/(api/v1|r0|v3|unstable)/presence/
 
   # User directory search requests
-  # Any worker can handle these, but we have a dedicated user_dir worker for this,
-  # so we'd like for other generic workers to not try and capture these requests.
-  # - ^/_matrix/client/(r0|v3|unstable)/user_directory/search$
+  - ^/_matrix/client/(r0|v3|unstable)/user_directory/search$
 
   # Additionally, the following REST endpoints can be handled for GET requests:
```

Looks like it works pretty good!

Tested on a server without matrix components (including synapse) and on a regular matrix server with synapse.

Fixes #895


spantaleev:
yamllint was complaining about some comments being indented incorrectly.

Instead of fixing it up manually (which the automation would undo.. until the automation gets fixed as well), I've decided to get rid of all the automation. It's a complicated to maintain and buggy awk thing, the generated output of which had to be fixed up manually later anyway.

I'm keeping tabs (RSS) on the `workers.md` file and will maintain the worker definitions as it changes.

PR #2235, opened 2022-11-05
stop gradle pacman error on ubuntu

drelephant:
On Ubuntu 18.04.6, when I run: `ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,ensure-matrix-users-created,start --ask-pass`

I get this error:
````
SSH password:
ERROR! couldn't resolve module/action 'community.general.pacman'. This often indicates a misspelling, missing collection, or incorrect module path.

The error appears to be in '/home/redacted/spant_matrix-docker-ansible-deploy/roles/custom/matrix-ma1sd/tasks/setup_install.yml': line 77, column 7, but may

    - name: Ensure gradle is installed for self-building (Archlinux)
      ^ here
````

This commit hopefully fixes that.

I'm not sure about the `update_cache: true` part.

### DIFF
--- a/roles/custom/matrix-ma1sd/tasks/setup_install.yml
+++ b/roles/custom/matrix-ma1sd/tasks/setup_install.yml
@@ -74,13 +74,8 @@
         msg: "Installing gradle on RedHat ({{ ansible_distribution }}) is currently not supported, so self-building ma1sd cannot happen at this time"
       when: ansible_os_family == 'RedHat'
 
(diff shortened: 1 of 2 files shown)


spantaleev:
It's generally faster without `update_cache`. Archlinux may be special here in that updating the cache may be better, but..

drelephant:
OMG my first ever pull request got merged!

PR #2241, opened 2022-11-06
Add ansible.posix collection as a requirement, latest release.

oscarcarlsson:
Otherwise roles/custom/matrix-synapse/tasks/import_media_store.yml won't run, taking the entire playbook with it.

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -14,3 +14,6 @@
 
 - src: git+https://github.com/devture/com.devture.ansible.role.playbook_runtime_messages.git
   version: f1c78d4e85e875129790c58335d0e44385683f6b
+
+- src: git+https://github.com/ansible-collections/ansible.posix.git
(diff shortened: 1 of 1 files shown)


spantaleev:
I wonder if we should be installing these collections ourselves or let the user's Ansible distribution do it. On Archlinux, the `ansible` package installs some 50+ collections already, so nothing else is necessary. On Alpinelinux (which handles our `devture/ansible` container, see `docs/ansible.md` for running Ansible in a container), it's the same - all these collections are included as part of the Ansible 6.5.0 distribution.

Us installing collections manually means we're overriding whatever is installed by the user on their system. We'd then need to maintain these collections (it's not just `ansible.posix`, I think) up to date. I don't know if these collections work on all versions of Ansible we support or if some releases work on newer versions only.

On the plus side, we'll know that all dependencies are installed and pinned to versions we know to work.

PR #2249, opened 2022-11-09
add command to download upstream roles according to 2022-11-04 changelog

mcnesium:
see https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/c3a7237de702b4f4ec26509cb1000f99051a8f83

### DIFF
--- a/inventory/scripts/ansible-all-hosts.sh
+++ b/inventory/scripts/ansible-all-hosts.sh
@@ -23,6 +23,9 @@ for host in "$root"/inventory/*.yml; do
     echo
 done
 
+# download upstream roles according to 2022-11-04 changelog
+make roles
+
 # run ansible on all captured passwords/hosts
 for host in "${!pws[@]}"; do
     ansible-playbook "$root"/setup.yml \


spantaleev:
I see the `ansible-all-hosts.sh` file as a wrapper to `ansible-playbook` - for running the playbook against multiple hosts.

I don't think it should be doing any prerequisite steps (installing role dependencies, etc) as part of this "run the playbook" job.

mcnesium:
Hm I guess you are right … :grimacing:

mcnesium:
For everyone who would have found this usefull, I have hacked yet another wrapper script that I simply called `update.sh`:

    #!/usr/bin/env bash
    #
    # Wrapper script for the entire update procedure including
    #   * updating git repo
    #   * downloading roles
    #   * running playbook
    #
    # This script first asks for the passwords for all the hosts and then pipes them
    # to the ansible playbook wrapper script, so that we don't need to wait for the
    # merging and downloading to finish, before being asked for the passwords.
    #
    
    # set playbook root path
    root=$(dirname "$(readlink -f "$0")")/../..
    
    # capture passwords for all hosts
    for host in "$root"/inventory/*.yml; do
        read -rp "sudo password for $(basename "$host"): " -s pw
        pipeinput+="$pw\n"
        echo
    done
    
    # merge upstream master branch
    git -C "$root" merge upstream/master
    
    # download upstream roles
    make -f "$root"/Makefile roles
    
    # run ansible-playbook on all hosts
    echo -e "$pipeinput" | bash "$root"/inventory/scripts/ansible-all-hosts.sh

@spantaleev do you think this is worth filing a merge request for?

spantaleev:
Calling it `update.sh` may cause people to think that that's what they need to use.. While in fact, it's a pretty niche thing. I don't know anyone who is using `ansible-all-hosts` even.

It's probably good to have this documentation around though.

---

I also fail to see why everyone seems so keen on:
- bundling `make roles` with the `ansible-playbook ..` command
- or asking for Ansible to run `make roles` for them

The next step would be to ask for "Ansible to even run `git pull`" for you, which.. your `update.sh` script seems to do ;)

A similar one would be "why can't Ansible install Ansible on my computer?"

I don't think Ansible should do any of this.. It makes sense to have Ansible just be Ansible and run the playbook, while it's on the user to:

- install Ansible
- have `git` on your computer or download the playbook in another way (ZIP)
- download the playbook the first time around and install its dependencies
- update the playbook (and its dependencies) when desired

---

It'd be great if we can have a `make` target that people could run easily and not have to worry about `make roles` and the latest invocation of the `ansible-playbook` command.

For example: `make setup-all`, which does `make roles` followed by `--tags=setup-all,ensure-matrix-users-created,start`.

However, there are a few problems with this:

- `make roles` is currently wasteful and shouldn't be called every single time. We need to make it only re-download roles if `requirements.yml` changed. We could `md5sum` the `requirements.yml` file each time we download roles. On next `make roles` invocations, we could compare the installed hash (`roles/galaxy/hash`) to the current `requirements.yml` file and only re-download if it's different

- some people need to run the playbook with various other arguments passed to it (`--ask-vault-pass`, `--ask-pass`, `--private-key`, etc). As far as I'm aware, one can't easily pass extra arguments to a `make` target and use them within the target, etc.

- people sometimes want to run other tags (not `--tags=setup-all,ensure-matrix-users-created,start`), so the `make setup-all` call would be useless to them.. If they run `ansible-playbook` directly, they may end up not being up to date on `make roles`.. And if we've simplified our documentation too much (to avoid mentioning the dreaded `make roles`), they may not even know of its existence. I think the Archlinux philosophy is better when it comes to technically difficult things. It's something we're trying to follow in our docs -- educating people of how and why things work a certain way, as opposed to keeping things as simple as possible (while keeping people ignorant and unable to help themselves in the future).

PR #2250, opened 2022-11-09
add how to integrate

felixx9:
describe, how to integrate a pad


spantaleev:
Maybe this should be named "How to use Etherpad widgets without an Integration Manager (like Dimension)".. to complement the other section that discusses how to use it with Dimension.

PR #2256, opened 2022-11-11
add matrix_hookshot_generic_enable_http_get option

alemairebe:
Allowing use of get request on the webhook config so specific hardware can trigger a hook.
https://matrix-org.github.io/matrix-hookshot/latest/setup/webhooks.html#configuration


spantaleev:
The `hookshot` role uses non-standard naming which is not consistent with the rest of the playbook. This faulty naming sems to have continued with this PR.

---

If we are to be consistent with the rest of the playbook, the variable should be named `matrix_hookshot_generic_enableHttpGet` instead of `matrix_hookshot_generic_enable_http_get`.

We should rename all other variables that don't comply with the naming at some point.

Also, using ` | to_json` in the YAML configuration is preferrable.

PR #2268, opened 2022-11-16
Signal Bridge Quickfix

madprogramer:
My journey to get all the bridges to work on Dendrite continues :)

Today, I noticed that some of the variables under [matrix-bridge-mautrix-signal](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/be166b6b378f1a63eeb16e583eacb447681a9c55) were blank!

image

Unfortunately, I am committing from the same branch I used 3 months ago and so it shows that "783 files" have changed, although I have really modified this one. If it's too risky to read through, can I ask you to at least commit yourself :)


spantaleev:
This PR should be rebased on top of our new commits. As it is right now, it's hard to read.

---

However, if the only thing that has changed is the part that you're showing on your screenshot (adjusting `matrix_mautrix_signal_homeserver_address` and `matrix_mautrix_signal_homeserver_domain`), you should note that:

- these variables are blank in `roles/custom/matrix-bridge-mautrix-signal/defaults/main.yml` itself
- .. but they are defined in the `group_vars/matrix_servers` file just like you're expecting: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/7a4918da15d7cc7263f1e254171169e361e57592/group_vars/matrix_servers#L564-L566

So changing the `roles/custom/matrix-bridge-mautrix-signal/defaults/main.yml` definitions would ultimately not change anything.

PR #2274, opened 2022-11-17
Sentry DNS setting in Jitsi jvb and jicofo

warrenbailey:
I've noticed that both the jvb and jicofo containers are no longer logging. This is due to the SENTRY_DNS environment value not being set correctly. I think they were a direct port from the docker compose file https://github.com/jitsi/docker-jitsi-meet
but when run as a systemctl service it doesn't work. 

The following is present in the logs:
```
Nov 17 14:51:54 ip-10-0-1-66.ec2.internal matrix-jitsi-jicofo[10706]: panic: render template error: template: /defaults/logging.properties:1:38: executing "/defaults/logging.properties" at <toBool>: error calling toBool: strconv.ParseBool: parsing "${JICOFO_SENTRY_DSN:-0}
Nov 17 14:51:54 ip-10-0-1-66.ec2.internal matrix-jitsi-jicofo[10706]: goroutine 1 [running]:
Nov 17 14:51:54 ip-10-0-1-66.ec2.internal matrix-jitsi-jicofo[10706]: main.templateExecute(0xc0000b44c0, 0x7fff4bf61d54, 0x1c)
Nov 17 14:51:54 ip-10-0-1-66.ec2.internal matrix-jitsi-jicofo[10706]: /Users/saghul/src/tpl/main.go:48 +0x3ca
Nov 17 14:51:54 ip-10-0-1-66.ec2.internal matrix-jitsi-jicofo[10706]: main.main()
```

Instead I've changed the jitsi sentry dns config to be configurable by ansible vars and I've tried to match how sentry dns configuration is named and set elsewhere in the playbook


roles/custom/matrix-jitsi/templates/jicofo/env.j2
```diff
@@ -27,7 +27,7 @@ JIGASI_SIP_URI
 JVB_BREWERY_MUC={{ matrix_jitsi_jvb_brewery_muc }}
 MAX_BRIDGE_PARTICIPANTS
 OCTO_BRIDGE_SELECTION_STRATEGY
-SENTRY_DSN="${JICOFO_SENTRY_DSN:-0}"
+SENTRY_DSN={{1 if matrix_jitsi_jicofo_sentry_dns else 0}}
```
spantaleev, line 30: Should this be a true/false value or the actual Sentry DSN address or something else?

warrenbailey: It's expecting it to be 0 or 1 which then gets converted to a boolean

warrenbailey: So basically it ends up setting this value: https://github.com/jitsi/docker-jitsi-meet/blob/c44c59e6cb5ccdadfe0e710b7c649545fb5904a8/jvb/rootfs/defaults/logging.properties#L1

I wonder if a better name would be "matrix_jitsi_jicofo_sentry_dns_enabled" with a default of "0"

spantaleev: I think that following the upstream names is more consistent with how we handle things in other roles. Would you revert [4146c48](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2274/commits/4146c48f6a2e71d1b0d3f58c767aea1b2f4f789c)?

warrenbailey: ok cool

warrenbailey:
@spantaleev reverted last commit as requested

roles/custom/matrix-jitsi/defaults/main.yml
```diff
@@ -205,6 +205,9 @@ matrix_jitsi_jicofo_component_secret: ''
 matrix_jitsi_jicofo_auth_user: focus
 matrix_jitsi_jicofo_auth_password: ''
 
+# Enable Sentry DSN
+matrix_jitsi_jicofo_sentry_dns: ''
```
spantaleev, line 209: Is this meant to be set by people to some connection string? At least that's my expectation when I see "DSN".

So, to enable the Sentry feature, one needs to set this to a non-empty string, right?

-----

Later on, we never seem to pass the connection string to Jitsi.. We only seem to pass `0` or `1` depending on this variable. How will Jitsi pick up the actual DSN connection string value then?

----

Your PR fixes the defaults, so that we'll be passing `SENTRY_DSN=0` to disable the Sentry feature (by default).

If people want to use Sentry though, they won't be able to do it, would they? I haven't confirmed this, but my understanding is that Jitsi expects either a string value (a valid Sentry DSN connection string) or `0` (to disable Sentry integration).

If that's the case, then maybe it's better to:

- set `matrix_jitsi_jicofo_sentry_dns` and `matrix_jitsi_jvb_sentry_dns` to `0` by default
- adjust the `env.j2` files to pass the `_sentry_dsn` value as is (no `if` stuff necessary)
- adjust the comments for these variables to say that people who wish to enable Sentry should override this and specify a valid Sentry DSN connection string value instead of this `0`

By the way, there's also a typo in the variable names (`_sentry_dns` -> `_sentry_dsn`).

spantaleev:
I've adjusted a few things! Thanks for fixing this up!

PR #2276, opened 2022-11-17
Enable location sharing in Element

NullIsNot0:
Element has a functionality for user to share a location on map, but it needs to be configured. I introduced new variable `matrix_client_element_location_sharing` which is `false` by default as you need to set up map tile server (which is out of scope of this project) to enable location sharing and provide map tile server.

https://element.io/blog/element-launches-e2ee-location-sharing/

image

image


roles/custom/matrix-client-element/tasks/validate_config.yml
```diff
@@ -8,6 +8,14 @@
   with_items:
     - "matrix_client_element_default_hs_url"
 
+- name: Fail if matrix_client_element_location_sharing is enabled, but matrix_client_element_location_map_tile_servers not defined
+  ansible.builtin.fail:
+    msg: >-
+      You need to define at least one map tile server.
+  when:
+    - matrix_client_element_location_sharing | bool
+    - matrix_client_element_location_map_tile_servers == []
```
spantaleev, line 17:
```suggestion
    - matrix_client_element_location_map_tile_servers | length == 0
```

roles/custom/matrix-client-element/defaults/main.yml
```diff
@@ -135,3 +135,21 @@ matrix_client_element_configuration_extension: "{{ matrix_client_element_configu
 # Holds the final Element configuration (a combination of the default and its extension).
 # You most likely don't need to touch this variable. Instead, see `matrix_client_element_configuration_default`.
 matrix_client_element_configuration: "{{ matrix_client_element_configuration_default | combine(matrix_client_element_configuration_extension, recursive=True) }}"
+
+# Enable Element Location sharing functionality
+# More info: https://element.io/blog/element-launches-e2ee-location-sharing/
+# How to host your own map tile server: https://matrix.org/docs/guides/map-tile-server
+matrix_client_element_location_sharing: false
```
spantaleev, line 142: Maybe `matrix_client_element_location_sharing_enabled` is a better (more consistent with the rest of the playbook) name for this.

roles/custom/matrix-client-element/tasks/setup_install.yml
```diff
@@ -69,6 +69,14 @@
     owner: "{{ matrix_user_username }}"
     group: "{{ matrix_user_groupname }}"
 
+- name: Ensure Element map style installed
+  ansible.builtin.copy:
```
spantaleev, line 73:
```suggestion
- name: Ensure Element map style installed
  when: matrix_client_element_location_sharing_enabled | bool
  ansible.builtin.copy:
```

Maybe we need an uninstall task as well

roles/custom/matrix-client-element/defaults/main.yml
```diff
@@ -135,3 +135,21 @@ matrix_client_element_configuration_extension: "{{ matrix_client_element_configu
...
+matrix_client_element_location_map_style: "{{ lookup('template', 'templates/style.json.j2') }}"
```
spantaleev, line 143: In the future we can consider adding a `matrix_client_element_location_map_style_extension`, which gets merged with the base template and provides a way for people to override some parts of the configuration easily. We use this approach in various roles.

See:

- `matrix_client_element_configuration_default`
- `matrix_client_element_configuration_extension_json`
- `matrix_client_element_configuration_extension`
- `matrix_client_element_configuration`

I wonder if we should rename `matrix_client_element_location_map_style` to `matrix_client_element_location_map_style_default` in preparation for this.

.. but.. if you'd like, you can keep things as they are. We probably shouldn't be overcomplicating things.

roles/custom/matrix-client-element/templates/style.json.j2
```diff
@@ -0,0 +1,18 @@
...
+    "layers": [
+        {
+            "id": "locallayer",
+            "source": "localsource",
+            "type": "raster"
+        }
+    ],
+    "sources": {
+        "localsource": {
+            "attribution": {{ matrix_client_element_location_map_attribution|to_json }},
```
spantaleev, line 11: It's more consistent if we name this `matrix_client_element_location_sharing_styles_sources_localsource_attribution`.

roles/custom/matrix-client-element/templates/style.json.j2
```diff
@@ -0,0 +1,18 @@
...
+            "tileSize": 256,
+            "tiles": {{ matrix_client_element_location_map_tile_servers|to_json }},
```
spantaleev, line 13: It's more consistent if we name this `matrix_client_element_location_sharing_styles_sources_localsource_tiles`.

spantaleev (changes requested):

roles/custom/matrix-client-element/defaults/main.yml
NullIsNot0: Renamed `matrix_client_element_location_sharing` to `matrix_client_element_location_sharing_enabled`.

roles/custom/matrix-client-element/tasks/setup_install.yml
NullIsNot0: If you set `matrix_client_element_location_sharing_enabled` to `false`, style file is not created on the next install. It's the same as with `config.json` file.

roles/custom/matrix-client-element/defaults/main.yml
NullIsNot0: Done. I have implemented this functionality for location sharing map style.

roles/custom/matrix-client-element/templates/style.json.j2
NullIsNot0: I have renamed it to `matrix_client_element_location_sharing_map_style_content_sources_localsource_tiles`.

roles/custom/matrix-client-element/templates/style.json.j2
NullIsNot0: I have renamed it to `matrix_client_element_location_sharing_map_style_content_sources_localsource_attribution`.

NullIsNot0 (commented):
All requested changes have been made.

roles/custom/matrix-client-element/templates/config.json.j2
```diff
@@ -36,6 +36,9 @@
 	"jitsi": {
         "preferredDomain": {{ matrix_client_element_jitsi_preferredDomain|to_json }}
     },
+	{% endif %}
+	{% if matrix_client_element_location_sharing_enabled %}
+	"map_style_url": "https://{{ matrix_server_fqn_element }}/style.json",
```
spantaleev, line 41: Can we rename this file to `map_style.json`? Seems like it's more self-explanatory and less likely to conflict with something else in the future.

NullIsNot0: Sure, I have renamed `style.json` to `map_style.json`. It works fine.
Also I found out that you can define `map_style_url` in /.well-known/matrix/client which takes precedence if is found there. And added the same url to `map_style.json` also there.

NullIsNot0:
Thanks for merging!

PR #2279, opened 2022-11-19
Restructure the list of services in the README.md

DennisCiba:
Hi,

as the list of services that this playbook manages is getting bigger and bigger, I find the unstructured list of installable services in the Readme to be increasingly harder to read. Additionally, some information in the list is superfluous (like marking every service as optional), while some are missing information like the link to their respective documentation.

In this PR I attempt to improve the situation by splitting the services into multiple categories and using a set of tables instead of an unordered list.

More precisely I changed these things: 

- Tried to split the list of services into distinct categories
- Use tables instead of a list
- Dropped "optional" flag, as every service is now optional anyways
- Added links to the documentation where docs exist and were missing before
- Split "Amazon S3" entry into two entries (Goofys and synapse-s3-storage-provider)
- Moved note about not needing all services to the top of the section

Note: I tried to keep the description of the services pretty much unchanged as not to change too much at once, which results in some entries (e.g. in the "Clients" section) having rather sparse and meaningless descriptions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,137 +15,144 @@ We run all services in [Docker](https://www.docker.com/) containers (see [the co
 
 ## Supported services
 
-Using this playbook, you can get the following services configured on your server:
+Using this playbook, you can get the following list of services configured on your server. Basically, this playbook aims to get you up-and-running with all the necessities around Matrix, without you having to do anything else.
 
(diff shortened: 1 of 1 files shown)


spantaleev:
This looks much better! Thank you!

PR #2290, opened 2022-11-23
Add dendrite captcha config to doc and hCaptcha

array-in-a-matrix:
Dendrite has support for Google's reCAPTCHA.

### DIFF
--- a/roles/custom/matrix-dendrite/defaults/main.yml
+++ b/roles/custom/matrix-dendrite/defaults/main.yml
@@ -101,6 +101,9 @@ matrix_dendrite_enable_registration_captcha: false
 matrix_dendrite_recaptcha_public_key: ""
 matrix_dendrite_recaptcha_private_key: ""
 matrix_dendrite_recaptcha_siteverify_api: "https://www.google.com/recaptcha/api/siteverify"
+matrix_dendrite_recaptcha_api_js_url: ""
+matrix_dendrite_recaptcha_form_field: ""
+matrix_dendrite_recaptcha_sitekey_class: ""
 
 # A list of additional "volumes" to mount in the container.
 # This list gets populated dynamically based on Dendrite extensions that have been enabled.
--- a/roles/custom/matrix-dendrite/defaults/main.yml
+++ b/roles/custom/matrix-dendrite/defaults/main.yml
@@ -97,13 +97,13 @@ matrix_dendrite_rate_limiting_cooloff_ms: 500
 matrix_dendrite_registration_disabled: true
 
 # reCAPTCHA API for validating registration attempts
-matrix_dendrite_enable_registration_captcha: false
-matrix_dendrite_recaptcha_public_key: ""
-matrix_dendrite_recaptcha_private_key: ""
-matrix_dendrite_recaptcha_siteverify_api: ""
-matrix_dendrite_recaptcha_api_js_url: ""
-matrix_dendrite_recaptcha_form_field: ""
-matrix_dendrite_recaptcha_sitekey_class: ""
+matrix_dendrite_client_api_enable_registration_captcha: false
+matrix_dendrite_client_api_recaptcha_public_key: ""
+matrix_dendrite_client_api_recaptcha_private_key: ""
+matrix_dendrite_client_api_recaptcha_siteverify_api: ""
+matrix_dendrite_client_api_recaptcha_api_js_url: ""
+matrix_dendrite_client_api_recaptcha_form_field: ""
+matrix_dendrite_client_api_recaptcha_sitekey_class: ""
 
 # A list of additional "volumes" to mount in the container.
 # This list gets populated dynamically based on Dendrite extensions that have been enabled.
--- a/roles/custom/matrix-dendrite/defaults/main.yml
+++ b/roles/custom/matrix-dendrite/defaults/main.yml
@@ -100,7 +100,7 @@ matrix_dendrite_registration_disabled: true
 matrix_dendrite_enable_registration_captcha: false
 matrix_dendrite_recaptcha_public_key: ""
 matrix_dendrite_recaptcha_private_key: ""
-matrix_dendrite_recaptcha_siteverify_api: "https://www.google.com/recaptcha/api/siteverify"
+matrix_dendrite_recaptcha_siteverify_api: ""
 matrix_dendrite_recaptcha_api_js_url: ""
 matrix_dendrite_recaptcha_form_field: ""
 matrix_dendrite_recaptcha_sitekey_class: ""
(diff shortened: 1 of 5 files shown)


array-in-a-matrix:
Dendrite also support's for [hCaptcha](https://github.com/matrix-org/dendrite/blob/main/dendrite-sample.monolith.yaml#L183), I'm not too sure how we can go about add this to the playbook.

array-in-a-matrix:
We could:
- Rename the existing variables from `recaptcha` to just `captcha` (i.e. `matrix_dendrite_captcha_siteverify_api`).
- Create new variables with `hcaptcha` (i.e. `matrix_dendrite_hcaptcha_siteverify_api`).

Regardless of which way we add this we would need to create new variables for these Dendrite options:

```yaml
recaptcha_api_js_url: "https://js.hcaptcha.com/1/api.js"
recaptcha_form_field: "h-captcha-response"
recaptcha_sitekey_class: "h-captcha"
```

The `recaptcha_siteverify_api` option is already in the playbook and defaults to:

```yaml
# default
matrix_dendrite_recaptcha_siteverify_api: "https://www.google.com/recaptcha/api/siteverify"

# to use hcaptcha instead
matrix_dendrite_recaptcha_siteverify_api: "https://hcaptcha.com/siteverify"
```

spantaleev:
The PR title says "Add hCaptcha", while this PR only does documentation. It doesn't seem to be adding support for any new variables, etc. Also, it seems to be adding ReCAPTCHA information, not anything related to hcaptcha..

So.. confusing :)

----

Otherwise, it seems like we can have support for both and people could enable whichever one they like. It doesn't hurt to document both (provided we have the right variables to toggle these options already).

array-in-a-matrix:
Sorry for the confusion, fixed title.

Should I rename the existing variable to add `hCaptcha`?

spantaleev:
Why rename the existing variable?

I thought Dendrite supports both and that we already have the config template + variables to control it for ReCAPTCHA.

We should keep it.

----

I suppose the better solution is to **add** the additional configuration for supporting hcaptcha to `roles/custom/matrix-dendrite/templates/dendrite/dendrite.yaml.j2` inspired from here: https://github.com/matrix-org/dendrite/blob/31f56ac3f43b692a56718fd1acfcb5a46e485c57/dendrite-sample.monolith.yaml#L183-L187

.. and to add a few additional variables for controlling these hcaptcha settings.

---

By default, all these captcha variables are set to an empty string, which tells Dendrite not to enable captcha.

roles/custom/matrix-dendrite/defaults/main.yml
```diff
@@ -101,6 +101,9 @@ matrix_dendrite_enable_registration_captcha: false
 matrix_dendrite_recaptcha_public_key: ""
 matrix_dendrite_recaptcha_private_key: ""
 matrix_dendrite_recaptcha_siteverify_api: "https://www.google.com/recaptcha/api/siteverify"
```
spantaleev, line 103: Maybe `matrix_dendrite_recaptcha_siteverify_api` should be defaulted to `''` to match the examples here: https://github.com/matrix-org/dendrite/blob/31f56ac3f43b692a56718fd1acfcb5a46e485c57/dendrite-sample.monolith.yaml#L183-L184 which say "keep it empty for reCAPTCHA".

roles/custom/matrix-dendrite/templates/dendrite/dendrite.yaml.j2
```diff
@@ -193,6 +193,9 @@ client_api:
   recaptcha_private_key: {{ matrix_dendrite_recaptcha_private_key|to_json }}
   recaptcha_bypass_secret: ""
   recaptcha_siteverify_api: {{ matrix_dendrite_recaptcha_siteverify_api|to_json }}
+  recaptcha_api_js_url: ""
+  recaptcha_form_field: ""
+  recaptcha_sitekey_class: ""
```
spantaleev, line 198:
```suggestion
  recaptcha_api_js_url: {{ matrix_dendrite_recaptcha_api_js_url | to_json }}
  recaptcha_form_field: {{ matrix_dendrite_recaptcha_form_field | to_json }}
  recaptcha_sitekey_class: {{ matrix_dendrite_recaptcha_sitekey_class | to_json }}
```

spantaleev (changes requested):

roles/custom/matrix-dendrite/defaults/main.yml
array-in-a-matrix: I see, thanks.

array-in-a-matrix:
@spantaleev should be good now.

spantaleev:
It looks good now. Actually, these variables (the old ones and the new ones that you've just introduced) under `client_api` should all be renamed to be more consistent with our naming.

It should be `matrix_dendrite_client_api_recaptcha_public_key`, instead of `matrix_dendrite_recaptcha_public_key` (note the `client_api` part in the name).

Renaming your new variables can be done without trouble even now - no one is using them yet.
If you'd like you can rename just those new variables and be done with it.

If you have time and would like to do things better - you can rename all other variables that you see under `client_api` in the `dendrite.yaml.j2` file. Doing that, we'd need to warn people by adding the old and new names of these variables to `roles/custom/matrix-dendrite/tasks/validate_config.yml`. There's already a **(Deprecation) Catch and report renamed settings** task for this in `validate_config.yml`. We just need to adjust its `with_items` part. For inspiration, you can look at how Synapse does it: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/34c01da9d2c7ac2d057506f68b221aab72dec746/roles/custom/matrix-synapse/tasks/validate_config.yml#L30-L61

array-in-a-matrix:
> It looks good now. Actually, these variables (the old ones and the new ones that you've just introduced) under `client_api` should all be renamed to be more consistent with our naming.
> 
> It should be `matrix_dendrite_client_api_recaptcha_public_key`, instead of `matrix_dendrite_recaptcha_public_key` (note the `client_api` part in the name).
> 
> Renaming your new variables can be done without trouble even now - no one is using them yet. If you'd like you can rename just those new variables and be done with it.
> 
> If you have time and would like to do things better - you can rename all other variables that you see under `client_api` in the `dendrite.yaml.j2` file. Doing that, we'd need to warn people by adding the old and new names of these variables to `roles/custom/matrix-dendrite/tasks/validate_config.yml`. There's already a **(Deprecation) Catch and report renamed settings** task for this in `validate_config.yml`. We just need to adjust its `with_items` part. For inspiration, you can look at how Synapse does it:
> 
> https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/34c01da9d2c7ac2d057506f68b221aab72dec746/roles/custom/matrix-synapse/tasks/validate_config.yml#L30-L61

would something like this work?

```yaml
    - {'old': 'matrix_dendrite_disable_tls_validation', 'new': '<superseded by matrix_dendrite_client_api_disable_tls_validation>'}
```

array-in-a-matrix:
I have renamed the old variables to add the new `client_api` part can you just double check if its all good?

spantaleev (approved):

spantaleev:
Some variables had been renamed, while they shouldn't have. I've fixed this up in [d2e0239](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2290/commits/d2e0239a1f19f6cbaf182e605736386aea392dc2).

Some variables were renamed in one or two places, but not in others throughout the playbook. I've fixed this up in [b2c4a39](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2290/commits/b2c4a390571123872c3db2db0f7f86416a670cf9).

Finally, I've added the validation logic in  cf7c83b and this is good to merge. Thank you!

PR #2293, opened 2022-11-24
Add service declaration to sample traefik config

jpdown:
Fixes the sample traefik config. Without these declarations, the routers will not be created, and traefik will log the following:
```
time="2022-11-24T00:27:09Z" level=error msg="Could not define the service name for the router: too many services" providerName=docker routerName=matrix-nginx-proxy
time="2022-11-24T00:27:09Z" level=error msg="Could not define the service name for the router: too many services" providerName=docker routerName=matrix-nginx-proxy-federation
```


spantaleev:
Thanks!

I'm using Traefik in a few places and I haven't had to define the service explicitly.

I suppose this came about with our recent reverse-proxying changes. We now have 2 different routers (`matrix-nginx-proxy` and `matrix-nginx-proxy-federation`) defined using labels on the same container (`matrix-nginx-proxy`). So it's probably getting confused somehow. Previously, the federation router was defined on the `matrix-synapse` container and no such confusion was arising.

PR #2306, opened 2022-11-30
Fix Docker installation typo in FAQ

MrAnno:


aaronraimist:
I don't think that's a typo. The question above that is "I already have Docker on my server. Can you stop installing Docker via the playbook?" so `false` seems appropriate.

aaronraimist:
Edit: Nevermind 🤦‍♂️

PR #2315, opened 2022-12-04
add prometheus-nginxlog-exporter role

ikkemaniac:
I ran into some issues one of my setups and needed a way to monitor nginx status codes, so i've put this together.

This role adds [prometheus-nginxlog-exporter](https://github.com/martin-helmich/prometheus-nginxlog-exporter) (nginxlog) container to the matrix deployment.

nginxlog listens for SYSLOG log messages on an udp port and converts them to prometheus type metrics which are served on the containers webserver at `/metrics`.
Have a look at [configuring-playbook-prometheus-nginxlog.md](https://github.com/ikkemaniac/matrix-docker-ansible-deploy/blob/4749e4a4bb5a4ab03640d560cbf47ef6fb1a71b0/docs/configuring-playbook-prometheus-nginxlog.md) for more info.

a summary of what this role does
    - pulls prometheus-nginxlog-exporter docker image
    - build prometheus-nginxlog-exporter config creating an SYSLOG udp socket and servering metrics via HTTP on `<container:port>/metrics`
    - adds SYSLOG type logging to both `matrix-nginx-proxy` and `matrix-synapse-reverse-proxy-companion`
    - adds configuration to `matrix-prometheus`
    - setup the 'standard' playbook's type of controlling the docker through a systemd service
    - copy custom dashboard to `matrix-grafana` if `matrix-grafana` is enabled
    - integrates with `matrix_nginx_proxy_proxy_matrix_metrics_enabled` to expose `/metrics/nginxlog`

Grafana dashboard:
grafana

prometheus-nginxlog-exporter metrics endpoint:
nginxlog-metrics

let me know what you think


roles/custom/matrix-nginx-proxy/templates/nginx/nginx.conf.j2
```diff
@@ -39,7 +39,18 @@ http {
 
 	{% if matrix_nginx_proxy_access_log_enabled %}
 	access_log /var/log/nginx/access.log main;
-	{% else %}
+	{% endif %}
+
+	{% if matrix_prometheus_nginxlog_exporter_enabled %}
```
spantaleev, line 44: We try to avoid referencing variables from other roles like this.

This should probably use `matrix_nginx_proxy_*` variables, which are defined in terms of `matrix_prometheus_nginxlog_exporter` variables via `group_vars/matrix_servers`.

Example `roles/custom/matrix-nginx-proxy/default/main.yml`:

```yaml
matrix_nginx_proxy_access_log_syslog_integration_enabled: false

matrix_nginx_proxy_access_log_syslog_integration_server: ''
matrix_nginx_proxy_access_log_syslog_integration_tag: 'matrix_nginx_proxy'
```

If `6514` is some port number, maybe we need (`matrix_nginx_proxy_access_log_syslog_integration_port` as well).

----

Then, from the `matrix-nginx-proxy` section in `group_vars/matrix_servers`, we can have something like this:

```yaml
matrix_nginx_proxy_access_log_syslog_integration_enabled: "{{ matrix_prometheus_nginxlog_exporter_enabled }}"

matrix_nginx_proxy_access_log_syslog_integration_server: "{{ matrix_prometheus_nginxlog_exporter_container_url if matrix_prometheus_nginxlog_exporter_enabled else '' }}"
```

roles/custom/matrix-prometheus-nginxlog-exporter/defaults/main.yml
```diff
@@ -0,0 +1,46 @@
+---
+# prometheus-nginxlog-exporter exports nginx logs into a prometheus usable format on a `/metrics/ endpoint
+# See: https://github.com/martin-helmich/prometheus-nginxlog-exporter/
+
+matrix_prometheus_nginxlog_exporter_enabled: true
+matrix_prometheus_nginxlog_exporter_container_url: 'matrix-prometheus-nginxlog-exporter'
```
spantaleev, line 6: Maybe `matrix_prometheus_nginxlog_exporter_container_hostname` is a better variable name?

roles/custom/matrix-synapse-reverse-proxy-companion/templates/nginx/nginx.conf.j2
```diff
@@ -39,10 +39,22 @@ http {
...
+	{% if matrix_prometheus_nginxlog_exporter_enabled %}
+	log_format prometheus_fmt 'matrix-synapse-reverse-proxy-companion $server_name - $upstream_addr - $remote_addr - $remote_user [$time_local] '
+					'$host "$request" '
+					'$status "$http_referer" '
+					'"$http_user_agent" "$http_x_forwarded_for"';
+
+	access_log syslog:server={{ matrix_prometheus_nginxlog_exporter_container_url }}:6514,tag=matrix_nginx_proxy_companion prometheus_fmt;
+	{% endif %}
+
+	{% if not matrix_synapse_reverse_proxy_companion_access_log_enabled and not matrix_prometheus_nginxlog_exporter_enabled %}
```
spantaleev, line 53: The same applies here - referencing variables from other roles is not something we encourage. See my other review comment for a detailed explanation and a proposed fix.

spantaleev:
Wow, this looks huge! I've given it a quick look and have provided some feedback

roles/custom/matrix-prometheus-nginxlog-exporter/defaults/main.yml
ikkemaniac: agreed, done

ikkemaniac:
btw, sounds like you're willing to accept this if things get fixed? lemme know otherwise I'll just leave it at is as this works for my own deployments.

spantaleev:
I think it could be useful to others, so let's fix it up and merge it

roles/custom/matrix-synapse-reverse-proxy-companion/templates/nginx/nginx.conf.j2
ikkemaniac: Done.

I've also update the additions to the prometheus role.

ikkemaniac:
Updated it.

Does this work for you?

roles/custom/matrix-prometheus-nginxlog-exporter/tasks/validate_config.yml
```diff
@@ -0,0 +1,31 @@
+---
+
+- name: Fail if Prometheus not enabled
+  ansible.builtin.fail:
+    msg: >
+      You need to enable `matrix_prometheus_enabled` as this roles requires Prometheus for data storage.
+  when: "not matrix_prometheus_enabled"
```
spantaleev, line 7: What if the user is using an external Prometheus server to scrape `/metrics/nginxlog`, etc.?

Perhaps we should remove this check, as it's not really a hard requirement.

ikkemaniac: fixed in 3acc2b7

ikkemaniac: but we can delete it all together. I prefer to leave it in, I believe quite a few will miss this otherwise and can't think of a downside for users who do.

spantaleev: I'm not sure I understand why you'd like to keep this in.

The downside is for users who use an external Prometheus server (hosted elsewhere, not on the Matrix server) and want to crawl their `https://matrix.DOMAIN/metrics/nginxlog`, along with other metrics from `/metrics/*`.

I doubt people would enable nginxlog-exporter and forget to add `matrix_prometheus_enabled`. The worst that could happen is them saying "I don't see the metrics in my Prometheus.. oh wait, I don't have a Prometheus".

ikkemaniac: > fixed in [3acc2b7](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/3acc2b7649b837a9d1babe30910c1ed834937dfe)

I guess you missed the commit 'above'.
The comment you reply should've been on this 'new commit' 3acc2b7. I think it's best to leave in with updated condition.
W/ the updated condition there's no downside.

Can remove if it's a hard 'no' for you.

spantaleev: Oh.. I saw that but didn't quite understood it. Now that I look at it again, it seems fine!

spantaleev:
Merged! Thank you!

PR #2316, opened 2022-12-04
fix s3 storage provider not ensuring data dir

qlyoung:
This path is accessed by the s3 storage provider stuff and needs to be ensured.

Broken by 7c5c3aedc


spantaleev:
Thanks for catching and fixing this! :bow:

PR #2319, opened 2022-12-06
matrix_ to devture_

felixx9:
I'm not sure, but this should be changed to devture_postgres_... !? 
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/CHANGELOG.md#2022-11-28

```yaml
matrix_postgres_process_extra_arguments: [
  "-c 'max_connections=200'"
]
```


spantaleev:
Sure! I've even fixed a few more in 9ab2a72e86f7ae0

PR #2323, opened 2022-12-07
Changes to allow a user to set the max participants on a jitsi conference

RhysRdm:
Hi Spantaleev,

I have added an option to set the max participants on a jitsi conference. The changes simply add another variable to the prosody env file based upon a user setting a specific INT variable in the playbook host-vars configuration file


spantaleev:
Thank you for your contribution!

`jitsi_max_participants` should be named `matrix_prosody_jitsi_max_participants` to follow playbook conventions (`matrix_` prefix, component name, name of setting that the variable controls)

PR #2324, opened 2022-12-07
Only delete playbook scripts from /usr/local/bin

thedanbob:
I had a script to run ansible in a docker container (similar to [this](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/ansible.md#running-ansible-in-a-container-on-the-matrix-server-itself)) that I put at the path `/usr/local/bin/matrix-install` to match the pattern of the other matrix maintenance scripts. It got deleted during my last update by [this new task](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/3d1ea3e79ea9a89c05d022ef6a75f2a9dc897a9b).

I probably should have thought through the name/location for my script a bit more, but this change will ensure the playbook only deletes scripts from the shared directory `/usr/local/bin` that it originally created.


spantaleev:
Makes sense! Thanks!

PR #2332, opened 2022-12-10
add 'not' to fail check for s3_storage_provider

mochman:
add 'not' to fail check for  matrix_synapse_ext_synapse_s3_storage_provider_config_endpoint_url


spantaleev:
Oh! Good catch

PR #2333, opened 2022-12-11
Use upstream Docker image for amd64 rather than self-build

### LINKED ISSUE #2294: Hydrogen image rebuilds every time (not idempotent)
Ref:
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5c3df22c3568371315585f386ecafc0edf4be714/roles/custom/matrix-client-hydrogen/tasks/setup_install.yml#L27-L69

1. "Ensure Hydrogen repository is present on self-build" has `force: "yes"`, which resets the working tree to the specified commit
2. "Ensure Hydrogen configuration installed" updates the working tree by adding `src/platform/web/assets/config.json`

(2) means that a subsequent run of the playbook *always* runs (1) and returns `changed: true`.

This in turn means that "Ensure Hydrogen Docker image is built" always runs (due to `force_source: "{{ matrix_client_hydrogen_git_pull_results.changed }}"`), taking up needless build time.

I'm not sure what the best solution here is.

mattcen:
I've just tested this, and confirmed that using the upstream Docker image for Hydrogen and bind-mounting in the config file works correctly. That is to say, it configures the Hydrogen client to use my own Matrix server etc.

I was a bit confused as to why I needed to change the case of the theme in the config file, but after doing so to make it match that of the file within the Docker image, it all seemed to behave itself.

Fixes #2294

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -2658,6 +2658,8 @@ matrix_client_element_jitsi_preferredDomain: "{{ matrix_server_fqn_jitsi if matr
 
 matrix_client_hydrogen_enabled: false
 
+matrix_client_hydrogen_container_image_self_build: "{{ matrix_architecture not in ['amd64'] }}"
+
 # Normally, matrix-nginx-proxy is enabled and nginx can reach Hydrogen over the container network.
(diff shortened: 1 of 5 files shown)


spantaleev:
Great! Thanks for saving us a lot of building time!

PR #2351, opened 2022-12-20
nginx sample config for fronting with own webserver

thigg:
This is the config i am using, it is working fine so far, but please be aware that i barely know what i am doing.

Would like to see this merged as I think this is actually quite common

Please feel free to edit this PR

### DIFF
--- a/docs/configuring-playbook-own-webserver.md
+++ b/docs/configuring-playbook-own-webserver.md
@@ -297,6 +297,7 @@ server {
         proxy_set_header Host $host;
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+		proxy_set_header X-Forwarded-Proto $scheme;
         client_max_body_size 0;
 
         # Websocket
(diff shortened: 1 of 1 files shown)


spantaleev:
If you're using nginx, you can just include the configuration generated by the playbook in `/matrix/nginx-proxy/conf.d`. Why are you trying to duplicate it manually?

thigg:
If this config is recommended, is this already mentioned in the documentation?

If one should just use the original proxy config, could this be done dynamically, thus allowing the playbook to update it?

spantaleev:
See: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-own-webserver.md#using-your-own-external-nginx-webserver

> Even with `matrix_nginx_proxy_enabled: false`, the playbook still generates some helpful files for you in `/matrix/nginx-proxy/conf.d`. Those configuration files are adapted for use with an external web server (one not running in the container network).
>
> You can most likely directly use the config files installed by this playbook at: `/matrix/nginx-proxy/conf.d`. Just include them in your own `nginx.conf` like this: `include /matrix/nginx-proxy/conf.d/*.conf`;

thigg:
Ah, I skipped section 1 because i read the section at the beginning:

> You should note, however, that the playbook's services work best when you keep using the integrated matrix-nginx-proxy webserver. 

as "it is easier to use option 2" and i thought, okay, then lets use option 2.

It looks like it might be beneficial to rephrase that part then?

spantaleev:
Ah, I see! I've tried to improve it.

Hopefully, d4a8435fa2f496 makes things clearer. If you've got other ideas, let me know!

thigg:
This is very good. thanks for helping, maintining this and being friendly :)

PR #2352, opened 2022-12-20
restructured introduction to own webserver deployments

thigg:
As a follow up to #2351.

This is a bit more opinionated (and maybe even wrong), but I just wrote it in the way I would have needed it.
For me, together with the changes in the first method this would have made the perfect match.


spantaleev:
Thanks! This PR makes things better.. However, it seems like the documentation was outdated.

There's no need to do separate things for worker and non-worker setups - since e9e84341a91fb00134, the playbook works equally well with both of these.

As such, I've simplified the documentation in 362954aeab35398e3cd6 by completely removing the warning.

PR #2366, opened 2022-12-28
bugfix: set the matrix public_baseurl

offbyone:
Instead of computing this from the `matrix_server_fqn_matrix` this
should use the pre-computed (or overridden!) `matrix_homeserver_url` so
that it is possible to set the homeserver port, if that isn't `443`


spantaleev:
The comments there seem to indicate that a trailing slash is necessary, but maybe that's not really required?

rltas:
Imo [this excerpt](https://spec.matrix.org/unstable/client-server-api/#well-known-uri) from the well-known spec applies here as well:
`It is important to note that the base_url value might include a trailing /. Consumers should be prepared to handle both cases.`

PR #2371, opened 2023-01-02
Added back variables which control which parts of the roles are run

Cadair:
These were the ones I ran into issues with, but I guess there are others I should add if this approach is acceptable.

---

We have discussed this a little on matrix, but it's easy for things to get lost in the busy room, so I will lay out what I am trying to achieve here to motivate this and then maybe we can figure out the best way forward.

My setup is that I don't use the `playbooks/matrix.yml` playbook, I include the roles into my own custom playbook. The playbook I run does all the setup and any restarting needed, meaning I never need to (or need to for any other ansible roles I use) set any tags to get my infrastructure in the state described by the inventory. 

I basically run `ansible-playbook site.yml -l matrix` and multiple playbooks are invoked, in order, which setup my matrix host and associated services (such as database roles on another host) to the state I expect them to be in.

The issue I see with having roles like the ones in this playbook, which *require* you to set tags for them to work, is that there is no way to set tags inside a playbook when invoking a role (at least using `role:` and `include_role` seems to have other issues relating to inventory [1]). So, unless I am missing something, there would be no way for me to avoid requiring setting tags on my invocation of `ansible-playbook`.

In my experience with using ansible for my homelab, it feels alien to be to have to use tags to make a role work (i.e. running all the tasks without tag filtering is illogical) and also alien to be to have to use tags to filter out related-but-optional functionality of a role (e.g. rust-state-compress). When I put a `role:` directive in my playbook I expect it to setup a thing based on the state of the inventory.

I am really grateful for this project, and all it makes super easy, and I appreciate that the way I use these roles isn't the expected way, but with the change of removing these `run_` variables, it's now become very difficult to impossible to include these roles in another playbook without also accepting that you need to use the CLI tags interface documented for the main `setup.yml` playbook in this repo, which is what I am trying to avoid having to do.

[1] It appears that using `include_role` doesn't mix the variables of all the roles together in the same way as `role:` so you end up with a load of missing variables making it effectively useless.

### DIFF
--- a/roles/custom/matrix-bridge-hookshot/defaults/main.yml
+++ b/roles/custom/matrix-bridge-hookshot/defaults/main.yml
@@ -2,6 +2,9 @@
 
 # A bridge between Matrix and multiple project management services, such as GitHub, GitLab and JIRA.
 # Project source code URL: https://github.com/matrix-org/matrix-hookshot
+#
+run_reset_encryption: true
+run_setup: true
 
 matrix_hookshot_enabled: true
 
--- a/roles/custom/matrix-bridge-hookshot/tasks/main.yml
+++ b/roles/custom/matrix-bridge-hookshot/tasks/main.yml
@@ -3,7 +3,7 @@
 - tags:
     - reset-hookshot-encryption
   block:
-    - when: matrix_hookshot_enabled | bool
+    - when: matrix_hookshot_enabled | bool and run_reset_encryption | bool
       ansible.builtin.include_tasks: "{{ role_path }}/tasks/reset_encryption.yml"
 
 - tags:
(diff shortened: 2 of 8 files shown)


Cadair:
@spantaleev Do you have any thoughts on this? Perpetually maintaining this branch, especially now with the same issue appearing in external roles like ntfy is making using these roles pretty untenable for me.

(It feels especially weird for standalone roles like ntfy to have to use tags).

spantaleev:
I'm not sure how this can be solved without adding such run variables to all roles like this.

Most of matrix-docker-ansible-deploy's roles are now published separately and also included in [mash-playbook](https://github.com/mother-of-all-self-hosting/mash-playbook), which for better or worse, also uses tags in the same way.

PR #2375, opened 2023-01-04
Matrix Authentication Support for Jitsi

### LINKED ISSUE #678: [Question] Is Jitsi OpenID Authentication via Matrix integrated?
It appears that Jitsi supports OpenID authentication, at least for widgets. The PR I'm referencing is here: https://github.com/vector-im/element-web/pull/15114 and the documentation/extra service required is here: https://github.com/matrix-org/prosody-mod-auth-matrix-user-verification

My question is: has openid auth via such a prosody mod been integrated into this build? Would it be possible to do so? My main goal is to be able to not have to manage jitsi users separately and instead pass auth to matrix so I can lock down jitsi to valid users.

I ask this question as I'm trying to find out if 1) it's been integrated here and 2) if it's worth building out an integration PR (I'm new to matrix, but know ansible).

jakicoll:
# Matrix Authentication Support for Jitsi

This extends the collection with support for seamless authentication at the Jitsi server using Matrix OpenID.

A minimal configuration to enable it would be:
```yaml
matrix_jitsi_enable_auth: true
matrix_jitsi_auth_type: "matrix"
matrix_user_verification_service_uvs_access_token: "$SOME_ADMIN_ACCESS_TOKEN"
```

## Motivation

Currently, users self-hosting Jitsi using this playbook collection can run their Jitsi server either withouth any access control or setup additional Jitsi accounts. While the former allows anybody that knows your Matrix room ids to access your conferences, the latter introduces overhead for managing additional credentials and disturbes the user experience by asking for extra credentials. Using this technique, we can have both, restriced access withouth additional credentials. 

Here is how it works: On joining a conference, Element will ask the user if the conference widged may authenticate them (see screenshot below). It then request an OpenID token from the matrix server (see [MSC1960](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1960-integrations-openid.md) ) and passes it to the Jitsi widget.

The Jitsi server can then ask the [Matrix User Verification Service](https://github.com/matrix-org/matrix-user-verification-service) if the submitted token is valid and if the user  is a member of the Matrix room the conference belongs to. This is done by a prosody plugin called [mod-auth-matrix-user-verification](https://github.com/matrix-org/prosody-mod-auth-matrix-user-verification). For a detailed view of the authentication flow, please refer to the diagram of that mod [here](https://github.com/matrix-org/prosody-mod-auth-matrix-user-verification/blob/2839499cb03894d8cfc3e5b2219441427cb133d8/widget_load.png).

## Implementation

Unless explicitly configured, the changes from this pull request have no effect on running or new installations.
 
This PR introduces the following changes:

1. New role for installing the [Matrix User Verification Service](https://github.com/matrix-org/matrix-user-verification-service)
2. Changes to Jitsi role: Installing Jitsi Prosody Mods and configuring Jitsi Auth
3. Changes to Jitsi and nginx-proxy roles: Serving .well-known/element/jitsi from jitsi.DOMAIN
4. We updated the Jitsi documentation on authentication and added documentation for the user verification service.

### Matrix User Verification Service

A new role installs the service using the docker images from `matrixdotorg/matrix-user-verification-service`. We adopted this playbook collection's approach of service setup and configuration.

By default the matrix-user-verification-service uses port 3000 over the docker network. If Jitsi is not managed by this collection and the user-verification service is manually enabled, port 3000 will be exposed on 127.0.0.1. However port 3000 may clash with grafana. We have elected to give grafana the right of way and change the port to 3003, if grafana is also enabled.

### Jitsi role

We added the installation of mod-auth-matrix-user-verification to the jitsi role in setup_jitsi_prosody_*. It clones a pinned version of the mod's repository to the server and copies the mod's files into the directory for prosody mods.

The already implemented 'internal' auth requires a running prosody server to set-up users. In the file [roles/custom/matrix-jitsi/tasks/util/setup_jitsi_auth.yml](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/f400093865e5e140c4dd3d25ebd51e90429ad601/roles/custom/matrix-jitsi/tasks/util/setup_jitsi_auth.yml), where this implemented, we found some comments indicating other authentication mechanisms should be implemented there as well. However, the 'matrix' machansism does not require Prosody to already run during set up. We have therefore refactored the implementation into a 'setup_jitsi_prosody_post_setup_hooks.yml', where tasks, that require a running prosody, may reside. These are not limited to authentication tasks, even though currently only the 'internal' authentication tasks reside there.

The post_setup_hook may require the prosody container be restarted (i.e. if the auth mechanism is to be changed from 'matrix' to 'internal'). We therefore introduced `matrix_jitsi_prosody_require_restart` as a flag to indicate such need.
This flag indicates changes, that require prosody to restart either before the post_setup_hooks (if they run) or after the install task is done, were made.

The Auth Flow requires GET parameters to be passed with the XMPP-Websocket request (/xmpp-websocket). We had to change the proxy configuration to allow this (matrix-jitsi.conf).
```diff
- proxy_pass $backend/xmpp-websocket;
+ proxy_pass $backend$request_uri;
```

In this regard, we were unsure, whether `$request_uri` or `$is_args$args` would be better suited.

### Jitsi Well-Known File

In order to request authentication, a file containing `{"auth": "openidtoken-jwt"}` needs to be served from https://jitsi.domain/.well-known/element/jitsi

We propose to add it to the existing 'static-files' directory (`matrix_static_files_base_path`) and create the file during the jitsi role.

This approach does come with some drawbacks:

* Unlike the matrix .well-known files, which are created in the 'matrix-base' role (where matrix_static_files_base_path is defined), the `.well-known/element/jitsi` file is managed from the jitsi role. We found this more fitting, since the jitsi role should be the 'configuration-owner' for the files content and also decide, whether to create/skip/delete it. **Drawback: the  jitsi role requires knowledge of `matrix_static_files_base_path`**, since the nginx-proxy role also relies on this knowlege, we found this to be ok.

* This also means all other domains (that server static files) will serve this file. (e.g. element.DOMAIN, matrix.DOMAIN ...) We don't consider that harmful. In a future pull request we would like to propose to further subdivide the 'static-files' directory. As this would require changes even for running installations (i.e. moving the current well-known files, changing the Nginx config, ..) we omited implementing this.

## Caveats

We are using a similiar setup on our production matrix server since +1y and confirmed this implementation to work on a fresh test server.

**Please note these caveats:**
- Not tested with clients other than Element and servers other than Synapse.
- This might breaks jitsi conferences for federated rooms. We did not test that. However, it's of course still possible to embed conferences from another Jitsi server in such rooms.
- We have not implemented self_build and removal for Matrix User Verification Service
- If one were to enable guest mode, the intened security collapses. I.e. any external party with knowlege of room names can join into running conferences without any authentication. In our test no lobby was enabled and we could join rightaway.
It seems possible to configure the requirement for a waiting-room by default (https://community.jitsi.org/t/how-to-how-do-i-use-the-new-lobby-feature/73100). We think, implementing this as a default option, when guest mode is enabled, could be beneficial.

## Additional Resources

**Implementation in Element**: https://github.com/vector-im/element-web/issues/7153
**MSC1960** seeking to standardise the implementation: https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1960-integrations-openid.md
**User Verification Service**: https://github.com/matrix-org/matrix-user-verification-service
**Prosody Mod auth-matrix-user-verification**: https://github.com/matrix-org/prosody-mod-auth-matrix-user-verification

## Linked issues
This fixes #678.


group_vars/matrix_servers
```diff
@@ -3166,3 +3176,52 @@ matrix_user_creator_users_auto: |
...
+
+
+######################################################################
+#
+# matrix-user-verification-service
+#
+######################################################################
+
+## FIXME: Needs to be updated when there is a proper release by upstream.
+matrix_user_verification_service_docker_image: "{{ matrix_user_verification_service_docker_image_name_prefix }}matrixdotorg/matrix-user-verification-service@sha256:d2aabc984dd69d258c91900c36928972d7aaef19d776caa3cd6a0fbc0e307270"
```
spantaleev, line 3188: Why not put these things in the role itself? Currently, the role has some version defined... which has no chance of working right now.. And then we patch it up here to make it work.

We'd better keep this in the role.

group_vars/matrix_servers
```diff
@@ -3166,3 +3176,52 @@ matrix_user_creator_users_auto: |
...
+        (['matrix-synapse.service'] if matrix_synapse_enabled else [])
+        +
+        ([devture_postgres_identifier ~ '.service'] if devture_postgres_enabled else [])
+      }}
+
+# If Jitsi is managed by this playbook we can use the docker network - no need to expose a port.
+# If Jitsi is not managed by this playbook, or you otherwise have a need for it, you can expose
+# matrix-user-verfification-services's client-server port to the local host.
+# Note: If grafana is also enabled, the exposed port is changed to 3003.
+matrix_user_verification_service_container_http_host_bind_port: "{{  '' if (matrix_jitsi_enabled | bool and matrix_jitsi_enable_auth | bool and matrix_jitsi_auth_type == 'matrix') else '127.0.0.1:' + ('3003' if matrix_nginx_proxy_proxy_grafana_enabled else '3000') }}"
```
spantaleev, line 3205: I think it's better to pick some free port for this and always use that one instead of trying to avoid Grafana, etc. We can pick 3003 and stick to that in all cases. It'd be nice to leave a comment as to why we're picking 3003 and not 3000.

group_vars/matrix_servers
```diff
@@ -3166,3 +3176,52 @@ matrix_user_creator_users_auto: |
...
+
+# URL exposed in the docker network
+matrix_user_verification_service_container_url: "http://{{  matrix_user_verification_service_container_name }}:3000"
+
+# Set the homeserver URL to the container name if synapse is managed by this collection
+matrix_user_verification_service_uvs_homeserver_url: "{{ matrix_homeserver_container_url if matrix_synapse_enabled }}"
```
spantaleev, line 3211: An `if` without an `else` makes certain (most? all?) Ansible/Jinja versions complain.

Mentioning synapse in the comment is not necessary. `matrix_homeserver_container_url` points to the homeserver, whatever it is. It may be Synapse or Dendrite or Conduit or something else in the future.

----

Most roles can safely assume `matrix-base` is around and can just use `matrix_homeserver_container_url` inside, so that `group_vars/matrix_servers` is smaller, but feel free to keep things more decoupled (like this) for this role.

group_vars/matrix_servers
```diff
@@ -3166,3 +3176,52 @@ matrix_user_creator_users_auto: |
...
+# If synapse is managed by this collection, we will connect via docker network, which is a private ip.
+# Therefore we need to disable IP checks
+matrix_user_verification_service_uvs_disable_ip_blacklist: "{{'true' if matrix_synapse_enabled else 'false'}}"
```
spantaleev, line 3214:
```suggestion
# We connect via the container network (private IPs), so we need to disable IP checks
matrix_user_verification_service_uvs_disable_ip_blacklist: true
```

docs/configuring-playbook-user-verification-service.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+In case Jitsi is also managed by this collection and 'matrix' authentication in Jitsi is enabled, this collection will automatically configure Jitsi to use the configured auth token.
+
+### (Optional) Federation
+
+In theory (however currently untested), UVS can handle federation. Simply set:
+
+```yaml
+matrix_user_verification_service_uvs_openid_verify_server_name: ~
```
spantaleev, line 84: This `matrix_user_verification_service_uvs_openid_verify_server_name` is somewhat messy.

Looks like it's not defined in `defaults/main.yml`, yet... `group_vars/matrix_servers` defines it, so it's always defined and various `matrix_user_verification_service_uvs_openid_verify_server_name is defined` checks will always pass.

`group_vars/matrix_servers` defines it in via some `if` statement, which has the chance to set `matrix_user_verification_service_uvs_openid_verify_server_name` to an empty string, but can never really set it to `null` due to how Ansible/Jinja if statements work.

Then there's this documentation here telling people to hardcode it as `null`. It would work, but.. the whole thing is messy.. undefined in the role, checks for undefined which will pretty much never work, possibly set to empty string, telling people about `null`.

It's better if `matrix_user_verification_service_uvs_openid_verify_server_name` is defined in `defaults/main.yml` and all this documentation is there (at least). We can then still keep the documentation here as well, but we can at least drop the `is defined` checks.

I think you can probably even make drop the `null` thing and hardcode `matrix_user_verification_service_uvs_openid_verify_server_name: "{{ matrix_domain }}"` in `defaults/main.yml` and clean up `group_vars/matrix_servers`.

Then your checks can be.. `if matrix_user_verification_service_uvs_openid_verify_server_name | length > 0` or `if matrix_user_verification_service_uvs_openid_verify_server_name != ''`

docs/configuring-playbook-jitsi.md
```diff
@@ -53,20 +63,35 @@ matrix_jitsi_prosody_auth_internal_accounts:
...
 
-The default authentication mode of Jitsi is `internal`, however LDAP is also supported. An example LDAP configuration could be:
+Using this authentication type require a [Matrix User Verification Service](https://github.com/matrix-org/matrix-user-verification-service).
+By default, this playbook creates and configures a user-verification-service to run locally, see [configuring-user-verification-service](configuring-playbook-user-verification-service.md).
+
+To enable set this configuration at host level:
+
+```yaml
+matrix_jitsi_enable_auth: true
+matrix_jitsi_auth_type: "matrix"
```
spantaleev, line 81: Instead of making `matrix_user_verification_service_enabled` try to be enabled in a smart way based on these few variables, it's better to be more explicit and just tell people here to add `matrix_user_verification_service_enabled: true` as well.

Then `group_vars/matrix_servers` can just have a default of `matrix_user_verification_service_enabled: false` (overriding the `true` value from the role).

docs/configuring-playbook-user-verification-service.md
```diff
@@ -0,0 +1,116 @@
+# Setting up Matrix User Verification Service (optional)
+
+**[Matrix User Verification Service](https://github.com/matrix-org/matrix-user-verification-service) (hereafter: UVS) can only be installed after Matrix services are installed and running.**
+If you're just installing Matrix services for the first time, please continue with the [Configuration](configuring-playbook.md) / [Installation](installing.md) flow and come back here later.
+
+Currently, the main purpose of this role is to allow Jitsi to authenticate matrix users and check if they are authorized to join a conference. Please refer to the documentation of the [Matrix User Verification Service](https://github.com/matrix-org/matrix-user-verification-service) to understand how it works.
+
+**Note**: enabling Matrix User Verification Service, means that the `openid` API endpoints will be exposed on the Matrix Federation port (usually `8448`), even if [federation](configuring-playbook-federation.md) is disabled.
+
+If the Jitsi server is also configured by this collection, all plugging of variables and secrets is handled in `group_vars/matrix_servers`.
```
spantaleev, line 10:
```suggestion
If the Jitsi server is also configured by this playbook, all plugging of variables and secrets is handled in `group_vars/matrix_servers`.
```

We're calling this a playbook everywhere, so I imagine 99% of people will be thinking "what collection?" when they see this.

There are many more instances of this in your changeset.

docs/configuring-playbook-user-verification-service.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Configuration
+
+The only required configuration variable is `matrix_user_verification_service_uvs_access_token` (see below).
+
+For a list of all configuration options see the role defaults [`roles/matrix-user-verification-service/defaults/main.yml`](../roles/custom/matrix-user-verification-service/defaults/main.yml).
+But be aware of all the plugging happening in `group_vars/matrix_servers`.
+
+In the default configuration, the UVS Server is only reachable via the docker network, which is fine if e.g. Jitsi is also running in a container on the host.
+However, it is possible to expose UVS via setting `matrix_user_verification_service_container_http_host_bind_port`. Be aware that the normally used port (3000) may collide with Grafana.
```
spantaleev, line 48: We'll be able to simplify this default port thing once we start using 3003 or whatever free port you pick.

group_vars/matrix_servers
```diff
@@ -3166,3 +3176,52 @@ matrix_user_creator_users_auto: |
...
+
+matrix_user_verification_service_uvs_auth_token: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'uvs.auth.token', rounds=655555) | to_uuid }}"
+
+# Pin UVS to only check openId Tokens for the matrix_server_name configured by this collection.
+# This is not the homeserverURL, but rather the domain in the matrix "user ID"
+matrix_user_verification_service_uvs_openid_verify_server_name: "{{ matrix_domain }}"
+matrix_user_verification_service_uvs_log_level: warning
```
spantaleev, line 3221: Why not use a sane log level default in the role itself and avoid messing with it here?

roles/custom/matrix-user-verification-service/tasks/setup_install.yml
```diff
@@ -0,0 +1,42 @@
...
+
+#- block:
+# TODO
+#  when: "matrix_user_verification_service_container_image_self_build|bool"
+
+- name: write env file
+  ansible.builtin.template:
+    src: "{{ role_path }}/templates/.env.j2"
+    dest: "{{ matrix_user_verification_service_config_env_file }}"
+    mode: 0644
```
spantaleev, line 35:
```suggestion
- name: Ensure Matrix User Verification Service env file installed
  ansible.builtin.template:
    src: "{{ role_path }}/templates/.env.j2"
    dest: "{{ matrix_user_verification_service_config_env_file }}"
    owner: "{{ matrix_user_username }}"
    group: "{{ matrix_user_groupname }}"
    mode: 0644
```

roles/custom/matrix-user-verification-service/tasks/setup_uninstall.yml
```diff
@@ -0,0 +1,35 @@
...
+
+- name: Ensure Matrix user-verification-service paths don't exist
+  file:
+    path: "{{ matrix_user_verification_service_base_path }}"
+    state: absent
+
+- name: Ensure user-verification-service Docker image doesn't exist
+  docker_image:
+    name: "{{ matrix_user_verification_service_docker_image }}"
+    state: absent
```
spantaleev, line 35:
```suggestion
- name: Check existence of matrix-user-verification-service service
  stat:
    path: "{{ matrix_systemd_path }}/{{ matrix_user_verification_service_systemd_service_name }}"
  register: matrix_user_verification_service_service_stat

- when: matrix_user_verification_service_service_stat.stat.exists | bool
  block:
    - name: Ensure matrix-user-verification-service is stopped
      service:
        name: "{{ matrix_user_verification_service_systemd_service_basename }}"
        state: stopped
        daemon_reload: yes
      register: stopping_result

    - name: Ensure matrix-user-verification-service.service doesn't exist
      file:
        path: "{{ matrix_systemd_path }}/{{ matrix_user_verification_service_systemd_service_name }}"
        state: absent

    - name: Ensure Matrix user-verification-service paths don't exist
      file:
        path: "{{ matrix_user_verification_service_base_path }}"
        state: absent
```

Using a `block` helps with readability and we don't have to repeat the same `when` over and over again.

Reloading systemd is intentionally skipped (like it is in all / most other roles) - we only aim to reload during things like `--tags=start` in the systemd service manager role.

Deleting the Docker image is also skipped, because it's mostly useless (normally, it changes often.. and cleaning up just 1 out of 20 (for example) doesn't help much).

roles/custom/matrix-user-verification-service/templates/.env.j2
```diff
@@ -0,0 +1,14 @@
...
+
+{% if matrix_user_verification_service_uvs_auth_token is defined and matrix_user_verification_service_uvs_auth_token|length %}
+  UVS_AUTH_TOKEN={{ matrix_user_verification_service_uvs_auth_token }}
+{% endif %}
+{% if matrix_user_verification_service_uvs_openid_verify_server_name is defined and matrix_user_verification_service_uvs_openid_verify_server_name|length %}
+  UVS_OPENID_VERIFY_SERVER_NAME={{ matrix_user_verification_service_uvs_openid_verify_server_name }}
+{% endif %}
+{% if matrix_user_verification_service_uvs_log_level is defined and matrix_user_verification_service_uvs_log_level|length %}
+  UVS_LOG_LEVEL={{ matrix_user_verification_service_uvs_log_level }}
+{% endif %}
```
spantaleev, line 13: Why `is defined`? Let's ensure all variables are defined in the role itself and have sane values.

`|length` checks are probably also unnecessary in at least a few of these (e.g. `matrix_user_verification_service_uvs_log_level`). Let's set a sane default value in the variable and just use it here without any `if`s at all.

Maybe this is also applicable to `matrix_user_verification_service_uvs_auth_token`.

I understand that for `matrix_user_verification_service_uvs_openid_verify_server_name` we may need some if (discussed in some other PR comments).

roles/custom/matrix-user-verification-service/defaults/main.yml
```diff
@@ -0,0 +1,81 @@
...
+# Enable by default. This is overwritten in provided group vars.
+matrix_user_verification_service_enabled: true
+
+# Fix version tag
+matrix_user_verification_service_version: "v2.0.0"
+
+# Paths
+matrix_user_verification_service_base_path: "{{ matrix_base_data_path }}/user-verification-service"
+# We need the docker src directory to be named user_verification_service. See: https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/588
+matrix_user_verification_service_docker_src_files_path: "{{ matrix_user_verification_service_base_path }}/docker-src/user-verification-service"
```
spantaleev, line 14: Seems like this has been copied from the `ma1sd` role and it likely doesn't apply here at all.

We can fix it up to be more consistent with other roles.

roles/custom/matrix-user-verification-service/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+
+- name: verify all necessary variables are present
+  assert:
+    that:
+      - matrix_user_verification_service_uvs_access_token is defined and matrix_user_verification_service_uvs_access_token|length
+      - matrix_user_verification_service_uvs_homeserver_url is defined and matrix_user_verification_service_uvs_homeserver_url|length
+    fail_msg: "Missing variable in {{ matrix_user_verification_service_ansible_name }} role"
```
spantaleev, line 8: It's better to move these validation checks into a `validate_config.yml` file here, like we do for other roles.

Also, the `validate_config.yml` tasks for the role should only run when the role is enabled (`matrix_user_verification_service_enabled`).

Dropped `is defined` checks (as discussed in other PR comments) would be better. All variables referenced by roles should have default values in `defaults/main.yml`. A variable being undefined should not be a possibility - it makes code like this ugly and discovery of variables hard. All variables should be defined and should have documentation around them showing what the default value is (and potentially why), what values they could take, what variables are related to them, links to upstream documentation (if applicable and useful), etc.

It's also a good time to think of any validation checks were forgotten to be added here.

I got the impression that this the matrix user verification service can only work with Synapse, not with other homeserver implementations? If so, it may make sense to add some check for `matrix_homeserver_implementation` here. If `matrix_homeserver_implementation != 'synapse'`, tell the user. This variable is part of `matrix-base` and we generally consider it OK to reach into `matrix-base` variables from any role. If you'd like to keep things more decoupled and pass it around through another variable (via `group_vars/matrix_servers`), that's fine too - it just makes things more complicated.

group_vars/matrix_servers
stift-n2: We decided to overwrite this in group_vars because the matrix-org/matrix-user-verification-service does not have any official releases after v2.0.0 (Feb 18, 2021) and we wanted to stick to a release tag in the role defaults. 
This does add some clutter to the group_vars and we should have clarified the comment such that it reads: 
```
# FIXME: When there is a new release for matrix-org/matrix-user-verification-service,
#   update matrix_user_verification_service_version in the role defaults and remove this.
```

We can also move the comment into the role defaults and use the working commit hash there.

In any case overwriting `matrix_user_verification_service_docker_image` is somewhat incorrect and we should only overwrite `matrix_user_verification_service_version`. I.e. `matrix_user_verification_service_version="sha256:d2aabc984dd69d258c91900c36928972d7aaef19d776caa3cd6a0fbc0e307270"`, which unfortunately does not work since docker pull requires an '@' instead of ':' as delimiter when using a digest. 

As a work around one could use `branch-master` as a version. We decided against this, since we did not want the container image to change without notice.

docs/configuring-playbook-jitsi.md
jakicoll: We changed that according to your suggestion.

Now someone might set `matrix_jitsi_auth_type: "matrix"`  and forget to enable the user_verification_service. We considered adding a check to the `validate_config.yml` in the matrix-jitsi role to catch the situation. We skipped that since `self_check_matrix_auth.yml` already checks if the user-verification-service is reachable for prosody.

jakicoll:
@spantaleev Thank you very much for reviewing. We've made the requested changes and will test the playbook again in our testing environment.

spantaleev:
I've fixed this up a bit and merged it!

Thank you for your hard work in bringing this large feature to the playbook!

crapo9000:
Hi,
can anyone confirm, that Jitsi Matrix auth is working? I've set everything up regarding the docs on a fresh deployment and when I try to connect to a Jitsi call (playbook managed) I just can't connect. On joining a lobby the jitsi auth is prompting:

image

No matter what I enter, I receive the following error message:

image

System logs are showing the following entries:

> matrix-jitsi-prosody[30101]: c2s55a5a6262a00                                              info        Client connected
matrix-jitsi-prosody[30101]: meet.jitsi:auth_matrix_user_verification                     info        REQUEST_COMPLETE reason:invalid_auth_token
matrix-jitsi-prosody[30101]: meet.jitsi:auth_matrix_user_verification                     warn        Error verifying membership err:bad-request, reason:JWT token must be provided with OpenID token and room ID
matrix-jitsi-prosody[30101]: c2s55a5a6366880                                              info        Client connected
matrix-jitsi-prosody[30101]: c2s55a5a6366880                                              info        Client disconnected: connection closed

jakicoll:
Are you joining the jitsi call via Element or are you trying to access Jitsi directly by navigating to jitsi.your-matrix-domain.com? The latter is not possible with Matrix User Authentication.

If you are doing the former: Does your synapse instance have a valid TLS certificate? Self-signed won't work here, since the user verification service by default requires a secure connection to your synapse server.

If none of these solve the issue, could you please:
(1) Set the `matrix_user_verification_service_uvs_log_level: debug`, apply that by running the playbook again including a restart of the service. Then try to join jitsi again and provide the relevant logs from the user verification service.
(2) Ensure that a JWT is passed to jitsi on loading. You could do this by opening the developer tools of your browser or Element desktop, then joining a jitsi conference and looking for the loading url in the network activity. It should look like this: https://jitsi.example.com/{{Your conference id}}?jwt={{the jwt authentification token, which is very long}}

crapo9000:
Thank you for your fast reply! Got it working now... 

> Are you joining the jitsi call via Element or are you trying to access Jitsi directly by navigating to jitsi.your-matrix-domain.com? The latter is not possible with Matrix User Authentication.

-->I tried joining it via Element (browser version) with a valid wildcard LE certificate

I didn't think about using the browser debugging console until your advice. With it I found out that the issue was located at the fronting reverse proxy that is handling SSL termination etc. The only issue I still have is that jitsi isn't bringing up the conference correctly. But this seems to be a websockets/firewall problem and is not related to UVS.

Best regards!

puhuri:
I'm not sure if this is related to the same issue (anyway, Jitsi authentication with _matrix_user_verification_service_enabled_ does not work). Even after adding `debug` on UVS, there were no any request logged. 

By testing at _matrix-jitsi-prosody_ host with wget, I get following error.
```
wget http://matrix-user-verification-service:3000  
--2023-04-14 08:34:43--  http://matrix-user-verification-service:3000/
Resolving matrix-user-verification-service (matrix-user-verification-service)... failed: Temporary failure in name resolution.
```

And when I look on running docker, I can see that the _matrix-jitsi-prosody_ is on _matrix-jitsi_ and _traefik_ networks and _matrix-user-verification-service_ is on _matrix_ only. There are now allow rules or NAT to enable access between those.

From prosody logs following can be found (obviously, JWT is invalid if it cannot communicate towards UVS).
```
mod_bosh                                                     info        New BOSH session, assigned it sid '83f3b2e9-c84b-413a-a32e-aa4442adb77b'
meet.jitsi:auth_matrix_user_verification                     info        REQUEST_COMPLETE reason:invalid_auth_token
meet.jitsi:auth_matrix_user_verification                     warn        Error verifying membership err:bad-request, reason:JWT token must be provided with OpenID token and room ID
```

The setup with Jitsi internal authentication worked without problems.

In addition, I had to remove `become` directives from _Checkout Prosody Auth Matrix User Verification Plugin Repo_ as there was no user for `jitsi_uid` the sudo to succeed.

crapo9000:
> I'm not sure if this is related to the same issue (anyway, Jitsi authentication with _matrix_user_verification_service_enabled_ does not work). Even after adding `debug` on UVS, there were no any request logged.
> 
> By testing at _matrix-jitsi-prosody_ host with wget, I get following error.
> 
> ```
> wget http://matrix-user-verification-service:3000  
> --2023-04-14 08:34:43--  http://matrix-user-verification-service:3000/
> Resolving matrix-user-verification-service (matrix-user-verification-service)... failed: Temporary failure in name resolution.
> ```
> 
> And when I look on running docker, I can see that the _matrix-jitsi-prosody_ is on _matrix-jitsi_ and _traefik_ networks and _matrix-user-verification-service_ is on _matrix_ only. There are no allow rules or NAT to enable access between those.
> 
> From prosody logs following can be found (obviously, JWT is invalid if it cannot communicate towards UVS).
> 
> ```
> mod_bosh                                                     info        New BOSH session, assigned it sid '83f3b2e9-c84b-413a-a32e-aa4442adb77b'
> meet.jitsi:auth_matrix_user_verification                     info        REQUEST_COMPLETE reason:invalid_auth_token
> meet.jitsi:auth_matrix_user_verification                     warn        Error verifying membership err:bad-request, reason:JWT token must be provided with OpenID token and room ID
> ```
> 
> The setup with Jitsi internal authentication worked without problems.
> 
> In addition, I had to remove `become` directives from _Checkout Prosody Auth Matrix User Verification Plugin Repo_ as there was no user for `jitsi_uid` the sudo to succeed.

That is something, I can confirm for a external hosted docker jitsi instance (non playbook managed). On joining a conference created via the jitsi plugin it seems that UVS is not generating any JWT token to authenticate the session although the UVS port is exposed via `matrix_user_verification_service_container_http_host_bind_port`

jakicoll:
> In addition, I had to remove become directives from Checkout Prosody Auth Matrix User Verification Plugin Repo as there was no user for jitsi_uid the sudo to succeed.

I guess that might be related to the refactoring of the jitsi installation, which has been moved into an external repository in 1d00d15482e8a2ec9e82a50624eda846ecf834d1 and 4546410f6a722084d9875034399137058bbbf139.

Anyway, thanks for reporting. I will take a careful look as soon as we update our own matrix installation and will rise an issue if it's a regression. It might take several weeks until I can have a look, but I eventually will. :)

crapo9000:
@jakicoll have you already been able to take a look at it?

jakicoll:
It's still on my list. Sorry, other things needed prioritization. Until then, you could of course continue to investiagate the authentication flow yourself using my hints above.

crapo9000:
@jakicoll I finally managed to get UVS working with the traefik setup and the new implemented jitsi role. (become commands still need to be commented out in `roles/galaxy/jitsi/tasks/util/setup_jitsi_auth_uvs_install.yml`)Editing the systemd service file like I mentioned above already solved the internal connection issues. I just had some additional firewall issues that triggered my errors, so no further changes are required to get UVS back running.

Here is the working systemd file (`/etc/systemd/system/matrix-user-verification-service.service`):


````

[Unit]
Description=Matrix User Verification Service
Requires=docker.service
After=docker.service
Requires=matrix-synapse.service
After=matrix-synapse.service
Requires=matrix-postgres.service
After=matrix-postgres.service
DefaultDependencies=no

[Service]
Type=simple
Environment="HOME=/root"
ExecStartPre=-/usr/bin/env sh -c '/usr/bin/env docker kill matrix-user-verification-service 2>/dev/null'
ExecStartPre=-/usr/bin/env sh -c '/usr/bin/env docker rm matrix-user-verification-service 2>/dev/null'


ExecStartPre=/usr/bin/env docker create --rm --name  matrix-user-verification-service\
                        --log-driver=none \
                        --user=998:1001 \
                        --cap-drop=ALL \
                        --read-only \
                        --network=matrix \
                        --mount type=bind,src=/matrix/user-verification-service/config/.env,dst=/app/.env,ro \
                        docker.io/matrixdotorg/matrix-user-verification-service@sha256:d2aabc984dd69d258c91900c36928972d7aaef19d776caa3cd6a0fbc0e307270

ExecStartPre=/usr/bin/env docker network connect traefik matrix-user-verification-service
ExecStart=/usr/bin/env docker start --attach matrix-user-verification-service

ExecStop=-/usr/bin/env sh -c '/usr/bin/env docker kill matrix-user-verification-service 2>/dev/null'
ExecStop=-/usr/bin/env sh -c '/usr/bin/env docker rm matrix-user-verification-service 2>/dev/null'
Restart=always
RestartSec=30
SyslogIdentifier=matrix-user-verification-service

[Install]
WantedBy=multi-user.target


````

Some additional changes are also required to fix the jitsi widget creation due do code changes in Element. Please change the preferredDomain variable from `preferredDomain` to `preferred_domain` in [https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/roles/custom/matrix-client-element/templates/config.json.j2](url) and [https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/7d1eb7e729b219ffc6800fecf28307e309e6848b/roles/custom/matrix-base/templates/static-files/well-known/matrix-client.j2#L21](url) + all related config templates

@spantaleev I'm not quite familiar with Github and it's functionality, could you please update the required files and implement the changes?

Best regards!

spantaleev:
@crapo9000, I've implemented these changes in 0a6b934a44a4f2a6a7d0f146b026d4aeb66024f7 and f3445c124c088b0122ee0bb8dd7ae327c96a7fb4.

I haven't tested them, but I'm hoping they work. Let me know if you hit any issues or corrections are necessary!

crapo9000:
> @crapo9000, I've implemented these changes in [0a6b934](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/0a6b934a44a4f2a6a7d0f146b026d4aeb66024f7) and [f3445c1](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/f3445c124c088b0122ee0bb8dd7ae327c96a7fb4).
> 
> I haven't tested them, but I'm hoping they work. Let me know if you hit any issues or corrections are necessary!

I've pulled your changes and  [f3445c1](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/f3445c124c088b0122ee0bb8dd7ae327c96a7fb4) seems to break the service file. I first have to check what's causing this, but I can't say yet whether I'll be able to do it this weekend. So until we fix this, we need to override the service file manually.

[0a6b934](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/0a6b934a44a4f2a6a7d0f146b026d4aeb66024f7) works as expected

spantaleev:
It may have been the missing space, which I've added in 06e2ab94fdebba0ad2.

crapo9000:
@spantaleev Sorry for my late response, but it still did not have any time to review your changes. Your adjustment in https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/06e2ab94fdebba0ad2be29307986f66828436548 don't seem to resolve the issue of breaking the service. Maybe there is a simple way to edit the service template just like I did in my example?

spantaleev:
What else is necesary to do to the service template to make it like your example?

If you can submit a PR with a working `matrix-user-verification-service.service`, that'd be helpful.

PR #2377, opened 2023-01-04
matrix_application

yassineaouam:
I cloned your repo on my local machine, and here's some remarks I had:

	- We made a condition on the matrix_synapse_ext_synapse_s3_storage_provider_config_endpoint_url variable so that if we give a value in the vars file, it's gonna be affected to it.
	If not, there won't be any need for the variable endpoints_url: '' to be shown with an empty field. 
	
	- I applied another condition on the --log-drive. If matrix_synapse_log_driver has a specific value, it should automatically be affected to the --log-drive. Else, it will get a none


rmalchow:
i think what he is trying to do is make more things conofigurable ... maybe this need a bit of extra work? making the log driver configurable is not too bad, i think ... and if i understand it correctly, the default behaviour would be the same as the current one. 

a bit messy, i agree ... but @yassineaouam can you clean this up?

spantaleev:
> making the log driver configurable is not too bad

Yes, it's not bad. All it needs is one more variable (`matrix_synapse_container_log_driver: none` in `defaults/main.yml`) and this:

```diff
---log-driver=none
+--log-driver={{ matrix_synapse_container_log_driver }}
```

Looks like this is done in a messy way and includes a bunch of other stuff.. Maybe you're maintaining your own docs (`docs/new_modifications.md`) in your own fork, etc... That's fine, but.. you shouldn't be publishing it here and wasting our time.

PR #2381, opened 2023-01-05
Allow the mautrix whatsapp relaybot to be enabled with a variable

MaybeJustJames:
This allows a user to enable the relaybot by setting a variable in `vars.yml` in the same way that the mautrix signal relaybot is configured.

### DIFF
--- a/roles/custom/matrix-bridge-mautrix-whatsapp/templates/config.yaml.j2
+++ b/roles/custom/matrix-bridge-mautrix-whatsapp/templates/config.yaml.j2
@@ -396,9 +396,9 @@ bridge:
     relay:
         # Whether relay mode should be allowed. If allowed, `!wa set-relay` can be used to turn any
         # authenticated user into a relaybot for that chat.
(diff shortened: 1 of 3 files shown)


roles/custom/matrix-bridge-mautrix-whatsapp/templates/config.yaml.j2
```diff
@@ -396,9 +396,9 @@ bridge:
     relay:
         # Whether relay mode should be allowed. If allowed, `!wa set-relay` can be used to turn any
         # authenticated user into a relaybot for that chat.
-        enabled: false
+        enabled: {{ matrix_mautrix_whatsapp_relaybot_enabled | default(false) }}
```
spantaleev, line 399: Judging by the setting this influences, it should probably be named `..._bridge_relay_enabled`

MaybeJustJames: Happy to change it. This name is consistent with the signal name though.

PR #2382, opened 2023-01-06
Hi again, sorry about the last pull request.

yassineaouam:
The last pull request was a little bit messy, so i decide to clean stuff and make it easier.
I added some more configuration options in the homeserver- and systemd unit templates. Default behaviour is unchanged, but:

matrix_synapse_log_driver - allows changing the log driver

matrix_synapse_config_search_all_users - allows changing the search behaviour in the homeserver.yaml

matrix_synapse_ext_synapse_s3_storage_provider_config_endpoint_url - changes the behaviour so that the endpoint url is optional (in case you're using vanillas S3)

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -390,6 +390,12 @@ See [How can I see the logs?](#how-can-i-see-the-logs).
 
 ### How can I see the logs?
 
+By default, we're configuring containers to use "none" as the logging driver, meaning that the output is simply discarded. By using 
+
+`matrix_synapse_log_driver=json-file`
+
+in your host vars, you can set it to something else (in this case, "json-file").
+
 We utilize [systemd/journald](https://www.freedesktop.org/software/systemd/man/systemd-journald.service.html#Description) for logging.
 
 To see logs for Synapse, run `journalctl -fu matrix-synapse.service`. You may wish to see the [manual page for journalctl](https://www.commandlinux.com/man-page/man1/journalctl.1.html).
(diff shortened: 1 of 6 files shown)


docs/faq.md
```diff
@@ -390,6 +390,12 @@ See [How can I see the logs?](#how-can-i-see-the-logs).
 
 ### How can I see the logs?
 
+By default, we're configuring containers to use "none" as the logging driver, meaning that the output is simply discarded. By using
```
spantaleev, line 393: Log data is not discarded. We disable Docker's own logging because systemd saves logs for us in systemd-journald (`journalctl -fu matrix-synapse`).

Using a log-driver other than `none` means that all logs are stored twice, which is wasteful. Also, Docker doesn't seem to do any log rotation, so using something like json-file is bound to cause trouble, especially if you enable debug logging for Synapse (which will generate tons of logs).

roles/custom/matrix-synapse/templates/goofys/systemd/matrix-goofys.service.j2
```diff
@@ -12,7 +12,11 @@ ExecStartPre=-{{ devture_systemd_docker_base_host_command_docker }} kill %n
 ExecStartPre=-{{ devture_systemd_docker_base_host_command_docker }} rm %n
 
 ExecStart={{ devture_systemd_docker_base_host_command_docker }} run --rm --name %n \
+			{% if matrix_synapse_log_driver %}
```
spantaleev, line 15: This variable is not defined in defaults/main.yml and it probably should be. An if won't be necessary at all then

spantaleev:
One PR should generally introduce just one change, not multiple unrelated changes.

The log-driver thing is not done correctly as of right now. Also, it's probably useless. Seems like you didn't know that logs are stored in systemd-journald. Now that you know how to access the logs there, it's likely that you don't want to do double logging via poorly-suited mechanisms such as Docker's json-file.

PR #2384, opened 2023-01-06
Change Synapse default room version from 9 to 10

FSG-Cat:
This PR bumps the default room version in accordance to [MSC3904](https://github.com/matrix-org/matrix-spec-proposals/pull/3904). This MSC is in FCP currently and is expected to pass.

All homeserver softwares that this playbook support support version 10. 

Dendrite and Conduit seem to have these hard coded or not in exposed config options that we currently use that is why this PR is Synapse only.


FSG-Cat:
This was implemented by https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2439

PR #2386, opened 2023-01-08
Add matrix-bot-chatgpt.

bertybuttface:


bertybuttface:
See: https://github.com/matrixgpt/matrix-chatgpt-bot

roles/custom/matrix-bot-chatgpt/defaults/main.yml
```diff
@@ -0,0 +1,78 @@
+---
+# chatgpt is a bot for chatting to openAI chatgpt matrix bot
+# Project source code URL: https://github.com/jakecoppinger/matrix-chatgpt-bot
```
spantaleev, line 3:
```suggestion
# Project source code URL: https://github.com/matrixgpt/matrix-chatgpt-bot
```

roles/custom/matrix-bot-chatgpt/defaults/main.yml
```diff
@@ -0,0 +1,78 @@
...
+# Once `MATRIX_BOT_ACCESS_TOKEN` is set this is no longer used.
+matrix_bot_chatgpt_matrix_bot_password: '' # MATRIX_BOT_PASSWORD=
+
+# Matrix Configurable Settings Defaults (optional)
+matrix_bot_chatgpt_matrix_default_prefix: "!chatgpt " # MATRIX_DEFAULT_PREFIX= Leave prefix blank to reply to all messages, trailing space matters
+matrix_bot_chatgpt_matrix_default_prefix_reply: false # MATRIX_DEFAULT_PREFIX_REPLY=
+
+# Matrix Access Control (optional)
+matrix_bot_chatgpt_matrix_blacklist: ''
+matrix_bot_chatgpt_matrix_whitelist: ''
```
spantaleev, line 52: Does this mean anyone (on any homeserver) can use the bot?

All our other bots default to "this homeserver only" and require manual configuration changes to make them work for external users. Perhaps we should do the same here.

spantaleev (commented):
Thanks for working on adding this bot!

I've submited some feedback. Also, it'd be nice to include some documentation (its own `docs/` page that says how to best set it up with regards to authentication, etc). It'd be nice to update `README.md` and `docs/configuring-playbook.md` as well with mentions of this new bot.

bertybuttface:
> Thanks for working on adding this bot!
> 
> I've submited some feedback. Also, it'd be nice to include some documentation (its own `docs/` page that says how to best set it up with regards to authentication, etc). It'd be nice to update `README.md` and `docs/configuring-playbook.md` as well with mentions of this new bot.

Thanks for the feedback, I'll get working on it. I'll also add some basic docs

roles/custom/matrix-bot-chatgpt/defaults/main.yml
bertybuttface: Yeah we could set this to :

```
matrix_bot_chatgpt_matrix_whitelist: ':{{ matrix_domain }}'
```

bertybuttface:
We also need to add the needed user to `ensure-matrix-users-created` playbook

roles/custom/matrix-bot-chatgpt/defaults/main.yml
spantaleev: Do we need `matrix_bot_chatgpt_matrix_blacklist: '*'` or something of that sort as well or is just the whitelist enough?

spantaleev:
> We also need to add the needed user to `ensure-matrix-users-created` playbook

Oh, yes.. That'd be great! Some `bot.chatgpt` username would be consistent with all the other bot names. We'd need to use the "password" field (`matrix_bot_chatgpt_matrix_bot_password` - auto-populated based on the generic secret), not the access token.

roles/custom/matrix-bot-chatgpt/defaults/main.yml
bertybuttface: Whitelist is enough, blacklisting * will stop everyone talking to the bot, including those on the whitelist.

bertybuttface:
> Oh, yes.. That'd be great! Some `bot.chatgpt` username would be consistent with all the other bot names.

That's easy enough, I'll make the change. 

>  We'd need to use the "password" field (`matrix_bot_chatgpt_matrix_bot_password` - auto-populated based on the generic secret), not the access token.

Ah, right now that won't work, logging in with a username and a password prints an access token, I can change that though if thats not very good expected behaviour.

spantaleev:
If you could do that, that'd be great and seamless!

Alternatively, we can ask users to register the bot manually, log in manually to obtain the access token, then populate the config.. and finally do `install-...` + `start`. It's more of a hassle, but we also have lots of roles that do it this way.

Another alternative may be to make the bot an appservice and it register itself.

bertybuttface:
> Alternatively, we can ask users to register the bot manually, log in manually to obtain the access token, then populate the config.. and finally do `install-...` + `start`. It's more of a hassle, but we also have lots of roles that do it this way.

Ok let's do this to get it merged and then I'll take a proper look at doing some cacheing.

spantaleev:
We're in no rush to merge this. Maybe it's better if we do it right from the beginning, so we won't have to redo it twice and potentially think of backward compatibility when we change things.

Looks like the documentation page hasn't been updated yet? It assumes that `ensure-matrix-users-created` would do anything, but it can't because `group_vars/matrix_servers` hasn't been updated. It also mentions `buscarron`, etc.

I think it'd be best if you could make it work with a password (and cached access token) upstream, and then for us to jump straight to that.

bertybuttface:
Not rushing, ratcheting. I'd rather have a completed solution now and a better solution later potentially than have 80% of a better solution but end up with nothing.

I've made the changes but I have no idea when my schedule will allow me to work on cacheing. I don't think we leave this open indefinitely until then.

bertybuttface:
Fixed the `ansible-lint` errors, quite a few instances of `yaml[comments]: Too few spaces before comment`

Do I need to do anything about the warnings e.g.

```
key-order[task]: You can improve the task key order to: tags, block (warning)
roles/custom/matrix-bot-chatgpt/tasks/main.yml:3 Task/Handler: block/always/rescue 

key-order[task]: You can improve the task key order to: tags, block (warning)
roles/custom/matrix-bot-chatgpt/tasks/main.yml:15 Task/Handler: block/always/rescue 

args[module]: Unsupported parameters for ansible.builtin.service module: daemon_reload. Supported parameters include: state, arguments (args), pattern, name, sleep, enabled, runlevel. (warning)
roles/custom/matrix-bot-chatgpt/tasks/setup_uninstall.yml:10 Task/Handler: Ensure matrix-chatgpt is stopped
```

I based this on the `matrix-bot-postmoogle` code so if this is wrong then that is also wrong.

spantaleev:
You can ignore the warnings. I've fixed up some additional things and merged this! Thank you!

PR #2388, opened 2023-01-09
Make borg 1 possible via variable

yncyrydybyl:
The current syntax only works for borg version 2. Which is currently still in beta.
This PR introduces a variable **matrix_backup_borg_remote_version** which defaults to version 2.


spantaleev:
This is a problem with how https://gitlab.com/etke.cc/borgmatic publishes releases - overwriting old Docker container image tags when it rebuilds images.

Borg 2 is not out. It's just that the republished borgmatic images use a newer borgmatic version which takes different command line flags. To fix it, SSH into the machine and repull your borgmatic image manually or use the `matrix_backup_borg_docker_image_force_pull: true`.

Ultimately, this is a problem that needs to be fixed upstream, so that `borgmatic:15` will keep meaning the same thing and not change at any given time in the future.

PR #2398, opened 2023-01-16
update borg image tag

etkecc:
New tags include postgres, borg and borgmatic versions

[ref](https://gitlab.com/etke.cc/borgmatic#docker-image-tags)


spantaleev:
Supersedes https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2388

PR #2405, opened 2023-01-19
Fix ChatGPT container

CyberShadow:
This container needs a writable $HOME, and will fail at startup if there isn't one.

Provide one by pointing HOME to a path under the mounted /data directory.

### DIFF
--- a/roles/custom/matrix-bot-chatgpt/templates/systemd/matrix-bot-chatgpt.service.j2
+++ b/roles/custom/matrix-bot-chatgpt/templates/systemd/matrix-bot-chatgpt.service.j2
@@ -25,6 +25,7 @@ ExecStart={{ devture_systemd_docker_base_host_command_docker }} run --rm --name 
 			--network={{ matrix_docker_network }} \
 			--env-file={{ matrix_bot_chatgpt_config_path }}/env \
 			--mount type=bind,src={{ matrix_bot_chatgpt_data_path }},dst=/data \
+			--env HOME=/data/home \
 			{% for arg in matrix_bot_chatgpt_container_extra_arguments %}
 			{{ arg }} \
 			{% endfor %}


bertybuttface:
@spantaleev 

I'm not convinced by this change, why has nobody else encountered this problem when I can see a large spike in container pulls since adding this package to `matrix-docker-ansible-deploy`?

Did anyone test these changes? What were the symptoms of the problem this was trying to solve?

spantaleev:
This PR coincides with a time when lots of users in our support room were complaining of the chatgpt bot not working, so this seemed legit. I don't use this component at all, so I'm trusting the community to maintain it.

Seems like we should undo it though. Would you send a PR for that?

CyberShadow:
@bertybuttface 

The software inside the container failed to start up because it did not have a writable home directory. It needs a writable home directory in order to install its npm dependencies. Without the ability to install the npm dependencies, the 

In [the Matrix room](https://matrix.to/#/!cNSQwPqhHKkIZdBnvt:devture.com?via=devture.com&via=matrix.org&via=tchncs.de), you can find multiple people who have tried getting the container working but failing. That is probably the reason for the container downloads you've been seeing. Here are a few reports of  other users who were trying to get the container to work:

- https://matrix.to/#/!cNSQwPqhHKkIZdBnvt:devture.com/$qAlFDERCzkt2GjF1OQsmhezojYNfyF44-ZQdqgHilfc?via=devture.com&via=matrix.org&via=tchncs.de
- https://matrix.to/#/!cNSQwPqhHKkIZdBnvt:devture.com/$3XhK9TBMUOjUdXLk-_k3k81ccfVDIn_OVNFSlgYoliI?via=devture.com&via=matrix.org&via=tchncs.de

You can find a log of the startup failure in one of these messages as well.

It's possible that the container worked at some point, but a change in this project caused that to regress. I see from the changelog that there has been some restructuring so that may have something to do with it.

bertybuttface:
Thanks for the detailed responses guys, appreciated, I will check through the Matrix Chat Room.

I wasn't aware there was a chat room for this project so I wasn't monitoring it. 

Puppeteer is a bit of a PITA to be honest, hopefully ChatGPT will have a real API soon.

PR #2451, opened 2023-02-08
Add Draupnir support to the project. 

FSG-Cat:
Draupnir is a mjolnir hardfork by its former lead maintainer and this PR should add all the required facilities to run Draupnir with the playbook. This PR DOES NOT touch the synapse module. Draupnir can be installed with the config of an existing mjolnir and be happy. The project is not yet guaranteed to be stable but this PR should help those who want to test it out test it. 

This PR i can not guarantee is perfect due to that this is my first time attempting to add a new role to the playbook. I based the changes i needed to make on the ChatGPT role and also what felt right.


FSG-Cat:
Update on the status of this PR. This PR is 1 commit away from what is running in production on feline.support currently. The last PR as of writing aka [ddcb173](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2451/commits/ddcb1735e2671344dec7742b6d5726edfaeeecba) is the only missing one and it exists to address a bug discovered in testing. Draupnir does not recognise its own name without it being in its config and this commit fixes that as it says on the tin.

PR #2449, opened 2023-02-07
change version matrix-bot-chatgpt 1.4.1->2.1.0

ste7la:

PR #2452, opened 2023-02-09
update matrix-chatgpt-bot

borisrunakov:


spantaleev:
Please see: https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2449#issuecomment-1422053318

borisrunakov:
What if we wanted to instruct the user to get an OpenAI API key _after_ the message indicating what values to replace?
Could you point me please to an example I can look into ?

spantaleev:
You can make a custom validation check task for any of the old variables (only for them), and have a custom message which explains the situation.

PR #2453, opened 2023-02-09
fix: only add element related entries to client well-known if element is enabled

Techno-Disaster:
As the title says, only add element related entries to client well-known if element is enabled.

### DIFF
--- a/roles/custom/matrix-base/defaults/main.yml
+++ b/roles/custom/matrix-base/defaults/main.yml
@@ -170,7 +170,7 @@ matrix_well_known_matrix_client_io_element_e2ee_secure_backup_setup_methods: []
 # `matrix_well_known_matrix_client_io_element_e2ee_secure_backup_required`
 # `matrix_well_known_matrix_client_io_element_e2ee_secure_backup_setup_methods`
 matrix_well_known_matrix_client_io_element_e2ee_entries_enabled: "{{ not matrix_well_known_matrix_client_io_element_e2ee_default or matrix_well_known_matrix_client_io_element_e2ee_secure_backup_required or matrix_well_known_matrix_client_io_element_e2ee_secure_backup_setup_methods | length > 0 }}"
-ta
+
 # Default `/.well-known/matrix/client` configuration - it covers the generic use case.
 # You can customize it by controlling the various variables inside the template file that it references.
 #
--- a/roles/custom/matrix-base/defaults/main.yml
+++ b/roles/custom/matrix-base/defaults/main.yml
@@ -151,18 +151,26 @@ matrix_client_element_jitsi_preferredDomain: ''  # noqa var-naming
 # Controls whether Element should use End-to-End Encryption by default.
 # Setting this to false will update `/.well-known/matrix/client` and tell Element clients to avoid E2EE.
 # See: https://github.com/vector-im/element-web/blob/develop/docs/e2ee.md
-matrix_client_element_e2ee_default: true
+matrix_well_known_matrix_client_io_element_e2ee_default: true
 
 # Controls whether Element should require a secure backup set up before Element can be used.
 # Setting this to true will update `/.well-known/matrix/client` and tell Element require a secure backup.
 # See: https://github.com/vector-im/element-web/blob/develop/docs/e2ee.md
-matrix_client_element_e2ee_secure_backup_required: false
+matrix_well_known_matrix_client_io_element_e2ee_secure_backup_required: false
 
 # Controls which backup methods from ["key", "passphrase"] should be used, both is the default.
 # Setting this to other then empty will update `/.well-known/matrix/client` and tell Element which method to use
 # See: https://github.com/vector-im/element-web/blob/develop/docs/e2ee.md
-matrix_client_element_e2ee_secure_backup_setup_methods: []
+matrix_well_known_matrix_client_io_element_e2ee_secure_backup_setup_methods: []
 
+# Controls whether element related entries should be added to the client well-known. Override this to false to hide
+# element related well-known entries.
+# By default if any of the following change from their default this is set to true:
+# `matrix_well_known_matrix_client_io_element_e2ee_default`
+# `matrix_well_known_matrix_client_io_element_e2ee_secure_backup_required`
+# `matrix_well_known_matrix_client_io_element_e2ee_secure_backup_setup_methods`
+matrix_well_known_matrix_client_io_element_e2ee_entries_enabled: "{{ not matrix_well_known_matrix_client_io_element_e2ee_default or matrix_well_known_matrix_client_io_element_e2ee_secure_backup_required or matrix_well_known_matrix_client_io_element_e2ee_secure_backup_setup_methods | length > 0 }}"
+ta
 # Default `/.well-known/matrix/client` configuration - it covers the generic use case.
 # You can customize it by controlling the various variables inside the template file that it references.
 #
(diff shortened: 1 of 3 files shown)


spantaleev:
I think that that `/.well-known/matrix/client` file can be consulted from other Element instances (hosted elsewhere), as well as `element-desktop`. So even if Element is not enabled, it may still make sense to adjust these settings.

-----

What I don't like is that the `matrix-base` role reaches into variables from the `matrix-client-element` role.. But.. that's a separate issue.

Techno-Disaster:
> 

Makes sense, can we change it to `matrix_client_element_e2ee_default` then?. The clients should have proper defaults according to this https://github.com/vector-im/element-web/blob/develop/docs/e2ee.md

Techno-Disaster:
SGTM, maybe a new variable to handle this then? (`matrix_client_element_e2ee_default` and `matrix_client_element_enabled` both won't work for this)

spantaleev:
I don't understand.. what's the current problem right now? Why can't you adjust `matrix_client_element_e2ee_default` as you see fit? Or would you like to **not** set a `default` key in the JSON?

Techno-Disaster:
This is just to cleanup the well-known entry. In case someone who hosts the server does not need element with disabled e2ee, these entries aren't really needed.

spantaleev:
We currently have `matrix_well_known_matrix_server_enabled` and other such variables in the `matrix-base` role, which control certain `.well-known` aspects.

Perhaps we could add something like `matrix_well_known_matrix_client_io_element_e2ee_entries_enabled: true` in `roles/custom/matrix-base/defaults/main.yml`.

Then, in `group_vars/matrix_servers`, in the `base` section, we can redefine it in terms of the whole playbook (involing variables from other roles), like this:

```yaml
matrix_well_known_matrix_client_io_element_e2ee_entries_enabled: "{{ matrix_client_element_enabled }}"
```

---

We could also introduce variables for each sub-entry:

```yaml
matrix_well_known_matrix_client_io_element_e2ee_default: "{{ matrix_client_element_e2ee_default }}"
matrix_well_known_matrix_client_io_element_e2ee_secure_backup_required: "{{ matrix_client_element_e2ee_secure_backup_required }}"
matrix_well_known_matrix_client_io_element_e2ee_secure_backup_setup_methods: "{{ matrix_client_element_e2ee_secure_backup_setup_methods }}"
```

^^ These are the `group_vars/matrix_servers` overrides we could add to the `base` section there.

These variables should also be defined in `matrix-base` with some hardcoded default values.

spantaleev:
That last commit seems to go in some other direction? What's your idea there?

Techno-Disaster:
> That last commit seems to go in some other direction? What's your idea there?

ah no, I pushed it before reading your comment and went afk. I'll mark this PR WIP for now

Techno-Disaster:
> We currently have `matrix_well_known_matrix_server_enabled` and other such variables in the `matrix-base` role, which control certain `.well-known` aspects.
> 
> Perhaps we could add something like `matrix_well_known_matrix_client_io_element_e2ee_entries_enabled: true` in `roles/custom/matrix-base/defaults/main.yml`.
> 
> Then, in `group_vars/matrix_servers`, in the `base` section, we can redefine it in terms of the whole playbook (involing variables from other roles), like this:
> 
> ```yaml
> matrix_well_known_matrix_client_io_element_e2ee_entries_enabled: "{{ matrix_client_element_enabled }}"
> ```
> 
> We could also introduce variables for each sub-entry:
> 
> ```yaml
> matrix_well_known_matrix_client_io_element_e2ee_default: "{{ matrix_client_element_e2ee_default }}"
> matrix_well_known_matrix_client_io_element_e2ee_secure_backup_required: "{{ matrix_client_element_e2ee_secure_backup_required }}"
> matrix_well_known_matrix_client_io_element_e2ee_secure_backup_setup_methods: "{{ matrix_client_element_e2ee_secure_backup_setup_methods }}"
> ```
> 
> ^^ These are the `group_vars/matrix_servers` overrides we could add to the `base` section there.
> 
> These variables should also be defined in `matrix-base` with some hardcoded default values.

@spantaleev thanks for the help here. First time playing with ansible scripts, so could you confirm if something like this is enough?  (basically if any of the key from changes from its default value we add the whole thing to the well-known)

``` matrix_well_known_matrix_client_io_element_e2ee_entries_enabled: "{{ true if not matrix_client_element_e2ee_default or if matrix_client_element_e2ee_secure_backup_required or if matrix_client_element_e2ee_secure_backup_setup_methods != [] else false}}" ```

**Comment by spantaleev** [comment]
For completely preventing these sub-entries, I've proposed a new variable called `matrix_well_known_matrix_client_io_element_e2ee_entries_enabled` above.

If that one is set to `true`, then all of the entries underneath will be defined and set. I suppose that's good enough. It lets people who wish to have "no entries at all" have it.. For others, we include all the entries and set them to some value.

-----

> `matrix_well_known_matrix_client_io_element_e2ee_entries_enabled: "{{ true if not matrix_client_element_e2ee_default or if matrix_client_element_e2ee_secure_backup_required or if matrix_client_element_e2ee_secure_backup_setup_methods != [] else false}}"`

This  doesn't seem to be valid syntax.

I suppose your idea is to "only expose (all) the sub-entries if at least one of them as a non-default value"?

You probably want something like this:

```yaml
matrix_well_known_matrix_client_io_element_e2ee_entries_enabled: "{{ not matrix_client_element_e2ee_default or matrix_client_element_e2ee_secure_backup_required or matrix_client_element_e2ee_secure_backup_setup_methods | length > 0 }}"
```

However.. As per my proposal above, we're introducing a new variable for each of these sub-entries (e.g. `matrix_well_known_matrix_client_io_element_e2ee_default`, `matrix_well_known_matrix_client_io_element_e2ee_secure_backup_required`, `matrix_well_known_matrix_client_io_element_e2ee_secure_backup_setup_methods`). This way:

- the `matrix-base` role remains clean and does not reference **any** foreign (`matrix-client-element`) variables. All the new variables have some sane defaults in the `matrix-base` role.
- these new variables can be defined in terms of `matrix-client-element` variables via `group_vars/matrix_servers`. See my examples above.

Doing that, we should probably go with:

```yaml
matrix_well_known_matrix_client_io_element_e2ee_entries_enabled: "{{ not matrix_well_known_matrix_client_io_element_e2ee_default or matrix_well_known_matrix_client_io_element_e2ee_secure_backup_required or matrix_well_known_matrix_client_io_element_e2ee_secure_backup_setup_methods | length > 0 }}"
```

------

Looking at things now, it seems like I incorrectly assumed that these variables are defined in the `matrix-client-element` role (due to their bad naming):

- `matrix_client_element_e2ee_default`
- `matrix_client_element_e2ee_secure_backup_required`
- `matrix_client_element_e2ee_secure_backup_setup_methods`

Looks like they're in the `matrix-base` role already. That's not a good place for us to be storing `matrix_client_element_`-prefixed variables.

Because of this, instead of introducing new variables (e.g. `matrix_well_known_matrix_client_io_element_e2ee_default`) which will be defined in terms of the old ones (e.g. `matrix_client_element_e2ee_default`) via `group_vars/matrix_servers`, we should instead **rename these old badly-named variables**.

When renaming, make sure to update `tasks/validate_config.yml` in the role, so that people who use these variables will be notified of their new name.

Techno-Disaster:
@spantaleev how about now?

spantaleev:
Merged! Thanks for tidying up!

Techno-Disaster:
uwu nice, also @spantaleev  huge thanks for all the help :3

Techno-Disaster:
meep @spantaleev if we wanted to stop generating well-knowns completely, would a `matrix_well_known_matrix_client_enabled` variable like `matrix_well_known_matrix_server_enabled` make sense? Or is this already something like this in place?   (I guess we could always just override the template to `{}`)

spantaleev:
What's the use-case for doing that? I haven't heard of anyone else requesting this

PR #2462, opened 2023-02-11
Upgrade Drapunir from 1.8.0 Beta to 1.8.0 release.

FSG-Cat:
https://github.com/Gnuxie/Draupnir/releases/tag/v1.80.0

Docker Image built and ready to go. Since Mjolnir did not build for ARM Draupnir does not either so we dont have to wait for ARM images.


FSG-Cat:
Converted to draft due to that there is a potential bugfix release in the pipeline.

FSG-Cat:
This being a draft was unnessecary it looks like. A cockup during testing caused Cat to have locked the version to the previous version making us think the release was bugged its fine according to further testing.

PR #2470, opened 2023-02-13
fix: missing endif

Techno-Disaster:
meep @spantaleev   that last MR kinda broke setup :p


spantaleev:
Oh, good catch! Thanks!

Fixes https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/2471, I believe

Techno-Disaster:
yep, just ran main for my server and it works  + no redundant entries in client well-known now so that can be closed :D

PR #2506, opened 2023-02-20
Fixed issue to allow coturn to use host networking

RhysRdm:
Currently the playbook does not allow a user to use host networking with co-turn.

In the current playbook we cannot select host networking for co-turn as it tries to create the network named 'host' which for obvious reasons fails.

I've added a check to not create the network if a user selects the name 'host' as the co-turn network.

The reason for selecting host networking is that on large port ranges docker' userland proxy runs out of memory - see https://github.com/moby/moby/issues/11185.

### DIFF
--- a/roles/custom/matrix-coturn/tasks/setup_install.yml
+++ b/roles/custom/matrix-coturn/tasks/setup_install.yml
@@ -66,6 +66,7 @@
   community.docker.docker_network:
     name: "{{ matrix_coturn_docker_network }}"
     driver: bridge
+  when: "'host' not in matrix_coturn_docker_network"
 
 - name: Ensure matrix-coturn.service installed
   ansible.builtin.template:


spantaleev:
The code is currently like this: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/30185a04e90d7a1b7964450edb9d9d8360829ac6/roles/custom/matrix-coturn/tasks/setup_install.yml#L65-L69

It skips the task when `matrix_coturn_docker_network` is empty or equal to `host`. I don't see 

Your PR seems to be adding the same thing.. in a kind of backward way, even: `when: "'host' not in matrix_coturn_docker_network"`

RhysRdm:
Apologies - ill close. I was initially working off an older version of the playbook that didn't have your fix in. 

Ill close PR - No issue

PR #2510, opened 2023-02-21
Add rageshake server

gnunicorn:


roles/custom/matrix-rageshake/tasks/setup_uninstall.yml
```diff
@@ -0,0 +1,24 @@
...
+        state: stopped
+        enabled: false
+        daemon_reload: true
+
+    - name: Ensure matrix-rageshake.service doesn't exist
+      ansible.builtin.file:
+        path: "{{ devture_systemd_docker_base_systemd_path }}/matrix-rageshake.service"
+        state: absent
+
+    - name: Ensure Matrix chatgpt paths don't exist
```
spantaleev, line 21: chatgpt reference

roles/custom/matrix-rageshake/templates/systemd/matrix-rageshake.service.j2
```diff
@@ -0,0 +1,42 @@
...
+ExecStartPre=-{{ devture_systemd_docker_base_host_command_sh }} -c '{{ devture_systemd_docker_base_host_command_docker }} rm matrix-rageshake 2>/dev/null || true'
+
+ExecStart={{ devture_systemd_docker_base_host_command_docker }} run --rm --name matrix-rageshake \
+			--log-driver=none \
+			--user={{ matrix_user_uid }}:{{ matrix_user_gid }} \
+			--cap-drop=ALL \
+			--read-only \
+			--tmpfs /tmp \
+			--network={{ matrix_docker_network }} \
+			-p 9110:9110 \
```
spantaleev, line 26: Why expose port `9110`?

roles/custom/matrix-nginx-proxy/tasks/setup_nginx_proxy.yml
```diff
@@ -129,6 +129,13 @@
     mode: 0644
   when: matrix_nginx_proxy_proxy_dimension_enabled | bool
 
+- name: Ensure Matrix nginx-proxy configuration for rageshake domain exists
+  ansible.builtin.template:
+    src: "{{ role_path }}/templates/nginx/conf.d/matrix-rageshake.conf.j2"
```
spantaleev, line 134: Where is `templates/nginx/conf.d/matrix-rageshake.conf.j2`? It doesn't seem to be part of this PR (yet).

roles/custom/matrix-rageshake/templates/systemd/matrix-rageshake.service.j2
gnunicorn: Don't we need that? I am honestly not sure I fully understand the complicated inner setup and interactions between all these services yet (and how to test changes properly - I often found myself in the situation that my changes weren't properly applied and/or systemd or the service not restarted with the new configuration) ...

If you say that from looking at the other files, this isn't needed, I happily remove it.

spantaleev: In the default configuraiton (matrix-nginx-proxy as a reverse proxy.. or Traefik as a reverse-proxy), none of the services expose ports locally at all.

spantaleev:
Thanks for contributing this useful service!

I've fixed up various lose ends and added native Traefik support. Some testing shows that it works well (at least with Traefik) - hopefully, I haven't introduced a regression for the `matrix-nginx-proxy` type of setup.

gnunicorn:
Thanks!

Still learning about all these loose ends in the project. I'll review your changes and figure out how to do them in the future.

@spantaleev two questions:
1. is it a requirement that all docker services allow for custom building? I copied that from an existing one, but it makes the code a lot more complex and I wonder how much that is really used for a niche project like this...
2. Is traefik support always required, too?

Essentially: what are the minimum implementation points for an additional service PR to be accepted to this project?

spantaleev:
Thank you for trying to learn and improve!

1. Custom building is not strictly a requirement. The playbook officially supports only `amd64`. ARM support is a nice bonus, if available, but it's not guaranteed to be available for every component. Nevertheless, we try to make the playbook work on ARM and it does work on such architectures for the most part. If roles don't provide container images for all architectures we care about (`amd64`, `arm32`, `arm64`), then we try to add self-building support so that the component can be built on the host. Self-building may also be useful in other cases - when one would like to build a custom repository or branch of some component. It's useful for customizing things
2. Since yestreday, [Traefik is now the default reverse-proxy for this playbook](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/CHANGELOG.md#traefik-is-the-default-reverse-proxy-now). I suggest that you switch your setup to Traefik and try adding native Traefik support to your PRs. Bothering with `matrix-nginx-proxy` is optional - this component will be removed soon (as soon as I managed to get the remaining components off of it).

PR #2511, opened 2023-02-22
Allow configure default country code in element welcome page.

igogold:
Element web client has ability to set default phone country code to use when login via phone number. By default it is 'GB' but can be changed by setting ['default_country_code'](https://github.com/vector-im/element-web/blob/a26be2665fc1ba55842a613a154cddab5bdb32e6/docs/config.md#default-settings) in 'config.json'. So I've added a new variable to configure it when needed.

### DIFF
--- a/roles/custom/matrix-client-element/defaults/main.yml
+++ b/roles/custom/matrix-client-element/defaults/main.yml
@@ -191,6 +191,9 @@ matrix_client_element_self_check_validate_certificates: true
 # don't show the registration button on welcome page
 matrix_client_element_registration_enabled: false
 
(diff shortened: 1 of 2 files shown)


spantaleev:
Thanks!

It should be noted that a dedicated variable was not strictly necessary. One could configure any option via `matrix_client_element_configuration_extension_json`, like this:

```yaml
matrix_client_element_configuration_extension_json: |
  {
     "default_country_code": "GB"
  }
```

igogold:
Thank you, I completely missed this style of configuration.

PR #2515, opened 2023-02-24
feat: add sliding-sync-proxy support

### LINKED ISSUE #2478: Support deploying the Sliding Sync proxy
Sliding sync is the future of the Matrix sync api, and might finally give matrix a vaguely okay new-device experience. Sadly, there are no plans for implementing it in Synapse yet, even as the SS-only Element X nears a first public release. The official word from Matthew himself is "install the proxy".

I would install it manually, but I'm unsure how to integrate it with the docker-based setup that this playbook installs, so it seems that the best solution would be for it to be installed by this playbook.

It seems the proxy is installed on it's own URL, which is then referenced in the .well-known file. See https://github.com/matrix-org/sliding-sync/blob/main/docs/Landing.md

https://github.com/matrix-org/sliding-sync

gnunicorn:
fixes #2478

### DIFF
--- a/roles/custom/matrix-base/templates/static-files/well-known/matrix-client.j2
+++ b/roles/custom/matrix-base/templates/static-files/well-known/matrix-client.j2
@@ -26,6 +26,11 @@
 		"preferredDomain": {{ matrix_client_element_jitsi_preferredDomain|to_json }}
 	}
 	{% endif %}
+	{% if matrix_sliding_sync_proxy_enabled %},
+	"org.matrix.msc3575.proxy": {
+		"url": "https://{{ matrix_server_fqn_sliding_sync_proxy }}/_matrix/client/unstable/org.matrix.msc3575/sync"
+	}
+	{% endif %}
 	{% if matrix_client_element_location_sharing_enabled %},
 	"m.tile_server": {
         "map_style_url": "https://{{ matrix_server_fqn_element }}/map_style.json"
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -344,6 +344,8 @@ devture_systemd_service_manager_services_list_auto: |
     +
     ([{'name': 'matrix-registration.service', 'priority': 4000, 'groups': ['matrix', 'registration']}] if matrix_registration_enabled else [])
     +
+    ([{'name': 'matrix-sliding-sync-proxy.service', 'priority': 4000, 'groups': ['matrix', 'sliding-sync-proxy']}] if matrix_sliding_sync_proxy_enabled else [])
+    +
     ([{'name': 'matrix-sygnal.service', 'priority': 800, 'groups': ['matrix', 'sygnal']}] if matrix_sygnal_enabled else [])
     +
     ([{'name': 'matrix-goofys.service', 'priority': 800, 'groups': ['matrix', 'goofys']}] if matrix_s3_media_store_enabled else [])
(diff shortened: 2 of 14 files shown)


roles/custom/matrix-base/templates/static-files/well-known/matrix-client.j2
```diff
@@ -26,6 +26,11 @@
 		"preferredDomain": {{ matrix_client_element_jitsi_preferredDomain|to_json }}
 	}
 	{% endif %}
+	{% if matrix_sliding_sync_proxy_enabled %},
+	"org.matrix.msc3575.proxy": {
+		"url": "https://{{ matrix_server_fqn_sliding_sync_proxy }}/_matrix/client/unstable/org.matrix.msc3575/sync"
```
Kab1r, line 31: I believe this should not include the path to the sync API, just the URL of sync proxy:
```json
{
    "org.matrix.msc3575.proxy": {
        "url": "https://{{ matrix_server_fqn_sliding_sync_proxy }}"
    }
}
```

See Matrix.org's client well-known as an example:
https://matrix.org/.well-known/matrix/client

Kab1r (changes requested):

roles/custom/matrix-base/templates/static-files/well-known/matrix-client.j2
gnunicorn: While, I'd generally agree, the way I read the [Section in the MSC3575](https://github.com/matrix-org/matrix-spec-proposals/blob/kegan/sync-v3/proposals/3575-sync.md#unstable-prefix) as it is now, it specifically talks about the URL the sync is to be found at. I'll ask there for clarification.

Kab1r: Hmm, the way I read that section was that both the well-known JSON key: `org.matrix.msc3575.proxy` and the path to the sync endpoint are prefixed.
One stabilized, the JSON key would just be `proxy`

FSG-Cat: The Proxy is a temporary measure im not sure it will actually become stable. As in its likely there wont be a proxy that is under a stable identifier as this will just become sync v4 endpoint wise

FSG-Cat: Clarification from Travis on the matter of the destiny of the proxy. It seems like its not going to become stabilised also as for what the /.well-known content should be to me it sounds like it should be 
```
{
    "org.matrix.msc3575.proxy": {
        "url": "https://{{ matrix_server_fqn_sliding_sync_proxy }}"
    }
}
``` 
as @Kab1r is suggesting.

gnunicorn:
Because of priority reshuffling I won't be working on this for the next few days yet. If anyone wants to take this and run it to completion, please feel free to do so and superseed my PR. The current state and necessary tasks from my point of view are:

 - [x] Sliding sync Docker image included
 - [x] Sliding sync proxy linked to postgresql service
 - [x] Sliding sync proxy running and exposed via nginx-proxy
 - [x] Sliding sync proxy-URL added to the well-known
 - [ ] ensure the self-build-part of docker works properly
 - [ ] switch from nginx to traefik as proxy
 - [ ] ensure the proxy-url in well-known is correct (seems to be only the fqdn)

spantaleev:
[Sliding Sync support is here](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/6c1a39e6e9efbe0c1637006c8be16238e70845b3/CHANGELOG.md#sliding-sync-proxy-element-x-support) now. Thanks to everyone who contributed to this!

PR #2526, opened 2023-02-27
Add example config + readme for fronting playbook's traefik with own nginx on same server

jalemann:


examples/nginx/README.md
```diff
@@ -0,0 +1,32 @@
...
+devture_traefik_config_entrypoint_web_secure_enabled: false
+
+devture_traefik_container_web_host_bind_port: '127.0.0.1:81'
+
+devture_traefik_additional_entrypoints_auto:
+  - name: matrix-federation
+    port: 8449
+    host_bind_port: '127.0.0.1:8449'
+    config: {}
+```
```
spantaleev, line 27: Duplicating this here may make it go out of sync at some point. Can we refer to the other document instead? Perhaps we should add sections to the other document, so we can link directly to that part.

jalemann: Sure, will do.

PR #2531, opened 2023-02-28
remove devture services in remove-all script

array-in-a-matrix:
Currently the `remove-all.j2` script (or `/matrix/bin/remove-all`) removes all matrix systemd services but leaves `devture-traefik.service`. If more devture services are added in the future they will also be left running.


array-in-a-matrix:
Looks like `/devture-traefik` also remains, what variable controls its location?

PR #2535, opened 2023-03-02
Set Traefik base path

Call-Me-G-Now:
set Traefik's base path to be in the matrix base path tree. This is in alignment with all other roles' data paths

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -3781,6 +3781,8 @@ matrix_user_verification_service_uvs_auth_token: "{{ '%s' | format(matrix_homese
 # See the comment there for more details about why we have both `devture_traefik_enabled` and `matrix_playbook_traefik_role_enabled`.
 devture_traefik_enabled: "{{ matrix_playbook_reverse_proxy_type == 'playbook-managed-traefik' }}"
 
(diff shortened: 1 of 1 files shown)


spantaleev:
Thinking about the Matrix playbook in isolation, it does make sense.. It also makes sense adding a `devture_traefik_identifier` variable to the role and setting it to `matrix-traefik` (so that the systemd service is called `matrix-traefik.service`, etc).

However, Traefik is a component that allows us to go beyond this Matrix playbook.. It's not Matrix's Traefik, it's your global Traefik instance for this server and it possibly reverse-proxies to all your other containers there. These other containers may be configured manually, via docker-compose or via our [Related](https://github.com/spantaleev/matrix-docker-ansible-deploy#related) playbooks.

All these other playbooks actually use the same Traefik role and install it to the same global location (`/devture-traefik`). Ideally, you would install it through just one playbook, but.. it's a global instance in any case.

One of the reasons for moving the Matrix playbook to Traefik was to align it with all these other playbooks and allow people to easily host **all** these other services on the same machine.

Call-Me-G-Now:
Understood, thanks for the explanation

PR #2540, opened 2023-03-04
Define missing matrix_bot_go_neb_identifier variable

phirz:
The playbook complained about a missing `matrix_bot_go_neb_identifier` variable when switching to `playbook-managed-traefik`. I hope this single line is the intended way to define it (in analogy to corresponding variables of other roles, though there usually seems to be another one in `group_vars/matrix_servers`). At least the setup works again now.


spantaleev:
This role doesn't really make use of it though. That reference you found sounds like a mistake that we should fix (remove) instead.

Thanks for catching this mistake!

PR #2561, opened 2023-03-10
add synapse_auto_compressor

etkecc:
moved from etke.cc as-is

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+# 2023-03-12
+
+## synapse-auto-compressor support
+
+Thanks to [Aine](https://gitlab.com/etke.cc) of [etke.cc](https://etke.cc/), the playbook can now set up [rust-synapse-compress-state](https://github.com/matrix-org/rust-synapse-compress-state)'s `synapse_auto_compressor` tool to run periodically.
+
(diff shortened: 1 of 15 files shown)


AndrewKvalheim:
A couple first impressions:

- The image took 15 minutes to compile on my server. What would it take to get `aarch64` prebuilt?
- A routine `run-docker-prune` promptly deleted the image, causing `matrix-synapse-auto-compressor.service` to fail. Can the latest image be filtered out of the prune?

spantaleev:
Building seems very resource intensive..

On some VM I've tested, building for its native architecture (which happens to be `amd64`) took some 7 minutes and lots of memory and CPU.

Cross-building for `arm64` on the same VM took 4GB of RAM and 4GB of swap and was stuck for a long time, effectively killing the server. I suppose that a more beefy machine would be able to cross-build this and offer multi-arch images, but.. I've given up on doing it myself.

Ideally, the upstream project would provide official multi-arch images. See https://github.com/matrix-org/rust-synapse-compress-state/issues/93 for details.

------

Pruning container images also potentially affects other things - currently `certbot`, for those still stuck on `matrix-nginx-proxy`, although not so catastrophically as the `certbot` image could just be pulled again, whereas self-building won't magically happen.

I suppose one should simply not run pruning *after* a setup, but rather before, like this:

- prune, either manually or via the playbook
- run `install-all` to ensure everything is in place

Updating the docs to indicate this "best practice" is probably worth it.

PR #2565, opened 2023-03-13
add dynamic thumbnails var for Dendrite

array-in-a-matrix:


roles/custom/matrix-dendrite/templates/dendrite/dendrite.yaml.j2
```diff
@@ -293,10 +293,10 @@ media_api:
   max_file_size_bytes: {{ matrix_dendrite_max_file_size_bytes|to_json }}
 
   # Whether to dynamically generate thumbnails if needed.
-  dynamic_thumbnails: false
+  dynamic_thumbnails: {{ matrix_dendrite_mediaapi_dynamic_thumbnails }}
```
spantaleev, line 296: ` | to_json` is preferable everywhere

PR #2580, opened 2023-03-17
Add docs on enabling e2b encryption for mautrix-signal

DominicBurkart:
- link to e2b docs
- include configuration snippet


spantaleev:
Maybe it's better to put this on the common mautrix bridge configuration page (`docs/configuring-playbook-mautrix-bridges.md`)?

Individual bridge pages may get their own **End-to-bridge encryption** section on their own pages, which links to the other one with all the details, I suppose.

DominicBurkart:
👋 Thanks for the pointer. I see that it's already [there](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/554915bd748cef1ed69e45e0834bd43611e494e5/docs/configuring-playbook-mautrix-bridges.md), we can close this pr then! :) 

Capture d’écran 2023-03-18 à 08 50 55

PR #2624, opened 2023-04-04
fix(traefik): Ignore federation endpoint if entrypoint with port already exists

Kab1r:
## Current Behavior
Docker binds ports `devture_traefik_config_entrypoint_web_port` (80), `devture_traefik_config_entrypoint_web_secure_port` (443), and `matrix_federation_public_port` (8448) for Traefik,
If `matrix_federation_public_port` is set to 443, docker attempts to bind to 443 twice and consequently fails.

## Expected Behavior
`matrix_federation_public_port` is bound only if it is not set to `devture_traefik_config_entrypoint_web_port` and `devture_traefik_config_entrypoint_web_secure_port`.
This way, Traefik uses the existing entrypoint.


spantaleev:
Makes sense! Thanks!

PR #2636, opened 2023-04-11
add SSH --key-file option to docs

array-in-a-matrix:
If ssh fails when using ssh keys using the `--key-file` flag may solve the issue.

### DIFF
--- a/docs/installing.md
+++ b/docs/installing.md
@@ -36,6 +36,8 @@ If you **don't** use SSH keys for authentication, but rather a regular password,
 
 If you **do** use SSH keys for authentication, **and** use a non-root user to *become* root (sudo), you may need to add `-K` (`--ask-become-pass`) to all Ansible commands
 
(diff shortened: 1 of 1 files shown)


spatterIight:
To throw my 2 cents in, another way is to to add a `ssh_config` file, relevant `ssh_keys` folder, and add a line to `ansible.cfg`

I think this is better than adjusting` ~/.ssh/config` since my suggested changes are localized to the repo

matrix-docker-ansible-deploy/ssh_config:
```
Host *
    IdentitiesOnly yes
    ServerAliveInterval 30
    UserKnownHostsFile ssh_known_hosts
    StrictHostKeyChecking accept-new

Host matrix.<matrix-domain>
     User ubuntu
     Hostname 172.28.12.5
     IdentityFile ssh_keys/matrix-key

Host *
     User root
```

matrix-docker-ansible-deploy/ansible.cfg
```
[defaults]
retry_files_enabled = False
stdout_callback = yaml

[ssh_connection]
ssh_args = -F ssh_config -o ControlMaster=auto -o ControlPersist=600s

[connection]
pipelining = True
```

In this way all the relevant ansible commands are the exact same, the only difference being a prompt for the key password (if there is one)

PR #2644, opened 2023-04-16
Fix recaptcha vars ansible-vault compatibility

kanlas-net:
Without conversion to string vault encrypted variables doesn't decrypt
Same conversion is used for database password https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/roles/custom/matrix-synapse/templates/synapse/homeserver.yaml.j2#L864

### DIFF
--- a/roles/custom/matrix-synapse/templates/synapse/homeserver.yaml.j2
+++ b/roles/custom/matrix-synapse/templates/synapse/homeserver.yaml.j2
@@ -1223,12 +1223,12 @@ oembed:
 # This homeserver's ReCAPTCHA public key. Must be specified if
 # enable_registration_captcha is enabled.
 #
-recaptcha_public_key: {{ matrix_synapse_recaptcha_public_key|to_json }}
+recaptcha_public_key: {{ matrix_synapse_recaptcha_public_key|string|to_json }}
 
 # This homeserver's ReCAPTCHA private key. Must be specified if
 # enable_registration_captcha is enabled.
 #
-recaptcha_private_key: {{ matrix_synapse_recaptcha_private_key|to_json }}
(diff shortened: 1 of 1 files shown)


spantaleev:
I believe the proper way to use encryption is to define them in terms of other variables.

Example: `some_var: "{{ vault_some_var }}` in vars.yml and then defining `vault_some_var` in your vault file.

Adding this string workaround to all places where sensitive values are used may mean we need to add it to hundreds of places. I believe we have it in more than a few places already, but it's probably not the best way to use encrypted variables anyway.

kanlas-net:
> I believe the proper way to use encryption is to define them in terms of other variables.

Why? Ansible documentation doesn't limit how encrypted variables should be stored. Any real reason?
https://docs.ansible.com/ansible/latest/vault_guide/vault_encrypting_content.html#encrypting-individual-variables-with-ansible-vault

> Adding this string workaround to all places where sensitive values are used may mean we need to add it to hundreds of places

Or remove it from every place) That argument doesn't make sense because you don't have convensions about variable processing and already using this workaround. I only PR two variables, if anybody needed more they can PR them too. I don't ask to change everything

spatterIight:
Hi @kanlas-net, it looks like there is some overlap in the problem space here as it relates to #2650, but I do not share your issue of the `string` operation being required for ansible-vault decryption to take place

Can you elaborate on your message "Without conversion to string vault encrypted variables doesn't decrypt", perhaps with the specific error message you encountered, your ansible version, and any other relevant information?

kanlas-net:
@spatterIight Hi, I use ansible 2.3.0, but I think newer version have this issue too. There was no errors, just vaulted variable was inserted as is without decryption. Seems like it cased by to_json operator, which is used in template

PR #2646, opened 2023-04-16
Added a nix flake

Naalunth:
I added a small flake for use with the Nix package manager. Also a .direnv file, which allows anyone that has nix installed to immediately run the tools in this repo without having to install any tools manually.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,19 @@
+{
+  inputs.nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
+
+  outputs = { self, nixpkgs, ... }:
+    let
+      pkgs = import nixpkgs { system = "x86_64-linux"; };
+    in
+    {
+      devShell.x86_64-linux = pkgs.mkShell {
+        buildInputs = with pkgs; [
+          just
+          python311Packages.ansible-core
+          python311Packages.passlib
+        ];
+        LC_ALL = "C.UTF-8";
+        LC_CTYPE = "C.UTF-8";
+      };
+    };
+}
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 /roles/**/files/scratchpad
 .DS_Store
 .python-version
+flake.lock
 
 # ignore roles pulled by ansible-galaxy
 /roles/galaxy/*


AndrewKvalheim:
Please don’t add direnv configuration to version control. I had an existing configuration that this overwrote. Furthermore you don’t need to pull all those package from unstable, and definitely don’t need to be changing my locale.

flake.nix
```diff
@@ -0,0 +1,19 @@
+{
+  inputs.nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
```
AndrewKvalheim, line 2: There’s no need to resort to unstable.

flake.nix
```diff
@@ -0,0 +1,19 @@
...
+
+  outputs = { self, nixpkgs, ... }:
+    let
+      pkgs = import nixpkgs { system = "x86_64-linux"; };
```
AndrewKvalheim, line 6: What’s rational for overriding the local architecture with a hard-coded value?

flake.nix
```diff
@@ -0,0 +1,19 @@
...
+    in
+    {
+      devShell.x86_64-linux = pkgs.mkShell {
+        buildInputs = with pkgs; [
+          just
+          python311Packages.ansible-core
```
AndrewKvalheim, line 12: There’s no need for specifically Python 3.11. Why not just use the `ansible` alias?

flake.nix
```diff
@@ -0,0 +1,19 @@
...
+          python311Packages.passlib
```
AndrewKvalheim, line 13: What is the purpose of this? The Ansible package already links it.

flake.nix
```diff
@@ -0,0 +1,19 @@
...
+        ];
+        LC_ALL = "C.UTF-8";
+        LC_CTYPE = "C.UTF-8";
```
AndrewKvalheim, line 16: Why are you changing the user’s locale?

.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
AndrewKvalheim, line 1: It’s not appropriate to include this directive in shared version control. You can’t assume that everyone will want to automatically use the flake, and certainly can’t assume that every user of direnv is also using Nix.

flake.nix
AndrewKvalheim: Resolved in #3259

flake.nix
AndrewKvalheim: Resolved in #3259

flake.nix
AndrewKvalheim: Resolved in #3259

flake.nix
AndrewKvalheim: Resolved in #3259

flake.nix
AndrewKvalheim: Resolved in #3259

PR #2650, opened 2023-04-17
Add support for backup LDAP servers

spatterIight:
This commit enables [matrix-synapse-ldap3](https://github.com/matrix-org/matrix-synapse-ldap3) to support N-number of backup servers, according to their documentation "[the]  first available server is used"

### DIFF
--- a/docs/configuring-playbook-ldap-auth.md
+++ b/docs/configuring-playbook-ldap-auth.md
@@ -8,7 +8,9 @@ If you decide that you'd like to let this playbook install it for you, you need 
 
 ```yaml
 matrix_synapse_ext_password_provider_ldap_enabled: true
-matrix_synapse_ext_password_provider_ldap_uri: "ldap://ldap.mydomain.tld:389"
+matrix_synapse_ext_password_provider_ldap_uri: 
+  - "ldap://ldap-01.mydomain.tld:389"
+  - "ldap://ldap-02.mydomain.tld:389"
 matrix_synapse_ext_password_provider_ldap_start_tls: true
 matrix_synapse_ext_password_provider_ldap_base: "ou=users,dc=example,dc=com"
 matrix_synapse_ext_password_provider_ldap_attributes_uid: "uid"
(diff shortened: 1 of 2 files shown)


spantaleev:
This breaks backwards compatibility and should be done in another way. Perhaps a new `uris` variable which is defined in terms of the old one would be better. It can even be added to the YAML file directly with `| to_json`, without needing to loop over it

spatterIight:
That is a good point, I did not think about backwards compatibility

Can you elaborate on what you mean by "a new `uris` variable which is defined in terms of the old one"?

It also may be of interest that this functionality was working previously. It used to be defined as `uri: {{ matrix_synapse_ext_password_provider_ldap_uri|to_json }}`  and changed to `uri: {{ matrix_synapse_ext_password_provider_ldap_uri | string|to_json }}` at some point. What is the purpose of the string operation here?

Thanks

spantaleev:
`defaults/main.yml`:

```yaml
matrix_synapse_ext_password_provider_ldap_uris: ["{{ matrix_synapse_ext_password_provider_ldap_uri | string }}"]
```

`homeserver.yaml.j2`:

```yaml
uri: {{ matrix_synapse_ext_password_provider_ldap_uris | to_json }}
```

------

Maybe we don't even need `matrix_synapse_ext_password_provider_ldap_uris` though and can let `matrix_synapse_ext_password_provider_ldap_uri` be either a string or a list of strings, like upstream supports it.

To put it in `homeserver.yaml`, we can use:

```diff
-uri: {{ matrix_synapse_ext_password_provider_ldap_uri | string | to_json }}
+uri: {{ matrix_synapse_ext_password_provider_ldap_uri | to_json }}
```

 and it should work for both string and list of string. Given that ` | string` was there before, I suppose someone wanted to store the `matrix_synapse_ext_password_provider_ldap_uri` variable in an encrypted Ansible vault, similar to what https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2644 does for some other variables. If you reuse the `matrix_synapse_ext_password_provider_ldap_uri` variable without introducing `matrix_synapse_ext_password_provider_ldap_uris`, it may stop working for these people. I suppose it's fine though - they should use Ansible vault in a better way anyway.

spatterIight:
Ok, check now -- with the `string` operation removed the variable supports both strings and lists as an input. I am not sure why adding `string` fixed #2644, in fact, I am using this variable exclusively in an ansible-vault and it works for me.

PR #2661, opened 2023-04-27
Update configuring-playbook.md

SimonBiggs:


spantaleev:
A lot of roles are installed into `roles/galaxy/`, so..it's more appropriate to use `roles/*/` for this.

PR #2674, opened 2023-05-08
Adds a self build option for dendrite

ibeckermayer:
Recently I needed to add an option to self-build dendrite in order to [apply a bug patch](https://github.com/matrix-org/dendrite/issues/2902#issuecomment-1489937931) that's not yet in any official release.

The actual build itself is done with a `ansible.builtin.command` rather than a `community.docker.docker_image`, because [the dendrite Docker image uses the BUILDPLATFORM variable](https://github.com/matrix-org/dendrite/blob/aa1bda4c58d20e7d14267f9c87fab8efd7ae36ad/Dockerfile#L6), which is a [BuildKit/buildx feature](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/), and according to the [`community.docker.docker_image` documentation](https://docs.ansible.com/ansible/latest/collections/community/docker/docker_image_module.html):

> Building images is done using Docker daemon’s API. It is not possible to use BuildKit / buildx this way.

Therefore we work around that by avoiding using Docker daemon's api and just using the docker cli via `ansible.builtin.command`.

### DIFF
--- a/roles/custom/matrix-dendrite/tasks/setup_install.yml
+++ b/roles/custom/matrix-dendrite/tasks/setup_install.yml
@@ -1,17 +1,21 @@
 ---
-
 - name: Ensure Dendrite paths exist
   ansible.builtin.file:
-    path: "{{ item }}"
+    path: "{{ item.path }}"
     state: directory
     mode: 0750
     owner: "{{ matrix_user_username }}"
     group: "{{ matrix_user_groupname }}"
   with_items:
-    - "{{ matrix_dendrite_config_dir_path }}"
-    - "{{ matrix_dendrite_bin_path }}"
-    - "{{ matrix_dendrite_ext_path }}"
-    - "{{ matrix_dendrite_nats_storage_path }}"
+    - { path: "{{ matrix_dendrite_config_dir_path }}", when: true }
+    - { path: "{{ matrix_dendrite_bin_path }}", when: true }
+    - { path: "{{ matrix_dendrite_ext_path }}", when: true }
+    - { path: "{{ matrix_dendrite_nats_storage_path }}", when: true }
+    - {
+        path: "{{ matrix_dendrite_docker_src_files_path }}",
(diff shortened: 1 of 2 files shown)


roles/custom/matrix-dendrite/tasks/setup_install.yml
```diff
@@ -72,6 +88,11 @@
     owner: "{{ matrix_user_username }}"
     group: "{{ matrix_user_groupname }}"
 
+- name: Ensure Dendrite Docker image is built
+  ansible.builtin.command:
+    cmd: "{{ devture_systemd_docker_base_host_command_docker }} build -t {{ matrix_dendrite_docker_image }} {{ matrix_dendrite_docker_src_files_path }}"
+  when: "matrix_dendrite_container_image_self_build | bool"
+
```
spantaleev, line 95: You may wish to get inspired from the `matrix-synapse` role, where we also check if the image has been built before trying to fire the `docker build` command: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/99f69822877a65d55b42a0755af34e124cea8e53/roles/custom/matrix-synapse/tasks/synapse/setup_install.yml#L21-L52

I'm not sure if this is strictly necessary though. If the files in `matrix_dendrite_docker_src_files_path` remain unchanged, it will likely not rebuild the image again anyway?

ibeckermayer: @spantaleev thanks, good call. I looked into it a bit, and while it's true that Docker will figure out that it needn't rebuild the images again, ansible will still run the command and mark the task as "changed". So this system gives us a more accurate ansible response, as well as makes the overall script more efficient. See https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2674/commits/bc709217cfebc49b729cec14364c62d280288875

spantaleev:
Looks good! Thank you!

PR #2702, opened 2023-05-25
add updated Caddyfile example

Kuchenmampfer:
Because I was so scared by touching my reverse proxy setup without having an example to follow, I postponed upgrading my matrix server until now. Today I finally decided to do the upgrade and discovered that guessing the correct cofiguration was actually not that hard.

To save others from this hurdle, I decided to add my Caddyfile as an example. Because the old example Caddyfile could still be of use, I did not remove it, but renamed it to `Caddyfile.deprecated`. To make clear what is what, I finally updated the `README.md` heavily inspired by the one of the nginx example.

### DIFF
--- a/docs/configuring-playbook-own-webserver.md
+++ b/docs/configuring-playbook-own-webserver.md
@@ -160,7 +160,7 @@ devture_traefik_additional_entrypoints_auto:
     #     # trustedIPs: ['IP-ADDRESS-OF-YOUR-REVERSE-PROXY']
 ```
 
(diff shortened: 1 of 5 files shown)


spantaleev:
Thanks for updating these!

I'm wondering about the base domain proxying configuration. Right now, it only handles `@wellknown` and ignores the rest. Shouldn't the example also include support for forwarding other requests as well? We can potentially serve some homepage or static website at the base domain.

Kuchenmampfer:
This could for sure be added. But I personally only use other subdomains, so my setup did not include additional base domain things. And I think we can assume that users who serve their own website already know how to do that. So I don't really find it necessary in the example, but maybe a hint in the README that their setup has to basically be merged with the `@wellknown` block would be good.

spantaleev:
Sure, makes sense! Let's leave it as it is

PR #2710, opened 2023-05-29
Update prerequisites.md

lukababu:
These ports need to be open in order to receive emails to the Postmoogle bridge

### DIFF
--- a/docs/configuring-playbook-bot-postmoogle.md
+++ b/docs/configuring-playbook-bot-postmoogle.md
@@ -5,25 +5,22 @@
 The playbook can install and configure [Postmoogle](https://gitlab.com/etke.cc/postmoogle) for you.
 
 It's a bot/bridge you can use to forward emails to Matrix rooms. 
-Postmoogle runs an email server through SMTP and allaws you to create mailboxes to the domain you define in the DNS settings.
+Postmoogle runs an SMTP email server and allows you to assign mailbox addresses to Matrix rooms.
 
 See the project's [documentation](https://gitlab.com/etke.cc/postmoogle) to learn what it does and why it might be useful to you.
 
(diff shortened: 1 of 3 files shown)


spantaleev:
This page only lists ports that need to be open for the default installation.

The playbook supports various other services which require various other ports. Most people will not care about these services and do not need to be burdened with this information on the "common prerequisites" page.

Each component's documentation page should list its own networking and other prerequisites. See how it's done for the Jitsi component in its own documentation here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-jitsi.md#prerequisites

A similar thing should be done for Postmoogle.

----

By the way, "TLS over TCP" is not a good description for `587/tcp`. Perhaps "TLS-encrypted SMTP over TCP" is better.

Well, SMTP is always "over TCP", so maybe we shouldn't even add the "over TCP" part anywhere.

lukababu:
> This page only lists ports that need to be open for the default installation.
> 
> The playbook supports various other services which require various other ports. Most people will not care about these services and do not need to be burdened with this information on the "common prerequisites" page.
> 
> Each component's documentation page should list its own networking and other prerequisites. See how it's done for the Jitsi component in its own documentation here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-jitsi.md#prerequisites
> 
> A similar thing should be done for Postmoogle.
> 
> By the way, "TLS over TCP" is not a good description for `587/tcp`. Perhaps "TLS-encrypted SMTP over TCP" is better.
> 
> Well, SMTP is always "over TCP", so maybe we shouldn't even add the "over TCP" part anywhere.

Made adjustments, let me know what you think!

PR #2711, opened 2023-05-29
Document how to use chatgpt-bot in encrypted rooms

anytimesoon:
added an extra step to creating the chatgpt bot so it can read encrypted messages

### DIFF
--- a/docs/configuring-playbook-bot-chatgpt.md
+++ b/docs/configuring-playbook-bot-chatgpt.md
@@ -20,9 +20,11 @@ ansible-playbook -i inventory/hosts setup.yml --extra-vars='username=bot.chatgpt
 ```
 
 
-## 2. Get an access token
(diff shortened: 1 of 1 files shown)


spantaleev:
Are you sure this will work? You're setting up keys in Element on your own device and then expecting that the bot would somehow fetch and use them.

I see some other encryption related FAQ entries here: https://github.com/matrixgpt/matrix-chatgpt-bot#i-get-error-decryption-failed-because-the-room-key-is-missing

anytimesoon:
All I can say for sure is that it worked for me. Prior to setting the encryption keys, the bot kept complaining that it couldn't read encrypted messages.

I tried following the instructions in the FAQ by removing the bot from my config, redeploying, and then deploying again with the bot, but kept getting the same error.

PR #2712, opened 2023-05-31
Fix prefix path when loading Hydrogen's theme-element.json

sergiodj:
When Hydrogen is configured using a prefix path different than `/`,
like:

```
matrix_server_fqn_hydrogen: "my.domain.com"
matrix_client_hydrogen_path_prefix: "/hydrogen"
```
an error is triggered on the web page because there's a missing
slash in the `config.json.j2` file.


spantaleev:
With this patch, it seems like if the path prefix is `/`, we'd end up with a double slash (`//assets`).

I believe we need a conditional: `{{ '' if matrix_client_hydrogen_path_prefix == '/' else matrix_client_hydrogen_path_prefix }}/assets/...`

sergiodj:
> With this patch, it seems like if the path prefix is `/`, we'd end up
> with a double slash (`//assets`).

Yes, that's correct.

> I believe we need a conditional: `{{ '' if
> matrix_client_hydrogen_path_prefix == '/' else
> matrix_client_hydrogen_path_prefix }}/assets/...`

Fair enough.  I thought about doing that but didn't know if it'd be
considered too much.  I'll update the PR accordingly.

Thanks for the review, BTW.

PR #2722, opened 2023-06-02
Create hooks for user search ability

lukababu:
YAML hooks to have all users searchable and prefer local user search.

This is to mimic the behaviour of Teams and Slack.

### DIFF
--- a/roles/custom/matrix-synapse/defaults/main.yml
+++ b/roles/custom/matrix-synapse/defaults/main.yml
@@ -334,6 +334,11 @@ matrix_synapse_limit_profile_requests_to_users_who_share_rooms: false
 # Defaults to 'true'.
 matrix_synapse_include_profile_data_on_invite: true
 
+
+# User search behaviour
+matrix_synapse_search_all_users: false
+matrix_synapse_prefer_local_users: false
+
 # Controls whether people with access to the homeserver can register by themselves.
 matrix_synapse_enable_registration: false
 # Controls whether people with access to the homeserver can register by themselves without verification (email/msisdn/token)
--- a/roles/custom/matrix-synapse/defaults/main.yml
+++ b/roles/custom/matrix-synapse/defaults/main.yml
@@ -336,8 +336,8 @@ matrix_synapse_include_profile_data_on_invite: true
 
 
 # User search behaviour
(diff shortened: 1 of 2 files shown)


roles/custom/matrix-synapse/defaults/main.yml
```diff
@@ -334,6 +334,11 @@ matrix_synapse_limit_profile_requests_to_users_who_share_rooms: false
 # Defaults to 'true'.
 matrix_synapse_include_profile_data_on_invite: true
 
+
+# User search behaviour
+matrix_synapse_search_all_users: false
+matrix_synapse_prefer_local_users: false
```
spantaleev, line 340: Given which actual settings these variables control, to keep consistency with the rest of the playbook, these variables should be renamed:

```suggestion
matrix_synapse_user_directory_search_all_users: false
matrix_synapse_user_directory_prefer_local_users: false
```

roles/custom/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -2630,7 +2630,7 @@ user_directory:
     # Uncomment to return search results containing all known users, even if that
     # user does not share a room with the requester.
     #
-    #search_all_users: true
+    search_all_users: {{ matrix_synapse_search_all_users|to_json }}
```
spantaleev, line 2633:
```suggestion
    search_all_users: {{ matrix_synapse_user_directory_search_all_users | to_json }}
```

roles/custom/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -2639,7 +2639,7 @@ user_directory:
     # Uncomment to prefer local over remote users in user directory search
     # results.
     #
-    #prefer_local_users: true
+    prefer_local_users: {{ matrix_synapse_prefer_local_users|to_json }}
```
spantaleev, line 2642:
```suggestion
    prefer_local_users: {{ matrix_synapse_user_directory_prefer_local_users | to_json }}
```

spantaleev (changes requested):

PR #2735, opened 2023-06-13
Support EC2 instance metadata for S3 storage provider

cbackas:
I have an EC2 instance that I deployed my stack to. I wanted to use the S3 storage provider but I'm not too keen on creating IAM users (and especially access keys) when I don't have to [so I assigned an IAM role directly to my EC2 instance. ](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-metadata.html)

The playbook just needed a few tiny tweaks so that it can optionally stop expecting the `matrix_synapse_ext_synapse_s3_storage_provider_config_access_key_id` and `...secret_access_key` variables. 

By setting the variable `matrix_synapse_ext_synapse_s3_storage_provider_config_ec2_instance_profile` to `true` in your host vars, the access key id/secret will not be expected by the playbook and will not be provided to the server, allowing instance metadata to get picked up. The s3_storage_provider uses boto3 so it already picks up the instance metadata without a hitch, after these few tweaks Synapse started uploading images to S3 without a problem.

Thanks for the very cool and comprehensive playbook

### DIFF
--- a/roles/custom/matrix-synapse/templates/synapse/ext/s3-storage-provider/env.j2
+++ b/roles/custom/matrix-synapse/templates/synapse/ext/s3-storage-provider/env.j2
@@ -1,3 +1,7 @@
+{% if not matrix_synapse_ext_synapse_s3_storage_provider_config_ec2_instance_profile|default(false)|bool %}
+AWS_ACCESS_KEY_ID={{ matrix_synapse_ext_synapse_s3_storage_provider_config_access_key_id }}
+AWS_SECRET_ACCESS_KEY={{ matrix_synapse_ext_synapse_s3_storage_provider_config_secret_access_key }}
+{% endif %}
 AWS_DEFAULT_REGION={{ matrix_synapse_ext_synapse_s3_storage_provider_config_region_name }}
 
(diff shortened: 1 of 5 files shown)


roles/custom/matrix-synapse/templates/synapse/ext/s3-storage-provider/env.j2
```diff
@@ -1,5 +1,7 @@
+{% if not matrix_synapse_ext_synapse_s3_storage_provider_config_ec2_instance_profile|default(false)|bool %}
```
spantaleev, line 1:
```suggestion
{% if not matrix_synapse_ext_synapse_s3_storage_provider_config_ec2_instance_profile | bool %}
```

spantaleev (changes requested):
I've submitted some suggestions for improvement.

It would also be good to update `docs/configuring-playbook-synapse-s3-storage-provider.md` with details about `matrix_synapse_ext_synapse_s3_storage_provider_config_ec2_instance_profile` and the other credentials not being strictly required in all cases.

cbackas:
> Thank you! I've fixed up the docs a bit as well and think it's good now

Thanks for the help 😃

PR #2736, opened 2023-06-14
Update configuring-playbook-synapse-s3-storage-provider.md

Tupsi:
added syntax to upload to wasabi (needs --endpoint param)


spantaleev:
I've adjusted the docs to not be so Wasabi-specific. I think it's good to merge now. Thank you!

PR #2739, opened 2023-06-15
Fix installing 'aux'-role on NTFS

### LINKED ISSUE #2738: Cannot install role "ansible-role-aux" when using NTFS
**Describe the bug**
When running `just roles`, the first requirement named "aux" fails to install with the error "Invalid argument".

**To Reproduce**
1. Clone the repo to a directory on an NTFS-formatted disk
2. Run `just roles`

**Expected behavior**
The roles should install without issues.

**Logs**
```
Starting galaxy role install process
- extracting aux to /some/path/matrix-docker-ansible-deploy/roles/galaxy/aux
[WARNING]: - aux was NOT installed successfully: Could not update files in /some/path/matrix-docker-ansible-deploy/roles/galaxy/aux: [Errno 22] Invalid argument: '/some/path/matrix-docker-ansible-deploy/roles/galaxy/aux'
```

**Matrix Server:**
 - OS: Linux 6.3.6
 - Architecture: x86_64

**Additional context**
The issue seems to be that "aux" is a reserved name on NTFS.
As the requirement is specified with `name: aux`, this tries to create `roles/galaxy/aux`, which fails like described above.

You can also reproduce this by navigating to an NTFS-formatted disk and executing `mkdir aux` - this will also fail.

SapuSeven:
As discussed in #2738 - this PR fixes installing the "ansible-role-aux" role on NTFS.

Tested using `just roles` and `just setup-all`.

### DIFF
--- a/playbooks/matrix.yml
+++ b/playbooks/matrix.yml
@@ -114,7 +114,7 @@
     - custom/matrix-nginx-proxy
     - custom/matrix-coturn
 
-    - role: galaxy/aux
+    - role: galaxy/auxiliary
 
     - role: galaxy/com.devture.ansible.role.postgres_backup
 
--- a/requirements.yml
+++ b/requirements.yml
@@ -2,7 +2,7 @@
 
 - src: git+https://github.com/mother-of-all-self-hosting/ansible-role-aux.git
   version: v1.0.0-1
-  name: aux
+  name: auxiliary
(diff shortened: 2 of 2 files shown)


spantaleev:
Thanks! I think we may need to grep through the docs - there may be some references to roles/galaxy/aux there.

SapuSeven:
You're right - on first glance I didn't find anything critical, but honestly I'm not too familiar with that part.

Issue #2721, opened 2023-06-02
Additional JVB: Make it work with traefik


aptiko:
Recently nginx became traefik. The documentation about additional JVBs, however, seems to have remained in nginx.

PR #2746, opened 2023-06-17
Fix errors when installing additional JVB

aptiko:
This pull requests contains five commits that fix four issues that I faced when attempting to install an additional JVB. It will be easier to review the commits one by one than all together—especially the first commit moves the playbooks and you need to read its message.

Please note that I had managed to work around the problems in February, but when later I started to write patches for them, I discovered that the change from nginx to traefik had broken it more, and the change into using an external jitsi role had broken it even more. Clearly this cannot continue. Either that functionality must be removed, or we must take care to not break it.

With the changes of this pull request, additional JVBs will (hopefully) work with nginx, but not (yet) with traefik (#2721).


spantaleev:
Thanks for fixing this up! :+1:

Issue #67, opened 2019-01-07
support travis's media repository


jcgruenhage:
https://github.com/turt2live/matrix-media-repo

turt2live:
`matrix-media-repo` now supports s3 (and s3-like) services by the way. With respect to it being invasive: it is, definitely. There are plans to make that less dangerous though: https://github.com/turt2live/matrix-media-repo/issues/101 and https://github.com/turt2live/matrix-media-repo/issues/60

spantaleev:
Good to hear you've made a lot of progress on this!

I'm guessing this will be a much better way to store files on S3-like services than what we have now (Goofys fuse filesystem with various hacks to make it work). I could see us migrating toward `matrix-media-repo` in the future, if not for other cases, then just for improving and hopefully simplifying the S3 case.

Will keep an eye on this!

aaronraimist:
1.0.0 is out now so it's probably a good time to revisit this.

One reason I would like to use matrix-media-repo even if the deduplication doesn't do that much for me is the ability to configure automatic deletion of remote content (and prevent my URL previews from expiring)
https://github.com/turt2live/matrix-media-repo/commit/0167c87e4278acb596d57164c8a52eef248b4d7d

MTRNord:
Any update on it? Would love to use it

ptman:
This is probably another thing that will be hit when scaling with workers

Forden:
any updates?

cvwright:
Can probably close this one now, with #2783 merged

PR #2783, opened 2023-07-12
Added matrix-media-repo role

Michael-Hollister:
I have added a role for the matrix-media-repo project (https://github.com/turt2live/matrix-media-repo) to the playbook for our own server deployments, and wanted to share changes upstream since there was previous interest in adding this to the playbook #67

This role supports out-of-box setup of the matrix-media-repo with any of the homeservers from this playbook, along with integrating with the other services in this playbook  like Postgres, NGINX, Prometheus, and Grafana.

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -2580,7 +2580,7 @@ matrix_media_repo_systemd_required_services_list: |
   {{
     (['docker.service'])
     +
-    ([devture_postgres_identifier ~ '.service'] if devture_postgres_enabled else [])
+    ([devture_postgres_identifier ~ '.service'] if devture_postgres_enabled and matrix_media_repo_database_hostname == devture_postgres_connection_hostname else [])
   }}
 
 ######################################################################
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -326,6 +326,8 @@ devture_systemd_service_manager_services_list_auto: |
     +
     ([{'name': 'matrix-ma1sd.service', 'priority': 2000, 'groups': ['matrix', 'ma1sd']}] if matrix_ma1sd_enabled else [])
     +
+    ([{'name': 'matrix-media-repo.service', 'priority': 4000, 'groups': ['matrix', 'media_store']}] if matrix_media_repo_enabled else [])
+    +
     ([{'name': 'matrix-mailer.service', 'priority': 2000, 'groups': ['matrix', 'mailer']}] if matrix_mailer_enabled else [])
     +
     ([{'name': 'matrix-nginx-proxy.service', 'priority': 3000, 'groups': ['matrix', 'nginx', 'reverse-proxies']}] if matrix_nginx_proxy_enabled else [])
@@ -395,7 +397,6 @@ devture_systemd_service_manager_services_list_auto: |
 ########################################################################
 
 
-
 ######################################################################
 #
 # com.devture.ansible.role.playbook_state_preserver
@@ -416,7 +417,6 @@ devture_playbook_state_preserver_commit_hash_preservation_dst: "{{ matrix_base_d
 # /com.devture.ansible.role.playbook_state_preserver
 #
 ######################################################################
-
 
 
 ######################################################################
@@ -2557,6 +2557,38 @@ matrix_ma1sd_database_password: "{{ '%s' | format(matrix_homeserver_generic_secr
 #
 ######################################################################
 
+######################################################################
+#
+# matrix-media-repo
+#
+######################################################################
+
+matrix_media_repo_enabled: false
+matrix_media_repo_identifier: matrix-media-repo
+matrix_media_repo_container_network: "{{ matrix_docker_network }}"
+
+matrix_media_repo_container_labels_traefik_enabled: false
+matrix_media_repo_container_labels_traefik_docker_network: "{{ matrix_playbook_reverse_proxyable_services_additional_network }}"
+matrix_media_repo_container_labels_traefik_entrypoints: "{{ devture_traefik_entrypoint_primary }}"
+matrix_media_repo_container_labels_traefik_tls_certResolver: "{{ devture_traefik_certResolver_primary }}"
+
+matrix_media_repo_database_hostname: "{{ devture_postgres_connection_hostname if devture_postgres_enabled else '' }}"
+matrix_media_repo_database_username: matrix_media_repo
+matrix_media_repo_database_password: "{{ '%s' | format(matrix_homeserver_generic_secret_key) | password_hash('sha512', 'mediarepo.db', rounds=655555) | to_uuid }}"
+matrix_media_repo_database_name: matrix_media_repo
+
+matrix_media_repo_systemd_required_services_list: |
+  {{
+    (['docker.service'])
+    +
+    ([devture_postgres_identifier ~ '.service'] if devture_postgres_enabled else [])
+  }}
+
+######################################################################
+#
+# /matrix-media-repo
+#
+######################################################################
 
 ######################################################################
 #
@@ -2638,6 +2670,10 @@ matrix_nginx_proxy_proxy_matrix_identity_api_enabled: "{{ matrix_ma1sd_enabled }
 matrix_nginx_proxy_proxy_matrix_identity_api_addr_with_container: "matrix-ma1sd:{{ matrix_ma1sd_container_port }}"
 matrix_nginx_proxy_proxy_matrix_identity_api_addr_sans_container: "127.0.0.1:{{ matrix_ma1sd_container_port }}"
 
+matrix_nginx_proxy_proxy_media_repo_enabled: "{{ matrix_media_repo_enabled }}"
+matrix_nginx_proxy_proxy_media_repo_addr_with_container: "matrix-media-repo:{{ matrix_media_repo_port }}"
+matrix_nginx_proxy_proxy_media_repo_addr_sans_container: "127.0.0.1:{{ matrix_media_repo_port }}"
+
 # By default, we do TLS termination for the Matrix Federation API (port 8448) at matrix-nginx-proxy.
 # Unless this is handled there OR Synapse's federation listener port is disabled, we'll reverse-proxy.
 matrix_nginx_proxy_proxy_matrix_federation_api_enabled: |-
@@ -2695,6 +2731,8 @@ matrix_nginx_proxy_systemd_wanted_services_list: |
     (['matrix-corporal.service'] if matrix_corporal_enabled else [])
     +
     (['matrix-ma1sd.service'] if matrix_ma1sd_enabled else [])
+    +
+    (['matrix-media-repo.service'] if matrix_media_repo_enabled else [])
     +
     (['matrix-client-cinny.service'] if matrix_client_cinny_enabled and matrix_playbook_reverse_proxy_type in ['playbook-managed-nginx', 'other-nginx-non-container'] else [])
     +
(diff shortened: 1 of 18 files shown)


roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
```diff
@@ -94,6 +94,96 @@
...
+		{% endif %}
+
+		# Make sure this matches your homeserver in media-repo.yaml
+		# You may have to manually specify it if using delegation or the
+		# incoming Host doesn't match.
+		proxy_set_header Host $host;
+
+		proxy_set_header X-Real-IP $remote_addr;
+		proxy_set_header X-Forwarded-For $remote_addr;
+	}
```
spantaleev, line 140: These seem like fairly generic API endpoints to capture and forward to matrix-media-repo.

Where is this coming from? I don't see it one their nginx example: https://docs.t2bot.io/matrix-media-repo/deployment/nginx.html

Michael-Hollister: Yeah, the proxy configuration documentation unfortunately does not mention configuring those endpoints. The full list of endpoints used by the application can be found here: https://github.com/turt2live/matrix-media-repo/blob/release-v1.2.8/api/webserver/webserver.go#L85

The reason for proxying the endpoints through the media repo however is for caching access tokens. Details regarding that can be found here: https://docs.t2bot.io/matrix-media-repo/access-token-cache/index.html

cvwright: The relevant quote about the access token cache:

> The access token cache is disabled by default, however if enabled it is strongly recommended to proxy the /logout and /logout/all endpoints through to the media repo. They'll be proxied through to the homeserver and update the cache internally.

roles/custom/matrix-media-repo/templates/media-repo/systemd/matrix-media-repo.service.j2
```diff
@@ -0,0 +1,52 @@
...
+ExecStartPre=-{{ devture_systemd_docker_base_host_command_sh }} -c '{{ devture_systemd_docker_base_host_command_docker }} rm matrix-media-repo 2>/dev/null || true'
+
+ExecStartPre={{ devture_systemd_docker_base_host_command_docker }} create \
+			--rm \
+			--name=matrix-media-repo \
+			--log-driver=none \
+			--user={{ matrix_user_uid }}:{{ matrix_user_gid }} \
+			--cap-drop=ALL \
+			--network={{ matrix_docker_network }} \
+			-p {{ matrix_media_repo_port }}:{{ matrix_media_repo_port }} \
```
spantaleev, line 26: Exposing the port seems unnecessary in most situations. `matrix-nginx-proxy` can reach the media repo over the container network. Exposing the media repo's port directly (and on all interfaces) doesn't seem ideal.

For other roles, we introudce a `_bind_port` variable (with a `""` default value) and have an `if` condition here. Only if the variable is explicitly set to something truthy, we expose the port.

roles/custom/matrix-media-repo/templates/media-repo/systemd/matrix-media-repo.service.j2
```diff
@@ -0,0 +1,52 @@
...
+			{% if matrix_media_repo_metrics_enabled %}
+			-p {{ matrix_media_repo_metrics_port }}:{{ matrix_media_repo_metrics_port }} \
```
spantaleev, line 28: Here, we may also wish to not always expose the metrics publicly on a given port (on all interfaces).

If the integrated (into the playbook) Prometheus server is used, then it can scrape metrics over the container network. `matrix_prometheus_services_connect_scraper_media_repo_static_configs_target` is already configured correctly for this.

Introducing another `_bind_port` variable (with a default `""` value) here and only exposing the port if it's truthy will be better.

roles/custom/matrix-media-repo/templates/media-repo/systemd/matrix-media-repo.service.j2
```diff
@@ -0,0 +1,52 @@
...
+			{% endif %}
+			-v {{ matrix_media_repo_base_path }}:/data:z \
```
spantaleev, line 30: It'd be better if we use `--mount` instead of `-v`. We probably don't need the magical `z` thing.

The better thing aobut using `--mount` is that `-v` is magical as well. If the target directory does not exist for some reason and you use `-v`, it will automatically create it (and likely with some `root:root` ownership).

In cases where the directory does not exist, something is probably wrong, so it's better to have the service fail and let you investigate instead of automatically creating directories.

------

We should probably introduce a new `matrix_media_repo_data_path` variable and use it for this data.

```yaml
matrix_media_repo_data_path: "{{ matrix_media_repo_base_path }}/data"
```

This frees up the base directory for being used for other things in the future (environment variables file, Traefik labels, other configuration, some migration files). Passing the whole "base path" as `/data` means that any other files/directories we create within it will be directly accessible by matrix-media-repo.

This is a bit of a breaking change though.

roles/custom/matrix-media-repo/templates/media-repo/media-repo.yaml.j2
```diff
@@ -0,0 +1,359 @@
+# General repo configuration
+repo:
+  bindAddress: {{ matrix_media_repo_bind_address }}
```
spantaleev, line 3: Adding ` | to_json` everywhere in this file is a good way to ensure that more special values do not cause breakage. It's probably unnecessary for most things, but it's a good best-practice which we try to follow in most roles.

roles/custom/matrix-media-repo/tasks/setup_install.yml
```diff
@@ -0,0 +1,74 @@
...
+    - path: "{{ matrix_media_repo_base_path }}"
+      when: true
+    - path: "{{ matrix_media_repo_docker_src_files_path }}"
+      when: "{{ matrix_media_repo_container_image_self_build }}"
+  when: "item.when | bool"
+
+- name: Ensure media-repo configuration installed
+  ansible.builtin.template:
+    src: "{{ role_path }}/templates/media-repo/media-repo.yaml.j2"
+    dest: "{{ matrix_media_repo_base_path }}/media-repo.yaml"
```
spantaleev, line 20: In other roles, we also try to keep the configuration:

- living in `config/`, separate from the data (living in `data/`)
- mounted into the container as read-only, if possible

Looking at https://github.com/turt2live/matrix-media-repo/blob/b7160248830520d5d3fbc401dbaa702b5e81696b/Dockerfile#L32 it looks like we may be able to:

- introduce a new `matrix_media_repo_config_path` variable (`matrix_media_repo_config_path: "{{ matrix_media_repo_base_path }}/config`) and create this new directory in the `Ensure media-repo paths exist` task above
- adjust this task to save the `media-repo.yaml` file in `matrix_media_repo_config_path`
- create a new `env` file in `{{ matrix_media_repo_base_path }}/env` which contains `REPO_CONFIG=/config/media-repo.yaml`. See `./roles/custom/matrix-sliding-sync/templates/env.j2` for an example
- add `--env-file={{ matrix_media_repo_base_path }}/env` to the systemd service
- add `--mount` for mounting `matrix_media_repo_config_path` to `/config` with `ro` (readonly)

group_vars/matrix_servers
```diff
@@ -2557,6 +2557,38 @@ matrix_ma1sd_database_password: "{{ '%s' | format(matrix_homeserver_generic_secr
...
 ######################################################################
 
+######################################################################
+#
+# matrix-media-repo
+#
+######################################################################
+
+matrix_media_repo_enabled: false
+matrix_media_repo_identifier: matrix-media-repo
```
spantaleev, line 2567: It may be better if this variable is moved to `defaults/main.yml`.

For other roles, that's where we usually store it, with some sane default value.

We also usually make roles use the identifier variable for:

- configuring the container name (`--name={{ matrix_media_repo_identifier }}`) in the systemd service
- configuring the base path that the role uses by default. `group_vars/*` usually overwrite this to something more "playbook-specific". Most of our Matrix roles directly use `matrix_base_path` though - there's no need for such stuff, I suppose.

-----

Generally, all `matrix_media_repo_*` variables are expected to have some default in the role itself. Looks like this one is an exception (only existing in group vars), which is somewhat confusing.

group_vars/matrix_servers
```diff
@@ -2638,6 +2670,10 @@ matrix_nginx_proxy_proxy_matrix_identity_api_enabled: "{{ matrix_ma1sd_enabled }
 matrix_nginx_proxy_proxy_matrix_identity_api_addr_with_container: "matrix-ma1sd:{{ matrix_ma1sd_container_port }}"
 matrix_nginx_proxy_proxy_matrix_identity_api_addr_sans_container: "127.0.0.1:{{ matrix_ma1sd_container_port }}"
 
+matrix_nginx_proxy_proxy_media_repo_enabled: "{{ matrix_media_repo_enabled }}"
+matrix_nginx_proxy_proxy_media_repo_addr_with_container: "matrix-media-repo:{{ matrix_media_repo_port }}"
```
spantaleev, line 2674:
```suggestion
matrix_nginx_proxy_proxy_media_repo_addr_with_container: "{{ matrix_media_repo_identifier }}:{{ matrix_media_repo_port }}"
```

roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
spantaleev: Makes sense!

The only trouble I see is that matrix-corporal (if enabled), will no longer be able to see these calls. They'll go to media-repo and then straight to the homeserver. Unless... we configure matrix-media-repo to forward requests to matrix-corporal (adjusting `csApi` in `matrix_media_repo_homeservers`).

Speaking of which.. for `csApi` in `matrix_media_repo_homeservers`, shouldn't we (couldn't we) use some container URL (`matrix_homeserver_container_url` or maybe `matrix_nginx_proxy_proxy_matrix_client_api_addr_with_container` which is matrix-corporal-aware) instead of a public endpoint? I'm not sure if that would work or if it introduces some other problems. I wonder if matrix-media-repo uses the `csApi` for some public stuff or just internally.

roles/custom/matrix-media-repo/defaults/main.yml
```diff
@@ -0,0 +1,681 @@
...
+matrix_media_repo_bind_address: '0.0.0.0'
+matrix_media_repo_port: 8000
+
+# Where to store the logs, relative to where the repo is started from. Logs will be automatically
+# rotated every day and held for 14 days. To disable the repo logging to files, set this to
+# "-" (including quotation marks).
+#
+# Note: to change the log directory you'll have to restart the repository. This setting cannot be
+# live reloaded.
+matrix_media_repo_log_directory: logs
```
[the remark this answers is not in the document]
Michael-Hollister: Tested and confirmed it works. Thanks for the suggestion.

group_vars/matrix_servers
Michael-Hollister: Huh, for some reason I could not find any of the matrix roles that has their identifiers declared in `defaults/main.yml` as you described. I did see that pattern for some of the galaxy devture roles however.

I agree though that its probably a good idea to have `matrix_media_repo_identifier` declared in `defaults/main.yml`. I'll follow the current pattern of re-declaring `matrix_media_repo_identifier` in `group_vars/matrix_servers` as its done for all other roles. If you prefer that  `matrix_media_repo_identifier` is only declared in `defaults/main.yml`, let me know and I will make further changes.

spantaleev: I think having a default identifier of `matrix-media-repo` and not redeclaring (to the same thing?) it is probably okay too.

roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
Michael-Hollister: I did not test with matrix-corporal enabled, but replacing `csApi` with `matrix_nginx_proxy_proxy_matrix_client_api_addr_with_container` seems to work without issue with it disabled, expanding to the proxy URL  `http://matrix-nginx-proxy:12080`

Michael-Hollister:
Thanks @spantaleev  for the feedback! I've implemented the suggested changes. Let me know if there is anything else you would like further modified.

spantaleev:
Thanks a lot for your excellent work in bringing such a major feature to the playbook! I'm sure many will be happy to see matrix-media-repo finally supported! :bow:

PR #2784, opened 2023-07-12
Explain the option of manually defining your public IP in the documentation for Coturn

QEDeD:
I was very surprised by the fact that a rather important configuration option for Coturn was "hidden" in the Hosts file, but not mentioned here. 

Therefore my suggestion is to explicitly mention it here, as I believe that is that natural for people to look.


QEDeD:
I also wanted to explicitly refer users to the relevant roles file so people know where to find the additional configuration options :)

PR #2791, opened 2023-07-20
use docker url for maubot to contact synapse server

ayushin:
This is more secure as most setups do not expose admin API endpoints to the outside world.

Also should not hurt to have `.idea/` in .gitignore


spantaleev:
This should use `{{ matrix_homeserver_container_url }}` instead of hardcoding a reference to Synapse.

This way, it would work with other homeserver implementations, etc.

Issue #2813, opened 2023-08-04
Invalid repository error with fresh BorgBase


altsalt:
**Describe the bug**

journald throws an error `/srv/repos/BORGBASEREPO/repo is not a valid repository. Check repo config.` when trying to deploy backups via borg.

**To Reproduce**
My `vars.yml` file looks like this:

```yaml
backup_borg_enabled: true
backup_borg_location_repositories:
  - ssh://BORGBASEREPO@BORGBASEREPO.repo.borgbase.com/./repo
backup_borg_storage_encryption_passphrase: "REDACTED"
backup_borg_location_exclude_patterns:
  - "postgres-backup"
  - "./postgres-backup"
  - "/postgres-backup"
  - "/matrix/postgres-backup"
backup_borg_retention_keep_hourly: 0
backup_borg_retention_keep_daily: 1
backup_borg_retention_keep_weekly: 2
backup_borg_retention_keep_monthly: 2
backup_borg_retention_keep_yearly: 2
backup_borg_ssh_key_private: |
  -----BEGIN OPENSSH PRIVATE KEY-----
  REDACTED
  -----END OPENSSH PRIVATE KEY-----
```

Running `systemctl start matrix-backup-borg` on the matrix server with a fresh BorgBase repository.

**Expected behavior**

A backup should be pushed to the BorgBase repository.

**Matrix Server:**
 - OS: Debian Bullseye
 - Architecture amd64

**Additional context**

I've also filed [an issue](https://github.com/borgbackup/borg/issues/7756) with borgbackup.

altsalt:
I just tried running the updated version with a fresh borgbase and got this error:

```
Aug 04 21:24:23 TLD systemd[1]: Starting Borg Backup...
Aug 04 21:24:29 TLD matrix-backup-borg[1501918]: Repository /repository does not exist.
Aug 04 21:24:30 TLD matrix-backup-borg[1501918]: Local Exception
Aug 04 21:24:30 TLD matrix-backup-borg[1501918]: Traceback (most recent call last):
Aug 04 21:24:30 TLD matrix-backup-borg[1501918]:   File "/usr/lib/python3.11/site-packages/borg/archiver.py", line 5213, in main
Aug 04 21:24:30 TLD matrix-backup-borg[1501918]:     exit_code = archiver.run(args)
Aug 04 21:24:30 TLD matrix-backup-borg[1501918]:                 ^^^^^^^^^^^^^^^^^^
[...]

PR #2816, opened 2023-08-04
Update borg role v1.2.4-1.7.15-1 -> v1.2.4-1.7.15-2

etkecc:
Adjusted to v1.7.15 config


altsalt:
Unfortunately this version didn't fix the issue that I've been tracking down and returns a new error. I don't know how it will interact with existing backups and suggest not merging yet.

See #2813

PR #2821, opened 2023-08-10
Advertise sliding-sync support when using conduit

meenzen:
Starting from version `0.6.0` conduit natively supports some sync v3 (sliding-sync) features.

https://gitlab.com/famedly/conduit/-/releases/v0.6.0
https://gitlab.com/famedly/conduit/-/merge_requests/501


spantaleev:
It's good to enable it, but I think it'd be better if all the logic related to it is **not** inside this file (with different `if` statements), but somewhere outside:

- some default value for a new `matrix_homeserver_sliding_sync_url` variable defined as `''` in `matrix-base`'s `defaults/main.yml`.. Along with some comments about it
- overrides for this variable in `group_vars/matrix_servers`, so that it's pointed correctly depending on homeserver implementation and whether `matrix_sliding_sync_enabled`

Then in this static file, we can just use an `if` like `{% if matrix_homeserver_sliding_sync_url %}` and then just use its value.

meenzen:
Damn, this leads to a  `recursive loop detected in template string` error

meenzen:
Changed as requested. I've also added a note to the sliding-sync docs explaining this functionality.

spantaleev:
Looks great! Thank you!

PR #2858, opened 2023-08-31
Improved README descriptions

array-in-a-matrix:
Specifically edited client and bridge descriptions.


spantaleev:
Much better than "Bridge for bridging .." ;) Thanks!

PR #2861, opened 2023-09-01
Prevent schildichat from being installed by default

gill6151:
Currently `matrix_client_schildichat_enabled` is set to true, installing Schildichat by default and causing the playbook to fail at SSL certificate retrieval if a DNS record for Schildichat has not been set (which I suspect is the case for majority of playbook users).

This PR sets that variable to false, requiring Schildichat to be manually enabled, as per the docs.


spantaleev:
All roles enable their respective components by default.

The playbook then potentially disables them via `group_vars/matrix_servers`. This should be fixed there

PR #2872, opened 2023-09-05
replace rust-synapse-compress-state docker image

aine-etke:
The mb-saces' https://gitlab.com/mb-saces/rust-synapse-compress-state/container_registry is empty.
This PR replaces the mb-saces' docker image with etke.cc's one that uses the original Dockerfile of the rust-synapse-compress-state project, plus contains tagged docker images.

It's not multi-arch, though, but it's better than nothing

### DIFF
--- a/roles/custom/matrix-synapse/defaults/main.yml
+++ b/roles/custom/matrix-synapse/defaults/main.yml
@@ -1008,8 +1008,9 @@ matrix_synapse_redaction_retention_period: 7d
 matrix_synapse_user_ips_max_age: 28d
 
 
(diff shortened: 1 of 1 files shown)


spantaleev:
Maybe https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2873 is better?

saces:
The container at registry.gitlab.com/mb-saces/rust-synapse-compress-state:latest is rebuild.
I was a bit hasty deleting them, sorry that.
The new container at registry.gitlab.com/mb-saces/rust-synapse-compress-state:stable (:stable means latest tagged version) contains also synapse-find-unreferenced-state-groups and is for arm64/amd64

Feel free to edit the PR to  :v0.0.1 :)

PR #2913, opened 2023-09-26
Merge changes

frost-cqre:


spantaleev:
This:

- changes so much, and even unrelated things
- it breaks the old (legacy) `password_providers` section in `homeserver.yml`
- the recommended way to inject SSO (OIDC) configuration is via `matrix_synapse_configuration_extension_yaml`, which allows you to inject anything. Adding custom variables for each OIDC provider is complicated and unnecessary (it will be too many variables to add and maintain for no good reason)

I suppose this PR may have been opened by mistake.

PR #2936, opened 2023-10-11
Update Admin-API link

nerdoc:
the old one is deprecated (says the page at the link target), and refers to this new one.


spantaleev:
I've fixed up the Synapse Admin APIs link and am merging it. Thanks!

PR #2963, opened 2023-10-23
Fix violation of privacy

MTRNord:
This is not only not really obvious to newcomers imho but also is a horrible default. It not only violates the privacy but also can easily leak information to the public.

Also it seems like a bad idea to change defaults as a middleman for a repo that itself wants to be used as a generalized server deployment tool.

### DIFF
--- a/roles/custom/matrix-synapse/defaults/main.yml
+++ b/roles/custom/matrix-synapse/defaults/main.yml
@@ -318,7 +318,7 @@ matrix_synapse_allow_public_rooms_without_auth: false
 # However, disabling federation completely (see `matrix_synapse_federation_enabled`) is a better way to make your server private,
 # instead of relying on security-by-obscurity -- federating with others, having your public rooms joinable by anyone,
 # but hiding them and thinking you've secured them.
-matrix_synapse_allow_public_rooms_over_federation: true
+matrix_synapse_allow_public_rooms_over_federation: false
 
 # Whether to require authentication to retrieve profile data (avatars,
 # display names) of other users through the client API. Defaults to


TheArcaneBrony:
+1, especially from the viewpoint of upstream *disabling this by default for user safety*.

spantaleev:
Setting up a Matrix server and making it federated.. Creating public rooms.. Allowing **any** Matrix server to join said rooms directly, without any access control.. Publishing these rooms to the "Room Directory".. And then complaining about the very minor part that "other Matrix servers can find my public rooms" seems like a minor problem.

Matrix is a federated network.  You have many ways to keep your rooms private:

- disable federation and run a fully private server
- or.. stop creating public rooms
- or.. should you create public rooms, prevent them from being joinable over federation (there's the `m.federate` `/createRoom` option and a nice UI for doing this)
- or.. should you ignore all of the above, at least avoid publishing said public rooms to the Rooms Directory

You're deciding to ignore all these options and are making your rooms fully public and federateable.. Then you're complaining that other Matrix servers can discover your rooms.

It doesn't make any sense to have Matrix be a federated network, but not-quite-federating. I don't think it's a violation of privacy to expose the list of public rooms for servers that are fully federating and that allow said rooms to be freely joined by anyone. You have **all of the above options, including changing playbook defaults** which are well-announced in the CHANGELOG and in the Matrix room for the playbook.

Next, people may complain that this playbook turns on Federation by default and may wish that we change `matrix_synapse_federation_enabled` to `false`. You're joining a federated network and the defaults match that - your public rooms are federated across the network.

There seems to be a vocal minority of possibly 3 users (out of thousands) who are complaining about this. You have options and are free to exercise them. I don't see a reason why we should all collectively continue to cripple Matrix federation and room discovery because a few users are unhappy with these defaults.

------

The playbook, as all installation tools (and software distributions) is opinionated in various ways. While, this playbook tends to stick to upstream defaults in most cases, this is not a guarantee.

Just like various Linux distributions are free to package (and patch and adjust) software as they see fit, so do we. If this is to be compared to a Linux distribution, it would probably be closest to [Arch Linux](https://archlinux.org/) in that:

- it tries to educate people in its huge documentation, so that they know how things work (to a reasonable extent)
- it tries to stick to upstream defaults, but.. may patch this or that when it makes sense to do it. I'm actually hoping that the Matrix-federation-regression introduced in Synapse v1.7.0 will be reverted upstream and we won't be deviating anymore. Perhaps this change (and its announcement in [TWIM](https://matrix.org/category/this-week-in-matrix/)) will provoke discussion and progress around this.

real-joshua:
> Also it seems like a bad idea to change defaults as a middleman for a repo that itself wants to be used as a generalized server deployment tool.

Important point, I'd expect this to be as secure and safe as the default. :D

hanthor:
This is a UX issue for Synapse+Element. An current solution is to make "public-for-only-my-server" rooms would be to make them joinable by members of a space and then make that space an auto-join room for new users.

Issue #2975, opened 2023-10-30
No such container: matrix-user-verification-service


ZelChief:
**Describe the bug**

How to debug error `No such container: matrix-user-verification-service`? 

```bash
TASK [galaxy/com.devture.ansible.role.systemd_service_manager : Ensure systemd services are started]******************
changed: [matrix.mydomain.com] => (item={'name': 'matrix-postgres.service', 'priority': 500, 'groups': ['matrix', 'postgres']})
failed: [matrix.mydomain.com] (item={'name': 'matrix-user-verification-service.service', 'priority': 800, 'groups': ['matrix', 'matrix-user-verification-service']}) => changed=false 
  ansible_loop_var: item
  item:
    groups:
    - matrix
    - matrix-user-verification-service
    name: matrix-user-verification-service.service
    priority: 800
  msg: |-
    Unable to start service matrix-user-verification-service.service: Job for matrix-user-verification-service.service failed because the control process exited with error code.
    See "systemctl status matrix-user-verification-service.service" and "journalctl -xe" for details.
changed: [matrix.mydomain.com] => (item={'name': 'matrix-synapse.service', 'priority': 1000, 'groups': ['matrix', 'homeservers', 'synapse']})
changed: [matrix.mydomain.com] => (item={'name': 'matrix-synapse-reverse-proxy-companion.service', 'priority': 1500, 'groups': ['matrix', 'homeservers', 'synapse', 'reverse-proxies']})
changed: [matrix.mydomain.com] => (item={'name': 'matrix-client-element.service', 'priority': 2000, 'groups': ['matrix', 'clients', 'element', 'client-element']})
changed: [matrix.mydomain.com] => (item={'name': 'matrix-mailer.service', 'priority': 2000, 'groups': ['matrix', 'mailer']})
changed: [matrix.mydomain.com] => (item={'name': 'matrix-container-socket-proxy.service', 'priority': 2900, 'groups': ['matrix', 'reverse-proxies', 'container-socket-proxy']})
changed: [matrix.mydomain.com] => (item={'name': 'matrix-nginx-proxy.service', 'priority': 3000, 'groups': ['matrix', 'nginx', 'reverse-proxies']})
changed: [matrix.mydomain.com] => (item={'name': 'matrix-traefik.service', 'priority': 3000, 'groups': ['matrix', 'traefik', 'reverse-proxies']})
[...]

Issue #2847, opened 2023-08-28
Unable to start service matrix-user-verification-service.service


ragnarblackmane:
**Describe the bug**
During installation of matrix-user-verification-service I have the following issue:
#2839 
then, after removing the following entries from setup_jitsi_auth_uvs_install.yml and running ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start:

  become: true
  become_user: "{{ jitsi_uid }}"

I have the following issue:

TASK [galaxy/com.devture.ansible.role.systemd_service_manager : Ensure systemd services are started] ****************************************************************************************
changed: [matrix.[REDACTEDDOMAIN]] => (item={'name': 'matrix-postgres.service', 'priority': 500, 'groups': ['matrix', 'postgres']})
failed: [matrix.[REDACTEDDOMAIN]] (item={'name': 'matrix-user-verification-service.service', 'priority': 800, 'groups': ['matrix', 'matrix-user-verification-service']}) => changed=false 
  ansible_loop_var: item
  item:
    groups:
    - matrix
    - matrix-user-verification-service
    name: matrix-user-verification-service.service
    priority: 800
  msg: |-
    Unable to start service matrix-user-verification-service.service: Job for matrix-user-verification-service.service failed because the control process exited with error code.
    See "systemctl status matrix-user-verification-service.service" and "journalctl -xeu matrix-user-verification-service.service" for details.

Checking through journalctl, the service starts and then stops after some seconds.
If I try to open a jitsi meet through Element, it does not authenticate.

**To Reproduce**
My `vars.yml` file looks like this:

matrix_domain: [REDACTED]
matrix_homeserver_implementation: synapse
matrix_homeserver_generic_secret_key: '[REDACTED]'
matrix_playbook_reverse_proxy_type: playbook-managed-traefik
devture_traefik_config_certificatesResolvers_acme_email: '[REDACTED]'
devture_postgres_connection_password: '[REDACTED]'
matrix_nginx_proxy_base_domain_serving_enabled: true
matrix_synapse_federation_enabled: false
jitsi_enabled: true
jitsi_enable_auth: true
jitsi_auth_type: matrix
matrix_user_verification_service_enabled: true
matrix_user_verification_service_uvs_access_token: "[REDACTED]"
matrix_user_verification_service_uvs_disable_ip_blacklist: true
jitsi_jvb_container_extra_arguments:
[...]

PR #2976, opened 2023-10-30
Fix matrix-user-verification-service service file

thiras:
Fixes #2375, #2975, and #2847

Credit should go to @crapo9000 as solved here https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2375#issuecomment-1655758296 . I've just created the PR according to their solution


spantaleev:
Thank you for wrapping this up! :bow:

PR #2979, opened 2023-11-01
draft encryption support for hookshot

HarHarLinks:
per the comment in main.yml, this requires some experimental synapse features to also be enabled, which the role does not do yet at the time of writing (94abf2d5bde63919c6b5597f3142eea5fed73815).
i manage that by setting (iirc) in my host vars:
```
matrix_synapse_configuration_extension_yaml: |
  experimental_features:
    msc2409_to_device_messages_enabled: true
    msc3202_device_masquerading: true
    msc3202_transaction_extensions: true
```

other than that, i tried this out at some point which worked for a while but eventually broke in a way that hookshot's messages were not decryptable anymore. i couldn't figure out what broke it and had no time to try it again, so i never contributed this patch. i think the feature is not quite stable yet, unless i made a mistake somewhere, which is for you to review. i am sharing it now out of interest from members in the hookshot room.

this PR is a draft for the above 2 reasons. feel free to take it as the base for something production ready.

### DIFF
--- a/docs/configuring-playbook-bridge-hookshot.md
+++ b/docs/configuring-playbook-bridge-hookshot.md
@@ -23,6 +23,11 @@ Other configuration options are available via the `matrix_hookshot_configuration
 
 Finally, run the playbook (see [installing](installing.md)).
 
+### End-to-bridge endcryption
+
+1. Enable by setting `: true`
+
+If the crypto store has become corrupted, reset it by running `ansible-playbook -i inventory/hosts setup.yml -K --tags=reset-hookshot-encryption`.
 
 ## Usage
 
--- a/docs/configuring-playbook-bridge-hookshot.md
+++ b/docs/configuring-playbook-bridge-hookshot.md
@@ -25,7 +25,7 @@ Finally, run the playbook (see [installing](installing.md)).
 
 ### End-to-bridge endcryption
 
-1. Enable by setting `: true`
+1. Enable by setting `matrix_hookshot_experimental_encryption_enabled: true`
 
 If the crypto store has become corrupted, reset it by running `ansible-playbook -i inventory/hosts setup.yml -K --tags=reset-hookshot-encryption`.
 
--- a/roles/custom/matrix-bridge-hookshot/tasks/reset_encryptioon.yml
+++ b/roles/custom/matrix-bridge-hookshot/tasks/reset_encryptioon.yml
@@ -1,0 +1,12 @@
+---
+- name: Resetting Hookshot's crypto store
+  ansible.builtin.command:
+    cmd: |
+      {{ devture_systemd_docker_base_host_command_docker }} run
+      --rm
+      --name={{ matrix_hookshot_container_url }}-reset-crypto
+      --user={{ matrix_user_uid }}:{{ matrix_user_gid }}
+      --cap-drop=ALL
+      -v {{ matrix_hookshot_base_path }}/config.yml:/config.yml
+      {{ matrix_hookshot_docker_image }} yarn start:resetcrypto
+  changed_when: false
--- a/roles/custom/matrix-bridge-hookshot/templates/config.yml.j2
+++ b/roles/custom/matrix-bridge-hookshot/templates/config.yml.j2
@@ -107,6 +107,14 @@ metrics:
   # (Optional) Prometheus metrics support
   #
   enabled: {{ matrix_hookshot_metrics_enabled | to_json }}
+{% if matrix_hookshot_experimental_encryption_enabled %}
+queue:
+  monolithic: true
+  port: 6379
+  host: matrix-redis
+experimentalEncryption:
+  storagePath: /data/encryption
+{% endif %}
 logging:
   # (Optional) Logging settings. You can have a severity debug,info,warn,error
   #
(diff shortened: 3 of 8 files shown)


roles/custom/matrix-bridge-hookshot/tasks/reset_encryptioon.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Resetting Hookshot's crypto store
+  ansible.builtin.command:
+    cmd: |
+      {{ devture_systemd_docker_base_host_command_docker }} run
+      --rm
+      --name={{ matrix_hookshot_container_url }}-reset-crypto
```
spantaleev, line 7: It's quite unfortunate that the container-name (and what we call `_ident` in our roles) is called `matrix_hookshot_container_url` here, making you think it's `http://something-something` -- a value unsuitable for a container name.

In actuality, `matrix_hookshot_container_url` is just `matrix-hookshot` - a container name / identifier.

roles/custom/matrix-bridge-hookshot/templates/config.yml.j2
```diff
@@ -107,6 +107,14 @@ metrics:
   # (Optional) Prometheus metrics support
   #
   enabled: {{ matrix_hookshot_metrics_enabled | to_json }}
+{% if matrix_hookshot_experimental_encryption_enabled %}
+queue:
+  monolithic: true
+  port: 6379
+  host: matrix-redis
```
spantaleev, line 114: These assumptions should not be hardcoded and should use variables (`matrix_hookshot_queue_host` and `matrix_hookshot_queue_port`).

The port could default to `6379`, since that's quite standard for Redis.

The host should default to empty. Perhaps you can only enable the queue (`if` statement) if the host is non-empty. Example: `{% if matrix_hookshot_queue_host %}`

The `matrix_hookshot_queue_host` and `matrix_hookshot_queue_port` variables can be set to the correct values (`matrix-redis`, ..) via `group_vars/matrix_servers` - the file we use for wiring the different roles together. It's better to explicitly wire things via this central file, instead of having roles randomly assume things about other roles or each into each other's variables.

`matrix_hookshot_queue_host` would be defined as such in `group_vars/matrix_servers`: 

```yml
matrix_hookshot_queue_host: "{{ redis_identifier if matrix_hookshot_experimental_encryption_enabled and redis_enabled else '' }}"
```

This: provides sane defaults for these variables, potentially overrides them via the group vars file.. and still allows people to use external Redis instances, if they want to.

roles/custom/matrix-bridge-hookshot/templates/config.yml.j2
```diff
@@ -107,6 +107,14 @@ metrics:
...
+experimentalEncryption:
+  storagePath: /data/encryption
```
spantaleev, line 116: This should be out of the `if` that deals with the `queue` stuff, since one may wish to enable the queue without enabling experimental encryption.

Experimental encryption does probably require the `queue` configuration, but I suppose Hookshot checks for this by itself. You may add a check in `validate_config.yml` if you're feeling like it though.

`experimentalEncryption` settings should be dependant on `{% if matrix_hookshot_experimental_encryption_enabled %}`.

roles/custom/matrix-bridge-hookshot/templates/systemd/matrix-hookshot.service.j2
```diff
@@ -30,6 +30,12 @@ ExecStart={{ devture_systemd_docker_base_host_command_docker }} run --rm --name
           {% endfor %}
           {{ matrix_hookshot_docker_image }}
 
+{% if matrix_hookshot_experimental_encryption_enabled %}
+ExecStartPre={{ devture_systemd_docker_base_host_command_docker }} network connect matrix-redis {{ matrix_hookshot_container_url }}
+{% endif %}
```
spantaleev, line 35: As discussed before, it's not great for roles to make assumptions about other roles.

The assumption that this role makes there is that when encryption is enabled, Redis will also be enabled via nothing else but `matrix-redis`.

-----

It's better to redo this by:

- adding a new  `matrix_hookshot_container_additional_networks` variable, which can connect the Hookshot container to any arbitrary additional network. In some roles, we even split this into multiple lists of networks (multiple variables) in `defaults/main.yml`:
  - `matrix_hookshot_container_additional_networks: "{{ matrix_hookshot_container_additional_networks_auto + matrix_hookshot_container_additional_networks_custom }}"`
  - `matrix_hookshot_container_additional_networks_auto: []`
  - `matrix_hookshot_container_additional_networks_custom: []`
  - With the above, `matrix_hookshot_container_additional_networks_auto` is supposed to be managed by the playbook, while `matrix_hookshot_container_additional_networks_custom` is supposed to be managed by the user
- `matrix_hookshot_container_additional_networks_auto` may be redefined in `group_vars/matrix_servers` to inject `redis_container_network` (equal to `matrix-redis`) into the list of networks under certain conditions (if `matrix_hookshot_queue_host == redis_identifier and redis_enabled`). You can find various examples of `_additional_networks` redefinitions in `group_vars/matrix_servers`
- redoing this systemd `.service` file, so that it loops over the list of networks (`matrix_hookshot_container_additional_networks`) and connects to them all

With this, the role is configurable and does not hardcode anything about `matrix-redis`. All wiring happens in `group_vars/matrix_servers` and people can do their own network wiring or use external Redis if they need to.

spantaleev (changes requested):

docs/configuring-playbook-bridge-hookshot.md
```diff
@@ -23,6 +23,11 @@ Other configuration options are available via the `matrix_hookshot_configuration
 
 Finally, run the playbook (see [installing](installing.md)).
 
+### End-to-bridge endcryption
```
roland-5, line 26: Typo: endcryption -> encryption

PR #2982, opened 2023-11-06
Remove mention of Android Element X being less feature complete than the iOS version

kuba-orlik:
Quoting upstream:

> Element X Android and Element X iOS apps are in a similar state.
> 
> https://github.com/vector-im/element-x-android/issues/911


spantaleev:
Given that the previous line (the one about Element X iOS) points to a place where you can download it (Test Flight), I believe we'd better do the same for Element X Android instead of removing all mentions of Element X Android.

kuba-orlik:
@spantaleev good point, changed that!

PR #2984, opened 2023-11-07
add self-build for matrix_user_verification

Chikage0o0:
Currently v3.0.0 tested with no issues.
So remove matrix_user_verification_service_docker_image from groups_vars.

/.npm must be writable or an error will be reported.


spantaleev:
Thank you!

Issue #2954, opened 2023-10-19
user_directory_search redirect not working anymore


janonym1:
**Describe the bug**
Hi!
We have `matrix_nginx_proxy_proxy_matrix_user_directory_search_enabled: true` set because we have to run our own search service (since we have a complicated LDAP environment), which runs as a service in the background and accepts search requests at port 8088. The search [backend service](https://github.com/maxmalek/ugreg/tree/master/src/maiden) runs fine and I can call the search API over curl as always. 

However, it seems that the routing/redirecting of the API with `/_matrix/client/r0/user_directory/search` is not working anymore. I am not sure when exactly this happened but it must have been in the last few months tops as it worked with the same config before that.

**To Reproduce**
My related config in the `vars.yml` file looks like this:

```yaml
#Controls whether proxying for the User Directory Search API (`/_matrix/client/r0/user_directory/search`) should be done (on the matrix domain).
# This can be used to forward the API endpoint to another service, augmenting the functionality of Synapse's own User Directory Search.
matrix_nginx_proxy_proxy_matrix_user_directory_search_enabled: true
matrix_nginx_proxy_proxy_matrix_user_directory_search_addr_sans_container: "127.0.0.1:8088"
matrix_nginx_proxy_proxy_matrix_user_directory_search_addr_with_container: "1.2.3.4:8088"
matrix_synapse_reverse_proxy_companion_container_client_api_host_bind_port: "1.2.3.4:8008"
# Controls whether the matrix-synapse-reverse-proxy-companion container exposes its HTTP Client-Server API port (tcp/8008 in the container).
# Takes an "<ip>:<port>" or "<port>" value (e.g. "127.0.0.1:8008"), or empty string to not expose.
matrix_synapse_reverse_proxy_companion_container_client_api_host_bind_port: "1.2.3.4:8008"
```
Note: we needed `client_api_host_bind_port` for our search workflow

Previously this worked fine but with the latest updates, this redirection to our search service broke somehow. I assume it has to do with the ongoing migration over to traefik and away from nginx but I am not sure, where exactly I should try define the new redirection. 
We are using traefik `matrix_playbook_reverse_proxy_type: playbook-managed-traefik` but there are still some services running using the nginx proxy (e.g. oidc, basic auth, metrics..)

[...]

PR #2985, opened 2023-11-08
Update matrix-domain.conf.j2 - trying to fix issue #2954

jolly-jump:
exchanged `^~` with `~` as a pattern matching in the location part. I am very sure, that it only works using `~`. I am not quite sure though, if this is the right way to do it, because `~` is probably more expensive than `^~`

the rewrite has to be behind the definition of the $backend. Otherwise nginx will fail to work. This is probably because "break" goes directly to the proxy_pass which uses $backend.

### DIFF
--- a/roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
+++ b/roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
@@ -119,7 +119,7 @@
 	# Redirect other endpoints registered by the media-repo to its container
 	# /_matrix/client/r0/logout
 	# /_matrix/client/r0/logout/all
-	location ^~ /_matrix/client/(r0|v1|v3|unstable)/(logout|logout/all) {
+	location ~ /_matrix/client/(r0|v1|v3|unstable)/(logout|logout/all) {
 		{% if matrix_nginx_proxy_enabled %}
 			{# Use the embedded DNS resolver in Docker containers to discover the service #}
 			resolver {{ matrix_nginx_proxy_http_level_resolver }} valid=5s;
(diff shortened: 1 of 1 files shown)


spantaleev:
I suppose we need the same `if` (and `rewrite` block) once again a bit below then, to cover the `else` statement?

roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
```diff
@@ -207,16 +209,18 @@
...
 		{% if matrix_nginx_proxy_enabled %}
 			{# Use the embedded DNS resolver in Docker containers to discover the service #}
 			resolver {{ matrix_nginx_proxy_http_level_resolver }} valid=5s;
 			set $backend "{{ matrix_nginx_proxy_proxy_matrix_3pid_registration_addr_with_container }}";
+   		    {% if matrix_nginx_proxy_proxy_matrix_3pid_registration_v3_to_r0_redirect_enabled %}
+            rewrite ^(.*?)/v3/(.*?)$  $1/r0/$2 break;
+	        {% endif %}
 			proxy_pass http://$backend;
 		{% else %}
+            {% if matrix_nginx_proxy_proxy_matrix_user_directory_search_v3_to_r0_redirect_enabled %}
```
jolly-jump, line 221: mistake, will fix

jolly-jump:
Hi. So I did what you suggested.
This PR works now as expected for user_directory search.
Do you want me to rebase and make a new PR?
Pretty messy six commits for the small change :shrug: ?

spantaleev:
Thanks for fixing it! I've squashed it, so it's clean enough now.

PR #3040, opened 2023-12-14
Update README.md: add mash-playbook, remove archived playbooks

joecool1029:
Notced the old playbooks were finally archived. Figured it would be a good time to remove them and replace with the successor in the README.md


spantaleev:
Makes sense to do it! I've reworded the text to make it flow better

PR #3042, opened 2023-12-14
Hookshot: Add support for encryption

real-joshua:
This is a completion/improvement of https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2979 by @HarHarLinks.

- Naming of variables is now more consistent with the rest of the Playbook
- Message queue can now be enabled seperately without enabling encryption support as well
- Networks are now a list which gets loop over


HarHarLinks:
Thank you for picking this up!

roles/custom/matrix-bridge-hookshot/templates/config.yml.j2
```diff
@@ -107,6 +107,16 @@ metrics:
   # (Optional) Prometheus metrics support
   #
   enabled: {{ matrix_hookshot_metrics_enabled | to_json }}
+{% if matrix_hookshot_queue_host is defined and matrix_hookshot_queue_host|d('')|length > 0 %} %}
+queue:
+  monolithic: true
+  port: {{ matrix_hookshot_queue_port | default('6379') }}
+  host: {{ matrix_hookshot_queue_host }}
+{& endif %}
```
spantaleev, line 115:
```suggestion
{% endif %}
```

roles/custom/matrix-bridge-hookshot/defaults/main.yml
```diff
@@ -30,6 +30,10 @@ matrix_hookshot_public_endpoint: /hookshot
 matrix_hookshot_appservice_port: 9993
 matrix_hookshot_appservice_endpoint: "{{ matrix_hookshot_public_endpoint }}/_matrix/app"
 
+# Controls whether the experimental end-to-bridge encryption support is enabled.
+# This requires that support is also enabled in the homeserver, see the hookshot docs.
```
spantaleev, line 34:
```suggestion
# This requires that:
# - support is also enabled in the homeserver, see the hookshot docs.
# - Hookshot is pointed to a Redis instance via the `matrix_hookshot_queue_*` variables.
```

roles/custom/matrix-bridge-hookshot/templates/config.yml.j2
```diff
@@ -107,6 +107,16 @@ metrics:
   # (Optional) Prometheus metrics support
   #
   enabled: {{ matrix_hookshot_metrics_enabled | to_json }}
+{% if matrix_hookshot_queue_host is defined and matrix_hookshot_queue_host|d('')|length > 0 %} %}
+queue:
+  monolithic: true
+  port: {{ matrix_hookshot_queue_port | default('6379') }}
+  host: {{ matrix_hookshot_queue_host }}
```
spantaleev, line 114:
```suggestion
{% if matrix_hookshot_queue_host != '' %}
queue:
  monolithic: true
  port: {{ matrix_hookshot_queue_port }}
  host: {{ matrix_hookshot_queue_host | to_json }}
```

----

Reasoning:

For the playbook, we always deal variables which are defined and have some default value (even if empty) in the `default/main.yml` file.

This makes `default/main.yml` (well, and sometimes `vars/vars.yml`) a source from which one can discover all variables for a given component.

It seems like `matrix_hookshot_queue_host` and `matrix_hookshot_queue_port` port are undefined by default and this code doing `is defined` if-checks. It's better if we **explicitly define these variables** (see the next part of my comment here) and adjust this code as suggested.

-----

We should **also add something like this to `defaults/main.yml`**:

```yml
# The variables below control the queue parameters and may optionally be pointed to a Redis instance.
# These are required when experimental encryption is enabled (`matrix_hookshot_experimental_encryption_enabled`).
matrix_hookshot_queue_host: ''
matrix_hookshot_queue_port: 6379
```
-----

We also need to adjust `group_vars/matrix_servers`, so that appropriate values are auto-assigned to these when Redis is available for use.

Perhaps by adding something lke this to the `# matrix-bridge-hookshot` section in  `group_vars/matrix_servers`:

```yml
# Hookshot's experimental encryption feature (and possibly others) may benefit from Redis, if available.
matrix_hookshot_queue_host: "{{ redis_identifier if redis_enabled else '' }}"

matrix_hookshot_container_additional_networks_auto: |
  {{
    ([redis_container_network] if redis_enabled and matrix_hookshot_queue_host == redis_identifier else [])
  }}
```

**Note**: I believe `matrix_hookshot_container_additional_networks_auto` does not exist in the Hookshot role yet, so we need to add it to `defaults/main.yml` and make use of it in the systemd service.

We need this, because Redis lives in its own separate container network (`matrix-redis`) tha Hookshot (`matrix-hookshot`), so they wouldn't be able to talk to one another.

----

Also, in `group_vars/matrix_servers`, we need to list Redis as a dependency for Hookshot:

```diff
 matrix_hookshot_systemd_wanted_services_list: |
   {{
     (['matrix-' + matrix_homeserver_implementation + '.service'])
     +
     (['matrix-nginx-proxy.service'] if matrix_nginx_proxy_enabled else [])
+    +
+    ([(redis_identifier + '.service')] if redis_enabled and matrix_hookshot_queue_host == redis_identifier else [])
   }}
```

-----

Also, given that `matrix_hookshot_queue_host` is required when encryption is enabled (`matrix_hookshot_experimental_encryption_enabled`), we should add validation checks in `tasks/validate_config.yml`. Perhaps something like this:

```yml
- name: Fail if no Redis queue enabled when Hookshot encryption is enabled
  ansible.builtin.fail:
    msg: >-
      You need to define a required configuration setting (`{{ item }}`) to enable Hookshot encryption.
  when: "matrix_hookshot_experimental_encryption_enabled and vars[item] == ''"
  with_items:
    - "matrix_hookshot_queue_host"
```

roles/custom/matrix-bridge-hookshot/templates/systemd/matrix-hookshot.service.j2
```diff
@@ -30,11 +28,18 @@ ExecStart={{ devture_systemd_docker_base_host_command_docker }} run --rm --name
           {% endfor %}
           {{ matrix_hookshot_docker_image }}
 
-ExecStop=-{{ devture_systemd_docker_base_host_command_docker }} stop --time={{ devture_systemd_docker_base_container_stop_grace_time_seconds }} {{ matrix_hookshot_container_url }}
-ExecStop=-{{ devture_systemd_docker_base_host_command_docker }} rm {{ matrix_hookshot_container_url }}
+{% if matrix_hookshot_container_additional_networks %}
+{% for network in matrix_hookshot_container_additional_networks %}ExecStartPre={{ devture_systemd_docker_base_host_command_docker }} network connect {{ network ]] {{ matrix_hookshot_container_ident }}{% endfor %}
+{% endif %}
```
spantaleev, line 33:
```suggestion
{% for network in matrix_hookshot_container_additional_networks %}ExecStartPre={{ devture_systemd_docker_base_host_command_docker }} network connect {{ network ]] {{ matrix_hookshot_container_ident }}{% endfor %}
```

It's simpler to not have the `{% if .. %}`.

As discussed before, all roles deal with variables that must be defined and have some default value in `defaults/main.yml`. I don't see `matrix_hookshot_container_additional_networks` defined yet.

Another review comment of mine discussed adjusting `group_vars/matrix_servers` to re-define `matrix_hookshot_container_additional_networks_auto`, so let's prepare these variables and their defaults by adding this to `defaults/main.yml`:

```yml
# Specifies additional networks that the Hookshot container would be connected to
matrix_hookshot_container_additional_networks: "{{ matrix_hookshot_container_additional_networks_auto + matrix_hookshot_container_additional_networks_custom }}"
matrix_hookshot_container_additional_networks_auto: []
matrix_hookshot_container_additional_networks_custom: []
```

You can add this new block somewhere around the other `matrix_hookshot_container_*` variables in `defaults/main.yml`

roles/custom/matrix-bridge-hookshot/tasks/main.yml
```diff
@@ -9,6 +9,12 @@
     - when: matrix_hookshot_enabled | bool
       ansible.builtin.include_tasks: "{{ role_path }}/tasks/inject_into_nginx_proxy.yml"
 
+- tags:
+    - reset-hookshot-encryption
+  block:
+    - when: matrix_hookshot_enabled | bool
+      ansible.builtin.include_tasks: "{{ role_path }}/tasks/reset_encryption.yml"
```
spantaleev, line 16: As of right now, your file is named `reset_encryptioon.yml` (double `o`), while you're including `reset_encryption.yml` here.

The file needs to be renamed.

roles/custom/matrix-bridge-hookshot/tasks/reset_encryptioon.yml
```diff
@@ -0,0 +1,12 @@
...
+- name: Resetting Hookshot's crypto store
+  ansible.builtin.command:
+    cmd: |
+      {{ devture_systemd_docker_base_host_command_docker }} run
+      --rm
+      --name={{ matrix_hookshot_container_ident }}-reset-crypto
+      --user={{ matrix_user_uid }}:{{ matrix_user_gid }}
+      --cap-drop=ALL
+      -v {{ matrix_hookshot_base_path }}/config.yml:/config.yml
+      {{ matrix_hookshot_docker_image }} yarn start:resetcrypto
```
spantaleev, line 11:
```suggestion
      --mount type=bind,src={{ matrix_hookshot_base_path }}/config.yml,dst=/config.yml
      {{ matrix_hookshot_docker_image }}
     yarn start:resetcrypto
```

`--mount` is less magical than `-v`, so it's safer to use. `-v` auto-creates source mount points (`{{ matrix_hookshot_base_path }}/config.yml`) as directories if they do not exist yet for some reason, which can be quite harmful. We should avoid using `-v`.

Also, we may add `,ro` to the `--mount` line, as I suppose the container will not need to modify the configuration.

docs/configuring-playbook-bridge-hookshot.md
```diff
@@ -23,6 +23,11 @@ Other configuration options are available via the `matrix_hookshot_configuration
 
 Finally, run the playbook (see [installing](installing.md)).
 
+### End-to-bridge endcryption
+
+1. Enable by setting `matrix_hookshot_experimental_encryption_enabled: true`
+
+If the crypto store has become corrupted, reset it by running `ansible-playbook -i inventory/hosts setup.yml -K --tags=reset-hookshot-encryption`.
```
spantaleev, line 30:
```suggestion
You can enable [experimental encryption](https://matrix-org.github.io/matrix-hookshot/latest/advanced/encryption.html) for Hookshot by adding `matrix_hookshot_experimental_encryption_enabled: true` to your configuration (`vars.yml`) and [re-running the playbook](installing.md).

If the crypto store ever becomes corrupted, reset it by running the `reset-hookshot-encryption` tag for this Ansible playbook (e.g. `ansible-playbook -i inventory/hosts setup.yml -K --tags=reset-hookshot-encryption`)
```

spantaleev (changes requested):

spantaleev:
Thank you for picking this up! I've posted a detailed review which should help make it ready for merge

roles/custom/matrix-bridge-hookshot/templates/config.yml.j2
real-joshua: thanks

roles/custom/matrix-bridge-hookshot/defaults/main.yml
real-joshua: thanks

roles/custom/matrix-bridge-hookshot/tasks/main.yml
real-joshua: thanks

roles/custom/matrix-bridge-hookshot/tasks/reset_encryptioon.yml
real-joshua: thanks

docs/configuring-playbook-bridge-hookshot.md
real-joshua: Thanks, I've added something like that with similar wording.

roles/custom/matrix-bridge-hookshot/templates/systemd/matrix-hookshot.service.j2
real-joshua: thanks

real-joshua:
> Thank you for picking this up! I've posted a detailed review which should help make it ready for merge

Thanks, @spantaleev! I think I've applied everything, let me know if there is still anything amiss. :)

spantaleev:
There were a lot of things missing and various things to fix - YAML syntax problems, etc.

It makes me wonder if the end result was tested at all or done blindly? The Hookshot container was not being connected to any network.. The syntax for connecting it to additional networks was broken, etc.

The reset-encryption command had a broken YAML syntax.

`matrix_hookshot_container_ident` was an undefined variable.

Network connectivity seems problematic for some reasons abnd Hookshot cannot connect to Redis (at least one my machine), hence me disabling it (dbf1a685bf41df1fac) for people who haven't explicitly enabled encryption.

This PR couldn't have possibly worked without all these fixups.. And even now, it still likely has some issues (network connectivity between Hookshot and Redis).

real-joshua:
As you can see, this is still a draft. I am working on implementing your feedback. My next step would be to get a VPS and DNS set up so I can test this without doing that on my production setup.

Once I am satisfied I will remove the "draft" marking and proceed.

PR #3045, opened 2023-12-16
MMR reverse proxy updates

Michael-Hollister:
Changes:
* Added Traefik support for the MMR role
* Fixed federation media download failures when using NGINX
* Changed internal MXC link format from `mxc://matrix.DOMAIN/hash` to `mxc://DOMAIN/hash`

### DIFF
--- a/roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
+++ b/roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
@@ -110,7 +110,7 @@
 		# Make sure this matches your homeserver in media-repo.yaml
 		# You may have to manually specify it if using delegation or the
 		# incoming Host doesn't match.
-		proxy_set_header Host $host;
+		proxy_set_header Host {{ matrix_domain }};
 
 		proxy_set_header X-Real-IP $remote_addr;
 		proxy_set_header X-Forwarded-For $remote_addr;
@@ -136,7 +136,7 @@
 		# Make sure this matches your homeserver in media-repo.yaml
 		# You may have to manually specify it if using delegation or the
 		# incoming Host doesn't match.
-		proxy_set_header Host $host;
+		proxy_set_header Host {{ matrix_domain }};
 
 		proxy_set_header X-Real-IP $remote_addr;
 		proxy_set_header X-Forwarded-For $remote_addr;
@@ -159,7 +159,7 @@
 		# Make sure this matches your homeserver in media-repo.yaml
 		# You may have to manually specify it if using delegation or the
 		# incoming Host doesn't match.
-		proxy_set_header Host $host;
+		proxy_set_header Host {{ matrix_domain }};
 
 		proxy_set_header X-Real-IP $remote_addr;
 		proxy_set_header X-Forwarded-For $remote_addr;
@@ -180,7 +180,7 @@
 		# Make sure this matches your homeserver in media-repo.yaml
 		# You may have to manually specify it if using delegation or the
 		# incoming Host doesn't match.
-		proxy_set_header Host $host;
+		proxy_set_header Host {{ matrix_domain }};
 
 		proxy_set_header X-Real-IP $remote_addr;
 		proxy_set_header X-Forwarded-For $remote_addr;
@@ -406,7 +406,7 @@ server {
 			# Make sure this matches your homeserver in media-repo.yaml
 			# You may have to manually specify it if using delegation or the
 			# incoming Host doesn't match.
-			proxy_set_header Host $host;
+			proxy_set_header Host {{ matrix_domain }};
 
 			proxy_set_header X-Real-IP $remote_addr;
 			proxy_set_header X-Forwarded-For $remote_addr;
@@ -432,7 +432,7 @@ server {
 			# Make sure this matches your homeserver in media-repo.yaml
 			# You may have to manually specify it if using delegation or the
 			# incoming Host doesn't match.
-			proxy_set_header Host $host;
+			proxy_set_header Host {{ matrix_domain }};
 
 			proxy_set_header X-Real-IP $remote_addr;
 			proxy_set_header X-Forwarded-For $remote_addr;
@@ -455,7 +455,7 @@ server {
 			# Make sure this matches your homeserver in media-repo.yaml
 			# You may have to manually specify it if using delegation or the
 			# incoming Host doesn't match.
-			proxy_set_header Host $host;
+			proxy_set_header Host {{ matrix_domain }};
 
 			proxy_set_header X-Real-IP $remote_addr;
 			proxy_set_header X-Forwarded-For $remote_addr;
@@ -476,7 +476,7 @@ server {
 			# Make sure this matches your homeserver in media-repo.yaml
 			# You may have to manually specify it if using delegation or the
 			# incoming Host doesn't match.
-			proxy_set_header Host $host;
+			proxy_set_header Host {{ matrix_domain }};
 
 			proxy_set_header X-Real-IP $remote_addr;
 			proxy_set_header X-Forwarded-For $remote_addr;
--- a/roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
+++ b/roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
@@ -390,6 +390,99 @@ server {
 		ssl_session_timeout {{ matrix_nginx_proxy_ssl_session_timeout }};
 	{% endif %}
 
+	{% if matrix_nginx_proxy_proxy_media_repo_enabled %}
+		# Redirect all media endpoints to the media-repo
+		location ^~ /_matrix/media {
+			{% if matrix_nginx_proxy_enabled %}
+				{# Use the embedded DNS resolver in Docker containers to discover the service #}
+				resolver {{ matrix_nginx_proxy_http_level_resolver }} valid=5s;
+				set $backend "{{ matrix_nginx_proxy_proxy_media_repo_addr_with_container }}";
+				proxy_pass http://$backend;
+			{% else %}
+				{# Generic configuration for use outside of our container setup #}
+				proxy_pass http://{{ matrix_nginx_proxy_proxy_media_repo_addr_sans_container }};
+			{% endif %}
+
+			# Make sure this matches your homeserver in media-repo.yaml
+			# You may have to manually specify it if using delegation or the
+			# incoming Host doesn't match.
+			proxy_set_header Host $host;
+
+			proxy_set_header X-Real-IP $remote_addr;
+			proxy_set_header X-Forwarded-For $remote_addr;
+
+			client_body_buffer_size {{ ((matrix_media_repo_max_bytes | int) / 4) | int }};
+			client_max_body_size {{ matrix_media_repo_max_bytes }};
+		}
+
+		# Redirect other endpoints registered by the media-repo to its container
+		# /_matrix/client/r0/logout
+		# /_matrix/client/r0/logout/all
+		location ~ ^/_matrix/client/(r0|v1|v3|unstable)/(logout|logout/all) {
+			{% if matrix_nginx_proxy_enabled %}
+				{# Use the embedded DNS resolver in Docker containers to discover the service #}
+				resolver {{ matrix_nginx_proxy_http_level_resolver }} valid=5s;
+				set $backend "{{ matrix_nginx_proxy_proxy_media_repo_addr_with_container }}";
+				proxy_pass http://$backend;
+			{% else %}
+				{# Generic configuration for use outside of our container setup #}
+				proxy_pass http://{{ matrix_nginx_proxy_proxy_media_repo_addr_sans_container }};
+			{% endif %}
+
+			# Make sure this matches your homeserver in media-repo.yaml
+			# You may have to manually specify it if using delegation or the
+			# incoming Host doesn't match.
+			proxy_set_header Host $host;
+
+			proxy_set_header X-Real-IP $remote_addr;
+			proxy_set_header X-Forwarded-For $remote_addr;
+		}
+
+		# Redirect other endpoints registered by the media-repo to its container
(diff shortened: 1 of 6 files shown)


roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2
```diff
@@ -390,6 +390,99 @@ server {
 		ssl_session_timeout {{ matrix_nginx_proxy_ssl_session_timeout }};
 	{% endif %}
 
+	{% if matrix_nginx_proxy_proxy_media_repo_enabled %}
+		# Redirect all media endpoints to the media-repo
+		location ^~ /_matrix/media {
```
spantaleev, line 395: Looks like you're adding serving for `/_matrix/media` endpoints in the vhost which serves federation traffic.

Why is this necessary?

Looking at thte [Server-Server API specs](https://spec.matrix.org/v1.9/server-server-api/), there are no `/_matrix/media` endpoints described. I wonder if this is here by mistake.

The same nginx `location` blocks appear in the other vhost in the same file (above), which handles client-server traffic. Those seems fine, but I'm not sure why all these client-related APIs are being added to the federation vhost here.

Michael-Hollister: Adding the endpoint redirection for the federation vhost is done to the fix issue where clients from other homeservers can not download media from the MMR. If not added, download requests from clients from other homeservers seem to be routed to the matrix server/synapse reverse proxy companion service instead, which results in the media not being found and downloads failing.

Per documentation of the MMR nginx setup, it states that the `/_matrix/media` redirection directive should be added to the `server` block for both client and federation requests: https://docs.t2bot.io/matrix-media-repo/deployment/nginx.html.

As you mention however it does look like there is no federation endpoints for media endpoints, so perhaps there may be another way to fix this issue. I'll take another look at this after fixing Traefik proxy setup with these changes.

venimus: can confirm that media federation is not working without changes

Michael-Hollister: Alright, this is ready for review now.

In regards to why `/_matrix/media` and related endpoints have to be added to the federation blocks, I think MMR adds the federation port when downloading remote media via federation: https://github.com/MatMaul/matrix-media-repo/blob/ae254aca90bd1074774e3b7b7fc2df24cce4416e/controllers/download_controller/download_resource_handler.go#L192C1-L199C62.

spantaleev:
I've considered doing a `group_vars/matrix_servers` change like this:

```diff
-matrix_nginx_proxy_proxy_media_repo_enabled: "{{ matrix_media_repo_enabled }}"
+matrix_nginx_proxy_proxy_media_repo_enabled: "{{ matrix_media_repo_enabled and matrix_playbook_reverse_proxy_type in ['playbook-managed-nginx', 'other-nginx-non-container'] }}"
```

.. which is consistent with what we do for other services which gain native Traefik support.
A service which is directly exposed at the Traefik level does not need all the old nginx configuration anymore.

----

However.. I think that **doing this for `matrix-media-repo` would break stuff**.

Various bridges and services which need access to the home server currently do it locally, by talking to `matrix_homeserver_container_url`.

As per its default definition of `matrix_homeserver_container_url`, communication with the homeserver usually goes through `matrix-nginx-proxy` (`http://matrix-nginx-proxy:12080`):

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/3da4c66b8541df09d3487740ece8fd9f79cec645/group_vars/matrix_servers#L450-L457

When `matrix-nginx-proxy` is disabled, it would go straight to the homeserver (e.g. `http://matrix-synapse-reverse-proxy-companion:8048` or `http://matrix-dendrite:PORT`, etc.).

That's all good when you need to talk to the homeserver directly, but.. `matrix-media-repo` needs to "steal" some media routes from the homeserver in all cases. It already has priority for these routes when `matrix.DOMAIN` is being accessed publicly (via Traefik), but if a bot/bridge talks directly to the homeserver, then `/_matrix/media` calls would, naturally, hit the homeserver.

Bots/bridges talking directly to the homeserver is not what we want to happen when it comes to media APIs. We want them to go through an intermediate place, where `matrix-media-repo` (and potentially other services) can steal the routes they care about. Other services which like stealing routes from the homeserver are [matrix-corporal](https://github.com/devture/matrix-corporal), identity servers (although those are supposedly going away, ..). For some of these services, bots/bridges potentially skipping and talking directly to the homeserver may be okay, but maybe not for others. Perhaps this "stealing routes" feature is important and here to stay.

If bots/bridges use the public address (`https://matrix.DOMAIN`), then we're fine as Traefik labels for `matrix-media-repo` would take priority, but:

- it's preferable if services talk locally, if possible
- using the public address sometimes causes routing issues, from what I've seen. I suppose that even with a public address, services may still be talking locally sometimes.

I believe we should **revisit this "public address" possibility, before proceding with stuff below**.

If we can't switch to public addresses, then this is a problem for us to solve. In the end, we'd like to completely get rid of `matrix-nginx-proxy` some day.. However, it seems like we'd need to introduce some new sort of intermediate service that sits in front of the homeserver and allows routes to be stolen from it.

----

I suppose we may need to create one new reverse-proxy instance (e.g. `matrix-homeserver`), which does part of the work `matrix-nginx-proxy` does for `matrix-domain.conf`.

That is, this new `matrix-homeserver` service would act as a front to whatever homeserver software is enabled right now - for exampe `matrix-synapse-reverse-proxy-companion` (which wraps Synapse and all its workers in a neat package) or to `matrix-dendrite`, etc.

It sucks to have to introduce an additional reverse-proxy and that MMR cannot be "Traefik only", unless briges/bots start talking publicly to the homeserver (discussed above).

The new `matrix-homeserver` reverse-proxy instance would most likely be an nginx instance as well.

The flow, at least initially, for Synapse would be: (user agent -> Traefik -> matrix-nginx-proxy -> matrix-homeserver -> matrix-reverse-proxy-companion -> matrix-synapse).
That's 4 reverse-proxies before ending up at Synapse.

`matrix-nginx-proxy` would ultimately disappear (perhaps as soon as `matrix-homeserver` is introduced), so we'd end up at the same place with regard to the total number of reverse-proxies.

----

Thoughts are welcome!

Michael-Hollister:
Is there still a need to have `matrix-synapse-reverse-proxy-companion` if the `matrix-homeserver` proxy is created? From what I understand about the reasoning why the proxy companion was created  (https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/CHANGELOG.md#2022-11-20), the playbook's front-facing proxy (Traefik/`matrix-nginx-proxy`/custom proxy) does not handle synapse worker endpoint routing, so the issue `matrix-synapse-reverse-proxy-companion` solves would not apply anymore. The ideal configuration then would be simpler with only two reverse proxies needed  (user agent -> Traefik -> matrix-homeserver).

spantaleev:
After I have thought about it some more over the past days, I am now imagining that `matrix-homeserver` would only be used internally - for bots and bridges to reach the homeserver over the container network.

And that all services would register themselves directly with Traefik for the least amount of "reverse-proxy overhead" and least complexity.

Thus, the flow (from outside the server) would become:

- (if media repo is enabled) Traefik -> matrix-media-repo for `/_matrix/media` routes, etc.
- (if an identity server is enabled) Traefik -> identity server for `/_matrix/identity` routes, etc.
- (if Synapse workers are enabled) Traefik -> matrix-synapse-reverse-proxy-companion for all (remaining) `/_matrix` routes -> specific matrix-synapse worker
- (if Synapse workers are not enabled) Traefik -> matrix-synapse for all (remaining) `/_matrix` routes
- (if e.g. Dendrite is used) Traefik -> matrix-dendrite for all (remaining) `/_matrix` routes
----

This leaves `matrix-homeserver` as a gateway for accessing the homeserver for traffic originating from bridges, etc., with the goal of providing a way to "steal some routes" from it.

Certain cervices (like MMR, matrix-corporal) would like to steal various routes and not let them reach the homeserver directly. For this reason, bots should not be connected to `matrix-synapse` (or `matrix-synapse-reverse-proxy-companion`) directly, but go through one such layer of indirection.

On the `matrix-homeserver` side, we could allow such route-stealing to happen. MMR could inject configuration which captures the `/_matrix/media` traffic and sends it to itself. Bridges that try to upload media to `matrix-homeserver/_matrix/media` would be transparently routed to the correct place.

The downside is: MMR would need both Traefik labels (to handle the public case, as described above) and some nginx `location` directives for `matrix-homeserver` (to allow route-stealing to happen there). It's annoying having to **duplicate routes** like this, but.. it's only for a few services (MMR, matrix-corporal likely some identity server, ..).

2 alternatives were considered:

1. Getting rid of `matrix-homeserver` and switching bots/bridges to talking to the homeserver over the public network (`https://matrix.DOMAIN`). Downsides are: SSL destroys performance a lot; routing issues happen sometimes (if `matrix.DOMAIN` is added to `/etc/hosts` with a value of `127.0.0.1`); going through the public network destroys performance once again. Benchmarks are below
2. Pointing bridges/bots to Traefik over the local container network (e.g. `https://matrix-traefik:8443` or a non-SSL variant) just like we do with `http://matrix-nginx-proxy:12080` now (`matrix_homeserver_container_url`). This cannot work, because Traefik only serves requests with the correct `Host` HTTP header. Bridges/bots cannot be made to send requests to some local container URL but also set custom HTTP Headers. A variation of this is 3) below.
3. Adding a `matrix-homeserver` service as described above, but not adding custom `location` blocks to it and thus avoiding the work duplication. Instead, re-routing everything to Traefik internally. nginx can send traffic to (e.g. `https://matrix-traefik:8443` or a non-SSL variant) and also force-set a `Host` header, so that Traefik will interpret the request corretly. The problem is, services register themselves with Traefik only on the `web-secure` entrypoint, which brings us to the huge SSL overhead once again. We could adjust services to register on more entrypoints, but that's likely just as ugly as the duplication discussed earlier

----

## Benchmarks

I've measured the `/_matrix/client/r0/versions` endpoint using `ab -n 100 -c 10`. I've noticed results like this for an existing flow (which will be gone soon):

1. public network, Traefik + SSL: 70 rps
2. `matrix-nginx-proxy:8008`: 600 rps
5. `matrix-nginx-proxy:12080` 850 rps
6. `matrix-synapse-reverse-proxy-companion:8008`: 1000 rps
7. `matrix-synapse:8008`: 1200 rps

Traefik was additionally benchmarked to see where the slowness comes from. Results are like this:
1. public network, Traefik + SSL: 70 rps
2. local (container) network, Traefik + SSL: 150 rps
3. local (container) network, Traefik without SSL: 500 rps
4. `matrix-nginx-proxy:8008`: 600 rps
5. Same as above when going directly to the next reverse-proxies

**Conclusions**:
- going over the public network destroys performance (150 -> 70 rps)
- SSL destroys performance (500 -> 150 rps) even over the local container network
- each nginx reverse proxy in the chain decreases performance by 200 rps
- Traefik being in the chain of reverse-proxies (without SSL) decreasess performance by 100 rps (better than nginx)

------

By the way, I believe various bots/bridges are now connected to the homeserver via `matrix_homeserver_container_url`, which is defined as `matrix-nginx-proxy:12080`.

`matrix-nginx-proxy:12080` is the "homeserver vhost" (`roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2`), which routes traffic to the homeserver but **does not support route-stealing**.

There's also `matrix-nginx-proxy:8008` (`roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2`), which is the public `matrix.DOMAIN`. It ultimately sends traffic to `matrix-nginx-proxy:12080`, but first allows for some route-stealing (to MMR, etc.) to happen.

Because services are pointed directly to `matrix-nginx-proxy:12080` right now, and thus directly go to the homeserver from there, I believe bots/bridges which try to interact with `/_matrix/media` APIs are **not** using MMR. To use MMR, they'd need to go through `matrix-nginx-proxy:8008` instead, as that's the only place route-stealing happens.

Due to this, I believe we have a problem with our MMR setup when bridges/bots using media are involved - unless said bridges/bots talking to the homeserver via the public `https://matrix.DOMAIN` URL, but not all of them are configured that way.

Michael-Hollister:
> Due to this, I believe we have a problem with our MMR setup when bridges/bots using media are involved - unless said bridges/bots talking to the homeserver via the public https://matrix.DOMAIN URL, but not all of them are configured that way.

Okay, would you be fine with the solution being to remove the MMR route-stealing from `roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-domain.conf.j2` and move it too `roles/custom/matrix-nginx-proxy/templates/nginx/conf.d/matrix-synapse.conf.j2` (and similarly for Conduit/Dendrite nginx template files)? I guess this also means that I should revert/disable Traefik changes since route-stealing would now occur in `matrix-nginx-proxy:12080`.

If the above is acceptable, I'll create a new PR to resolve the bot/bridge issue.

spantaleev:
I'm doing some major reworking (removing `matrix-nginx-proxy` completely), in the [bye-bye-nginx-proxy](https://github.com/spantaleev/matrix-docker-ansible-deploy/tree/bye-bye-nginx-proxy) branch. Maybe it's better to fix it there instead of doing more work on `matrix-nginx-proxy`.

I'm porting all services from `matrix-nginx-proxy` to their `matrix-homeserver-proxy` equivalent (etc.), so.. I think I'll fix it myself during this migration.

This `bye-bye-nginx-proxy` branch is far from being generally usable and still far from actually removing `matrix-nginx-proxy`, so.. It's very much not ready yet.

joshsimmons:
Just wanted to pipe up and say thank you for your work on this! The MMR reverse proxy updates ended up being relevant to an issue I was having, and I expect other folks migrating from EMS would be running into it, as well. I've just plucked the updates to matrix-domain.conf for now.

Thanks also for the broader reworking that'll help bots/bridges communicate with MMR as part of the bye-bye-nginx-proxy work :) I can appreciate the level of effort that'll take.

PR #3047, opened 2023-12-16
Updated MMR docs with updated fields in main.yaml

Michael-Hollister:
The MMR v1.3.3 update resulted in some config variable name changes, so updating docs to reflect that. Additional changes:

* ~~Added `matrix_media_repo_database_password` variable in the quickstart for users to change~~
* Removed placeholder password value from defaults
* Changed default value of `matrix_media_repo_datastore_s3_opts_temp_path` since temp directory is not created by the playbook

### DIFF
--- a/docs/configuring-playbook-matrix-media-repo.md
+++ b/docs/configuring-playbook-matrix-media-repo.md
@@ -18,6 +18,7 @@ Add the following configuration to your `inventory/host_vars/matrix.DOMAIN/vars.
 
 ```yaml
 matrix_media_repo_enabled: true
+matrix_media_repo_database_password: "your_password"
 
 # (optional) Turned off by default
 # matrix_media_repo_metrics_enabled: true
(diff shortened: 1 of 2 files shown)


docs/configuring-playbook-matrix-media-repo.md
```diff
@@ -18,6 +18,7 @@ Add the following configuration to your `inventory/host_vars/matrix.DOMAIN/vars.
 
 ```yaml
 matrix_media_repo_enabled: true
+matrix_media_repo_database_password: "your_password"
```
spantaleev, line 21: Why explicitly ask for a database password here, when passwords for all components are auto-generated via `group_vars/matrix_servers`?

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/64db27c7faf3032341c665f8abf7aa66c7aa7aca/group_vars/matrix_servers#L2706

Michael-Hollister: Ah good catch. Forgot about that while looking over the default values. Removed the placeholder value from the docs and defaults.

spantaleev:
Thank you! :bow:

PR #3050, opened 2023-12-17
Update prerequisites.md

real-joshua:
Document that sudo is required.

### DIFF
--- a/docs/prerequisites.md
+++ b/docs/prerequisites.md
@@ -42,4 +42,6 @@ If your distro runs within an [LXC container](https://linuxcontainers.org/), you
   - the range `49152-49172/udp`: TURN over UDP
   - potentially some other ports, depending on the additional (non-default) services that you enable in the **configuring the playbook** step (later on). Consult each service's documentation page in `docs/` for that.
 
+- `sudo`, make sure it is installed. Some distributions, like a minimal Debian net install, do not include it by default.
+
 When ready to proceed, continue with [Configuring DNS](configuring-dns.md).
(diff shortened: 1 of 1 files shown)


real-joshua:
To me it's not clear that sudo is required, only that access as root is required.

real-joshua:
It does not say `and` but `or`, it is ambiguous.

spantaleev:
Maybe the docs page should say:

> Root access or sudo installed, so the playbook can ultimately elevate to root

.. instead of adding a separate line asking for sudo in all cases (even when you have direct root access)

real-joshua:
Well, without `sudo` the Playbool fails at a point where it uses it, even if you run as root. So it'd be good to add that somewhere specifically.

spantaleev:
You're right - I've tried with direct `root` access and it seems like Ansible requires `sudo` even then during the "Gathering facts" task:

```
TASK [Gathering Facts] ***************************************************************************************************************************************************************************************************************
fatal: [HOST]: FAILED! => changed=false 
  ansible_facts: {}
  failed_modules:
    ansible.legacy.setup:
      failed: true
      module_stderr: |-
        /bin/sh: 1: sudo: not found
      module_stdout: ''
      msg: |-
        MODULE FAILURE
        See stdout/stderr for the exact error
      rc: 127
  msg: |-
    The following modules failed to execute: ansible.legacy.setup
```

I've relocated the `sudo` requirement line to be next to the other server-related requirements in the `docs/prerequisites.md` file and have reworded the text a bit.

Thank you for bringing this up!

Issue #3074, opened 2024-01-02
Signal bridge docker tag is not following the version


pm-McFly:
**Describe the bug**
The latest merge including the signal bridge fixed the architecture detection but also freezes the image tag:

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/c744d29567f84c52d4c6be3e52d9cec88e0dda51/roles/custom/matrix-bridge-mautrix-signal/defaults/main.yml#L16

**To Reproduce**
Use any value to update the version of the bridge, then check the signal service file to verify the version in use.

```yaml
# inventory/host_vars/<custom>/vars.yml
matrix_mautrix_signal_version: "latest"
```

**Expected behavior**
Docker tag to be aligned with the declared version.

PR #3075, opened 2024-01-02
[#3074] Docker tag follows the version properly

pm-McFly:
Here is a fix proposal for issue: https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3074


spantaleev:
Good catch! Thanks for fixing it

PR #3076, opened 2024-01-03
Add keepalive on worker upstreams and use persistent connections

cvwright:
This PR adds `keepalive` on each of the Synapse worker upstreams, and uses HTTP 1.1 with persistent connections.  This allows the reverse proxy companion to re-use its connections to the workers, reducing the need for TCP round-trips for 3-way handshake and teardown.

This is in keeping with Nginx best practices https://www.nginx.com/blog/avoiding-top-10-nginx-configuration-mistakes/#no-keepalives

The expectation is that this change should improve performance when the server is under heavy load.  Load test results can be provided in a day or two to support or refute this hypothesis.

### DIFF
--- a/roles/custom/matrix-synapse-reverse-proxy-companion/templates/nginx/conf.d/matrix-synapse-reverse-proxy-companion.conf.j2
+++ b/roles/custom/matrix-synapse-reverse-proxy-companion/templates/nginx/conf.d/matrix-synapse-reverse-proxy-companion.conf.j2
@@ -12,6 +12,7 @@
 {% macro render_worker_upstream(name, workers) %}
 {% if workers | length > 0 %}
 	upstream {{ name }} {
+		keepalive {{ workers | length * 2 }};
 		{% for worker in workers %}
 			server "{{ worker.name }}:{{ worker.port }}";
 		{% endfor %}
@@ -24,6 +25,8 @@
 	location ~ {{ location }} {
 		proxy_pass http://{{ upstream_name }}$request_uri;
 		proxy_set_header Host $host;
+		proxy_http_version 1.1;
+		proxy_set_header Connection "";
 	}
 	{% endfor %}
 {% endmacro %}
@@ -39,6 +42,7 @@
 		# ensures that requests from the same client will always be passed
 		# to the same server (except when this server is unavailable)
 		hash $http_x_forwarded_for;
+		keepalive {{ generic_workers | length * 2 }};
 
 		{% for worker in generic_workers %}
 			server "{{ worker.name }}:{{ worker.port }}";


cvwright:
Here are some load testing results courtesy of @Michael-Hollister.  The test system was (IIRC) an 8-core / 32 GB dedicated CPU VM on DigitalOcean SFO3.  It's a very simple test: the clients are all idle -- they only `/sync` and do not send any messages.  We gradually ramp up the number of clients from 0 to 20k.

The server begins to struggle at around 15k connected clients, causing the sync response time to increase above the 30 sec timeout requested by the clients.

Without keepalive, the errors per second start to spike after all 20k clients are connected, and more requests begin to hit the client side 60 sec timeout.

With keepalive enabled, the error rate stays very low.  The 95th percentile response time *just* manages to stay below the 60 sec cutoff.

keepalive

spantaleev:
Thank you for making this scale better! :+1: 

------

I wonder if it would help to add keepalive to other reverse-proxies in the chain.

As discussed in https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/3045, we currently have a long chain of reverse-proxies, as documented [in this comment](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/3045#issuecomment-1871830463) (Traefik -> `matrix-nginx-proxy:8008` -> `matrix-nginx-proxy:12080` -> `matrix-synapse-reverse-proxy-companion:8008` -> `matrix-synapse:8008`).

All these other reverse-proxies are going away in the `bye-bye-nginx-proxy` branch (not ready for general usage yet), so there's no point in fixing it anyway. In the end, it would just be like this when looked at **from the outside**:

- with workers enabled: Traefik -> `matrix-synapse-reverse-proxy-companion:8008` -> `matrix-synapse:8008`
- without workers: Traefik -> `matrix-synapse:8008`

For addons services (bridges, bots) which communicate with the homeserver over the local container network, we'll have one more reverse-proxy (`matrix-homeserver-proxy`). I have **not yet decided on the final flow**, but it would be one of these:

1. (bridge / bot) -> `matrix-homeserver-proxy:8008` -> Synapse (either directly to `matrix-synapse:8008` or via `matrix-synapse-reverse-proxy-companion:8008`) - requires some duplication of routes: e.g., matrix-media-repo will need to register routes both with Traefik (for public route-stealing) and with `matrix-homeserver-proxy`. No extra hops in the reverse-proxy chain though.
2. (bridge / bot) -> `matrix-homeserver-proxy:8008` -> a new non-SSL-enabled Traefik endpoint -> Synapse (either directly to `matrix-synapse:8008` or via `matrix-synapse-reverse-proxy-companion:8008`) - less duplication of work (reusing existing Traefik labels), but one more hop (via Traefik)

Both of these potential flows have pros and cons, so I'm still debating which one to go with. I'm somewhat leaning toward the second one right now, but let's see what tomorrow will bring.

**In any case**, `matrix-homeserver-proxy` may be another candidate to benefit from `keepalive`, to optimize local connections (from bots/bridges) to the homeserver.

For small homeservers, it probably doesn't matter. For large ones with more people using bots and bridges (and passing more and more messages through them), these services also start hiting the homeserver frequently.

I'm not sure what a suitable `keepalive` value would be for this service though.

cvwright:
Thanks Slavi!

> In any case, matrix-homeserver-proxy may be another candidate to benefit from keepalive, to optimize local connections (from bots/bridges) to the homeserver.

Yes I think that makes sense.

> I'm not sure what a suitable keepalive value would be for this service though.

I have no idea.  The Nginx docs suggest 2x the number of servers in an upstream block.  But here the situation is a lot more complicated.

I've also seen people in `#synapse-extreme-tuning:fostered.uk` suggest using a fixed number, like 10 or 20.

The number is the maximum number of *idle* connections to leave open, so there's probably no harm in setting it to the low double digits when your upstream is another reverse proxy.

PR #3086, opened 2024-01-09
Inventory as a git submodule

cvwright:
This PR moves the `inventory/` subdirectory out of this git repo and makes it a git submodule.

The idea is that users could keep their inventory in a separate repo of their own, which would be forked from the base inventory repo.  This should make it easier for users to keep their `host_vars` under revision control, without worrying about changes getting lost when merging in the latest version of the playbook from upstream, and without risk of inventory files accidentally being included in public repos or PRs etc.

To make this PR production ready, there are probably a bunch of other things that need to happen:
* Creating a `just` recipe to hook up the user's own `inventory` repo
* Creating a `just` recipe to update the inventory repo
* Make the inventory a dependency for other `just` recipes (?)
* Decide whether the base `inventory` repo should ship as a submodule of the playbook by default, or if it's better to leave it out until the user attaches their own inventory
* Others?


cvwright:
Sounds like it's going to be much easier to simply remove `inventory` from this repo's version control, and let users plug in whatever they want, similar to what the MASH playbook does.

reudy14:
The installation guide will need updating

PR #3100, opened 2024-01-16
Room workers

cvwright:
This PR adds support for the more specialized workers approach proposed in @tcpipuk's [Synapse homeserver guide](https://tcpipuk.github.io/synapse/index.html).  The general idea is that each (type of) worker should do one task, and do it well.

Here we create 4 new kinds of workers that handle the endpoints previously handled by the generic workers:
* Room workers
* Sync workers
* Client readers
* Federation readers

NOTE: The Synapse Homeserver Guide referenced above contains many recommendations for tuning various aspects of the system.  This PR focuses solely on enabling the new worker types, doing so with as minimal a change set as is possible.  Once the specialized workers are enabled in the playbook, it may make sense to incorporate other improvements as well, e.g. cache auto-tuning.

# New worker types

## Room workers
Room workers handle all URLs that contain a room id.  For a given room, we route all requests (client and federation) for that room to the same room worker.  Then that worker can make the most effective use of its cache, keeping the cache as hot as possible and minimizing the need to hit the database.  This reduces the load on the database, leading to faster performance and better scalability for handling large rooms and/or large numbers of rooms.

## Sync workers
Sync workers handle the `/sync` endpoint.  For a given user, we route all sync requests from all the user's devices to the same sync worker.

## Client readers
Client readers handle the remaining client-server endpoints that can be handled by a worker, but do not contain a room id.

## Federation readers
Federation readers handle the remaining federation endpoints that can be handled by a worker, but do not contain a room id.

# New preset `room-workers`
We add a new worker preset `room-workers`, which is similar to the existing `one-of-each` preset, but with one of each new worker type and zero generic workers.

# Backwards compatibility
This PR does not remove the old `generic_worker` type or modify either of the existing presets.  Users of the `one-of-each` and `little-federation-helper` presets should see no change in their systems.

Users can upgrade to room workers by changing the preset to `room-workers` and then tweaking the counts for each worker just like before.


cvwright:
Looks like this needs #3102 in order to re-enable workers under Traefik

akdk7:
> Looks like this needs #3102 in order to re-enable workers under Traefik

Yep, my setup is having trouble with worker setup and it's showing a lot of these error messages.

```
Jan 16 16:31:17 matrix-srv matrix-synapse[75863]: 2024-01-16 15:31:17,578 - synapse.http.server - 140 - ERROR - PUT-12735 - Failed handle request via 'RoomTypingRestServlet': <X>
Jan 16 16:31:17 matrix-srv matrix-synapse[75863]: Traceback (most recent call last):
Jan 16 16:31:17 matrix-srv matrix-synapse[75863]:   File "/usr/local/lib/python3.11/site-packages/synapse/http/server.py", line 326, in _async_render_wrapper
Jan 16 16:31:17 matrix-srv matrix-synapse[75863]:     callback_return = await self._async_render(request)
Jan 16 16:31:17 matrix-srv matrix-synapse[75863]:                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Jan 16 16:31:17 matrix-srv matrix-synapse[75863]:   File "/usr/local/lib/python3.11/site-packages/synapse/http/server.py", line 538, in _async_render
Jan 16 16:31:17 matrix-srv matrix-synapse[75863]:     callback_return = await raw_callback_return
Jan 16 16:31:17 matrix-srv matrix-synapse[75863]:                       ^^^^^^^^^^^^^^^^^^^^^^^^^
Jan 16 16:31:17 matrix-srv matrix-synapse[75863]:   File "/usr/local/lib/python3.11/site-packages/synapse/rest/client/room.py", line 1225, in on_PUT
Jan 16 16:31:17 matrix-srv matrix-synapse[75863]:     raise Exception("Got /typing request on instance that is not typing writer")
Jan 16 16:31:17 matrix-srv matrix-synapse[75863]: Exception: Got /typing request on instance that is not typing writer
```

cvwright:
> Yep, my setup is having trouble with worker setup and it's showing a lot of these error messages.
> 
> ```
> ...
> Jan 16 16:31:17 matrix-srv matrix-synapse[75863]:     raise Exception("Got /typing request on instance that is not typing writer")
> Jan 16 16:31:17 matrix-srv matrix-synapse[75863]: Exception: Got /typing request on instance that is not typing writer
> ```

hmm I don't think that's #3102.  That's a bug in my routing for the new workers.  Working on a fix... Thanks @akdk7

cvwright:
Ok I now have this working on a test server with the following config:
* 2 room workers
* 2 sync workers
* 2 client readers
* 2 federation readers
* 0 generic workers
* 1 of every other worker type

I have tested most of the normal things: creating rooms, sending messages, inviting local and remote users, setting avatars, etc.  We did have some small hiccups with E2E encryption upon first inviting federated users, although as far as I can tell this seems to be more down to the clients more than the servers.

The Element-Web invitation UI includes user search, so that is tested working too.

I have also tested with the existing `one-of-each` preset and the new `room-workers` preset, and verified that it worked as expected in each case.

More scientific load tests are still to come, but so far it feels reasonably snappy on a cheap Hetzner Cloud VM.

akdk7:
I've updated my configuration and set up multiple workers on my testbench. Everything is running fine so far.

```
matrix_synapse_workers_enabled: true
matrix_synapse_workers_preset: room-workers
matrix_synapse_workers_room_workers_count: 2
matrix_synapse_workers_sync_workers_count: 2
matrix_synapse_workers_client_reader_workers_count: 2
matrix_synapse_workers_federation_reader_workers_count: 2
```

spantaleev:
I've also tested it and it seems to work well.

I've renamed the `room-workers` preset to `specialized-workers`, which I think is a better name for it. See [84446e5](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/84446e52e983f9cf1a1441f980d55773406b0a8c).

I've also added documentation and a changelog announcement.

---

Once again, thanks to [Tom Foster](https://github.com/tcpipuk) for his [Synapse homeserver guide](https://tcpipuk.github.io/synapse/index.html) and to [Charles Wright](https://github.com/cvwright) for porting these ideas to the playbook! :bow:

PR #3102, opened 2024-01-16
Fixed Traefik routing when using Synapse workers

Michael-Hollister:
Traefik seems to have routing priority conflicts when using the reverse proxy companion (and likely MMR as well, though I did not have an issue on my server). With the [default priority](https://doc.traefik.io/traefik/routing/routers/#priority) rules  being used, I have seen requests being routed directly to the Synapse container instead of the reverse proxy companion container.

I went with disabling registration of Synapse routes instead of adjusting priority values  if the reverse proxy companion is enabled. Tested the changes for correct behavior of the reverse proxy companion and MMR routing, but I don't use any of the bridges/bots, so those roles have not been tested.

### DIFF
--- a/roles/custom/matrix-synapse/templates/synapse/labels.j2
+++ b/roles/custom/matrix-synapse/templates/synapse/labels.j2
@@ -82,7 +82,7 @@ traefik.http.routers.matrix-synapse-public-client-api.tls.certResolver={{ matrix
 
 
 
-{% if matrix_synapse_container_labels_internal_client_api_enabled %}
+{% if matrix_synapse_container_labels_internal_client_api_enabled and not matrix_synapse_reverse_proxy_companion_enabled %}
 ############################################################
 #                                                          #
 # Internal Client-API (/_matrix)                           #
--- a/roles/custom/matrix-synapse/defaults/main.yml
+++ b/roles/custom/matrix-synapse/defaults/main.yml
@@ -192,7 +192,7 @@ matrix_synapse_container_labels_public_client_root_redirection_url: ""
 # Controls whether labels will be added that expose the Client-Server API on a public Traefik entrypoint.
 # Regardless of whether this is enabled, it may or may not take effect due to the value of other variables.
 # See `matrix_synapse_container_labels_traefik_enabled` or `matrix_synapse_container_labels_matrix_related_labels_enabled`
-matrix_synapse_container_labels_public_client_api_enabled: true
+matrix_synapse_container_labels_public_client_api_enabled: "{{ not matrix_synapse_reverse_proxy_companion_enabled }}"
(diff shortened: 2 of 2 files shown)


roles/custom/matrix-synapse/templates/synapse/labels.j2
```diff
@@ -82,7 +82,7 @@ traefik.http.routers.matrix-synapse-public-client-api.tls.certResolver={{ matrix
 
 
 
-{% if matrix_synapse_container_labels_internal_client_api_enabled %}
+{% if matrix_synapse_container_labels_internal_client_api_enabled and not matrix_synapse_reverse_proxy_companion_enabled %}
```
Michael-Hollister, line 85: The check was added here instead of `roles/custom/matrix-synapse/defaults/main.yml` since `matrix_synapse_container_labels_internal_client_api_enabled` is configured in `group_vars/matrix_servers` depending on the proxy setup being used.

spantaleev:
There's no need for disabling labels for `matrix-synapse` individually based on whether `matrix-synapse-reverse-proxy-companion` is enabled or not.

The companion is only enabled (at least automatically) when workers are enabled: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/17859ecccab07399db9e210e86010dbf57e6663c/group_vars/matrix_servers#L4046-L4046

When workers are enabled, `matrix-synapse` knows it cannot route to them appropriately so it automatically disables **all** Matrix-related routers (`/_matrix` and `also `/_synapse/*`), courtesy of `matrix_synapse_container_labels_matrix_related_labels_enabled`: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/17859ecccab07399db9e210e86010dbf57e6663c/roles/custom/matrix-synapse/defaults/main.yml#L181-L191

Some of these routes (e.g. `/_synapse/client`) may or may not be load-balanced to workers, so Synapse could potentially keep them enabled, but.. we disable them all just in case.

I can see a situation where someone keeps workers **disabled**, but force-enables the companion explicitly (for whatever reason). In such a case, both components would end up registering the same labels. It shouldn't hurt though, because whichever services ends up receiving these routes, can handle them correctly. This is because:

- a non-workers Synapse can handle routing on its own (everything goes to `matrix-synapse:8008`)
- `matrix-synapse-reverse-proxy-companion`, when enabled, in combination with a non-workers Synapse sends everything to `matrix-synapse:8008` anyway. It's one extra (unnecessary) hop, but it should work.

So.. the current configuraton should be fine. However, I had forgotten to make use of `matrix_synapse_container_labels_matrix_related_labels_enabled` in the `labels.j2` template for `matrix-synapse`, so it wasn't fine. This has been remedied in 94378a77. Thank you for catching the problem!

PR #3110, opened 2024-01-16
Fixed docker network access for Prometheus Postgres exporter

Michael-Hollister:
With the updates to the proxy and Docker networks, the Prometheus Postgres exporter no longer has access to the Postgres Docker network. This PR addresses that issue.


spantaleev:
I've fixed it in a cleaner way in [0315d03](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/0315d03cdb1212d20d3c3f57bcc8bc98948a9081).

Thank you for finding and working on this!

PR #3118, opened 2024-01-19
Updated nginx proxy fronting with NPM guide

SirHazza:
TL;DR: Updated the 'nginx reverse-proxy fronting' documentation with a guide for Nginx Proxy Manager, as you can't use the pre-existing nginx matrix.conf

---

I use Nginx Proxy Manager because its nice and simple, but also because it allows for snippets of custom nginx configuration when things get more advanced. Whilst setting up matrix with NPM via fronting the internal proxy, the playbook configuration was very easy (amazing documentation btw). But trying to translate a full [nginx config](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/8f06e2bf05cf543da6b0317873a5f89d2638109f/examples/reverse-proxies/nginx/matrix.conf) into the web interface of NPM was very confusing (NPM has very little guidance on how its web ui translates to nginx configuration + doesn't have a config import tool).

So after a couple of hours, and coming across a few help threads of people asking "NPM+ matrix playbook how?" it all made sense in the end. And so here's a write up of the basic setup for others to follow, so no one else has to have the headache I had.


spantaleev:
Thank you for contributing back and making the documentation even better!

I wonder if it wouldn't be better to have a dedicated directory for this (e.g. `examples/reverse-proxies/nginx-proxy-manager/`)? That way, people skimming the directory list an immediately see that we support it and can go to the right place. I suppose most people may click **nginx** right now, hoping that it mentions nginx-proxy-manager as well, but.. why not dedicate its own page to it?

SirHazza:
@spantaleev done!

spantaleev:
Thanks! Merged!

PR #3119, opened 2024-01-20
Mautrix-Telegram bridge config update

pm-McFly:
After reviewing the configuration of the Mautrix-Telegram bridge, I noticed that a lot of improvements were missing... Here there are!


spantaleev:
We were quite out of date, it seems! Thank you for correcting this!

PR #3120, opened 2024-01-20
Added docu on how to host another server behind traefik.

darksnakezero:
Added docu on how to host another server behind traefik to the docu file about configuring traefik.

All configuration options were succsessfully tested in my own deployment.

### DIFF
--- a/docs/configuring-playbook-traefik.md
+++ b/docs/configuring-playbook-traefik.md
@@ -51,20 +51,22 @@ devture_traefik_configuration_extension_yaml: |
 
 ## Reverse-proxying another service behind Traefik
 
-The preferred way to reverse-proxy additional services behind Traefik would be to add the service as another container, configure the container with the corresponding traefik labels, and add it to the `traefik` network. Some services are also already available via the compatible [mash-playbook](https://github.com/mother-of-all-self-hosting/mash-playbook), but take a look at the minor [interoperability adjustments](https://github.com/mother-of-all-self-hosting/mash-playbook/blob/main/docs/interoperability.md).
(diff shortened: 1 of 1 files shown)


spantaleev:
Oh, I thought this would be a guide that tells people how to:

- add container labels to existing services on the host
- connect these to the `traefik` network

----

Instead, it seems like it's a guide for reverse-proxying to services which either run on another machine on the network (e.g. `http://192.168.1.40:3333`) or to services on localhost which potentially do not run in a container (e.g. `http://127.0.0.1:3333`).

I suppose we should at least mention both possibilities and encourage people to host services in a container if they're hosting them on the same machine. If services are hosted somewhere else, then there's no other optional than doing what your current guide says you should do.

Perhaps we should also reiterate our `README.md` text about our sibling project ([mash-playbook](https://github.com/mother-of-all-self-hosting/mash-playbook)) and it being interoperable with this playbook.

darksnakezero:
I am sorry that I was not specific enough, maybe that could have helped earlier. Nevertheless, I learnt a lot and just wanted to share my results :-)

While I agree that it might be a good idea to run the other services in a container (and encourage other people to do so), my nextcloud instance does indeed run on another machine.

Shall I adjust my proposed changes to include the recommendation to run other services in the container, add a link to the MASH playbook and clear up that the guide indeed is ment for reverse-proxying other services?

Should I then separate it into another documentation file or keep it in the same file with the traefik configuration?

spantaleev:
I think keeping the guide in the current Traefik documentation page is a good idea.

If you could redo it to mention the 2 approaches and MASH, etc., that'd be great! I suppose we could later expand on that other method (adding container labels; connecting services to the `traefik` network). If you'd like to start a section on that even now, go ahead!

darksnakezero:
Kept it in the same page, mentioned the other approaches as preffered, recommended mash, some rephrasing and minor adjustments to sound better.

Unfortunately, I do not know enough about the docker and label way to write documentation on it. On a glance, it seems easy enough, but I ll leave the details to someone who will actually do the testing.

Thanks for the work on the playbooks nevertheless, they are a blessing!

spantaleev:
I've improved the wording a bit and merged it. Thanks!

PR #3124, opened 2024-01-23
fix setting root path because the script moved one level up in 2f457b2a

mcnesium:


spantaleev:
Oh,  I thought that I've adjusted this when relocating, but it seems like I haven't. Thanks!

PR #3126, opened 2024-01-23
Remove redundant nginx config

tacheometry:


spantaleev:
People who run Synapse with workers (which enables `matrix-synapse-reverse-proxy-companion`, which is based on nginx) may benefit from this nginxlog exporter.

PR #3130, opened 2024-01-24
add missing ' in config

jalemann:


spantaleev:
I believe it's OK (valid YAML) even without this, but I can see how not having quotes can be confusing in this specific case.

jalemann:
Yes, shouldn't make a difference. I thought it might be the cause for a bug, it wasn't. Still wanted to add it just to be sure as `:` is also used for assignments.

PR #3131, opened 2024-01-24
Added extra systemd service arguments to synapse workers and proxy companion

Michael-Hollister:
* Added extra systemd service arguments  to matrix-synapse-reverse-proxy-companion and synapse workers
* Added extra arguments to the master synapse service that only apply to that container

One use-case we are using this for is to limit docker container CPU utilization for specific synapse processes and other services in the playbook.


spantaleev:
Makes sense! Thanks!

PR #3155, opened 2024-02-04
Small doc fix

ingydotnet:
`mkdir` with multiple subdirs needs `-p`


spantaleev:
Thank you!

Background: `inventory/host_vars/` used to exist thanks to `.gitkeep` files and multi-level-directory-creation was not necessary. Recently, the `inventory` directory was removed, so that people can create and manage it themselves as they see fit (as a git repository, etc).

PR #3163, opened 2024-02-08
Bring config.yaml.j2 in line with upstream

needo37:


Peetz0r:
This PR made all my WA contacts change their name from "Example (WA)" to "Example (WhatsApp)". No big deal to me, but it would have been nice to read about this in CHANGELOG.md before it happened :)

PR #3164, opened 2024-02-08
Update signal config.yaml.j2 merging upstream changes

needo37:


spantaleev:
It seems like your d2345bd1 commit dropped the `{{` and `}}` used by Golang templates. I've "added it back" (in a way), using `{% raw %}` and `{% endraw %}`.

PR #3166, opened 2024-02-08
Update Signal config.yaml.j2

needo37:
Not sure why but the endraw is not working.


spantaleev:
Not sure why that would be, but.. let's go for not using raw/endraw anyway.

PR #3169, opened 2024-02-09
allow to configure whatsapp polls via extev_polls parameter.

kumarunster:
currently extev_polls parameter is hardcoded to false.

This change allows to overwrite it via custom variable. Would be great, if that change could be merged into the master/main branch.

Thank you


spantaleev:
You could have overriden the default value using `matrix_mautrix_whatsapp_configuration_extension_yaml`, but OK - let's add a dedicated variable for this.

PR #3195, opened 2024-02-21
Add Synapse module auto accept invite to rooms and direct messages

xvsun:
Hi,

here's my attempt to implement this [Synapse module](https://github.com/matrix-org/synapse-auto-accept-invite).
I don't know how important the last two points from the current feature list are - maybe someone can give me advice how to do it. Thanks for the project! 

## Feature/Todo list:
- [x] Add module to homeserver.yaml configuration
- [x] Integrate synapse-auto-accept-invite module in Synapse container 
- [x] Only accept invites to direct messages (1:1 rooms) 
- [x] Adjust Synapse logging (disable existing loggers: false)
- [x] Change load in workerised Synapse deployments


spantaleev:
You seem to be on the right path. Thanks for making progress on it!

-----

For loading it on a worker-enabled Synapse, you may introduce a new variable (e.g. `matrix_synapse_ext_synapse_auto_accept_invite_worker_to_run_on`), similar to `matrix_synapse_ext_synapse_auto_accept_invite_only_direct_messages`, which would be injected into the module configuration (at `worker_to_run_on`).

It can be unset (null or empty string) by default and you may pass it (via `roles/custom/matrix-synapse/tasks/ext/synapse-auto-accept-invite/setup_install.yml`) in all cases. Looking at [module's code](https://github.com/matrix-org/synapse-auto-accept-invite/blob/d673c6767862e0340e58f6e9947c7d90ec190201/synapse_auto_accept_invite/__init__.py), it may be better if you pass it as `None` if it's empty.

Perhaps something like this:

```yml
- ansible.builtin.set_fact:
    matrix_synapse_modules: |
      {{
        matrix_synapse_modules | default([])
        +
        [
          {
            "module": "synapse_auto_accept_invite.InviteAutoAccepter",
            "config": {
              "accept_invites_only_for_direct_messages": matrix_synapse_ext_synapse_auto_accept_invite_only_direct_messages,
              "worker_to_run_on": (matrix_synapse_ext_synapse_auto_accept_invite_worker_to_run_on if matrix_synapse_ext_synapse_auto_accept_invite_worker_to_run_on else None),
            },
        }]
      }}
````

-----

You may also add a validation task (`roles/custom/matrix-synapse/tasks/ext/synapse-auto-accept-invite/validate_config.yml` which would need to be hooked from `roles/custom/matrix-synapse/tasks/ext/synapse-auto-accept-invite/main.yml`), to validate that `matrix_synapse_ext_synapse_auto_accept_invite_worker_to_run_on` is non-empty when `matrix_synapse_workers_enabled`.

See `roles/custom/matrix-synapse/tasks/ext/s3-storage-provider/*` for an example.

------

You can inject additional loggers via `matrix_synapse_additional_loggers_auto`, like some of the other extensions are doing it.

Take a look at `roles/custom/matrix-synapse/tasks/ext/encryption-disabler/setup_install.yml`

The README for the auto-accept-invite module advocates for `disable_existing_loggers: False`, but this is probably not necessary to do if you add an explicit logger for the new module.

You may even add a new variable (e.g. `matrix_synapse_ext_synapse_auto_accept_invite_logging_level`) to make the logging level configurable. For `encryption-disabler`, it's hardcoded to `INFO`, but you may wish to do better.

-----

Also, I'd rename `matrix_synapse_ext_synapse_auto_accept_invite_only_direct_messages` to `matrix_synapse_ext_synapse_auto_accept_invite_accept_invites_only_for_direct_messages`.

It's longer, but it makes sense. The first part (`matrix_synapse_ext_synapse_auto_accept_invite_accept_`) is the prefix for this extension in the playbook. The second part (`accept_invites_only_for_direct_messages`) matches the actual configuration key used by the extension. We always try to name variables this way.

----

Also, indentation in  `roles/custom/matrix-synapse/tasks/ext/synapse-auto-accept-invite/setup_install.yml` seems a little off. My example code-block in this comment (above) possibly corrects it.

xvsun:
Thanks for the input. Will try to include worker config and logging. Renaming direct_messages sounds reasonable.

> You may also add a validation task (`roles/custom/matrix-synapse/tasks/ext/synapse-auto-accept-invite/validate_config.yml` which would need to be hooked from `roles/custom/matrix-synapse/tasks/ext/synapse-auto-accept-invite/main.yml`), to validate that `matrix_synapse_ext_synapse_auto_accept_invite_worker_to_run_on` is non-empty when `matrix_synapse_workers_enabled`.
> 
> See `roles/custom/matrix-synapse/tasks/ext/s3-storage-provider/*` for an example.

I'm not sure if it's necessary. Because if not set on a worker main process should be used. Right?

xvsun:
Hi,

@spantaleev I added logging and the synape workers config option. `None` didn't work for the synapse worker, it has to be `null`. 

I chose `matrix-synapse-worker-generic-0` as an example for the docs. Hope choosing this worker is fine.

Logging is called `matrix_auto_accept_invite` (how to check if it works?) and the level is also hardcoded to `INFO`.  Thoughts on anything else or issues?

spantaleev:
I've fixed it up a bit and think it's good now. Merged! Thank you!

PR #3213, opened 2024-03-07
Enable ephemeral events in mautrix-meta registration

6502mos:


spantaleev:
Thanks! I had removed this part, thinking it's some a from the past, but it seems like it's actually necessary.

Related to: https://github.com/mautrix/meta/issues/32

PR #3218, opened 2024-03-09
Add setting to enable ipv6 in docker networks

tilosp:
For this to work the docker daemon has to be configured for ipv6: https://docs.docker.com/config/daemon/ipv6/

### DIFF
--- a/roles/custom/matrix-appservice-draupnir-for-all/tasks/setup_install.yml
+++ b/roles/custom/matrix-appservice-draupnir-for-all/tasks/setup_install.yml
@@ -77,6 +77,7 @@
 
 - name: Ensure matrix-appservice-draupnir-for-all container network is created
   community.general.docker_network:
+    enable_ipv6: "{{ matrix_ipv6_enabled }}"
     name: "{{ matrix_appservice_draupnir_for_all_container_network }}"
     driver: bridge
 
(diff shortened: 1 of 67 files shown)


spantaleev:
I've reworked this to be based on `devture_systemd_docker_base_ipv6_enabled`, which is not part of the [systemd-docker-base](https://github.com/devture/com.devture.ansible.role.systemd_docker_base) role - see [here](https://github.com/devture/com.devture.ansible.role.systemd_docker_base/blob/c11a526bb8e318b42eb52055056377bb31154f13/defaults/main.yml#L14-L31).

I've also added support for auto-reconfiguring our Docker role ([geerlingguy/ansible-role-docker](https://github.com/geerlingguy/ansible-role-docker)) based on the value of `devture_systemd_docker_base_ipv6_enabled`.

Still, existing installations will probably need to follow [the instructions for deleting existing container networks](https://github.com/devture/com.devture.ansible.role.systemd_docker_base/blob/c11a526bb8e318b42eb52055056377bb31154f13/defaults/main.yml#L26-L30) to have changes to `devture_systemd_docker_base_ipv6_enabled` take effect. New installations with `devture_systemd_docker_base_ipv6_enabled: true` should work seamlessly - Docker should be auto-reconfigured and all container networks should get IPv6 support.

It should also be noted that not all roles used by this playbook respect the `devture_systemd_docker_base_ipv6_enabled` variable yet. Many roles are external (installed via ansible-galaxy into `roles/galaxy`) - none of those respect the `devture_systemd_docker_base_ipv6_enabled` variable yet, but can be adjusted to do so. They all implicitly depend on the [systemd-docker-base](https://github.com/devture/com.devture.ansible.role.systemd_docker_base) role, so they can directly start referencing the new `devture_systemd_docker_base_ipv6_enabled` variable.

PR #3236, opened 2024-03-24
feat: Add global option to configure all bridges encryption default

gardar:


spantaleev:
Makes sense! Thank you!

PR #3279, opened 2024-04-17
Update configuring-playbook-bridge-mautrix-meta-messenger.md

ThellraAK:
Add " ' " around the username for matrix_admin.

### DIFF
--- a/docs/configuring-playbook-bridge-mautrix-meta-messenger.md
+++ b/docs/configuring-playbook-bridge-mautrix-meta-messenger.md
@@ -62,7 +62,7 @@ matrix_mautrix_meta_messenger_bridge_permissions_default:
   '{{ matrix_admin }}': admin
 ```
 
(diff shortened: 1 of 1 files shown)


spantaleev:
Why? Are you using a username with special characters that break the YAML parser without quotes?

ThellraAK:
Without the single quotes I get an error

```
ERROR! We were unable to read either as JSON nor YAML, these are the errors we got from each:
JSON: Expecting value: line 1 column 1 (char 0)

Syntax Error while loading YAML.
  found character that cannot start any token

The error appears to be in '/share/matrix-docker-ansible-deploy/inventory/host_vars/matrix.relativelyprivate.com/vars.yml': line 58, column 15, but may

matrix_mautrix_meta_messenger_enabled: true
matrix_admin: @ThellraAK:mydomain.com
              ^ here
```

That's with `matrix_admin: @ThellraAK:mydomain.com`

However with `matrix_admin: '@ThellraAK:mydomain.com'    `


It works as expected.

PR #3288, opened 2024-04-22
[Element] Fix translation on welcome screen

krassle:
- Provoked by: [Migrate translations to keys and switch to Localazy](https://github.com/element-hq/element-web/commit/c525b633bdb4801086999048295d0d4e5e296d6f#diff-3bbccf505f97c801dfb746d577c2ded291ce3f2ba9fdec55ac856565608a93e4)
- Schildi-Chat: Not affected yet (still using Weblate)


spantaleev:
Seems like we should only merge this once v1.11.65 is actually released.

PR #3353, opened 2024-06-05
Change the hardcoded 'docker.service' to `devture_systemd_docker_base_docker_service_name` variable

cksit:
First, thank you for adding a new variable to make it easier to install the project on Synology DSM. However, it still asks for the existence of docker.service during the installation.

So, I made changes to this file, and it works without any issues so far.

I'm not sure why this file was ignored during the previous commit 9f2eff2. Could you please let me know if there was any reason for this?

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -2873,7 +2873,7 @@ backup_borg_location_exclude_patterns: |
 
 backup_borg_systemd_required_services_list: |
   {{
-    ['docker.service']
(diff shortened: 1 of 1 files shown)


spantaleev:
Looks like I've missed a few references in 9f2eff2ac799b9144cc2b1809760e1681a8a5b9f!

Thanks!

There were actually a few more that I've handled in 42247411308bd301b1c093ed132be0512368b885

Issue #3359, opened 2024-06-09
Sliding Sync Metrics


HarHarLinks:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
https://github.com/matrix-org/sliding-sync/tree/main/grafana
should work also with external collection

**Describe alternatives you've considered**
config extension, but that's meh

**Additional context**
should work like the other metrics

HarHarLinks:
done in #3367, at least for my purposes

PR #3367, opened 2024-06-14
sliding sync metrics support

HarHarLinks:
- tested with external collection
- NOT tested with playbook prom/grafana stack


spantaleev:
Fixes https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3359

Thank you! :bow:

PR #3385, opened 2024-06-27
Allow configuring synapse `auto_join_mxid_localpart`

derhagen:
`auto_join_mxid_localpart` defines the local part of the user id which is used to create auto-join rooms. The variable needs to be set to invite new users to any auto-join rooms which are set to invite-only.

### DIFF
--- a/roles/custom/matrix-synapse/defaults/main.yml
+++ b/roles/custom/matrix-synapse/defaults/main.yml
@@ -537,11 +537,17 @@ matrix_synapse_account_threepid_delegates_msisdn: ''
 
 # Users who register on this homeserver will automatically be joined to these rooms.
 # Rooms are to be specified using addresses (e.g. `#address:example.com`)
+# If any auto-join rooms are invite-only, you need to define `matrix_synapse_auto_join_mxid_localpart`.
 matrix_synapse_auto_join_rooms: []
 
 # Controls whether auto-join rooms (`matrix_synapse_auto_join_rooms`) are to be created
 # automatically if they don't already exist.
 matrix_synapse_autocreate_auto_join_rooms: true
+
+# The local part of the user id which is used to create auto-join rooms if `matrix_synapse_autocreate_auto_join_rooms` is true.
+# Defaults to the initial user account that registers.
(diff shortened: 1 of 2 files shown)


roles/custom/matrix-synapse/defaults/main.yml
```diff
@@ -537,12 +537,18 @@ matrix_synapse_account_threepid_delegates_msisdn: ''
...
 # Rooms are to be specified using addresses (e.g. `#address:example.com`)
+# If any auto-join rooms are invite-only, you need to define `matrix_synapse_auto_join_mxid_localpart`.
 matrix_synapse_auto_join_rooms: []
 
 # Controls whether auto-join rooms (`matrix_synapse_auto_join_rooms`) are to be created
 # automatically if they don't already exist.
 matrix_synapse_autocreate_auto_join_rooms: true
 
+# The local part of the user id which is used to create auto-join rooms if `matrix_synapse_autocreate_auto_join_rooms` is true.
+# Defaults to the initial user account that registers.
```
spantaleev, line 548: What is this supposed to mean?

spantaleev: Oh, now I saw it! Thanks!

roles/custom/matrix-synapse/templates/synapse/homeserver.yaml.j2
```diff
@@ -1541,7 +1541,7 @@ autocreate_auto_join_rooms: {{ matrix_synapse_autocreate_auto_join_rooms|to_json
 # Note that, if the room already exists, this user must be joined and
 # have the appropriate permissions to invite new members.
 #
-#auto_join_mxid_localpart: system
+auto_join_mxid_localpart: {{ matrix_synapse_auto_join_mxid_localpart | to_json }}
```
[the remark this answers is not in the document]
spantaleev: I've checked the [Synapse source code](https://github.com/element-hq/synapse/blob/4af654f0da31072bf0e0cac33202c621369ee823/synapse/config/registration.py#L112) and it seems like an empty string should not be causing problems for how the `mxid_localpart` variable is being used.

PR #3388, opened 2024-06-28
Update maintenance-migrating.md

Kuchenmampfer:
Hi,
I am currently migrating to a new server and very happy about the awesome documentation that you provide :)
However, I ran into an issue in the last step where ansible could not pull the exim-relay image. The cause for that was that docker was not yet installed on the server. Therefore, I added that step to the documentation.


spantaleev:
By default, the playbook installs Docker by itself.

You shouldn't need to do this, unless you have explicitly disabled Docker installation (because you're installing Docker yourself some other way). If you have disabled Docker installation, then you're supposedly aware of what you're doing and will remember to install Docker yourselt.

Issue #3308, opened 2024-05-06
draupnir can't reach the synapse admin API


HarHarLinks:
**Describe the bug**

Per the docs, draupnir can poll abuse reports from the synapse admin API. This used to work back in matrix-nginx-proxy times :older_man:.

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-bot-draupnir.md#abuse-reports

According to the docs, what is needed, is to enable the admin API (e.g. by configuring directly, or indirectly be enabling synapse-admin), the darupnir user being an admin, and the draupnir config extension.

However it does not work: draupnir bot will spam the control room:
> `‼ | failed to poll events: Error: Error during MatrixClient request GET /_synapse/admin/v1/event_reports?dir=f&from=0: 404 Not Found -- "404 page not found\n"`

So I dug around.

**Finding the issue**
With my config, which should be basically default in that regard...

draupnir config includes:
```
homeserverUrl: http://matrix-traefik:8008
rawHomeserverUrl: http://matrix-traefik:8008
```
which is just the main reverse proxy traefik container's name

---

draupnir service connects networks:
```
--network=matrix-addons
```
```
docker network connect traefik matrix-bot-draupnir
```

---

traefik service connects networks:
```
--network=traefik \
```
which connects it to draupnir such that it should be able to reach `matrix-traefik`, and
```
docker network connect matrix-container-socket-proxy matrix-traefik
```
which is just a firewall between traefik and the host docker socket

---

traefik.yml defines as only entryPoint on 8008:
```
  matrix-internal-matrix-client-api:
    address: :8008
```

---

synapse reverse proxy companion is the only router on that entrypoint i can find:
```
traefik.http.routers.matrix-synapse-reverse-proxy-companion-internal-client-api.entrypoints=matrix-internal-matrix-client-api
traefik.http.routers.matrix-synapse-reverse-proxy-companion-internal-client-api.rule=PathPrefix(`/_matrix`)
```
but it's on the wrong path

meanwhile, the `_synapse/admin_` API is only on
```
traefik.http.routers.matrix-synapse-reverse-proxy-companion-public-client-synapse-admin-api.entrypoints=web-secure
```

---

buuuut you can really skip reading most of this if instead we read the draupnir docs at https://github.com/the-draupnir-project/Draupnir/blob/main/config/default.yaml
```
[...]

PR #3389, opened 2024-06-28
Internal Admin API and Draupnir Hjack Command Config

FSG-Cat:
This PR was made to resolve https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3308 as it popped up again and on the draupnir side with hjacking of rooms in issue https://github.com/the-draupnir-project/Draupnir/issues/463. 

Now this PR is admittedly not my best work but its something. I simply am way too out of my depth here to be sure this is the right way forward. 

I threw something together as well the Labels are claimed to work in https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3308#issuecomment-2096834915 so i assume they are good enough for a start.

Docs for this are matter that can be resolved after the primary problem in this PR is resolved.

Now this PR wont fix the polling. Why? Polling doesnt work in Draupnir in general currently it looks like.


roles/custom/matrix-bot-draupnir/defaults/main.yml
```diff
@@ -75,6 +75,13 @@ matrix_bot_draupnir_raw_homeserver_url: ""
 # Its Exposed here because its common enough to be valid to expose.
 matrix_bot_draupnir_disable_server_acl: "false"
 
+# Used to control if the Synapse Admin API is exposed internally to the containers and therefore giving Draupnir Access.
+matrix_bot_draupnir_admin_api_enabled: ""
+
+# Controls if the draupnir room hijack command is activated or not. This also automatically enables the internal admin API
+# in the process of activation.
+matrix_bot_draupnir_room_hijack_enabled: "false"
```
spantaleev, line 83: It's odd that this appears to a boolean, but its default value in the role is the string `"false"`.

```suggestion
matrix_bot_draupnir_room_hijack_enabled: false
```

roles/custom/matrix-bot-draupnir/defaults/main.yml
```diff
@@ -75,6 +75,13 @@ matrix_bot_draupnir_raw_homeserver_url: ""
 # Its Exposed here because its common enough to be valid to expose.
 matrix_bot_draupnir_disable_server_acl: "false"
 
+# Used to control if the Synapse Admin API is exposed internally to the containers and therefore giving Draupnir Access.
+matrix_bot_draupnir_admin_api_enabled: ""
```
spantaleev, line 79:
```suggestion
matrix_bot_draupnir_admin_api_enabled: false
```

It's odd that this appears to be boolean, but its default value in the role is an empty string.

group_vars/matrix_servers
```diff
@@ -4307,6 +4309,7 @@ matrix_synapse_container_labels_public_client_root_redirection_enabled: "{{ matr
 matrix_synapse_container_labels_public_client_root_redirection_url: "{{ (('https://' if matrix_playbook_ssl_enabled else 'http://') + matrix_server_fqn_element) if matrix_client_element_enabled else '' }}"
 
 matrix_synapse_container_labels_public_client_synapse_admin_api_enabled: "{{ matrix_synapse_admin_enabled }}"
+matrix_synapse_container_labels_internal_client_synapse_admin_api_enabled: "{{ matrix_bot_draupnir_admin_api_enabled }}"
```
spantaleev, line 4312: The `matrix_synapse_container_labels_internal_client_synapse_admin_api_enabled` variable is written to here, but it's not defined in `defaults/main.yml` in the `matrix-synapse` role.

There should be a definition with some default value (`false`) there.

group_vars/matrix_servers
```diff
@@ -4466,6 +4469,7 @@ matrix_synapse_reverse_proxy_companion_container_labels_traefik_hostname: "{{ ma
 
 matrix_synapse_reverse_proxy_companion_container_labels_public_client_synapse_client_api_enabled: "{{ matrix_synapse_container_labels_public_client_synapse_client_api_enabled }}"
 matrix_synapse_reverse_proxy_companion_container_labels_public_client_synapse_admin_api_enabled: "{{ matrix_synapse_container_labels_public_client_synapse_admin_api_enabled }}"
+matrix_synapse_reverse_proxy_companion_container_labels_internal_client_synapse_admin_api_enabled: "{{ matrix_synapse_container_labels_internal_client_synapse_admin_api_enabled }}"
```
spantaleev, line 4472: Like `matrix_synapse_container_labels_internal_client_synapse_admin_api_enabled`, the `matrix_synapse_reverse_proxy_companion_container_labels_internal_client_synapse_admin_api_enabled` variable is written to here, but it's not defined in `defaults/main.yml` in the `matrix-synapse-reverse-proxy-companion` role.

There should be a definition with some default value (`false`) there.

roles/custom/matrix-synapse-reverse-proxy-companion/templates/labels.j2
```diff
@@ -119,6 +119,44 @@ traefik.http.routers.matrix-synapse-reverse-proxy-companion-public-client-synaps
...
 {% endif %}
 
+{% if matrix_synapse_reverse_proxy_companion_container_labels_internal_client_synapse_admin_api_enabled %}
+############################################################
+#                                                          #
+# Internal Synapse Admin API (/_synapse/client)            #
+#                                                          #
+############################################################
+
+traefik.http.routers.matrix-synapse-reverse-proxy-companion-internal-client-synapse-client-api.rule=PathPrefix(`/_synapse/client`)
```
spantaleev, line 129: It'd be better if the rule is coming from a variable (`matrix_synapse_reverse_proxy_companion_container_labels_internal_client_synapse_admin_api_traefik_rule`), defined in `default/main.yml`  in the `matrix-synapse-reverse-proxy-companion` role.

Here's an example: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/e000cbf0f42874bec05611d57a2922259da292bc/roles/custom/matrix-synapse-reverse-proxy-companion/defaults/main.yml#L78

You may even consider adding a `_priority` variable ([like this](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/e000cbf0f42874bec05611d57a2922259da292bc/roles/custom/matrix-synapse-reverse-proxy-companion/defaults/main.yml#L79)), getting its value from the `matrix-synapse` role. In the `matrix-synapse` role, a priority of `0` is a good default, I suppose.

roles/custom/matrix-synapse-reverse-proxy-companion/templates/labels.j2
```diff
@@ -119,6 +119,44 @@ traefik.http.routers.matrix-synapse-reverse-proxy-companion-public-client-synaps
...
+
+
+traefik.http.routers.matrix-synapse-reverse-proxy-companion-internal-client-synapse-client-api.service=matrix-synapse-reverse-proxy-companion-client-api
+traefik.http.routers.matrix-synapse-reverse-proxy-companion-internal-client-synapse-client-api.entrypoints=matrix-internal-matrix-client-api
```
spantaleev, line 133: It'd be better if the entrypoints are coming from a variable (`matrix_synapse_reverse_proxy_companion_container_labels_internal_client_synapse_admin_api_traefik_entrypoints`), defined in `default/main.yml`  in the `matrix-synapse-reverse-proxy-companion` role.

Then, `group_vars/matrix_servers` can pass the correct internal entrypoint like this:

```yml
matrix_synapse_reverse_proxy_companion_container_labels_internal_client_synapse_admin_api_traefik_entrypo: "{{ matrix_playbook_internal_matrix_client_api_traefik_entrypoint_name }}"
```

Here's an example: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/e000cbf0f42874bec05611d57a2922259da292bc/group_vars/matrix_servers#L4320

roles/custom/matrix-synapse-reverse-proxy-companion/templates/labels.j2
```diff
@@ -119,6 +119,44 @@ traefik.http.routers.matrix-synapse-reverse-proxy-companion-public-client-synaps
...
+#                                                          #
+# Internal Synapse Admin API (/_synapse/admin)             #
+#                                                          #
+############################################################
+
+traefik.http.routers.matrix-synapse-reverse-proxy-companion-internal-client-synapse-admin-api.rule=PathPrefix(`/_synapse/admin`)
+
+
+traefik.http.routers.matrix-synapse-reverse-proxy-companion-internal-client-synapse-admin-api.service=matrix-synapse-reverse-proxy-companion-client-api
+traefik.http.routers.matrix-synapse-reverse-proxy-companion-internal-client-synapse-admin-api.entrypoints=matrix-internal-matrix-client-api
```
spantaleev, line 152: Same as for `/_synapse/client` above, it'd be better if these are not using hardcoded values, but the values are coming from variables.

roles/custom/matrix-synapse-reverse-proxy-companion/templates/labels.j2
```diff
@@ -119,6 +119,44 @@ traefik.http.routers.matrix-synapse-reverse-proxy-companion-public-client-synaps
 ############################################################
 {% endif %}
 
+{% if matrix_synapse_reverse_proxy_companion_container_labels_internal_client_synapse_admin_api_enabled %}
```
spantaleev, line 122: It seems like `matrix_synapse_reverse_proxy_companion_container_labels_internal_client_synapse_admin_api_enabled` actually enables 2 things right now: `/_synapse/client` and `/_synapse/admin`.

For the "public" exposure of these, we have dedicated variables: `matrix_synapse_reverse_proxy_companion_container_labels_public_client_synapse_client_api_enabled` and `matrix_synapse_reverse_proxy_companion_container_labels_public_client_synapse_admin_api_enabled`.

So it's odd to have a single variable (`matrix_synapse_reverse_proxy_companion_container_labels_internal_client_synapse_admin_api_enabled`) exposing both right now.

I suggest to create an additional variable (`matrix_synapse_reverse_proxy_companion_container_labels_internal_client_synapse_client_api_enabled`) and use that one to decide whether to expose `/_synapse/client`.

Of course, you can wire the new variable (`matrix_synapse_reverse_proxy_companion_container_labels_internal_client_synapse_client_api_enabled`) in `group_vars/matrix_servers` the same way (based on `matrix_bot_draupnir_room_hijack_enabled`).

I don't know if Draupnir room hijacking requires both `/_synapse/admin` and `/_synapse/client` though. If it only requires the former, then there's no need to enable the latter based on `matrix_bot_draupnir_room_hijack_enabled`.

-----------

You've added labels to `matrix-synapse-reverse-proxy-companion`, which handles the workers-being-enabled case.

Similar labels should be added to `roles/custom/matrix-synapse/templates/labels.j2` as well.
Along with those, variables for each of these should be defined in the `matrix-synapse` role. Example:

- `matrix_synapse_container_labels_internal_client_synapse_admin_api_enabled`
- `matrix_synapse_container_labels_internal_client_synapse_admin_api_traefik_rule`
- `matrix_synapse_container_labels_internal_client_synapse_admin_api_entrypoints`
- `matrix_synapse_container_labels_internal_client_synapse_client_api_enabled`
- `matrix_synapse_container_labels_internal_client_synapse_client_api_traefik_rule`
- `matrix_synapse_container_labels_internal_client_synapse_client_api_entrypoints`

The `matrix-synapse-reverse-proxy-companion` role tends to inherit (via `group_vars/matrix_overrides`) various variable values from `matrix-synapse` ([here](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/e000cbf0f42874bec05611d57a2922259da292bc/group_vars/matrix_servers#L4473-L4474) is an example), so you can wire the new `matrix-synapse-reverse-proxy-companion` variables (enabled, entrypoints, etc.) the same way.

FSG-Cat: I should note that when it comes to this PR i am blindly taking the labels from the issue in question. 

So like it makes sense that these flaws exist. And i need to thank you for this very thuroughly written review. 

I will try to take a look at this during the week to see how much of this stuff i can fix on my own based on all this valuable feedback and trying to look at other roles more closely to understand how they are plumbed. (Primarily Synapse admin i think i will need to look closer at as i think it helps a lot.)

As for the exact API surface Draupnir needs for its admin commands i will come back with an answer on that front as i honestly dont know but i know i can find out.

spantaleev: That's understandable! Thank you for looking into it!

FSG-Cat: So that as a long week /s. Ye this got pushed down the priority list until it dropped of the radar. Cat has finally gotten back to it and started to do some work. I still know theres more work needed based on the last comment. But atleast stuff is moving again and its rebased.

This stuff is hard for me that i will say. But recent developments have gotten me back into having a reason to push this forward.

FSG-Cat: Hopefully this PR should be mostly reviewable now that Cat has hopefully adressed all the feedback given.

spantaleev:
We should be all good now! Thanks! :bow:

PR #3410, opened 2024-07-14
rewrite `just update` command to provide a one-line command to update everything

etkecc:
That's a common problem in the support room when people update the playbook, but don't update roles.
So, let's suggest a new command that will do everything at once - update the playbook and the roles at the same time.

That command is a modified version of etke.cc/ansible's update, that proved to be working flawlessly and (in this version) considers ansible-galaxy and agru existence in the system.

To simplify repo maintenance, that command supports `-u` flag to update requirements.yml file (only if you have agru installed) - this flag is considered for developers only.

So, here are examples of the new options:

* `just update` - similar to `git pull && just roles` before that PR, we should suggest using that command everywhere
* `just update -u` - similar to `git pull && just update` before that PR, only developers should use it to keep the dependencies in sync


spantaleev:
I've remaed `update-self` to `update-playbook-only` to make it clearer what it does. I don't think anyone (or many) will actually call this one directly though.

PR #3413, opened 2024-07-16
doc: mention HTTP/3 in port configuration

bfabio:


spantaleev:
HTTP/3 is enabled by default, so "if you enable it" is somewhat misleading.

bfabio:
@spantaleev you're right, I made the change after casually reading the changelog but before actually updating my server and realizing there was nothing to configure :) 

Fix incoming.

PR #3428, opened 2024-07-17
Update configuring-playbook-synapse.md

throny:
https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3429
Give hints in documentation, that synapse workers won't work with sliding sync


throny:
If confirmed, there should be a note [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-sliding-sync-proxy.md) too

throny:
probably was wrong here, sorry

PR #3469, opened 2024-08-09
Automated MMR signing key generation process

Michael-Hollister:
Automated signing key setup process described in https://docs.t2bot.io/matrix-media-repo/v1.3.5/installation/signing-key/

Homeserver backup key is kept alongside the merged signing key in case server administrator needs to revert.

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -3642,6 +3642,11 @@ matrix_media_repo_systemd_required_services_list_auto: |
     ([devture_postgres_identifier ~ '.service'] if devture_postgres_enabled and matrix_media_repo_database_hostname == devture_postgres_connection_hostname else [])
   }}
 
+matrix_media_repo_generate_signing_key: "{{ matrix_homeserver_implementation == 'synapse' or matrix_homeserver_implementation == 'dendrite'}}"
+matrix_media_repo_homeserver_signing_key: "{{ matrix_media_repo_synapse_signing_key if matrix_homeserver_implementation == 'synapse' else (matrix_media_repo_dendrite_signing_key if matrix_homeserver_implementation == 'dendrite' else '') }}"
+matrix_media_repo_synapse_signing_key: "{{ matrix_synapse_config_dir_path }}/{{ matrix_server_fqn_matrix }}.signing.key"
+matrix_media_repo_dendrite_signing_key: "{{ matrix_dendrite_config_dir_path }}/{{ matrix_server_fqn_matrix }}.signing.pem"
+
 # Auto configured server setup by the playbook
 matrix_media_repo_homeservers_auto:
   -  # Keep the dash from this line.
@@ -3666,7 +3671,7 @@ matrix_media_repo_homeservers_auto:
     # The signing key to use for authorizing outbound federation requests. If not specified,
     # requests will not be authorized. See https://docs.t2bot.io/matrix-media-repo/v1.3.5/installation/signing-key/
     # for details.
-    signingKeyPath: ""
+    signingKeyPath: "{{ '/config/' + matrix_media_repo_identifier + '.signing.key' if matrix_media_repo_generate_signing_key else '' }}"
 
 matrix_media_repo_homeserver_federation_enabled: "{{ matrix_homeserver_federation_enabled }}"
 
--- a/roles/custom/matrix-media-repo/tasks/setup_install.yml
+++ b/roles/custom/matrix-media-repo/tasks/setup_install.yml
@@ -77,6 +77,58 @@
       changed_when: true
       when: "matrix_media_repo_git_pull_results.changed | bool or matrix_media_repo_docker_image_check_result.stdout == ''"
 
+- name: Check existence of media-repo signing key
+  ansible.builtin.stat:
+    path: "{{ matrix_media_repo_config_path }}/{{ matrix_media_repo_identifier }}.signing.key"
+  register: matrix_media_repo_signing_key_stat
+
+- when: "matrix_media_repo_generate_signing_key | bool and not (matrix_media_repo_signing_key_stat.stat.exists | bool)"
+  block:
+    - name: Generate media-repo signing key
+      ansible.builtin.command:
+        cmd: |
+          {{ devture_systemd_docker_base_host_command_docker }} run
+          --rm
+          --name={{ matrix_media_repo_identifier }}
+          --user={{ matrix_synapse_uid }}:{{ matrix_synapse_gid }}
+          --cap-drop=ALL
+          --mount type=bind,src={{ matrix_media_repo_config_path }},dst=/config
+          --workdir='/config'
+          --entrypoint='generate_signing_key'
+          {{ matrix_media_repo_docker_image }}
+          -output {{ matrix_media_repo_identifier }}.signing.key
+        creates: "{{ matrix_media_repo_config_path }}/{{ matrix_media_repo_identifier }}.signing.key"
+
+    - name: Merge media-repo signing key with homeserver signing key
+      ansible.builtin.command:
+        cmd: |
+          {{ devture_systemd_docker_base_host_command_docker }} run
+          --rm
+          --name={{ matrix_media_repo_identifier }}
+          --user={{ matrix_synapse_uid }}:{{ matrix_synapse_gid }}
+          --cap-drop=ALL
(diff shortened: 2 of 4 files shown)


roles/custom/matrix-media-repo/tasks/setup_install.yml
```diff
@@ -77,6 +77,58 @@
...
+          --name={{ matrix_media_repo_identifier }}
+          --user={{ matrix_synapse_uid }}:{{ matrix_synapse_gid }}
+          --cap-drop=ALL
+          --mount type=bind,src={{ matrix_media_repo_config_path }},dst=/config
+          --mount type=bind,src={{ matrix_base_data_path }},dst=/matrix
+          --workdir='/config'
+          --entrypoint='combine_signing_keys'
+          {{ matrix_media_repo_docker_image }}
+          -format {{ matrix_homeserver_implementation }} -output {{ matrix_media_repo_homeserver_signing_key }}.merged {{ matrix_media_repo_homeserver_signing_key }} {{ matrix_media_repo_identifier }}.signing.key
+        creates: "{{ matrix_media_repo_homeserver_signing_key }}."
```
spantaleev, line 116: You're mounting the base path as `/matrix`, but then are accessing some file (`matrix_media_repo_homeserver_signing_key`).

If someone used a different path (e.g. `/xyz`), the file path in `matrix_media_repo_homeserver_signing_key` (which respects `matrix_base_data_path`) would be something like `/xyz/synapse/config/...`.

There's a mismatch been what's available in this container (the base path being exposed at `/matrix`) and what would be accessed (`/xyz/synapse/config/...`).

-----

The `creates` part of this Ansible command invocation seems off. Not sure what your intention was. Maybe it should have a `.merged` suffix?

-----

It's probably easiest if you do it like this:

```suggestion
          --mount type=bind,src={{ matrix_media_repo_config_path }},dst=/config
          --mount type=bind,src={{ matrix_media_repo_homeserver_signing_key | dirname }},dst=/homeserver-signing-key-dir,ro
          --workdir='/config'
          --entrypoint='combine_signing_keys'
          {{ matrix_media_repo_docker_image }}
          -format {{ matrix_homeserver_implementation }} -output /homeserver-signing-key-dir/{{ matrix_media_repo_homeserver_signing_key | basename }}.merged /homeserver-signing-key-dir/{{ matrix_media_repo_homeserver_signing_key | basename }} {{ matrix_media_repo_identifier }}.signing.key
        creates: "{{ matrix_media_repo_homeserver_signing_key }}."
```

group_vars/matrix_servers
```diff
@@ -3642,6 +3642,11 @@ matrix_media_repo_systemd_required_services_list_auto: |
     ([devture_postgres_identifier ~ '.service'] if devture_postgres_enabled and matrix_media_repo_database_hostname == devture_postgres_connection_hostname else [])
   }}
 
+matrix_media_repo_generate_signing_key: "{{ matrix_homeserver_implementation == 'synapse' or matrix_homeserver_implementation == 'dendrite'}}"
+matrix_media_repo_homeserver_signing_key: "{{ matrix_media_repo_synapse_signing_key if matrix_homeserver_implementation == 'synapse' else (matrix_media_repo_dendrite_signing_key if matrix_homeserver_implementation == 'dendrite' else '') }}"
```
spantaleev, line 3646: It seems like these 2 should have definitions in `defaults/main.yml` in the role, given that they are used within the role.

`matrix_media_repo_synapse_signing_key` and `matrix_media_repo_dendrite_signing_key` seem like helper variables used by the playbook, so they're OK to remain in `group_vars/matrix_servers` only.

roles/custom/matrix-media-repo/tasks/setup_install.yml
```diff
@@ -77,6 +77,58 @@
       changed_when: true
       when: "matrix_media_repo_git_pull_results.changed | bool or matrix_media_repo_docker_image_check_result.stdout == ''"
 
+- name: Check existence of media-repo signing key
+  ansible.builtin.stat:
+    path: "{{ matrix_media_repo_config_path }}/{{ matrix_media_repo_identifier }}.signing.key"
+  register: matrix_media_repo_signing_key_stat
+
+- when: "matrix_media_repo_generate_signing_key | bool and not (matrix_media_repo_signing_key_stat.stat.exists | bool)"
+  block:
```
spantaleev, line 86: All the magic happens in this guard block, but.. The `when` will be satisfied early on (in the first sub-task) and there are more tasks after that.

This means that if the playbook dies after the first task creates the file and someone re-runs the playbook later on, we would never again enter this `when` block again. The job would effectively be half-done, without a chance to make it fully done except for figuring out what happened and deleting the `{{ matrix_media_repo_config_path }}/{{ matrix_media_repo_identifier }}.signing.key` file.

-----

A potential solution is:

1. Continue guarding on ` {{ matrix_media_repo_config_path }}/{{ matrix_media_repo_identifier }}.signing.key` as you do now
2. Generate  `{{ matrix_media_repo_config_path }}/{{ matrix_media_repo_identifier }}.signing.key.TEMPORARY` instead of  `{{ matrix_media_repo_config_path }}/{{ matrix_media_repo_identifier }}.signing.key`
3. Work with `{{ matrix_media_repo_config_path }}/{{ matrix_media_repo_identifier }}.signing.key.TEMPORARY` in the `block`
4. Move `{{ matrix_media_repo_config_path }}/{{ matrix_media_repo_identifier }}.signing.key.TEMPORARY` to `{{ matrix_media_repo_config_path }}/{{ matrix_media_repo_identifier }}.signing.key` as a last step in this block

spantaleev (changes requested):
Other than what's suggested in the review (which I have not tested at all):

1. I wonder if this should be enabled by default, or if we should have some way for the user to control it. Right now, people can force-set `matrix_media_repo_generate_signing_key` to `false` to disable this behavior, but they must do it before it has already been done. Thoughts?

2. Given that the procedure is somewhat invasive, I wonder if we should mention it (and the [upstream docs page](https://docs.t2bot.io/matrix-media-repo/v1.3.5/installation/signing-key/) and playbook behavior - merging, backup, etc.) on our matrix-media-repo docs page?

3. Someone removing matrix-media-repo and deleting all its files (from `matrix_media_repo_base_path`) and re-enabling it would probably end up with a signing key which contains 3 keys (homeserver original key + 1st matrix-media-repo signing key + 2nd matrix-media-repo signing key). This may not be as bad as it sounds, I guess.

Michael-Hollister:
Thanks for the feedback, updated PR with your suggestions.

> 1. I wonder if this should be enabled by default, or if we should have some way for the user to control it. Right now, people can force-set `matrix_media_repo_generate_signing_key` to `false` to disable this behavior, but they must do it before it has already been done. Thoughts?
> 2. Given that the procedure is somewhat invasive, I wonder if we should mention it (and the [upstream docs page](https://docs.t2bot.io/matrix-media-repo/v1.3.5/installation/signing-key/) and playbook behavior - merging, backup, etc.) on our matrix-media-repo docs page?

Key generation is enabled by default and can be disabled by setting `matrix_media_repo_generate_signing_key` to false as you said. I think having the MMR docs page explaining the process would be sufficient. I've updated the MMR docs page explaining the process in more detail, and feel free to let me know if any further changes you'd like to see. 

> 3. Someone removing matrix-media-repo and deleting all its files (from `matrix_media_repo_base_path`) and re-enabling it would probably end up with a signing key which contains 3 keys (homeserver original key + 1st matrix-media-repo signing key + 2nd matrix-media-repo signing key). This may not be as bad as it sounds, I guess.

Yeah true, if a user manually deletes the MMR directory and re-runs the setup role, old keys will not be removed. If the user operates outside the playbook, I don't think this is a case we can easily protect against.

We could potentially auto remove old keys if the user runs the setup role with MMR disabled. However perhaps its best if left to the user to manually remove the key for the following reasons:
1. The playbook could restore from the `.backup` homeserver signing key. This could be destructive however since if the user modifies the signing key file after running the MMR setup role, those changes will be lost.
2. The playbook could instead search for the MMR key id in the homeserver signing key file before removal in the uninstall task. This could work, though a bit more code to maintain.
3. Its recommended to update your homeserver config when revoking/expiring old keys (`old_signing_keys` field for Synapse and `old_private_keys` for Dendrite). This would be harder to automate, and changes to the config would be lost when re-running the homeserver setup roles.

Feel free to discuss if you think otherwise.

PR #3548, opened 2024-09-25
synapse-admin v0.10.3-etke17: add `Contact support` menu item

aine-etke:
This allows users to quickly find support in case of issues. The link is configurable via config.json - `supportURL`

image

### DIFF
--- a/roles/custom/matrix-synapse-admin/defaults/main.yml
+++ b/roles/custom/matrix-synapse-admin/defaults/main.yml
@@ -14,7 +14,7 @@ matrix_synapse_admin_container_image_self_build: false
 matrix_synapse_admin_container_image_self_build_repo: "https://github.com/etkecc/synapse-admin.git"
 
 # renovate: datasource=docker depName=ghcr.io/etkecc/synapse-admin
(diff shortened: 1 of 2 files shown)


roles/custom/matrix-synapse-admin/defaults/main.yml
```diff
@@ -175,6 +175,9 @@ matrix_synapse_admin_configuration: "{{ matrix_synapse_admin_configuration_defau
 # restricts the homeserver(s), so that the user can no longer define a homeserver manually during login.
 matrix_synapse_admin_config_restrictBaseUrl: "{{ [matrix_homeserver_url] }}"  # noqa var-naming
 
+# Controls the supportURL configuration setting, which, if defined, replaces the default link to the Synapse Admin GitHub repository.
```
spantaleev, line 178: This comment seems to be inaccurate.

aine-etke: [why?](https://github.com/etkecc/synapse-admin/blob/main/src/components/AdminLayout.tsx#L4)

PR #3569, opened 2024-10-03
Revert Mjolnir version to 1.7.0 due to severe breakage.

FSG-Cat:
github.com/matrix-org/mjolnir/issues/531

Any user not using E2EE support and therefore using User password auth is unable to launch updated Mjolnir. Due to this severe breakage its recomended to revert to before this bug is introduced until its fixed.

### DIFF
--- a/roles/custom/matrix-bot-mjolnir/defaults/main.yml
+++ b/roles/custom/matrix-bot-mjolnir/defaults/main.yml
@@ -5,7 +5,7 @@
 matrix_bot_mjolnir_enabled: true
 
 # renovate: datasource=docker depName=matrixdotorg/mjolnir
-matrix_bot_mjolnir_version: "v1.8.1"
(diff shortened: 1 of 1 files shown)


MichaelSasser:
The current workaround would be setting `encryption.use` to `false` (as a default) until https://github.com/matrix-org/mjolnir/issues/534 was resolved.

FSG-Cat:
I agree that we should now have acceptable workarounds to the bug.

MichaelSasser:
The issue was picked up already and resolved: https://github.com/matrix-org/mjolnir/pull/535

I'm guessing there will be a new release soon.

MichaelSasser:
I've tried out the new release [v1.8.3](https://github.com/matrix-org/mjolnir/releases/tag/v1.8.3) and can confirm, it is working without the workaround. So the issue I described in https://github.com/matrix-org/mjolnir/issues/531 has been resolved.

PR #3573, opened 2024-10-04
Update Cinny nginx.conf.j2 to include service worker rewrite

hanthor:
Cinny has added a service worker to support authenticated media and requires /sw,js to resolve to the serviceworker js file


hanthor:
currently all thubnails and images ore broken for MDAD deployments of Cinny

hanthor:
tried this fix manually and it fixed my media in Cinny

spantaleev:
Thanks!🙏🏻

PR #3592, opened 2024-10-12
Update configuring-playbook-synapse-admin.md

ilterugur:
Fixed compatibility misinformation


spantaleev:
It doesn't work with Dendrite at all?

I thought the admin API is somewhat compatible and that synapse-admin works to some extent (even if it's not 100% usable).

aine-etke:
it doesn't

PR #3617, opened 2024-10-16
Use a common expression for suggesting to uncomment and adjust configuration

luixxiul:
It seems to be that `Uncomment and adjust this part if you'd like to…` is a common expression to suggest uncommenting a line below it and adjust the configuration. Having the same pattern will surely help us in some way :-)


spantaleev:
Agreed! 👍🏻

Issue #3625, opened 2024-10-17
Adopt special-use domain names such as example.com on documentation


luixxiul:
**Describe the bug**
On documentation files there has been fairly a lot of various ways to indicate domains. Here is a list of examples:

- DOMAIN
- YOUR_DOMAIN
- `<example.com>`
- domain.com
- mydomain.com
- mydomain.tld
- domain.tld
- `<your-domain>`
- …

**To Reproduce**
Irrelevant

**Expected behavior**
For documentation there are special domain names reserved for documentations (see: https://datatracker.ietf.org/doc/html/rfc6761); using example.com, example.net, or example.org would help us to minimize the possibility of confusion.

**Matrix Server:**

**Ansible:**

**Client:**

**Additional context**

PR #3628, opened 2024-10-17
Adopt special-use domain names for documentation purpose

luixxiul:
Closes https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3625

This PR intends to replace strings which indicate homeserver's domain such as `YOUR_DOMAIN`, `domain.com`, `mydomain.com`, `domain.tld`, and `server.com`, with special-use domain names such as example.com, example.net, and example.org, replacing hard-coded values with the template `{{ matrix_domain }}`.

Hopefully adopting them should make it easier to set up the instance as you just would need to replace all `example.com` strings with your own domain name once for all.

If the placeholder clearly indicates the homeserver, it is replaced with example.com. If it is clear that the placeholder indicates another server, then it is replaced with either example.net or example.org.

To be fair I personally like the expression `DOMAIN` as it stands out, but sticking to the technical standard should be preferred.


docs/configuring-playbook-conduit.md
```diff
@@ -36,7 +36,7 @@ You will have to manually register appservices using the the [register-appservic
 Find the `registration.yaml` in the `/matrix` directory, for example `/matrix/mautrix-signal/bridge/registration.yaml`, then pass the content to Conduit:
 
 
-    @conduit:your.server.name: register-appservice
+    @conduit:example.com: register-appservice
```
luixxiul, line 39: The change should be fine; `@conduit:matrix.example.com` should be incorrect.

spantaleev: Sounds correct

docs/configuring-well-known.md
```diff
@@ -50,15 +50,15 @@ matrix_static_files_file_matrix_support_enabled: true
...
   - matrix_id: "@admin2:{{ matrix_domain }}"
-    email_address: admin2@domain.tld
+    email_address: admin2@example.com
     role: m.role.admin
-  - email_address: security@domain.tld
+  - email_address: security@example.com
     role: m.role.security
 
-matrix_static_files_file_matrix_support_property_m_support_page: "https://example.domain.tld/support"
+matrix_static_files_file_matrix_support_property_m_support_page: "https://matrix.example.com/support"
```
luixxiul, line 61: This change should also be fine, though I am not perfectly sure.

spantaleev: This property points to the "support page".

The example URL points to the base domain (or some other domain) which hosts your organization's support page (if any).

More details can be found on this MSC page: https://github.com/matrix-org/matrix-spec-proposals/blob/hs/proposal-admin-contact-1/proposals/1929-admin-contact.md

So I'd say `example.com/support` is a better address, to clearly indicate that it's not the Matrix homeserver that is hosting this page. You need to host it yourself somewhere.

roles/custom/matrix-conduit/templates/conduit.toml.j2
```diff
@@ -10,7 +10,7 @@
 # and room IDs. Examples: matrix.org, conduit.rs
 
 # The Conduit server needs all /_matrix/ requests to be reachable at
-# https://your.server.name/ on port 443 (client-server) and 8448 (federation).
+# https://example.com/ on port 443 (client-server) and 8448 (federation).
```
luixxiul, line 13: The change should also be fine, I think.

luixxiul:
I checked the whole change repeatedly and it should be fine now, while it is possible that other strings which have been used to describe domain names are left unchanged :crossed_fingers:

spantaleev:
Thank you for your diligence! :bow:

PR #3646, opened 2024-10-21
Update documentation for components about adjusting the URL and the DNS records

luixxiul:
This adopts the structure of `docs/configuring-playbook-matrix-authentication-service.md` which was recently created.

Here is the template which this PR follows, based on that file.

````
- …
- Adjusting the playbook configuration
  - … 
  - Adjusting the (service name here) URL
  - …
- Adjusting DNS records
- Installing
- …
````

The details of the instructions are slightly different between the documentations for components which require subdomain settings by default (such as `buscarron`) and the ones which do not (such as `matrix-alertmanager-receiver`). Each has been handled by the individual commits separately for easier review.

This also adds the description about the subdomain of Rageshake.


luixxiul:
Indeed, I'll see what can be improved here :-)

Issue #3631, opened 2024-10-18
Create a style guide for playbook developers


luixxiul:
**Is your feature request related to a problem? Please describe.**

Since there has been a lot of cases where inconsistent usage of common terms on documentation, a style guide for playbook developers should make it easier for us to create a new playbook component, keeping the style consistent.

**Describe the solution you'd like**

Create a document after finishing these:

- Define how typical and common terms should be written; e.g. use special-use domains (example.com, example.net. etc.) instead of arbitrary ones
- Define capitalization; e.g. proper nouns (Matrix, Element, Mjolnir, etc.) should be capitalized
- Define a common structure of documentation; e.g. a documentation for bridge should include sections for "prerequisites", "adjusting the playbook configuration", "DNS", "Installing", "Usage", etc.

**Describe alternatives you've considered**
Creating a lint tool should enforce the tool, though it must be overkill.

**Additional context**

spantaleev:
An initial version of the style guide landed in 924c471497d507269f083e6c9782ab5ac2798a4b: [docs/style-guide.md](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/style-guide.md).

It codifies the conventions from this issue's checklist (example values based on the Matrix specification, capitalization, line break and list numbering rules, the common documentation page structure, README.md and docs/container-images.md registration) plus development conventions around variable naming, deprecation, and where new components should live (this playbook, an external MASH role, or mash-playbook).

A companion AGENTS.md file (cb574c88ee4daa6df4b2329ee8752d484bf8a27e) points AI coding agents at the style guide too, since these days they write a good share of the contributions.

@luixxiul this is meant as a starting point rather than the final word: please iterate on it as you see fit.

PR #3653, opened 2024-10-21
Update room IDs and alias with ones on the Matrix Specification

luixxiul:
Related: https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3631

This PR intends to update room IDs and alias with ones on the Matrix Specification.

See:
- https://spec.matrix.org/v1.12/#room-structure where the room ID `!qporfwt:matrix.org` is used as an example.
- https://spec.matrix.org/v1.12/#room-aliases where the room ID `!aaabaa:matrix.org` is used as an example.
- https://spec.matrix.org/v1.12/#room-aliases where the room alias `#matrix:example.org` is used as an example.

If I am honest, I do not think those are nice choices... they should be something which looks more as example, but using ones on the specs should help us to keep consistency when they are updated on a future spec version.

### DIFF
--- a/docs/configuring-playbook-alertmanager-receiver.md
+++ b/docs/configuring-playbook-alertmanager-receiver.md
@@ -25,12 +25,12 @@ matrix_alertmanager_receiver_config_matrix_access_token: ''
 # Optionally, configure some mappings (URL-friendly room name -> actual Matrix room ID).
 #
 # If you don't configure mappings, you can still deliver alerts using URLs like this:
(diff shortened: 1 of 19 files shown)


spantaleev:
I'm also a bit worried that using something like `!qporfwt:example.com` may be less visible (and easier to miss when adjusting configs) than `!your-room-id:your-server.com`.

It's probably fine though.

luixxiul:
I guess something such as `!example:example.com`, `!example:example.net`, and `#example:example.org` could be better, though there are too much "example" here..

I'll open an issue to track the issue.

PR #3665, opened 2024-10-22
Update docs/configuring-playbook-bridge-appservice-slack.md: fix indentation

luixxiul:


luixxiul:
It does not make visual difference, but patching it should be better than not done from the viewpoint of strictness :-D

PR #3674, opened 2024-10-24
Fix capitalization: Sliding Sync proxy, rageshake

luixxiul:
rageshake does seem to be natural if it is capitalized (Rageshake), but it apparently isn't: https://github.com/matrix-org/rageshake?tab=readme-ov-file#rageshake-


spantaleev:
`baibot` is the same way - we prefer lower-case. I believe the playbook does it correctly though.

Capitalization.. it can sometimes fill you with rage ;)

luixxiul:
Reflected this one on https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3631.

PR #3677, opened 2024-10-24
optional role for matrix-fluffygate

ayushin:


docs/configuring-playbook-fluffygate.md
```diff
@@ -0,0 +1,147 @@
...
+4. Configure the notification templates and options as needed through the playbook variables
+
+### Debugging
+
+If you need to troubleshoot issues:
+
+1. Enable debug logs by setting:
+```yaml
+matrix_fluffygate_debug_logs: true
+```
```
luixxiul, line 126:
```suggestion
    ```yaml
    matrix_fluffygate_debug_logs: true
    ```
```

docs/configuring-playbook-fluffygate.md
```diff
@@ -0,0 +1,147 @@
...
+
+2. Check the container logs:
+```bash
+docker logs matrix-fluffygate
+```
```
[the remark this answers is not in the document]
spantaleev: `docker logs` will not work for any of the services configured via this playbook.

All containers are started via systemd services, which automatically log to systemd-journald (logs are exposed via `journalctl -fu matrix-SERVICE`).

Because logging already happens via systemd-journald, we explicitly pass  `--log-driver=none` to `docker run` (and `docker create`) commands, so that we won't do double-logging.

So.. this section's `docker logs` command is inaccurate. You can adjust it to point people to `journalctl` instead, but it's probably even better to link to `docs/docs/maintenance-and-troubleshooting.md` which discusses logging in a more generic way.

roles/custom/matrix-fluffygate/.claudesync/config.local.json
```diff
@@ -0,0 +1,7 @@
+{
+  "active_provider": "claude.ai",
+  "active_organization_id": "4c1644e0-e630-48f7-bb87-4d8387ef9cd8",
+  "active_project_id": "1d6c1d5c-5270-439c-9885-aff7168cb24d",
+  "active_project_name": "matrix-fluffygate",
+  "local_path": "/Users/alexis/Work/witi/witi-matrix-ansible/roles/custom/matrix-fluffygate"
+}
```
spantaleev, line 7: We don't need this.

docs/configuring-playbook-fluffygate.md
```diff
@@ -0,0 +1,147 @@
+# Setting up Fluffygate (optional)
+
+The playbook can install and configure [Fluffygate](https://github.com/krille-chan/fluffygate), a simple Push Gateway for Fluffychat.
+
+See the project's documentation to learn what it does and why it might be useful to you.
+
+**Note**: most people don't need to install their own gateway. This optional playbook component is only useful to people who develop/build their own Matrix client applications themselves, as you'll need access to your own Firebase/FCM and APNS credentials.
```
spantaleev, line 7: It may be useful to also mention Sygnal here (`docs/configuring-playbook-sygnal.md`) here and mention Fluffygate on the Sygnal docs page as well.

If this "mention" is also a brief comparison as to when one might use one or the other, it'd be even better. It seems like Sygnal is supports direct delivery via APNS , while Fluffygate only supports FCM? Maybe this makes setup easier as all platform-specific setup would happen on the FCM side?

docs/configuring-playbook-fluffygate.md
```diff
@@ -0,0 +1,147 @@
...
+      mutable-content: 1
+```
+
+For a complete list of available configuration options, see the `defaults/main.yml` file in the role.
+
+### Required Configuration
+
+The following settings are required and must be defined:
+- `matrix_fluffygate_hostname`
+- `matrix_fluffygate_path_prefix`
```
spantaleev, line 56: Should we consider automatically configuring `matrix_fluffygate_hostname` to point to `matrix_server_fqn_matrix` and `matrix_fluffygate_path_prefix` to be something like `/fluffygate`, etc?

This way we'll have a reasonable default.

Have you confirmed that hosting it under a path prefix works as expected?

ayushin: I'll clean things up bit later - very little time these days, yet wanted to contribute to the upstream.

I tried hosting on the same host but that did not work for some reason and I went with a dedicated hostname instead. Can try again later.

luixxiul:
It would be great if you would add an entry about Fluffygate to `README.md` (around [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/509542ccaf8667e759866c988baf6ea6c7dff994/README.md?plain=1#L182)) and `container-images.md` (around [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/509542ccaf8667e759866c988baf6ea6c7dff994/docs/container-images.md?plain=1#L150)) :smiley:

docs/configuring-playbook-fluffygate.md
```diff
@@ -0,0 +1,147 @@
...
+matrix_fluffygate_container_labels_traefik_rule: "Host(`{{ matrix_fluffygate_container_labels_traefik_hostname }}`)"
+matrix_fluffygate_container_labels_traefik_priority: 0
+matrix_fluffygate_container_labels_traefik_entrypoints: web-secure
+```
+
+## Adjusting DNS records
+
+You will need to configure your DNS records to point the Fluffygate hostname to your server. This typically involves creating either:
+- an A record pointing to your server's IPv4 address
+- a CNAME record pointing to your server's hostname
```
luixxiul, line 94: Per existing lines for Sygnal [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/509542ccaf8667e759866c988baf6ea6c7dff994/docs/configuring-playbook-sygnal.md?plain=1#L70-L76).

Please update the default CNAME record `push`, when addressing the [comment](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/3677/files#r1818803261).

```suggestion
## Adjusting DNS records

Once you've decided on the domain and path, **you may need to adjust your DNS** records to point the Fluffygate domain to the Matrix server.

By default, you will need to create a CNAME record for `push`. See [Configuring DNS](configuring-dns.md) for details about DNS changes.

If you've decided to reuse the `matrix.` domain, you won't need to do any extra DNS configuration.
```

docs/configuring-playbook-fluffygate.md
luixxiul: @ayushin I'd second- could you please consider it?

PR #3686, opened 2024-10-27
Fixup variable name thats install nginx grafana dashboard

ginta1337:
As for now nginx grafana dashboard is installed in provisioning config directory, not in dashboards one.
To fix this [new configuration option was introduced](https://github.com/mother-of-all-self-hosting/ansible-role-grafana/pull/2) in grafana role.

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -5421,7 +5421,7 @@ grafana_dashboard_download_urls: |
     (matrix_synapse_usage_exporter_dashboard_urls if matrix_synapse_usage_exporter_enabled else [])
   }}
 
-grafana_provisioning_dashboard_template_files: |
+grafana_dashboard_files: |
   {{
     ([{
         'path': 'roles/custom/matrix-prometheus-nginxlog-exporter/templates/grafana/nginx-proxy.json',


spantaleev:
Has something changed in newer Grafana versions to make this necessary? Why do we need to do this?

I have not heard of people complaining about dashboards missing, but it may be that no one has reported it or I've missed it.

ginta1337:
The command-line argument `cfg:default.paths.provisioning=` controls where Grafana searches for the provisioning configuration directory structure, and by default, this is `/etc/grafana/provisioning` inside the container. The dashboard search path is defined in `/etc/grafana/provisioning/dashboards/default.yaml` as `/etc/grafana/dashboards`.

The Ansible variable `grafana_provisioning_dashboard_providers` controls the content of `default.yaml`, while the variable `grafana_provisioning_dashboard_template_files` allows additional provisioning configuration files to be placed in the same directory. The dashboard provisioning syntax is described [here](https://grafana.com/docs/grafana/latest/administration/provisioning/#dashboards).

It is incorrect to place exported dashboards in the provisioning configuration path. I hope this clarifies the origin of this obvious issue. Additionally, redefining the `grafana_dashboard_files` variable can help users add their custom dashboards more effectively.

PR #3688, opened 2024-10-28
Export Traefik metrics to Prometheus and add Grafana dashboard for them

ginta1337:
Introduce __matrix_prometheus_traefik_exporter_enabled__ variable. Setting it to `true` enables export [Traefik metrics](https://doc.traefik.io/traefik/observability/metrics/prometheus) to Prometheus and adds Grafana dashboard for them.

### DIFF
--- a/docs/configuring-playbook-prometheus-grafana.md
+++ b/docs/configuring-playbook-prometheus-grafana.md
@@ -16,7 +16,7 @@ prometheus_node_exporter_enabled: true
 prometheus_postgres_exporter_enabled: true
 
 # You can remove this, if unnecessary.
-matrix_prometheus_nginxlog_exporter_enabled: true
+matrix_prometheus_traefik_exporter_enabled: true
 
 grafana_enabled: true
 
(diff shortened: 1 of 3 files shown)


spantaleev:
I'm not sure what you're doing here and why.

You seem to be replacing `matrix_prometheus_nginxlog_exporter_enabled` which toggles the [matrix-prometheus-nginxlog-exporter](https://github.com/spantaleev/matrix-docker-ansible-deploy/tree/dc461004b472ce38e980d52899e1ad65121c7947/roles/custom/matrix-prometheus-nginxlog-exporter) Ansible role (which installs the [martin-helmich/prometheus-nginxlog-exporter](https://github.com/martin-helmich/prometheus-nginxlog-exporter/) component) with some new other variable (`matrix_prometheus_traefik_exporter_enabled`) which does other things.

While Traefik is the reverse-proxy that the playbook uses nowadays, nginx is still used in one or more services (like `matrix-synapse-reverse-proxy-companion` - auto-enabled if Synapse workers are enabled). Using `prometheus-nginxlog-exporter` provides a way to observe this service.

This PR seems to be adding something additional (the ability to observe Traefik metrics via Prometheus), but:

- it breaks  `prometheus-nginxlog-exporter` installation
- it introduces a `matrix_prometheus_traefik_exporter_enabled` variable which I'm not sure is necessary

I suppose there's no need for `matrix_prometheus_traefik_exporter_enabled`. We can tell people to flip `traefik_config_metrics_prometheus_enabled` instead and to have that automatically cascade into modifying `matrix_prometheus_services_connect_scraper_traefik_enabled`, etc.

For other components, `grafana_dashboard_download_urls` is redefined in `group_vars/matrix_servers` and various dashboards are auto-enabled based on "metrics enabled" variables in the various Ansible roles: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/dc461004b472ce38e980d52899e1ad65121c7947/group_vars/matrix_servers#L5409-L5422

The same can be done for Traefik. If `traefik_config_metrics_prometheus_enabled`, then some `_dashboard_urls` variable (containing Traefik-related dashboards) can be injected into `grafana_dashboard_download_urls`. You have currently defined these dashboard URLs in `matrix_traefik_dashboard_download_*` variables. While it's nice that this allows extra configurability as to whether dashboard download should happen or not (which is not the case for the other components), it would probably be nice if the Traefik dashboard URL lived (was exported by) the Traefik role itself. This also makes it easier for people using these roles independently (or in the context of [mash-playbook](https://github.com/mother-of-all-self-hosting/mash-playbook)) to easily add dashboards to their Grafana instance.

mash-playbook does not automatically populate `grafana_dashboard_download_urls` like we do for matrix-docker-ansible-deploy. For mash-playbook, we've explicitly decided to avoid such wiring magic. The only wiring magic there is the auto-configuration of Postgres/MariaDB. It's up to the user to wire other services together as necessary. Still, it 's helpful if roles expose `_dashboard_urls*` variables that can be used for people wishing to do this wiring.

PR #3696, opened 2024-10-28
Recategorize Postmoogle

luixxiul:


spantaleev:
This one is confusing, because Postmoogle is:

- kind of a bot in the sense that:
  - it uses the [Client-Server API](https://spec.matrix.org/v1.12/client-server-api), not the [application service API](https://spec.matrix.org/v1.12/application-service-api/))
  - it doesn't create manage any puppet users - users only interact with a single Postmoogle bot user
- but also kind of a bridge, because it effectively bridges Matrix to the email protocol

Because of the latter, I'd call it a bridge more-so than call it a bot, I guess.

aine-etke:
Postmoogle is a bridge that doesn't use AS API, but rather uses CS API:

* If by "bridge" you mean "a software that uses AS API", then Postmoogle can be considered a bot.
* **But** if the meaning of "bridge" is "a software that bridges a different protocol(-s) into matrix", then Postmoogle is a bridge.

luixxiul:
I see, I will re-categorize it as a bridge instead of a bot.

luixxiul:
I am not quite sure if it would be safe to replace variables `matrix_bot_postmoogle_*` with `matrix_postmoogle_*`. Apparently it must cause backward compatibility breaks.

spantaleev:
Thanks!

Indeed, changing the variables would be a compatibility break. If done silently and people don't adapt their configuration on time, they will find Postmoogle uninstalled from their server.

By adding validation tasks, the playbook can catch usage of the old `matrix_bot_postmoogle_*` variables automatically and ask people to migrate to the new `matrix_postmoogle_*` variables. The playbook would refuse to run until the migration in `vars.yml` is performed.

We have some similar things in the past (see `roles/custom/matrix_playbook_migration/tasks/validate_config.yml`) and it may be worth doing it again for Postmoogle.

Here's an example: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/ecd4fc028d071116423bdc104755027d3018608c/roles/custom/matrix_playbook_migration/tasks/validate_config.yml#L211-L224

The validation task is guarded by a variable (`matrix_playbook_migration_matrix_jitsi_migration_validation_enabled`), which allows it to be skipped. Certain playbook users (like https://etke.cc/ but possibly others) may wish to disable the validation task temporarily to allow for a more graceful migration to happen (wiring new variables in terms of the old variables; eliminating the old variables later on).

PR #3698, opened 2024-10-28
Rename variables of Postmoogle to handle it as a bridge

luixxiul:
Follow-up to https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/3696

Though I tried to make changes sensible as much as I could, I'm not quite convinced if they indeed make sense. Reviews would be appreciated pretty much. Also please note that the change is not tested!


roles/custom/matrix_playbook_migration/tasks/validate_config.yml
```diff
@@ -223,6 +223,20 @@
           We found usage of the following variables: {{ matrix_playbook_migration_jitsi_migration_vars.keys() | join(', ') }}
       when: "matrix_playbook_migration_jitsi_migration_vars | length > 0"
 
+- when: matrix_playbook_migration_matrix_jitsi_migration_validation_enabled | bool
```
spantaleev, line 226: This `when` variable should be adjusted

luixxiul: Done → https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/3698/commits/67205297ba40af3d8be471f157872a52e47b8818

luixxiul:
I really hope this will not break anything...

spantaleev:
With a bunch of additional last-minute fixes (see the commits above), I think it should be fine.

luixxiul:
Thanks for taking care of it!

Issue #3681, opened 2024-10-25
docs/configuring-playbook-bridge-appservice-slack.md is outdated in that creation of classic Slack app was discontinued


luixxiul:
**Describe the bug**

docs/configuring-playbook-bridge-appservice-slack.md mentions a classic Slack app [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/b63d8a568739af53b82a63b8cf70e63fb15b307b/docs/configuring-playbook-bridge-appservice-slack.md?plain=1#L26-L27) and [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/b63d8a568739af53b82a63b8cf70e63fb15b307b/docs/configuring-playbook-bridge-appservice-slack.md?plain=1#L48-L54), but it seems no longer possible to create a new classic app since the last June. See: https://github.com/matrix-org/matrix-appservice-slack/issues/789

https://github.com/matrix-org/matrix-appservice-slack/issues/789#issuecomment-2172947787 says:

> Currently the best (and only) option for new installations is to use the webhook bridging.

Because `matrix-appservice-webhooks` is no longer maintained as we claim [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/b63d8a568739af53b82a63b8cf70e63fb15b307b/docs/configuring-playbook-bridge-appservice-webhooks.md?plain=1#L5), it seems `matrix-hookshot` should be suggested for new installations.

**To Reproduce**

**Expected behavior**

- Add the note to docs/configuring-playbook-bridge-appservice-slack.md that it has been unavailable to use for new installations since the last June.
- Add the link to the documentation to  docs/configuring-playbook-bridge-hookshot.md, claiming that the webhook is the only effective solution for new installations for now.

**Matrix Server:**

**Ansible:**

**Client:**

**Additional context**

I am wondering if the discontinuation of a new classic app would possibly affect the other two bridges for Slack, [Mautrix Slack bridging](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/b63d8a568739af53b82a63b8cf70e63fb15b307b/docs/configuring-playbook-bridge-mautrix-slack.md) and [MX Puppet Slack bridging](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/b63d8a568739af53b82a63b8cf70e63fb15b307b/docs/configuring-playbook-bridge-mx-puppet-slack.md) as well.

PR #3706, opened 2024-10-31
Update documentation about Slack bridges

luixxiul:
Closes https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3681

- Add a note that `matrix-appservice-slack` is not usable for new installation
- Edit notes that `matrix-appservice-webhooks` has been deprecated


docs/configuring-playbook-bridge-appservice-slack.md
```diff
@@ -1,6 +1,8 @@
...
+**Notes**:
+- Bridging to [Slack](https://slack.com) can also happen via the [mx-puppet-slack](configuring-playbook-bridge-mx-puppet-slack.md) and [mautrix-slack](configuring-playbook-bridge-mautrix-slack.md) bridges supported by the playbook.
+- Currently (as of November, 2024) this component is not available for new installation unless you have already created a classic Slack application (which the bridge makes use of in order to enable bridging between Slack and Matrix), because the creation of classic Slack applications has been discontinued since June 4 2024. The author of the bridge claims [here](https://github.com/matrix-org/matrix-appservice-slack/issues/789#issuecomment-2172947787) that he plans to support the modern Slack application and until then "the best (and only) option for new installations is to use the webhook bridging". The playbook supports the Webhook with [matrix-hookshot](configuring-playbook-bridge-hookshot.md). While the playbook supports `matrix-appservice-webhooks`, it has been deprecated in favor of matrix-hookshot.
```
spantaleev, line 5: I think Hookshot's webhook payload was not very compatible with Slack's payload (nowadays).

When they say "webhook bridging", I don't know if they mean using a a webhook service (bridge) like Hookshot or rather a Slack bridge which relies on "webhook technology" to do its work. Because of this, I don't think referencing Hookshot and appservice-webhooks is of much use here.

PR #3716, opened 2024-11-01
Add labels to issue templates

luixxiul:
Please note that `suggestion` label needs to be created first as it does not exist for the repository.


spantaleev:
Created, thank you!

luixxiul:
Confirmed; it should work as expected!

PR #3718, opened 2024-11-01
Update docs/container-images.md: create tables following README.md

luixxiul:
This is essentially an imitation of 59a01da for container images. It copies the entries order and descriptions on the table of README.md

There are several services which require self building, and `matrix-synapse-rest-auth`, `matrix-synapse-shared-secret-auth`, `matrix-synapse-ldap3`, `synapse-s3-storage-provider`, `synapse_auto_accept_invite`, and `synapse-simple-antispam` do not seem to have its container image.


spantaleev:
Great - this is much cleaner! :bow:

PR #3726, opened 2024-11-02
Update docs/maintenance-postgres.md: add a note about upgrading PostgreSQL

luixxiul:


docs/maintenance-postgres.md
```diff
@@ -73,6 +73,8 @@ Restoring a backup made this way can be done by [importing it](importing-postgre
 
 ## Upgrading PostgreSQL
 
+**Note**: before upgrading PostgreSQL, please take extra precautions such as checking service's PostgreSQL version requirement. There is a possibility that some services do not support the latest version yet. If the service on your homeserver is incompatible with the upgraded version, you will have to wait until the incompatibility is addressed, address it by yourself, or revert the PostgreSQL version to keep using the service.
```
spantaleev, line 76: I believe the only time compatibility issues have arisen until now is with Borgmatic.

As far as I know, all other services have always been compatible as soon as the new Postgres version was released.

luixxiul: That's fair- understood!

PR #3739, opened 2024-11-04
Update docs/installing.md: playbook tags about starting and restarting services

luixxiul:
It is no longer correct that running tasks with "setup-all" and "install-all" tags would not start or restart services.

### DIFF
--- a/docs/installing.md
+++ b/docs/installing.md
@@ -15,11 +15,11 @@ The general command syntax is: `ansible-playbook -i inventory/hosts setup.yml --
 
 Here are some playbook tags that you should be familiar with:
 
-- `setup-all` - runs all setup tasks (installation and uninstallation) for all components, but does not start/restart services
+- `setup-all` - runs all setup tasks (installation and uninstallation) for all components, and starts/restarts services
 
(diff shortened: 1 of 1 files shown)


spantaleev:
There's a confusing difference between the `install-all` just recipe (and `Makefile` target) and the `install-all` playbook tag.

The recipe (e.g. `just install-all`) automatically restarts services at the end of execution, because it actually runs `just run-tags install-all,start` (note the `start` tag) under the hood.

The playbook tag (e.g. `just run-tags install-all`) does **not** restart services. Only the `start` tag does it.

------

The situation is the same for `setup-all` and others edited by this PR.

So the changes proposed in this PR are not correct, but maybe we should clear up the confusion somehow.

luixxiul:
I see, I did not know the difference between the just recipe and the playbook tag as I'm a user of `ansible-playbook` myself and have not used `just` at all.

The difference indeed should be worth noting as I'm not quite sure how many of us are quite accustomed to it.

luixxiul:
Closing in favor of https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3741.

PR #3744, opened 2024-11-05
Update docs/configuring-dns.md: remove an incorrect instruction to the next step

luixxiul:
The right next step of docs/configuring-dns.md is docs/getting-the-playbook.md, as shown on the last line.

### DIFF
--- a/docs/configuring-dns.md
+++ b/docs/configuring-dns.md
@@ -24,8 +24,6 @@ Be mindful as to how long it will take for the DNS records to propagate.
 
 If you are using Cloudflare DNS, make sure to disable the proxy and set all records to `DNS only`. Otherwise, fetching certificates will fail.
 
-When you're done configuring DNS, proceed to [Configuring the playbook](configuring-playbook.md).
-
 ## DNS settings for optional services/features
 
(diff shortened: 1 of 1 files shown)


docs/configuring-dns.md
```diff
@@ -24,8 +24,6 @@ Be mindful as to how long it will take for the DNS records to propagate.
 
 If you are using Cloudflare DNS, make sure to disable the proxy and set all records to `DNS only`. Otherwise, fetching certificates will fail.
 
-When you're done configuring DNS, proceed to [Configuring the playbook](configuring-playbook.md).
```
spantaleev, line 27: You're right - this is incorrect.

The bottom of this docs page does link to the correct place (`getting-the-playbook.md`), but there's so much content between "here" and "there", that it's probably easy to miss and not know how to proceed.

I'll merge this anyway, but we should consider how to improve the situation. We may relocate that (or duplicate that) section here. Or we should use some emojis or something to "always make the next step easy to see and jump to".

luixxiul: I've got an idea: creating a breadcrumb on the header above the heading. I've tried it locally and it seems to be helpful.

PR #3745, opened 2024-11-05
Add hard-coded breadcrumbs to documentation files about installation procedure for easier navigation

luixxiul:
Please see: https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/3744#discussion_r1829003610 and the commit comment.

### DIFF
--- a/docs/configuring-dns.md
+++ b/docs/configuring-dns.md
@@ -1,5 +1,7 @@
 # Configuring your DNS server
 
+<sup>[Prerequisites](prerequisites.md) > Configuring your DNS server > [Getting the playbook](getting-the-playbook.md) > [Configuring the playbook](configuring-playbook.md) > [Installing](installing.md) </sup>
+
 To set up Matrix on your domain, you'd need to do some DNS configuration.
 
 To use an identifier like `@<username>:example.com`, you don't actually need to install anything on the actual `example.com` server.
--- a/docs/configuring-dns.md
+++ b/docs/configuring-dns.md
@@ -1,6 +1,6 @@
 # Configuring your DNS server
 
-<sup>[Prerequisites](prerequisites.md) > Configuring your DNS server > [Getting the playbook](getting-the-playbook.md) > [Configuring the playbook](configuring-playbook.md) > [Installing](installing.md) </sup>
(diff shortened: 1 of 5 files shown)


docs/configuring-dns.md
```diff
@@ -1,5 +1,7 @@
 # Configuring your DNS server
 
+<sup>[Prerequisites](prerequisites.md) > Configuring your DNS server > [Getting the playbook](getting-the-playbook.md) > [Configuring the playbook](configuring-playbook.md) > [Installing](installing.md) </sup>
```
spantaleev, line 3: What do you think about adding more context here? Here are some ideas:

- a prefix to this line, like **(Quick Start)**, possibly linking to some dedicated "Quick Start" docs page which explains that the docs are extensive and you can follow "these" steps for a quick start guide. The main README (via the "Getting started" sectiong and `docs/README.md` can link to the "Quick Start" page somehow
- perhaps an emoji before the "Quick Start" prefix, which makes the whole line easier to spot on each docs page. Not sure which emoji would be best. Maybe ⚡️?

spantaleev: We may even add an emoji to each step, unless it makes it too busy.

When an emoji is added to each step, perhaps we don't need the ⚡️ prefix emoji, though I have not experimented to see what looks good.

------

Here are some emoji ideas for the different steps, as proposed by gpt-4o:

**Prerequisites:**
1. 📋 (maybe this one is best?)
2. 📚
3. ✅

**Configuring your DNS server:**
1. 🌐 (maybe this one is best?)
2. 🔧
3. 🛠️

**Getting the playbook:**
1. 📥 (maybe this one is best?)
2. 📄
3. 💾

**Configuring the playbook:**
1. 📝 (maybe this one is best?)
2. ⚙️
3. ✏️

**Installing:**
1. 🚀 (maybe this one is best?)
2. 🛠️
3. ⚡

luixxiul: I got a kind of feeling that using them for the breadcrumbs would make the UI too funky :nerd_face: Still, I like the thunder symbol with "Quick Start" being put before the "Prerequisites".

spantaleev:
Looks good!

PR #3748, opened 2024-11-05
Update docs/registering-users.md: add configuration for enabling public user registration on Dendrite

luixxiul:


docs/registering-users.md
```diff
@@ -100,10 +100,18 @@ If you'd rather **keep your server private** (public registration closed, as is
...
+For Synapse:
+
 ```yaml
 matrix_synapse_enable_registration: true
 ```
 
+For Dendrite:
+
+```yaml
+matrix_dendrite_client_api_registration_disabled: true
```
spantaleev, line 112: Shouldn't this say `false`?

PR #3751, opened 2024-11-05
Improve readability -  configuring-playbook-own-webserver.md

ikkemaniac:


docs/configuring-playbook-own-webserver.md
```diff
@@ -186,12 +186,14 @@ matrix_playbook_public_matrix_federation_api_traefik_entrypoint_config_custom:
 ```
 
 Such a configuration would expose all services on a local port `81` and Matrix Federation on a local port `8449`.
+Your reverse-proxy configuration needs to send traffic to these ports.
+[`examples/reverse-proxies`](../examples/reverse-proxies/) contains examples for various webservers such as Apache2, Caddy, HAproxy, nginx and Nginx Proxy Manager.
```
luixxiul, line 190: Despite being lack of the style guide, I'm not convinced if splitting them with with single line breaks would be good idea, because it introduces inconsistency between the source text and the rendered result, as you would see those sentences rendered as one single paragraph.

ikkemaniac: that's intentional. this should be 1 paragraph with L188 as "these ports" refer to the ports mentioned in 188 and the examples in 190 refer to the configuration mentioned in 189. Therefore all 3 lines should be 1 paragraph.

luixxiul, line 190:
```suggestion
Such a configuration would expose all services on a local port `81` and Matrix Federation on a local port `8449`. Your reverse-proxy configuration needs to send traffic to these ports. [`examples/reverse-proxies`](../examples/reverse-proxies/) contains examples for various webservers such as Apache2, Caddy, HAproxy, nginx and Nginx Proxy Manager.
```

PR #3761, opened 2024-11-08
Update docs/installing.md

luixxiul:
The main intention of this PR is:

- Reduce the amount of information on `docs/installing.md` by moving the nicely detailed section for playbook tags into a dedicated page
- Add a hint of finalizing the installation by explaining the simplest way of configuring Service Discovery, which is to use the variable `matrix_static_files_container_labels_base_domain_enabled` on the server dedicated for the Matrix installation.

  Currently this information cannot be found until you dig the article `docs/configuring-well-known.md` and then `docs/configuring-playbook-base-domain-serving.md`.

### DIFF
--- a/docs/installing.md
+++ b/docs/installing.md
@@ -86,16 +86,6 @@ ansible-playbook -i inventory/hosts setup.yml --tags=start
 
 Proceed to [Maintaining your setup in the future](#2-maintaining-your-setup-in-the-future) and [Finalize the installation](#3-finalize-the-installation)
 
-
-## 2. Maintaining your setup in the future
-
-Feel free to **re-run the setup command any time** you think something is off with the server configuration. Ansible will take your configuration and update your server to match.
-
-Note that if you remove components from `vars.yml`, or if we switch some component from being installed by default to not being installed by default anymore, you'd need to run the setup command with `--tags=setup-all` instead of `--tags=install-all`. See [Playbook tags introduction](#playbook-tags-introduction)
-
-A way to invoke these `ansible-playbook` commands with less typing in the future is to use [just](https://github.com/casey/just) to run them: `just install-all` or `just setup-all`. See [our `justfile`](../justfile) for more information.
-
-
 ## 3. Finalize the installation
 
 Now that services are running, you need to **finalize the installation process** (required for federation to work!) by [Configuring Service Discovery via .well-known](configuring-well-known.md).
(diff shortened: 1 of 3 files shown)


docs/installing.md
```diff
@@ -84,26 +64,25 @@ You can now:
...
+Now that services are running, you need to **finalize the installation process** (required for federation to work!) by [Configuring Service Discovery via .well-known](configuring-well-known.md).
 
-A way to invoke these `ansible-playbook` commands with less typing in the future is to use [just](https://github.com/casey/just) to run them: `just install-all` or `just setup-all`. See [our `justfile`](../justfile) for more information.
+If you need the base domain for anything else (such as hosting a website), you have to configure it manually, following the procedure described on the linked documentation.
 
+However, if you do not need the base domain (`example.com`) for anything else, the easiest way of configuring it is to [serve the base domain](configuring-playbook-base-domain-serving.md) from the integrated web server. It will enable you to use a Matrix user identifier like `@<username>:example.com` while hosting services on a subdomain like `matrix.example.com`.
```
luixxiul, line 73: I think this should be correct according to [this line](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/7a77d84276ca3f608487757738e1a4821a3eb078/docs/configuring-well-known.md?plain=1#L16), but I am not perfectly sure.

docs/installing.md
```diff
@@ -84,26 +64,29 @@ You can now:
 ansible-playbook -i inventory/hosts setup.yml --tags=start
```
spantaleev, line 64: I wonder if we should add `ensure-matrix-users-created` here to be consistent with what we propose above.

In most cases, `ensure-matrix-users-created` may not be necessary, because people would be importing an existing database which should contain all bridge/bot users, as expected.

However, we don't really know what kind of database dump users are importing and whether or not they have decided to enable some extra services during their initial installation. Perhaps the user enabled some extra bot.

In that case, it would be safer to add `ensure-matrix-users-created` here.

```suggestion
ansible-playbook -i inventory/hosts setup.yml --tags=ensure-matrix-users-created,start
```

docs/installing.md
```diff
@@ -84,26 +64,29 @@ You can now:
...
-## 3. Finalize the installation
+```yaml
+matrix_static_files_container_labels_base_domain_enabled: true
+```
 
-Now that services are running, you need to **finalize the installation process** (required for federation to work!) by [Configuring Service Discovery via .well-known](configuring-well-known.md).
+After configuring the playbook, run the installation command:
 
+```sh
+ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start
```
spantaleev, line 84: I wonder if this one should suggest `install-all`, instead of `setup-all`. After all, we don't expect that users uninstalled stuff, so the quicker `install-all` tag would be better and more consistent with what we propose above.

```suggestion
ansible-playbook -i inventory/hosts setup.yml --tags=install-all,start
```

spantaleev (commented):
Thanks! This is much tidier!

PR #3789, opened 2024-11-17
Update docs/installing.md: add "Create your user account" section before "Finalize the installation"

luixxiul:


docs/installing.md
```diff
@@ -66,9 +66,31 @@ You can now:
...
+
+# Example: `ansible-playbook -i inventory/hosts setup.yml --extra-vars='username=john password=secret-password admin=yes' --tags=register-user`
+```
+
+For more information, see the documentation for [registering users](registering-users.md).
+
 ## Finalize the installation
 
-Now that services are running, you need to **finalize the installation process** by [delegating (redirecting) the server](howto-server-delegation.md) to your base domain (`example.com`).
+Now you've configured Matrix services and your user account, you need to **finalize the installation process** by [delegating (redirecting) the server](howto-server-delegation.md) to your base domain (`example.com`).
```
spantaleev, line 93:
```suggestion
Now you've configured Matrix services and your user account, you need to **finalize the installation process** by  [setting up Matrix delegation (redirection)](howto-server-delegation.md), so that your Matrix server (`matrix.example.com`) can present itself as the base domain (`example.com`) in the Matrix network.
```

Your original sentence ("delegating the server to the base domain") is kind of misleading, I think.

I've also tried to avoid using words that may indicate that "the base domain is redirected to the Matrix server", because people may think they need to redirect all of `example.com` to `matrix.example.com` or something, whereas in reality they only need to set up a few file redirects (or a DNS SRV record).

luixxiul: The word `delegation` is really a technical term, and I indeed did not quite get it right when I was translating the word. I would say the term deserves an FAQ entry ("People are talking about 'delegation', but what is it?")

PR #3791, opened 2024-11-17
Update docs/registering-users.md: add details about delegation

luixxiul:


spantaleev:
This **Things to do after registering users** section is in `registering-users.md`, because `installing.md` was leading people there.

Now that we have a "registering users" section on `installing.md` and people are less likely to jump to `registering-users.md`, this whole section is likely unnecessary.

Still, maybe it's better to keep it around.

luixxiul:
I'd agree. It seems it does not make sense to keep it there.

Issue #3746, opened 2024-11-05
Create "Quick start" page


luixxiul:
**Is your feature request related to a problem? Please describe.**
This is a follow-up ticket to https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/3745

The PR implemented breadcrumbs to pages which describe install procedure, and it also added a link to "Quick start" page on each of them. Currently it is linked to `docs/README.md`, which should be replaced with an actual quick start page like `docs/quickstart.md` which simply explains what to be done to complete installation.

**Describe the solution you'd like**
See above.

**Describe alternatives you've considered**

**Additional context**

PR #3801, opened 2024-11-21
Add "Quick start" guide

luixxiul:
Since I am not quite sure whether the links to prerequisites.md should be replaced in favor of the links to quick-start.md, this PR leaves them as they are for now.

Closes https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3746


docs/quick-start.md
```diff
@@ -0,0 +1,180 @@
...
+ansible-playbook -i inventory/hosts setup.yml --extra-vars='username=YOUR_USERNAME_HERE password=YOUR_PASSWORD_HERE admin=<yes|no>' --tags=register-user
+
+# Example: ansible-playbook -i inventory/hosts setup.yml --extra-vars='username=john password=secret-password admin=yes' --tags=register-user
+```
+
+## Finalize server installation
+
+<sup>This section is optimized for this quick-start guide and is derived from the following full-documentation page: [Server Delegation](howto-server-delegation.md)</sup>
+
+Now you've configured Matrix services and your user account, you need to **finalize the installation process** by [setting up Matrix delegation (redirection)](howto-server-delegation.md), so that your Matrix server (`matrix.example.com`) can present itself as the base domain (`example.com`) in the Matrix network.
```
spantaleev, line 126:
```suggestion
Now that you've configured Matrix services and your user account, you need to **finalize the installation process** by [setting up Matrix delegation (redirection)](howto-server-delegation.md), so that your Matrix server (`matrix.example.com`) can present itself as the base domain (`example.com`) in the Matrix network.
```

docs/quick-start.md
```diff
@@ -0,0 +1,180 @@
...
+
+**This is required for federation to work!** Without a proper configuration, your server will effectively not be part of the Matrix network.
+
+There are several ways to configure the delegation, but the easiest one is to [serve the base domain](configuring-playbook-base-domain-serving.md) from the integrated web server.
```
spantaleev, line 130: It feels like this section focuses on the "base domain serving" option, which is nice and quick, but..

People who cannot afford to point the base domain to the playbook will need other options.

Above, we already link to `howto-server-delegation.md` twice, but maybe we should also be more explicit as to what your options are if you cannot serve the base domain on the Matrix server.

Perhaps a sentence like this can be added somewhere:

> If you're using the base domain for other purposes and cannot point it to the Matrix server (and thus cannot "serve the base domain" from it), you most likely need to [Manually install well-known files on the base domain's server](configuring-well-known.md#manually-installing-well-known-files-on-the-base-domains-server), but feel free to familiarize yourself with all [Server delegation (redirection)](howto-server-delegation.md) options.

-------

More restructuring could be done as well. For example (just a quick draft below):

> You have 2 options:
>
> - If you can afford to point the base domain at the Matrix server, follow the instructions below which guide you into [serving the base domain](configuring-playbook-base-domain-serving.md) from the Matrix server
> - Alternatively, if you're using the base domain for other purposes and cannot point it to the Matrix server (and thus cannot "serve the base domain" from it), you most likely need to [Manually install well-known files on the base domain's server](configuring-well-known.md#manually-installing-well-known-files-on-the-base-domains-server), but feel free to familiarize yourself with all [Server delegation (redirection)](howto-server-delegation.md) options.

docs/quick-start.md
```diff
@@ -0,0 +1,180 @@
...
+
+After the command finishes, you can also check whether your server federates with the Matrix network, by using the [Federation Tester](https://federationtester.matrix.org/).
+
+If you think something is off with the server configuration, feel free to re-run the setup command any time.
+
+## Log in to your user account
+
+Finally, let's make sure that you can log in to the created account with the specified password.
+
+You should be able to log in to it with your own Element Web client which you have set up at `element.example.com` by running the playbook. Open the URL (`https://element.example.com`) on a web browser, and enter the credentials to log in.
```
spantaleev, line 152:
```suggestion
You should be able to log in to it with your own [Element Web](configuring-playbook-client-element-web.md) client which you have set up at `element.example.com` by running the playbook. Open the URL (`https://element.example.com`) in a web browser and enter your credentials to log in.
```

The main suggestion here is `on` -> `in`, but I felt like adding a link for additional reading is always helpful.

docs/quick-start.md
```diff
@@ -0,0 +1,180 @@
...
+
+**If you successfully logged in to your account, installing and configuring procedure is complete**🎉
```
spantaleev, line 154:
```suggestion
**If you successfully logged in to your account, installing and configuring is complete**🎉
```

luixxiul:
Going to add the breadcrumb header…

luixxiul:
It should be fine.

docs/quick-start.md
```diff
@@ -0,0 +1,197 @@
...
+<sup>This section is optimized for this quick-start guide and is derived from the following full-documentation page: [Registering users](registering-users.md)</sup>
+
+As you have configured your brand new server and the client, you need to **create your user account** on your Matrix server.
+
+To register a user via this Ansible playbook, run the command below on your local computer.
+
+**Notes**:
+- Make sure to adjust `YOUR_USERNAME_HERE` and `YOUR_PASSWORD_HERE`
+- For `YOUR_USERNAME_HERE`, use a plain username like `john`, not your full identifier (`@user:example.com`)
+- Use `admin=yes` or `admin=no` depending on whether you wish to make the user an administrator of the Matrix server
```
luixxiul, line 126: I have two questions here:

- Is it clear which effect making a user an admin of the Matrix server has? (I guess you should never make a user admin casually)
- Should you make yourself (= the first user of the Matrix server) an administrator? Is it rather optional?

spantaleev: This section is named "Create your user account" which may not be great, because you may wish to create multiple accounts here. While it's possible, it doesn't seem obvious.

Yes, you will most likely wish to create your first user account as an administrator. The person who sets up the Matrix server will definitely wish to create their account as an administrator.

You're right that we don't explain what the effect is and that one shouldn't make users administrators willy-nilly.

luixxiul: > This section is named "Create your user account" which may not be great, because you may wish to create multiple accounts here. While it's possible, it doesn't seem obvious.

I would add: `For an instruction to add multiple user accounts, see [Registering users](docs/registering-users.md)`, because creating other accounts than your user account is out of the scope of this document, whose objects are listed at the top.

Beryesa:
Can't it use `just install-all` on the update section instead? Simpler.

luixxiul:
`quick-start.md` intends to be read for those who have never used the playbook to set up a server. so it is safer to think that it is not clear to them that what the `just` "recipes" are made of, ie. it takes some time and experience until someone understands simplicity of them.

For beginners, I believe that we should prefer the basics over simplicity, so that they will get familiar with those commands, on which the `just` "recipes" are based, from the educational point of view.

If someone feels tired of using the same command repetitively, then the person will have been already well accustomed to the way how the playbook works and how the server is supposed to be maintained, and the person can use the just "recipes" with confidence, so that he/she can explain the difference between the playbook tags and the `just` "recipes", for example, `just install-all` and `ansible-playbook -i inventory/hosts setup.yml --tags=install-all`. Such level of familiarity and experience should not be expected on the quick start guide.

Basically, right now we lack the dedicated documentation about the `just` "recipes". It should be created for intermediate level users (based on [this entry](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/f1712cec736e253eae9cbbdc50c94db2aeab6fcd/CHANGELOG.md?plain=1#L1282-L1300) on CHANGELOG.md), yet it should be avoided that those who have never used the playbook already are familiar with the `just` program itself, since the program is not as well known as Ansible itself. The documentation can be created with this PR or be deferred to another.

docs/quick-start.md
```diff
@@ -0,0 +1,200 @@
...
+matrix_static_files_container_labels_base_domain_enabled: true
+```
+
+After configuring the playbook, run the installation command and wait until it finishes:
+
+```sh
+ansible-playbook -i inventory/hosts setup.yml --tags=install-all,start
+```
+
+After the command finishes, you can also check whether your server federates with the Matrix network, by using the [Federation Tester](https://federationtester.matrix.org/).
```
spantaleev, line 174:
```suggestion
After the command finishes, you can also check whether your server federates with the Matrix network, by using the [Federation Tester](https://federationtester.matrix.org/) and running a check using your base domain (`example.com`), not the `matrix.example.com` subdomain.
```

We often have people doing this wrong - using `matrix.example.com` in the Federation Tester. It's probably good to explicitly clarify.

docs/quick-start.md
```diff
@@ -50,7 +50,7 @@ As the table illustrates, you need to create 2 subdomains (`matrix.example.com`
 
 It might take some time for the DNS records to propagate after creation.
 
-**Note**: if you are using Cloudflare DNS, make sure to disable the proxy and set all records to "DNS only"
+**💡 Note**: if you are using Cloudflare DNS, make sure to disable the proxy and set all records to "DNS only"
```
luixxiul, line 53: Off-topic (as a memo): shall we create a emoji guide for consistency on a style guide or something?

- Warning emoji with `**Warning**:`
- Light bulb emoji with `**Note(s)**:`

We might need some time for consideration.

PR #3805, opened 2024-11-21
Edit README.md: move up the section "Getting started" above "Self-hosting or Managed / SaaS"

luixxiul:


spantaleev:
The **Self-hosting or Managed / SaaS** section was intentionally moved higher up in the README recently, so that people know what the alternatives are before **Getting Started** (with self-hosting themselves).

luixxiul:
I see, that's reasonable.

PR #3806, opened 2024-11-22
Add .github/workflows/close-stale-issues.yml: close stale issues automatically

luixxiul:
The workflow itself should be fine. Perhaps `debug-only: true` should be enabled for testing, though I am not really sure how it works.

### DIFF
--- a/.github/workflows/close-stale-issues.yml
+++ b/.github/workflows/close-stale-issues.yml
@@ -1,0 +1,27 @@
+---
+name: 'Close stale issues'
+on:  # yamllint disable-line rule:truthy
+  schedule:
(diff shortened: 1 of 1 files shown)


spantaleev:
Thank you for figuring this out! Let's give it a try!

luixxiul:
This is the log of the first run: https://github.com/spantaleev/matrix-docker-ansible-deploy/actions/runs/11983371962/job/33412765868

No issues were labelled with `stale` because apparently adding labels to all open issues have reset the stale counter for all of them. We should see the label `stale` is started to be added in the next month or in January 2025.

Issue #3741, opened 2024-11-04
Describe difference between just "recipe" and playbook tags on docs/installing.md


luixxiul:
**Is your feature request related to a problem? Please describe.**
Context: https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/3739

The documentation file `docs/installing.md` should describe difference between just "recipe" and the playbook tags, as for most of people who have never heard of `just`, it is almost impossible to see the difference between `just install-all` and `just run-tags install-all` without knowing what `just` is, let alone the documentation itself explains what to be done by using `just` commands [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/bcc6c4022d371f1d9134ab96c148ead2d7da4144/docs/installing.md?plain=1#L5-L7), which compounds the confusion.

**Describe the solution you'd like**
See above.

**Describe alternatives you've considered**

**Additional context**
Of course it should be clear how `just` works for those who know what `just` is, but it does mean that most people know what `just` is. As documentation to explain installing, it should be one of the most important articles and therefore the confusing description should be avoided as much as possible.

PR #3811, opened 2024-11-23
Add docs/just.md

luixxiul:
Closes https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3741

### DIFF
--- a/docs/just.md
+++ b/docs/just.md
@@ -10,15 +10,17 @@ For some recipes such as `just update`, our `justfile` recommends to install [ag
 
 Here are some examples of shortcuts:
 
-- `just roles` - trigger `make roles` to install the necessary Ansible roles pinned in [`requirements.yml`](../requirements.yml)
-- `just update` - run `git pull` to update the playbook and `just roles`
-- `just install-all` - run `ansible-playbook -i inventory/hosts setup.yml --tags=install-all,ensure-matrix-users-created,start`
-    - `just install-all --ask-vault-pass` - commands also support additional arguments (`--ask-vault-pass` will be appended to the above installation command)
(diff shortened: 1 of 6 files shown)


luixxiul:
Since this documentation perhaps will be linked from quick-start.md (see: https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/3801), I would like to see this merged earlier.

docs/just.md
```diff
@@ -0,0 +1,37 @@
...
+
+The `just` utility executes shortcut commands (called as "recipes"), which ultimately run `ansible-galaxy`. The targets of the recipes are defined in [`justfile`](../justfile). It defines commands, most of which are not available on our `Makefile`.
+
+For some recipes such as `just update`, our `justfile` recommends to install [agru](https://github.com/etkecc/agru#where-to-get) to speed up the process.
+
+Here are some examples of shortcuts:
+
+| Shortcut                                      | Effect                                                                                                         |
+|-----------------------------------------------|----------------------------------------------------------------------------------------------------------------|
+| `just roles`                                  | Run `make roles` to install the necessary Ansible roles pinned in [`requirements.yml`](../requirements.yml)    |
```
spantaleev, line 15: This makes it sound like `just roles` would invoke `make roles`, which does some things.

None of our `justfile`-defined recipes invoke `make`.

`just roles` may also use `agru` to install roles (if available) and will fall back to `ansible-galaxy` if not.

While `make roles` exists, it's more simplistic - it only uses `ansible-galaxy`. While it can probably be extended, the `Makefile` is not really maintained (by most of us). The goal is to only use `just`, but.. given that `just` is not very ubiquitous, we cannot pull this off yet and have a minimal old `Makefile` still around.

luixxiul: Right, it is my misunderstanding. Fixing…

luixxiul:
Thanks! Looks good to me :smiley:

PR #3816, opened 2024-11-24
Edit docs/maintenance-migrating.md

luixxiul:
This PR intends to improve migration guide. While the current instructions do work (I've followed the instruction to migrate successfully), I think the numbering list can be improved by replacing it with sections.

### DIFF
--- a/docs/maintenance-migrating.md
+++ b/docs/maintenance-migrating.md
@@ -44,7 +44,7 @@ Having adjusted DNS records, replace the old server's external IP address on the
 
 ## Create `matrix` user and group on the new server
 
-After updating `inventory/hosts` file, run the command below to create the `matrix` user and group on the new server:
+After updating `inventory/hosts` file, run the command below on your local computer to create the `matrix` user and group on the new server:
 
 ```sh
 ansible-playbook -i inventory/hosts setup.yml --tags=setup-system-user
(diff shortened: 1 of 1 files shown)


docs/maintenance-migrating.md
```diff
@@ -1,14 +1,71 @@
...
+
+```sh
+ansible-playbook -i inventory/hosts setup.yml --tags=stop
+```
+
+Alternatively, you can log in to the old server and run the command (you might have to `cd` to `/etc/systemd/system/` first):
+
+```sh
+systemctl disable --now matrix*
+```
```
spantaleev, line 29: Only the `systemctl` commands makes sure services won't be starting again.

Running `--tags=stop` merely stops them once. If you reboot the server, they will start again.

Perhaps we should make a note of that.. and we'd probably better recommend running the `systemctl` command on the server as the better option. Maybe it should even be mentioned first.

---------

`cd`-ing into `/etc/systemd/system/` is also highly recommended, as the wildcard (`matrix*`) will likely not work properly otherwise.

Perhaps we should make the command be `cd /etc/systemd/system/ && systemctl disable --now matrix*` to make it easier to miss.

docs/maintenance-migrating.md
```diff
@@ -1,14 +1,71 @@
...
+ansible-playbook -i inventory/hosts setup.yml --tags=setup-system-user
+```
+
+**Note**: because the `matrix` user and group are created dynamically on each server, the user/group ID may differ between the old and new server. We suggest that you adjust ownership of `/matrix` files. To adjust the ownership, log in to the new server and run the command:
+
+```sh
+chown -R matrix:matrix /matrix
+```
+
+## Start all services on the new server
```
spantaleev, line 59:
```suggestion
## Install and start all services on the new server
```

PR #3820, opened 2024-11-25
Update docs/faq.md and docs/prerequisites.md: add a note about installing both ansible-core and ansible

luixxiul:


spantaleev:
I have tried to research this topic a bit.

---------

For example, on Archlinux, most people would install [ansible](https://archlinux.org/packages/extra/any/ansible/), which is the "Official assortment of Ansible collections". This package depends on `ansible-core`, which is "the Ansible program itself".

Interestingly, the [Ansible page on ArchWiki](https://wiki.archlinux.org/title/Ansible) recommends `ansible-core` and then says:

> Additionally, you can install the [ansible](https://archlinux.org/packages/?name=ansible) package, which provides a range of community curated collections.

.. so it's possible that some people only install `ansible-core` and skip this "additional step".

On the other hand, it makes more sense to just install `ansible` and most people will even blindly do `pacman -Syu ansible` without even checking this ArchWiki page. Doing that will do the right thing (pulling both `ansible` and its `ansible-core` dependency).

----------------

On Ubuntu (24.04) and Debian (12) the situation is similar - `ansible` depends on `ansible-core`. You installing `ansible` will do the right thing.

-----------------

On Rocky Linux, there's only `ansible-core` in the base repositories.

If you install `epel-release` (Extra Packages for Enterprise Linux (EPEL) repository), then this extra repository provides:

- `ansible`, which contains various collections (and depends on `ansible-core`)
- various `ansible-collection-*` packages (e.g. `ansible-collection-community-docker` providing the `community.docker` Ansible collection), etc.

Interestingly, the `ansible` package seems to provide the `community.docker` plugin as well (verified by running `rpm -ql ansible | grep community/docker`)

So, installing `ansible` will do the right thing (pulling `ansible-core` as a dependency), but it requires that you have enabled EPEL (by installing `epel-release`) first.

--------------

I have not tested other popular distros here, but the ones I've tested probably cover most people's use-cases.

In all these distros, simply installing `ansible` will pretty much do the right thing (pulling `ansible-core` as a dependency).

It's curious to see what the situation is on Mac / Windows when it comes to this.

luixxiul:
Thanks for digging! Since it looks we need more investigation about this, especially about the situation on Windows and macOS, I'd like to close this in favor of an issue :wink:

PR #3827, opened 2024-11-26
Replace ansible-community/ansible-lint-action with ansible/ansible-lint

luixxiul:


spantaleev:
I'm not sure if this can be gotten to work that easily.

I've adjusted it to define a `requirements_file` (`requirements.yml`), so that roles could be installed. Roles are installed now, as [seen here](https://github.com/spantaleev/matrix-docker-ansible-deploy/actions/runs/12034620118/job/33551601947?pr=3827#step:3:191), but they are installed to `/home/runner/.ansible/roles`, not under `roles/galaxy` as the playbook expects.

The playbook's `justifle` and `Makefile` install into a custom path (`roles/galaxy`) by passing it as an additional argument to the `ansible-galaxy` tool. `agru` seems to default to using `roles/galaxy`.

We may look for ways to make this Github Action install roles into `roles/galaxy`. I've tried defining an `ANSIBLE_ROLES_PATH` environment variable, as discussed [here](https://docs.ansible.com/ansible/latest/galaxy/user_guide.html#setting-where-to-install-roles), and it seems like it's taken into account, but roles are now installed into `roles`, not `roles/galaxy`. Perhaps additional tweaks are necessary (maybe switching the order of the 2 entries).

----------

Another issue is the lack of "Ansible collection" modules, like `community.docker.docker_image` as [seen here](https://github.com/spantaleev/matrix-docker-ansible-deploy/actions/runs/12034763113/job/33552099724?pr=3827#step:3:281).

luixxiul:
Thanks to https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/f8142a0c377d929798d18770db5a34933d8cc09a, ansible-lint has been implemented :+1:

PR #3837, opened 2024-11-28
Update documentation of setting up double puppeting with bridges

luixxiul:


luixxiul:
Hm, since the double puppeting with Appservice Double Puppet is enabled automatically, it seems better to mention it as optional prerequisite.

PR #3842, opened 2024-11-30
Edit descriptions about installation of components

luixxiul:
The main aims here are:

- Add both `ansible-playbook` commands and shortcuts (mostly `just install-all` and `just setup-all`) for quick installation
    - Add `just run-tags install-all,start` or `just run-tags setup-all,start` instead, if the raw ansible-playbook commands do not require `ensure-matrix-users-created`
- Add/remove `ensure-matrix-users-created` according to components
- Update capitalization and punctuation of list items, according to https://www.purchase.edu/editorial-style-guide/general-style-preferences/punctuation/bulleted-and-numbered-lists/

Though these changes do seem to be really repetitive, it will not matter to readers as they read documentation selectively, when they want/need to.

--

I chose the raw ansible-playbook commands as main ones since they are basic and conservative, but the `just` shortcut commands could be displayed as main and the raw ansible-playbook commands could be displayed as hint in notes conversely, if this is better.


luixxiul:
Thanks for the review!

PR #3843, opened 2024-12-01
Update docs/configuring-playbook-client-schildichat-web.md: remove warning notification about lagging versions

luixxiul:


luixxiul:
Since SchildiChat's development is getting as active as before, it seems to be fine to remove the notification. Let's revert the change if it often gets lagging again.

PR #3862, opened 2024-12-09
Use common expressions for user IDs, bridge usage, etc.

luixxiul:
This PR intends to adopt common expressions for user IDs, bridge usage, etc., to make it easier to manage documentation in bulk, fixing other stuff like incorrect examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 This [Ansible](https://www.ansible.com/) playbook is meant to help you run your own [Matrix](http://matrix.org/) homeserver, along with the [various services](#supported-services) related to that.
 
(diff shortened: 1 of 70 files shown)


README.md
```diff
@@ -6,7 +6,7 @@
 
 This [Ansible](https://www.ansible.com/) playbook is meant to help you run your own [Matrix](http://matrix.org/) homeserver, along with the [various services](#supported-services) related to that.
 
-That is, it lets you join the Matrix network using your own `@<username>:example.com` identifier, all hosted on your own server (see [prerequisites](docs/prerequisites.md)).
+That is, it lets you join the Matrix network using your own user ID like `@alice:example.com`, all hosted on your own server (see [prerequisites](docs/prerequisites.md)).
```
luixxiul, line 9: Consistency is key, so it does not matter whether the placeholder is `@alice`, `@user`, or `@<username>`.

PR #3911, opened 2025-01-04
Update docs: misc edits for consistency

luixxiul:


docs/configuring-playbook-etherpad.md
```diff
@@ -4,6 +4,8 @@
 
 When enabled together with the Jitsi audio/video conferencing system (see [our docs on Jitsi](configuring-playbook-jitsi.md)), it will be made available as an option during the conferences.
 
+**Note**: the component is not managed by this repository but its [own repository](https://github.com/mother-of-all-self-hosting/ansible-role-etherpad).
```
spantaleev, line 7: This is true for many components listed in https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/requirements.yml

Most of these used to live in the playbook, but were extracted out, so they can be used in other playbooks more easily. It's just Matrix-specific roles that continue to live in this playbook.

1. Are you thinking of adding a similar note to each component that's included from outside?

2. Do you think most people care about this internal technical detail or have anything useful to do with this additional information? Perhaps this tells them that it's better to open issues (and feature requests, etc.) in the `ansible-role-etherpad` repository instead of this one? Of course, some issues may be due to our integration of the role with the rest of the playbook components and not with the role itself, so the issue may as well belong to this repository here.

luixxiul: My intuition (based on my experience) is that some people would look for files in this repository to configure variables with `_extension_` with no luck, only to find out later that they are managed in externally. Probably a note on `configuring-playbook.md` would be better than adding a similar one on each document.

PR #3929, opened 2025-01-11
Update docs/configuring-playbook-synapse-s3-storage-provider.md: fix the path to media-store directory

luixxiul:


spantaleev:
Good catch!

PR #3936, opened 2025-01-14
Update docs for DNS settings etc.

luixxiul:


spantaleev:
Much better to have the Configuring DNS page be simpler! :bow:

luixxiul:
Now that the large table is gone, it also should be sensible to move the section for the DNS configuration for default services above the one for the SRV record!

spantaleev:
That'd be a nice improvement. Almost no one uses DNS SRV records for Matrix Server Delegation, so making it less prominent is a good idea. Perhaps this section may only remain on other docs pages, not on "Configuring DNS", but for completeness we can still keep it on the "Configuring DNS" page as well (just.. at the bottom).

PR #3946, opened 2025-01-16
Remove an obsolete variable: backup_borg_retention_prefix

luixxiul:
Please let me know if the variable is still useful somehow. Thanks in advance!


spantaleev:
I don't see it being used anywhere, so.. it's good for removal!

PR #3954, opened 2025-01-17
Update docs/configuring-playbook-etherpad.md

luixxiul:


docs/configuring-playbook-etherpad.md
```diff
@@ -43,6 +43,30 @@ After changing the domain, **you may need to adjust your DNS** records to point
 
 If you've decided to reuse the `matrix.` domain, you won't need to do any extra DNS configuration.
 
+### Configure the default text (optional)
+
+You can also edit the default text on a new pad with the variable `etherpad_default_pad_text`.
+
+To do so, add the following configuration to your `inventory/host_vars/matrix.example.com/vars.yml` file (adapt to your needs):
```
spantaleev, line 50: Didn't you recently adjust all such instructions to only say `vars.yml` instead of a full path like this?

luixxiul: Yes, nice catch! :smile:

PR #3984, opened 2025-01-21
Change matrix_synapse_forgotten_room_retention_period  from null to 28d

FSG-Cat:
As we automatically forget rooms on leave in the playbook this option working at all is probably heavily desired.

Timing is copied from upstream example.

Credit goes to Anoa for making me even check this.

https://element-hq.github.io/synapse/latest/usage/configuration/config_documentation.html#forgotten_room_retention_period for the documentation on this matter.


spantaleev:
Good catch! Thanks!

PR #4039, opened 2025-02-06
Correct the name of matrix-bot-baibot's configuration file

hardye:
Commit 30dad8ba277bc renamed the config file for baibot from `config.yml` to `config.yaml`. While this increases consistency, it actually broke my baibot setup: the service `matrix-bot-baitbot.service` failed to start because the baibot container apparently expects the file's name to be `config.yml` (without the 'a').

This PR reverts part of the changes introduced in 30dad8ba277bc to ensure the container gets the file name it expects.


spantaleev:
Thanks for reporting this and working on a fix! :bow:

PR #4046, opened 2025-02-08
Update docs/configuring-playbook-prometheus-grafana.md

luixxiul:
This PR intends to improve docs/configuring-playbook-prometheus-grafana.md by mainly splitting configuration blocks into two (for Prometheus and Grafana). In the end it will be synchronized with docs/configuring-playbook-prometheus-nginxlog.md to improve maintainability of those documents which are referred rather infrequently.

Since this PR includes copied text from `ansible-role-grafana`, I'd create a PR there to address it for fairness and my conscience.
→ Blocks on https://github.com/mother-of-all-self-hosting/ansible-role-grafana/pull/3

### DIFF
--- a/docs/configuring-playbook-prometheus-grafana.md
+++ b/docs/configuring-playbook-prometheus-grafana.md
@@ -8,6 +8,8 @@ By default, this playbook installs Grafana web user-interface on the `stats.` su
 
 When setting, replace `example.com` with your own.
 
+**Note**: It is possible to install Prometheus without installing Grafana. This case it is not required to create the CNAME record.
+
 ## Adjusting the playbook configuration
 
 ### Configure Prometheus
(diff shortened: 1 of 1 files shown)


docs/configuring-playbook-prometheus-grafana.md
```diff
@@ -8,37 +8,69 @@ By default, this playbook installs Grafana web user-interface on the `stats.` su
 
 When setting, replace `example.com` with your own.
 
+**Note**: It is possible to install Prometheus without installing Grafana. This case it is not required to create the CNAME record.
```
spantaleev, line 11:
```suggestion
**Note**: It is possible to install Prometheus without installing Grafana. In this case it is not required to create the CNAME record.
```

PR #4059, opened 2025-02-12
Update docs/configuring-playbook-backup-borg.md: replace the instruction with the reference to the document copied to the MASH project

luixxiul:
Now that the document for setting up BorgBackup has been moved to the MASH project (https://github.com/mother-of-all-self-hosting/ansible-role-backup_borg) it is sensible to remove the details in favor of it.

Because it may as well be noted that the document becomes available locally once the Ansible role is fetched, I think the document itself should not be deleted from the repository.


spantaleev:
This is an interesting new direction for managing the docs, but it does make sense and avoids duplication!

luixxiul:
Hopefully this should help to make it easier for other projects to implement the role on the MASH project directly and configure it for their own use case more easily :smile:

PR #4066, opened 2025-02-14
Make docs/configuring-playbook-etherpad.md slim

luixxiul:
- Add links to the role at MASH project and its document
- Replace common instructions with the links to the role at MASH project

### DIFF
--- a/docs/configuring-playbook-etherpad.md
+++ b/docs/configuring-playbook-etherpad.md
@@ -48,9 +48,17 @@ etherpad_enabled: true
 
 As the most of the necessary settings for the role have been taken care of by the playbook, you can enable Etherpad on your Matrix server with this minimum configuration.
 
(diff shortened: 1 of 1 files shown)


docs/configuring-playbook-etherpad.md
```diff
@@ -30,12 +32,14 @@ To enable Etherpad, add the following configuration to your `inventory/host_vars
 
 ```yaml
 etherpad_enabled: true
-
-# Uncomment and adjust this part if you'd like to enable the admin web UI
-# etherpad_admin_username: YOUR_USERNAME_HERE
-# etherpad_admin_password: YOUR_PASSWORD_HERE
```
spantaleev, line 36: It seems like these are getting lost and they could be quite important for protecting your instance from anonymous users using it.

luixxiul: Hopefully https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/4066/commits/91d991cac6d497f0383a31a2c0e760953716b427 should address this :slightly_smiling_face:

PR #4084, opened 2025-02-20
Rename getting-the-playbook.md to getting-playbook.md

luixxiul:
The conventional practice is not to include an article to the file name, and we do not have to make an exception here, though the file has existed since years.


spantaleev:
`getting-playbook` seems kind of incomplete to me, but maybe it's fine and I can get used to it.

Sometimes, I've been thinking of renaming all other `configuring-playbook-*.md` files, because the `configuring-playbook-` prefix is quite long.

This is all a backward compatibility break though. We'll break all our own links (which we can fix more easily), but we'll also break links for people linking from outside.

Another idea may be to move the configuration-related documentation to a `configuration/` subdirectory. There may even be multiple levels for better organization and "fewer doc files per directory", instead of the current state (lots of files all in the same place). Examples:

- `docs/configuration/bridge/mautrix-whatsapp`
- `docs/tooling/ansible.md`
- `docs/tooling/just.md`
- `docs/maintenance/synapse.md`

luixxiul:
Closing as obsolete.

PR #4129, opened 2025-03-01
Fix GitHub issue template selector

luixxiul:
Apparently those Markdown files are not expected to put something before the three hyphens on the top.


luixxiul:
Fixed by https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/70655e898163f0e0078ca6bb611ff7b943f02df3.

PR #4166, opened 2025-03-12
Update D4A Configuration

FSG-Cat:
D4A had some breaking config changes so this commit fixes them and gets us back into compliance with upstream. 

Now this PR also deals with a second matter. It updates the SPDX headers in the affected files to be correct by only crediting Cat in addition to the named other contributors as well Cat is the only mdad side contributor to these files as far as git blame is concerned.

Now Draupnir doesnt use AGPL and all the config is taken straight from Draupnir so atleast for production-bots.yaml.j2 it could be argued that it should be Gnuxie and CC0 as thats the upstream licence and Credit. And Draupnir is also doing Reuse. 

Anyways Cat isnt an expert at the reuse stuff. The code it self runs as its what runs on my System currently and was required to unbreak D4A.

And yes Room state variable is a carbon copy of the bot mode config as its the same option and i cba writing new text.

### DIFF
--- a/roles/custom/matrix-appservice-draupnir-for-all/defaults/main.yml
+++ b/roles/custom/matrix-appservice-draupnir-for-all/defaults/main.yml
@@ -52,7 +52,7 @@ matrix_appservice_draupnir_for_all_systemd_wanted_services_list: []
 # Note: Draupnir is fairly verbose - expect a lot of messages from it.
 # This room is diffrent for Appservice Mode compared to normal mode.
 # In Appservice mode it provides functions like user management.
-matrix_appservice_draupnir_for_all_config_adminRoom: "" # noqa var-naming
+matrix_appservice_draupnir_for_all_config_adminRoom: ""  # noqa var-naming
 
 # Controls if the room state backing store is activated.
 # Room state backing store makes restarts of the bot lightning fast as the bot does not suffer from amnesia.
--- a/roles/custom/matrix-appservice-draupnir-for-all/defaults/main.yml
+++ b/roles/custom/matrix-appservice-draupnir-for-all/defaults/main.yml
@@ -52,12 +52,7 @@ matrix_appservice_draupnir_for_all_systemd_wanted_services_list: []
 # Note: Draupnir is fairly verbose - expect a lot of messages from it.
 # This room is diffrent for Appservice Mode compared to normal mode.
 # In Appservice mode it provides functions like user management.
-matrix_appservice_draupnir_for_all_master_control_room_alias: ""
-
-# Placeholder Remenant of the fact that Cat belived Master Control Room to be separated from Access Control Policy List.
-# The alias of the Policy list used to control who can provision a bot for them selfs.
-# This should be a room alias - not a matrix.to URL.
-# matrix_appservice_draupnir_for_all_management_policy_list_alias: ""
+matrix_appservice_draupnir_for_all_config_adminRoom: "" # noqa var-naming
 
 # Controls if the room state backing store is activated.
 # Room state backing store makes restarts of the bot lightning fast as the bot does not suffer from amnesia.
--- a/roles/custom/matrix-appservice-draupnir-for-all/defaults/main.yml
+++ b/roles/custom/matrix-appservice-draupnir-for-all/defaults/main.yml
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: 2024 - 2025 MDAD project contributors
+# SPDX-FileCopyrightText: 2024 - 2025 Catalan Lover <catalanlover@protonmail.com>
 # SPDX-FileCopyrightText: 2024 - 2025 Slavi Pantaleev
 # SPDX-FileCopyrightText: 2024 Suguru Hirahara
 #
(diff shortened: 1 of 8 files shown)


roles/custom/matrix-appservice-draupnir-for-all/defaults/main.yml
```diff
@@ -58,6 +58,11 @@ matrix_appservice_draupnir_for_all_master_control_room_alias: ""
 # This should be a room alias - not a matrix.to URL.
 # matrix_appservice_draupnir_for_all_management_policy_list_alias: ""
 
+# Controls if the room state backing store is activated.
+# Room state backing store makes restarts of the bot lightning fast as the bot does not suffer from amnesia.
+# This config option has diminished improvements for bots on extremely fast homeservers or very very small bots on fast homeservers.
+matrix_appservice_draupnir_for_all_enable_room_state_backing_store: "false"
```
spantaleev, line 64: Why not a bool (`"false"` -> `false`)?

Furthermore, this seems to control a configuration property at the `roomStateBackingStore.enabled` path. To be consistent with other roles (and to have less surprising naming as far as people are concerned), this variable should be named in one of these ways:

- `matrix_appservice_draupnir_for_all_room_state_backing_store_enabled`
- `matrix_appservice_draupnir_for_all_roomStateBackingStore_enabled`
- `matrix_appservice_draupnir_for_all_config_room_state_backing_store_enabled` (explicit `_config_` prefix before the actual property name)
- `matrix_appservice_draupnir_for_all_config_roomStateBackingStore_enabled`  (explicit `_config_` prefix before the actual property name)

When `roomStateBackingStore` is used, ansible-lint/yamllint would complain about variable naming and you'd need to define it like this:

```yml
matrix_appservice_draupnir_for_all_roomStateBackingStore_enabled: false  # noqa var-naming
```

------

It seems like the other configuration settings don't follow these rules as well, but it's never too late to start and to change the existing variables later on (in other PRs).

Changing them would involve replacing all references and creating a task in `tasks/validate_config.yml` that detects usage of old variables & tells people about the change.

Like this: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/2f30886b192dceb760b2f818a1dedc5685676fd7/roles/custom/matrix-alertmanager-receiver/tasks/validate_config.yml#L20-L26

roles/custom/matrix-appservice-draupnir-for-all/templates/production-appservice.yaml.j2
```diff
@@ -23,3 +23,9 @@ adminRoom: "{{ matrix_appservice_draupnir_for_all_master_control_room_alias }}"
 # This is a web api that the widget connects to in order to interact with the appservice.
 webAPI:
   port: 9000
+
+# The directory the bot should store various bits of information in
+dataPath: "/data"
+
+roomStateBackingStore:
+  enabled: {{ matrix_appservice_draupnir_for_all_enable_room_state_backing_store }}
```
spantaleev, line 31: Using `to_json` is safer and more consistent with how we do things.

```diff
-  enabled: {{ matrix_appservice_draupnir_for_all_enable_room_state_backing_store }}
+  enabled: {{ matrix_appservice_draupnir_for_all_enable_room_state_backing_store | to_json }}
```

Of course, also see my other comment about renaming this variable. The code suggestion above is

roles/custom/matrix-appservice-draupnir-for-all/defaults/main.yml
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: 2024 - 2025 MDAD project contributors
+# SPDX-FileCopyrightText: 2024 - 2025 Catalan Lover <catalanlover@protonmail.com>
```
spantaleev, line 1: > It updates the SPDX headers in the affected files to be correct by only crediting Cat in addition to the named other contributors as well Cat is the only mdad side contributor to these files as far as git blame is concerned.

I'd say:

1. I don't know why your name isn't there if you're on the `git blame`. Maybe it's an omission and you should have been added. Perhaps @luixxiul can say

2. I think "MDAD project contributors" is there as a catch-all, because roles are typically not created from scratch, but rather based on an existing role. Roles have evolved over time and lots of people have influenced the role you copied from. It's probably impossible to trace it all and credit the correct people now. We do the best we can, which is credit the "MDAD project contributors" group. Now, I don't know if `2024 - 2025` is correct here, given that the role design (and actual YAML files) start their life much earlier, but it would probably be odd to also add `2019` there.

roles/custom/matrix-appservice-draupnir-for-all/defaults/main.yml
FSG-Cat: This specific example should hopefully be fixed now and i will take a look at the rest of the config to align it with this Schema.

Should im not as sure as to when do Explicit `_config_` names. As in if it should be for optional stuff or everything that touches a config file. Tbh doing it for everything that goes into a config file makes sense.

roles/custom/matrix-appservice-draupnir-for-all/templates/production-appservice.yaml.j2
FSG-Cat: `to_json` is now used. The rest of this is addressed separately ofc but is hopefully addressed.

roles/custom/matrix-appservice-draupnir-for-all/defaults/main.yml
spantaleev: Thanks for addressing everything!

You need to use 2 spaces before `#noqa var-naming` to make yamllint happy.

FSG-Cat: That was indeed a silly error for me to make. It has now been resolved.

And as for the `_config_` schema. Stuff that isnt database string generally speaking got the config treatment in the AS was the logic i applied.

Will apply the same type of logic to bot mode when i work on migrating that next.

PR #4169, opened 2025-03-13
Add ensure-users-created to the list of available tags on playbook-tags.md

luixxiul:
The tag has been introduced with 54babc5de0664622edb13e2a07ad099e68b95c80.

### DIFF
--- a/docs/playbook-tags.md
+++ b/docs/playbook-tags.md
@@ -27,7 +27,7 @@ Here are some playbook tags that you should be familiar with:
 
 - `stop` — stops all systemd services
 
-- `ensure-matrix-users-created` and `ensure-users-created` — a special tag which ensures that all special users needed by the playbook (for bots, etc.) are created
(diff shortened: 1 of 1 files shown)


docs/playbook-tags.md
```diff
@@ -27,7 +27,7 @@ Here are some playbook tags that you should be familiar with:
 
 - `stop` — stops all systemd services
 
-- `ensure-matrix-users-created` — a special tag which ensures that all special users needed by the playbook (for bots, etc.) are created
+- `ensure-matrix-users-created` and `ensure-users-created` — a special tag which ensures that all special users needed by the playbook (for bots, etc.) are created
```
spantaleev, line 30:
```suggestion
- `ensure-matrix-users-created` or its alias `ensure-users-created` — a special tag which ensures that all special users needed by the playbook (for bots, etc.) are created
```

Maybe this makes it clearer that the 2 tags are identical?

PR #4176, opened 2025-03-14
Fix bug in marking Pantal and Rust Crypto as mutually exclusive.

FSG-Cat:
So it was reported that when configuring rust crypto the playbook can complain about the fact that Pantal use isnt null.

This PR hopefully moves to a method of detecting this conflict that works for this specific situation. This PR is a draft as i dont know if the logic is sound.

This logic is copied from https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/e073685632b97d0ad837c2119f773e81a756ee44/roles/custom/matrix-synapse/tasks/validate_config.yml#L144-L153

Like to me it looks like a similar situation.


spantaleev:
This looks okay, I guess. Anything else you'd like to add?

FSG-Cat:
> This looks okay, I guess. Anything else you'd like to add?

I have no other changes to the Draupnir and D4A role that i am aware of being needed in the pipeline that arent separate PRs.

PR #4221, opened 2025-03-30
Rework Draupnir report interception to accommodate other Web API uses.

FSG-Cat:
<del>This PR is marked as a draft as i have not yet gone and tested this on my production machine but i wanted to push a PR so that it can be reviewed before it has gone thru validation on my dev box.<del>

Edit: This successfully deployed on my dev box without changes and abuse reporting flows work as expected.


spantaleev:
Looks good! I haven't tested it though.

PR #4229, opened 2025-04-04
Update docs for maintenance: add the description about providers blocking the federation port

luixxiul:
This PR intends to update the migration guide based on my own experience. During migration I've scratched my head for the federation API port not being available, and it turned out the server provider blocked ports other than basic ones (for SSH, HTTPS, etc.)

This kind of stuff can sometimes be difficult to realize, and I think it deserves to be noted for the sake of troubleshooting.


luixxiul:
Closing as obsolete.

PR #4235, opened 2025-04-07
Add lock-threads.yml

luixxiul:
The action will be used to lock issues and PRs to prevent necro-posting on closed ones which have not had any activity in the past year.

### DIFF
--- a/.github/workflows/lock-threads.yml
+++ b/.github/workflows/lock-threads.yml
@@ -1,0 +1,37 @@
+# SPDX-FileCopyrightText: 2017 - 2023 Armin Sebastian
+#
+# SPDX-License-Identifier: MIT
+
+---
+name: 'Lock Threads'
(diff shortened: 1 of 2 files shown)


CyberShadow:
> to prevent necro-posting

Why is this bad?

luixxiul:
Because it is most likely to be unrelated to the bug, and often the context has been long forgotten. Creating a new issue helps  everyone to keep focus on the actual issue.

CyberShadow:
> Because it is most likely to be unrelated to the bug

When this happens, we can move the comment to a new issue (or just ask the user to do this) and hide the original comment:

image

> Creating a new issue helps everyone to keep focus on the actual issue.

I can see how it helps maintainers, who already receive emails on all activity. But for users, this is a negative change, because now it becomes impossible to add valuable context to existing discussions, or collaborate on a long-standing problem which the maintainers don't have the resources to investigate.

PR #4236, opened 2025-04-08
Remove comments before locking threads

tulir:
Necro-posting and notifying all subscribers for the purpose of preventing necro-posting is somewhat ironic


spantaleev:
Agreed! I've received hundreds, if not thousands of emails.

PR #4247, opened 2025-04-15
680 synapse configuration

rbartl:
requirements für etckeeper role hinzugefügt.


spantaleev:
You probably didn't mean to do this.

PR #4267, opened 2025-04-23
Add continuwuity support

Virkkunen:
[continuwuity](https://continuwuity.org) is a continuation of conduwuit, supported by the conduwuit community. At the moment, it is a 100% drop-in replacement from conduwuit, so simply changing the variables and container links should work.


Virkkunen:
As an added note, since continuwuity is a drop-in replacement for conduwuit, it should be possible to have a tag to migrate from one to another, by copying the `conduwuit/data` folder to `continuwuity/data` and `conduwuit/config/conduwuit.toml` to `continuwuity/config/continuwuity.toml`

PR #4284, opened 2025-04-29
Add support for synapse-http-antispam and add Draupnir Integration

### LINKED ISSUE #4180: Add support for maunium/synapse-http-antispam
**Is your feature request related to a problem? Please describe.**
The fact that the Mjolnir Anti-spam module simply isnt good at its job. And has not seen any real maintenance in years.

**Describe the solution you'd like**
Add support for [synapse-http-antispam](https://github.com/maunium/synapse-http-antispam) to the playbook.

This would enable Draupnir to use the module and help a future project to add [meowlnir](https://github.com/maunium/meowlnir) support to the project.

Draupnir will currently require the module for features like Room Takedown where if a room gets a takedown policy issued against it that room will be purged from your homeserver via admin API call upon detection. This will be used to combat the current plauge of spam invites to abuse rooms as one example use case.

**Describe alternatives you've considered**
In this case there are not rly any viable alternatives because this module doesnt have competitors for this exact functionality.

**Additional context**
Support for this module needs a tiny bit of configuration in the roles that want to utilise it.

Draupnir needs the following added to its configuration at time of writing.
https://github.com/the-draupnir-project/Draupnir/blob/4ca8d239384ba9485a365b2fef04089fffe2fa7c/config/default.yaml#L283-L287

FSG-Cat:
Fixes https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/4180

This took well a fair bit more work than Cat likes to admit but well here we go. This branch deploys successfully on my test system and well invite blocking works confirming the module works as far as i understand. 

I would argue that what is left now is things like refining the documentation if that is needed and some other detail work like that around playbook conventions that Cat might not be aware of.

### DIFF
--- a/roles/custom/matrix-bot-draupnir/defaults/main.yml
+++ b/roles/custom/matrix-bot-draupnir/defaults/main.yml
@@ -138,6 +138,25 @@ matrix_bot_draupnir_web_url: 'http://matrix-bot-draupnir'
 matrix_bot_draupnir_config_web_synapseHTTPAntispam_enabled: false  # noqa var-naming
 matrix_bot_draupnir_config_web_synapseHTTPAntispam_authorization: ''  # noqa var-naming
 
+# This controls the URL that the module targets in Draupnir.
+matrix_bot_draupnir_synapse_http_antispam_config_base_url: "{{ matrix_bot_draupnir_web_url }}:{{ matrix_bot_draupnir_config_web_port }}/api/1/spam_check"
+
+# These variables control the configuration of the Synapse module as the configuration is highly consumer dependent.
+# Therefore the module is configured from Draupnir because the consumer of the module determines what settings are relevant.
+
+matrix_bot_draunpir_synapse_http_antispam_config_enabled_callbacks_yaml: |
+  - check_event_for_spam
+  - user_may_invite
+  - user_may_join_room
+
+matrix_bot_draupnir_synapse_http_antispam_config_fail_open_yaml: |
+  check_event_for_spam: true
+  user_may_invite: true
+  user_may_join_room: true
+
+matrix_bot_draupnir_synapse_http_antispam_config_async_yaml: |
+  check_event_for_spam: true
+
 # Default configuration template which covers the generic use case.
 # You can customize it by controlling the various variables inside it.
 #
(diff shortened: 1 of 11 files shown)


roles/custom/matrix-bot-draupnir/defaults/main.yml
```diff
@@ -129,6 +129,34 @@ matrix_bot_draupnir_config_admin_enableMakeRoomAdminCommand: false  # noqa var-n
...
+  - user_may_invite
+  - user_may_join_room
+
+matrix_bot_draupnir_synapse_http_antispam_config_fail_open_yaml: |
+  check_event_for_spam: true
+  user_may_invite: true
+  user_may_join_room: true
+
+matrix_bot_draupnir_synapse_http_antispam_config_async_yaml: |
+  check_event_for_spam: true
```
spantaleev, line 158: Why use YAML here, instead of a list for the first and maps for the others?

It should be possible to pass these without having to define YAML strings and using `from_yaml`

FSG-Cat: Cat wasnt sure how to achive the final result that is needed so this was the solution Cat picked to get something that works until a better solution could be found.

https://the-draupnir-project.github.io/draupnir-documentation/bot/synapse-http-antispam#configuration contains a config example that shows how all of this needs to look once it lands in the Synapse config.

Your suggestion is probably the way to go.

FSG-Cat: So Cat sent a message on matrix relating to this but i think its good to also provide a update here. So i have been attempting to address this feedback in https://github.com/FSG-Cat/matrix-docker-ansible-deploy/tree/FSG-Cat/Draupnir-HTTP-Antispam-Dev but well that branch doesnt deploy.

```
TASK [custom/matrix-synapse : ansible.builtin.set_fact] *******************************************************************************************************************
fatal: [meow.meow.meow]: FAILED! => {"msg": "An unhandled exception occurred while templating '{{ matrix_synapse_ext_synapse_http_antispam_config_yaml | from_yaml }}'. Error was a <class 'yaml.scanner.ScannerError'>, original message: while scanning a simple key\n  in \"<unicode string>\", line 4, column 1\ncould not find expected ':'\n  in \"<unicode string>\", line 5, column 1"}
```

Is the failure generated. This stems from that the templating is failing in the synapse side. And well im unsure how to fix this. The failure comes from the changes im attempting to make to https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/4284/files#diff-b3a271cf06b93ee34ea213f82d47568832064446a50c3f2fed1cd3b8b5f12f66R1437-R1442 in order to accomodate integrating this feedback.

FSG-Cat:
Thank you on behalf of my self and the Draupnir project as i am quite sure Gnuxie is going to be happy with these great news.

PR #4317, opened 2025-05-13
perf(ansible): add ControlPersist for SSH connection reuse

gardar:
Should speed up playbook execution significantly for most users.
While pipelining is already enabled and reduces per-task work inside each SSH session, ControlPersist reduces the cost of establishing SSH sessions by reusing the master connection.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -4,3 +4,6 @@ result_format = yaml
 
 [connection]
 pipelining = True
+
+[ssh_connection]
+ssh_args = -o ControlMaster=auto -o ControlPersist=60s


spantaleev:
Not sure if there's anything special in my environment (not that I know of), but doing this doesn't seem to have any impact.

`time just run-tags install-all` shows about ~300 seconds for a remote server I manage.

gardar:
Perhaps you already have controlpersist in your ssh config?

aine-etke:
the impact depends on how fast you establish a connection from your device (like laptop) to a server. With a fast internet connection and low ping to the server, you may not see any difference. 
So, while having that option enabled by default is a good idea, you won't notice any changes with proper internet connection and server location

spantaleev:
Ansible's default `ssh_args` is `-C -o ControlMaster=auto -o ControlPersist=60s`, so ControlPersist connection reuse is already enabled by default. The only effective change here is dropping the `-C` (compression) flag — which also explains why I couldn't measure any difference back when I tested. Closing.

PR #4352, opened 2025-06-02
Enable pre-commit and run codespell

luixxiul:
This PR intends to add pre-commit hooks by pre-commit.com to `.github/workflows/matrix.yml`, so that basic errors like misspells can be automatically detected by GitHub Action and REUSE can be executed along with it. Please note autofix is not enabled, and linting is not enforced either.


spantaleev:
Does this intentionally remove the `.github/workflows/reuse.yml` file?

luixxiul:
Yes, on CI reuse is going to be executed with the pre-commit hook instead, as described at https://reuse.software/dev/#pre-commit-hook.

PR #4392, opened 2025-06-27
Update Draupnir role Synapse-http-antispam config to match upstream.

FSG-Cat:
The title should say it all. But i will say i am not 100% confident in this PR from a data structure perspective. Like i hope i didnt magically break shit so i would love some feedback on if this even looks sane.

The problem is how i now comment out all lines in the async config as it only had one entry.


FSG-Cat:
Well im not too sure on how to fix this linting error considering my initial attempt just made it worse so feel free to fix it for me or send a suggested fix.

spantaleev:
Maybe it's better to make it an empty list (`[]`) and avoid comments.

FSG-Cat:
> Maybe it's better to make it an empty list (`[]`) and avoid comments.

Sounds like a sensible suggestion.

Hopefully the fix i applied works. Going to go and attempt to run the playbook with that setting just to be sure.

FSG-Cat:
Cat can now say that Cat applied this patch via overriding the variables in my vars.yml and well it works. As in its deployed on my production Synapse deployment and nothing has blown up.

PR #4501, opened 2025-08-14
Stabilize Matrix Authentication Service integration for Synapse

spantaleev:
Related to https://github.com/element-hq/synapse/pull/18759

Currently problematic (leading to failures to start for Synapse) because of: https://github.com/element-hq/synapse/pull/18759#issuecomment-3172744530


aine-etke:
https://github.com/element-hq/synapse/releases/tag/v1.137.0rc1 has the fix, so the PR probably should be merged with the next stable release

PR #4534, opened 2025-08-29
Enable openapi endpoint, when element call is setup - even without fe…

### LINKED ISSUE #4232: Support livekit-jwt without federation
The docs claim that federation support is required for JWT. 
https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/0aeb9a6b14b420731a5e1d510fd9bb6eab798b00/docs/configuring-playbook-element-call.md?plain=1#L19

The openid verification endpoint is indeed part of the federation API. However synapse also supports [configuring a separate listener](https://element-hq.github.io/synapse/latest/usage/configuration/config_documentation.html?highlight=listeners#listeners) for this endpoint, so it can also be exposed anywhere else without enabling full federation, e.g. over the client-server-api. This is even recommended by the [docs of UVS](https://element-hq.github.io/synapse/latest/usage/configuration/config_documentation.html?highlight=listeners#listeners), which is livekit-jwt's spiritual predecessor from jitsi times.

razzeee:
…deration

This is untested and I'm not very familiar with ansible.

Closes #4232

### DIFF
--- a/docs/configuring-playbook-matrix-rtc.md
+++ b/docs/configuring-playbook-matrix-rtc.md
@@ -16,7 +16,6 @@ The Matrix RTC stack is a set of supporting components ([LiveKit Server](configu
 ## Prerequisites
 
 - A [Synapse](configuring-playbook-synapse.md) homeserver (see the warning below)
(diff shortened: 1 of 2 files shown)


spantaleev:
I was initially thinking of enabling the `openid` listener where the Client-Server API is, but doing it this way (exposing `openid` on the federation port):

- makes more sense, as [`openid` routes](https://spec.matrix.org/v1.15/server-server-api/#openid) are part of the [Server-Server API spec](https://spec.matrix.org/v1.15/server-server-api/)
- is simpler, as we already have it implemented and just needed to adjust the config (as done here)

----

It'd be nice to get it tested though!

Even with this, I suppose `https://BASE_DOMAIN/.well-known/matrix/server` still needs to be configured and still needs to point to `matrix.DOMAIN:8448`. Otherwise, the federation base URL (where the `openid` API route is served) will not be determined correctly. Perhaps we need an additional note about this.

PR #4585, opened 2025-09-20
fix: displayname_template lost during parsing to dict

TheBestDestroyer:
Coming from https://github.com/mautrix/whatsapp/issues/565 , I noticed that the variable displayname_template was missing complete in the final configuration file. Therefore the bridge was overwriting it with a default value. (Might be another discussion if these fields should be left out anyway for the most recent defaults to be taken from the official bridge repo). As the creation of the final file is done here via casting to a dict and back to a yaml I removed the raw comment text as a) it is not used anymore and does not appear in the final file and b) this lead to the key being recognised as comment because the "endraw" statement lead to the # being infront of the entry for the parser.

Small fix, but good if anyone wants to customize their displaynames via the playbook.

### DIFF
--- a/roles/custom/matrix-bridge-mautrix-whatsapp/templates/config.yaml.j2
+++ b/roles/custom/matrix-bridge-mautrix-whatsapp/templates/config.yaml.j2
@@ -16,12 +16,6 @@ network:
     proxy_only_login: false
 
     # Displayname template for WhatsApp users.
-    # {% raw %}
-    # {{.PushName}}     - nickname set by the WhatsApp user
-    # {{.BusinessName}} - validated WhatsApp business name
-    # {{.Phone}}        - phone number (international format)
-    # {{.FullName}}     - Name you set in the contacts list
-    # {% endraw %}
     displayname_template: {{ matrix_mautrix_whatsapp_network_displayname_template | to_json }}
 
     # Should incoming calls send a message to the Matrix room?


spantaleev:
Can't say I can reproduce this.

`grep displayname_template -B 5 /matrix/mautrix-whatsapp/config/config.yaml`:

```yml
    # Displayname template for WhatsApp users.
    # {{.PushName}}     - nickname set by the WhatsApp user
    # {{.BusinessName}} - validated WhatsApp business name
    # {{.Phone}}        - phone number (international format)
    # {{.FullName}}     - Name you set in the contacts list
    displayname_template: "{{or .BusinessName .PushName .Phone}} (WA)"
```

julinodev:
Can you please recheck by altering the displayname_template via a hostvar? And are you running the most recent version of the bridge? The bridge itself just inserts this during startup when there is nothing defined. Maybe a more refined way to check: Debug log on  matrix_mautrix_whatsapp_configuration  dictionary and check if it contains the displayname_template entry.


If this is not reproducible it is still a question if the comment should be kept in the template when it is not even used. In the final config file, the bridge itself inserts this part on first start, it is not produced by the playbook in the current form.

spantaleev:
You're right!

There were 2 issues here:

- yours is a legit template issue that was causing `displayname_template` to be disregarded. As a result, `displayname_template` was missing from the final config file

- mautrix-whatsapp was **not** running with `--no-update` like our other bridges (fixed in 2ea438484), which was causing mautrix-whatsapp update the configuration and add `displayname_template`, [making me think](https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/4585#issuecomment-3314965035) that everything was OK

Now that we pass `--no-update` to mautrix-whatsapp, it's easy to see that the templating issue is there and that your fix actually works.

Merged! Thank you! :bow:

PR #4635, opened 2025-10-10
Update s3 storage provider version to 1.6.0, add renovate

bblacher:
https://github.com/matrix-org/synapse-s3-storage-provider/releases/tag/v1.6.0

I don't know if this is the preferred way to have renovate handle non-docker updates.
The update to 1.6.0 is needed for https://github.com/element-hq/synapse/releases/tag/v1.140.0rc1


spantaleev:
Thanks! Looks like we had to use `depName` instead of `packageName`, but I've fixed that up.

spantaleev:
`s3-storage-provider=1.6.0` requires Synapse v1.140.0, which hasn't been released yet.
Ref: https://github.com/matrix-org/synapse-s3-storage-provider/pull/134#issuecomment-3396609289

So... I've reverted us to `s3-storage-provider=1.5.0`.

PR #4680, opened 2025-10-25
Enable mautrix-slack relay mode with `matrix_bridges_relay_enabled`

derhagen:
I found a comment that mautrix-slack didn't support relay mode. However, relay mode has been supported since the [september release](https://github.com/mautrix/slack/commit/8179a5d5f630dac6ab22e245268a2c9fcdd78837).

This commit enables mautrix-slack relay mode when `matrix_bridges_relay_enabled` is set and updates the docs.


derhagen:
Oh wow, that was quick :rocket:

PR #4694, opened 2025-10-29
Bump synapse, remove some recently stabilized features

bblacher:


bblacher:
I also took a look at some other vars (which control already merged MSCs), some of those seem like they're still required, even though the MSC is merged. Synapse docs on this should really be improved.

Issue #3867, opened 2024-12-10
Remove `matrix-bridge-mx-*` roles which have alternative


luixxiul:
**Is your feature request related to a problem? Please describe.**

@spantaleev reports at https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/CHANGELOG.md#backward-compatibility-break-synapse-now-defaults-to-enabling-authenticated-media:

> Old and unmaintained bridges (like all mx-puppet bridges, etc.) do not support authenticated media. Other software may be similarly affected. If you experience issues with some Matrix-related software, you may wish to disable authenticated media and contact the software maintainers to let them know.

I assume it would be acceptable to remove these roles if fixing incompatibilities require him (or other persons) so much work, especially of mx bridges which have alternatives, such as:

- matrix-bridge-mx-puppet-discord (alternatives: matrix-bridge-appservice-discord, matrix-bridge-mautrix-discord)
- matrix-bridge-mx-puppet-instagram (alternative: matrix-bridge-mautrix-meta-instagram)
- matrix-bridge-mx-puppet-slack (alternative: matrix-bridge-mautrix-slack. Note: matrix-bridge-appservice-slack is available only if classic app has been created)
- matrix-bridge-mx-puppet-twitter (alternative: matrix-bridge-mautrix-twitter)

There is also a report that the current documentation for mx-puppet-steam is not correct: https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/1371 (although it is not sure whether the role itself has become outdated or not)

As the role for Google Hangouts is scheduled to be removed from the playbook (see [here](https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3863#issuecomment-2527185217)), I assume the same kind of plan can be applied to the mx bridges removal. Besides, if the issues will be fixed by the upstream project, it should be relatively easy to re-implement the bridges by reverting the change.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
(This suggestion is per communication with Slavi)

transcaffeine:
👍 Those bridges are in addition to not supporting authenticated media also not very active, with last activity somewhere in 2024. I think suggesting to use the alternatives and removing the roles is a good idea

PR #4707, opened 2025-11-04
Remove MX Puppet bridges for Discord, Instagram, Slack, and Twitter

luixxiul:
Closes https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3867

Since almost one year after https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3867 was created, not one of these bridges became active, so it is sensible to call them unmaintained.

We can just revert the commits if they become active again.

Because MX Puppet Steam bridge is claimed to work and MX Puppet GroupMe bridge does not have an alternative, this PR does not intend to remove them. We can revisit them later if necessary.

It is not necessary to merge this just away; though I checked things are correctly removed, double-checking would be really appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,28 +1,22 @@
-# 2025-11-04
+# 2025-11-05
 
-## The MX Puppet Discord bridge has been removed from the playbook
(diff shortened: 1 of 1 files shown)


spantaleev:
It looks good! Nice to see all this cruft removed! :bow:

PR #4714, opened 2025-11-07
Update matrix-steam-bridge - Fix DisplayName Templating

jasonlaguidice:
I recently fixed DisplayName template handling in the bridge and subsequently bumped the version tag to v1.0.8

This PR adds that change and adjusts the default template to use `.PersonaName` - which will be the Steam user's chosen display name. I added extra comments by that field for the other acceptable values.


spantaleev:
I've "manually cherry-picked" this change in 1861faf31d5490c6604efdd34d83073ec3850d88

PR #4716, opened 2025-11-08
Remove Go-NEB

luixxiul:
It does not feel nice to remove components in a row, but it does not make sense to keep what's been deprecated by the upstream in the playbook, and reducing tech debt will improve quality of the playbook down the road.


bblacher:
A bit off-topic, but does it still make sense to keep the sliding sync proxy around?

luixxiul:
If there is not something which uses it then it should be fine to remove it, but I am not sure if there is one or not.

spantaleev:
Yes, sliding-sync should be fine to remove. It's long been replaced with native sliding sync.

luixxiul:
I'm taking care of it soon!

PR #4717, opened 2025-11-08
Remove sliding sync proxy

luixxiul:
cf.: https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/4716#issuecomment-3506091074


spantaleev:
Thank you and RIP! :smile:

PR #4776, opened 2025-12-02
Stop using more deprecated `vars` variables

HarHarLinks:
My playbook threw these warnings to I copied what was done in 430f350733e4695affccad06464a1ce73ceb1bb3


HarHarLinks:
a `rg 'vars\[item'` says there are quite some more left as well :/

PR #4787, opened 2025-12-07
Matrix.to: Use matrix_base_data_path as base

der-domi:
This PR is done to put Matrix.to's data into `matrix_base_data_path`.

### DIFF
--- a/roles/custom/matrix-matrixto/defaults/main.yml
+++ b/roles/custom/matrix-matrixto/defaults/main.yml
@@ -12,7 +12,7 @@
 matrix_matrixto_enabled: true
 
 matrix_matrixto_identifier: matrix-matrixto
-matrix_matrixto_base_path: "/{{ matrix_matrixto_identifier }}"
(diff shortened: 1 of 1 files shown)


spantaleev:
The role default is OK and is consistent with what other roles already do - keeping the role independent of other roles' variables as much as possible.

Such re-wiring in terms of playbook variables is then typically accomplished in `group_vars/matrix_servers`. I've done it in https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/88dcfbdaa7dcbca6e38343c240e58185e2e27f30

der-domi:
Hey. Okay, I'm wondering. I took the example from the first role I found: [`roles/custom/matrix-alertmanager-receiver`](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/roles/custom/matrix-alertmanager-receiver/defaults/main.yml#L25).
But okay, the main thing is, that the data is stored in the central folder. :-)

Issue #4820, opened 2025-12-20
Disable matrix-matrixto for Watchtower


der-domi:
**Is your feature request related to a problem? Please describe.**
This request is related to #4818 - a bit.

If you are using [Watchtower](https://watchtower.nickfedor.com/) with following configuration
```yaml
environment:
  - "WATCHTOWER_REMOVE_VOLUMES=true"
  - "WATCHTOWER_CLEANUP=true"
  - "WATCHTOWER_SCHEDULE=0 11 */4 * * *"
  - "WATCHTOWER_NO_STARTUP_MESSAGE=true"
  - "WATCHTOWER_NOTIFICATIONS=email"
```

you will get continuously following message by Watchtower

> Digest retrieval failed, falling back to full pull | container=matrix-matrixto error=registry responded with invalid HEAD request: status 404 Not Foundimage=shirahara/matrixto:1.2.17-1

This is expected since the container is deployed locally and not available at any registry.

**Describe the solution you'd like**
Watchtower provides the possibility to disable watching for a certain image by adding following label

```yaml
matrix-matrixto:
  labels:
    - "com.centurylinklabs.watchtower.enable=false"
```

This will deactivate the notification for all users automatically. See [configuration](https://watchtower.nickfedor.com/latest/configuration/arguments/#enable_label_filter) and [example](https://watchtower.nickfedor.com/latest/advanced-features/http-api/#example).

**Describe alternatives you've considered**
Alternatively each user (which is using Watchtower) has to [disable the specific container](https://watchtower.nickfedor.com/latest/configuration/arguments/#disable_specific_containers) by its own.

**Additional context**
-/-

PR #4821, opened 2025-12-20
Disable matrix-matrixto for Watchtower

der-domi:
This solves #4820.
- Add the additional label for Matrix.to to disable to disable Watchtower's notification.

### DIFF
--- a/roles/custom/matrix-matrixto/defaults/main.yml
+++ b/roles/custom/matrix-matrixto/defaults/main.yml
@@ -104,7 +104,7 @@ matrix_matrixto_container_labels_traefik_additional_response_headers_custom: {}
 # matrix_matrixto_container_labels_additional_labels: |
 #   my.label=1
 #   another.label="here"
-matrix_matrixto_container_labels_additional_labels: ""
+matrix_matrixto_container_labels_additional_labels: "com.centurylinklabs.watchtower.enable=false"
 
 # A list of extra arguments to pass to the container (`docker run` command)
 matrix_matrixto_container_extra_arguments: []


spantaleev:
This works, but.. `matrix_matrixto_container_labels_additional_labels` is meant to be used by people to inject additional labels. If they start using this new variable, they will find that this special label is no longer being applied.

It's better to use something else - perhaps even a special `matrix_matrixto_container_labels_watchtower_skip_enabled: true` value which injects the correct label (perhaps the label can be configurable too, via `matrix_matrixto_container_labels_watchtower_skip_label: com.centurylinklabs.watchtower.enable=false`).

This frees up `matrix_matrixto_container_labels_additional_labels` for usage by others.

-----

An alternative is to split `matrix_matrixto_container_labels_additional_labels` into `matrix_matrixto_container_labels_additional_labels_default`, `matrix_matrixto_container_labels_additional_labels_auto` and `matrix_matrixto_container_labels_additional_labels_custom`. We have never done this for the "additional labels", but at least it's consistent with what we do with other variables.

These variables can then be merged into `matrix_matrixto_container_labels_additional_labels`, which will be used in the `labels.j2` file as it currently is.

The watchtower-related label can be injected into `matrix_matrixto_container_labels_additional_labels_default`. Perhaps it should be conditional, via other variables (perhaps `matrix_matrixto_container_labels_watchtower_skip_enabled` and `matrix_matrixto_container_labels_watchtower_skip_label`).

luixxiul:
> An alternative is to split ...

Off-topic, but probably `*_container_extra_arguments_auto`, `_default`, `_custom` too? Some have already these, but not all do.

spantaleev:
These could also work, but it's better to have the labels in the dedicated labels file, instead of dirtying up the systemd service file and making the container start command more complicated.

der-domi:
Okay, thanks for the feedback. I will have a deeper look to it. But I need a while...

spantaleev:
Thanks for the report and the patch! As discussed above, pre-filling `matrix_matrixto_container_labels_additional_labels` would make the label silently disappear for anyone overriding that variable, so this landed via a dedicated `matrix_matrixto_container_labels_watchtower_skip_enabled` variable (enabled by default) in 7d2a2c40c. Closing.

der-domi:
Currently I'm playing around with the Renovate-Bot and [Doco-CD](https://doco.cd). So, my future plan is to update my containers with those services. Then I won't use Watchtower anymore - or only for some specific containers.

That's why I got stuck here.

PR #4851, opened 2026-01-10
Matrix media repo redis fix

akdk7:
If Redis is enabled for the Matrix media repository, a connection to Valkey cannot be established due to a configuration error.

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -3649,7 +3649,7 @@ matrix_media_repo_container_additional_networks: |
       +
       ([matrix_playbook_reverse_proxyable_services_additional_network] if (matrix_playbook_reverse_proxyable_services_additional_network and matrix_media_repo_container_labels_traefik_enabled) else [])
       +
-      ([valkey_container_network] if valkey_enabled else [])
+      ([valkey_container_network] if valkey_enabled and matrix_media_repo_redis_enabled else [])
     ) | unique
   }}
 


spantaleev:
It's probably better to do these things in `group_vars/matrix_servers`:

```yml
matrix_media_repo_redis_enabled: "{{ valkey_enabled }}"

# Use next redis index since Synapse is on 0. You can chose between index 0 and 15.
matrix_media_repo_redis_database_number: 1

matrix_media_repo_redis_shards: |
  {{
    ([{
          'name': 'valkey',
          'addr': (valkey_identifier + valkey_container_http_port | string),
    }])
    if valkey_enabled and matrix_media_repo_redis_enabled
    else []
  }}
```

Then, the role itself can keep the defaults as before but maybe `matrix_media_repo_redis_shards` better default to empty (`[]`):

```yml
# The database number to use. Leave at zero if using a dedicated Redis instance.
matrix_media_repo_redis_database_number: 0

# The Redis shards that should be used by the media repo in the ring. The names of the
# shards are for your reference and have no bearing on the connection, but must be unique.
matrix_media_repo_redis_shards: []
```

This leaves the role generic and independent enough, and wiring between different components is moved to `group_vars/matrix_servers`, which is consistent with how this project operates.

akdk7:
Sure. If that is a better fitting approach for the project.

akdk7:
Just saw something. Isn't it missing a colon tho?

(valkey_identifier + valkey_container_http_port | string),

spantaleev:
Good catch! It was missing. Fixed in https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/8a02d791ea1bd89e2c54b4fedde6460f15be70ea

akdk7:
Thanks!

PR #4880, opened 2026-01-27
Upgrade Coturn (4.6.2-r11 -> 4.8.0-r0)

sidm-finno:
Major version bump to 4.8.0. This update tracks the [latest upstream](https://github.com/coturn/coturn/blob/docker/4.8.0-r0/docker/coturn/CHANGELOG.md) security fixes and minor protocol improvements.


spantaleev:
Thanks! It seems like Renovate was (and still is) sleeping on this already quite old update!

I'll do some changes to the `versioning` value which would hopefully help.

PR #4898, opened 2026-02-06
Remove ma1sd

luixxiul:
This intends to remove ma1sd. Not tested because ma1sd had not been installed on my machine..


tiran133:
Why was this removed? I'm aware it's very old, but it's working for us as an email invite feature!

Is there any alternative? 

Or is there a way to add it back locally?

PR #4910, opened 2026-02-10
Synapse register user: wait for synapse to be available

aine-etke:
This PR adds wait-for functionality to the Synapse register-user script.
Basically, it waits until Synapse container is up and ready to accept http connections


aine-etke:
discarded - current paybook behavior with calling `docker exec` directly from ansible considered good enough

PR #4914, opened 2026-02-12
Fix regression introduced in a77a875

krejcar25:
`matrix_synapse_systemd_service_post_start_delay_seconds` is [assigned a string value](https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/a77a8753d9e50f7587e4ccf35203111c502c6e6d#diff-250fee5433c247b39e35580376e6e53e3b229255269e453a092ac9b3b12848b7R4472), and setup fails while creating the service file. It is impossible to compare str and int.

---

Edit: meta: I can't find a contributing guide. If DCO or `*.license` file should be updated please let me know and I will amend that. Otherwise it's such a miniature change that I don't believe it necessary.

### DIFF
--- a/roles/custom/matrix-synapse/templates/synapse/systemd/matrix-synapse.service.j2
+++ b/roles/custom/matrix-synapse/templates/synapse/systemd/matrix-synapse.service.j2
@@ -74,7 +74,7 @@ ExecStart={{ devture_systemd_docker_base_host_command_docker }} start --attach m
 ExecStartPost={{ matrix_synapse_systemd_healthcheck_command }}
 {% endif %}
 
-{% if matrix_synapse_systemd_service_post_start_delay_seconds > 0 %}
+{% if matrix_synapse_systemd_service_post_start_delay_seconds | int > 0 %}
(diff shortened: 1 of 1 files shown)


spantaleev:
Thank you!

Explanation by Claude as to why this is necessary **for some setups**:

The `matrix_synapse_systemd_service_post_start_delay_seconds` variable is defined in `group_vars/matrix_servers` as a `"{{ expr }}"` Jinja2 expression. Whether this evaluates to a native Python `int` or a `str` depends on the `jinja2_native` Ansible setting:

- **ansible-core 2.18+** defaults `jinja2_native` to `True`, so `"{{ expr }}"` preserves the native int type. The template comparison `matrix_synapse_systemd_service_post_start_delay_seconds > 0` works fine (`int > int`).
- **ansible-core 2.17 and older** defaults `jinja2_native` to `False`, so `"{{ expr }}"` always produces a string (e.g., `"3"` or `"0"`). Then in the template, `"3" > 0` is a `str > int` comparison, which raises a `TypeError` in Python 3.

The `| int` filter fix ensures the comparison works regardless of the `jinja2_native` setting and the Ansible version.

PR #4922, opened 2026-02-13
migrate FluffyChat to SWS; add config mounting with correct server by default

aine-etke:


spantaleev:
`matrix_client_fluffychat_data_path: "{{ matrix_base_data_path }}/client-fluffychat"` is a bad variable name.

We call this a `base_path` elsewhere.

Then we usually have data path (data/) and config path (config/), which are derived from the base path.

This config.json typically should go to the config path (config/config.json).
`labels` is typically at the base path for other roles.

-----

This can be done without compatibility issues right now, since it's mostly a variable naming change and the labels file remains where it is.

When done, the config.json would go to config/config.json. Since no one has deployed this yet, it's OK to do this without a migration as well.

PR #4961, opened 2026-02-24
fix: add livekit_server_container_http_host_bind_port for nginx-on-ho…

AlexanderBrolin:
…st setups

The upstream galaxy/livekit_server role never publishes port 7880 (HTTP signaling) to the host — it assumes a Traefik container on the same Docker network. This breaks setups where nginx runs on a separate host and proxies to the Matrix server over LAN.

Changes:
- Fork galaxy/livekit_server → roles/custom/livekit_server/ so the patch survives `make roles` (which wipes roles/galaxy/)
- Add livekit_server_container_http_host_bind_port variable (defaults to "") in defaults/main.yml; when set, publishes the HTTP port via -p flag
- Add {% if livekit_server_container_http_host_bind_port %} block in the systemd service template to conditionally map the port
- Guard network-creation tasks with `and livekit_server_container_network != 'host'` to prevent 403 errors when network: host is configured
- Update setup.yml: galaxy/livekit_server → custom/livekit_server

Usage in vars.yml:
  livekit_server_container_http_host_bind_port: '192.168.90.83:7880'

https://claude.ai/code/session_01SbRgWsRDKpTZ2TTbWDhtEB

### DIFF
--- a/docs/configuring-playbook-livekit-jwt-service.md
+++ b/docs/configuring-playbook-livekit-jwt-service.md
@@ -16,3 +16,42 @@ Take a look at:
 
 - `roles/custom/matrix-livekit-jwt-service/defaults/main.yml` for some variables that you can customize via your `vars.yml` file
 - `roles/custom/matrix-livekit-jwt-service/templates/env.j2` for the component's default configuration.
+
+## Using with an external nginx reverse-proxy
+
+If you use nginx (running on a separate host or outside Docker) to reverse-proxy the LiveKit JWT Service, the location block **must include a trailing slash** after the path prefix.
+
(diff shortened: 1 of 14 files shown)


AlexanderBrolin:
1

spantaleev:
Thanks. The `livekit_server` role is maintained externally at [mother-of-all-self-hosting/ansible-role-livekit-server](https://github.com/mother-of-all-self-hosting/ansible-role-livekit-server), so we can't vendor it into `roles/custom/`. The vendored copy in this PR is also already stale: based on LiveKit v1.9.11 (upstream is now v1.11.0) and missing features that have landed since, such as host network mode, TURN/TLS behind Traefik, container cleanup/grace timings, TURN relay range configuration, etc. The `container_network == 'host'` guards you added are also already handled upstream in a cleaner way.

That said, the underlying need is legitimate: exposing LiveKit's HTTP signaling port (7880) on a host interface so an off-host reverse-proxy can reach it. I've taken just that idea and added it upstream:

- [mother-of-all-self-hosting/ansible-role-livekit-server@6f7d663](https://github.com/mother-of-all-self-hosting/ansible-role-livekit-server/commit/6f7d663) adds `livekit_server_container_http_listen_interface`, following the existing `_rtc/_turn/_metrics_listen_interface` convention (takes an IP or empty string; port is implicit from `livekit_server_config_port`).
- Released as `v1.11.0-1` and already pinned in the playbook: 675338fc1.

Your use case after `just roles`:

```yaml
livekit_server_container_http_listen_interface: "192.168.90.83"
```

Closing.

PR #4979, opened 2026-02-26
Add agru to flake.nix

jack-wines:
The [nixpgks pr for agru](https://github.com/NixOS/nixpkgs/pull/492883) is still simmering so I figured I'd add it here directly in the meantime. That means that it'll download and compile agru on first run of `nix develop`. It doesn't take that long though, so I'd judge it worth it. 

This pr also updates the nix dependencies. The existing version of ansible fails when determining the existing postgres docker version. I think it's using an old docker api or something.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,15 +1,31 @@
 {
   "nodes": {
+    "agru-src": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1772059918,
+        "narHash": "sha256-o3ZNyJjDEY2RCwIiBWSjWTo6/DqAVMK+ElTuwAVdZrg=",
+        "owner": "etkecc",
+        "repo": "agru",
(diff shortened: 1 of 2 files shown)


jack-wines:
to clarify, if/when agru makes it into nixpkgs I'll make another PR to just fetch it from there instead

spantaleev:
Merged (with a small conflict resolution: master's flake had gained `mise` in the meantime). Renovate's nix support with flake.lock maintenance has also been enabled in aa97075f1, which should keep the agru-src and nixpkgs inputs fresh going forward (agru is at v0.2.1 already). The `version` label in flake.nix will drift from the locked revision over time, but that's cosmetic. Looking forward to the follow-up that switches to the nixpkgs package once your nixpkgs PR lands. Thanks!

PR #4985, opened 2026-02-28
fix(synapse s3 storage): make prefix used

parisni:
currently the s3 prefix is not reused


spantaleev:
Thanks! Announced as potential backward compatibility break.

[(Potential BC Break) Synapse S3 media prefix is now applied consistently](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/46321552b7fd18e30cf43638fca932773cd5c8a5/CHANGELOG.md#potential-bc-break-synapse-s3-media-prefix-is-now-applied-consistently)

PR #4997, opened 2026-03-03
Add commet webclient support

D4GU:
This allows setting up the commet web client, in a similar way as the other already existing webclients. The client still in development but already quite good and currently there is no available package to use on the main repository (https://github.com/commetchat/commet). I'm using one created in Fork of it and have a runner setup to build it.


luixxiul:
I am not really sure why we would need `--build-arg` here.

If they are not necessary, let's remove them to omit `Check Commet git remote configuration` and `Capture git hash` tasks, which introduce ansible-lint errors.

PR #5008, opened 2026-03-05
Fix commet client missing defintion

D4GU:
There's an issue with the current version when pulling the commet client image, its asking for this definition which does not exist. +Default enabled is probably not the way.


spantaleev:
Enabled by default is what we do for all components in role defaults. We then override this in `group_vars/matrix_servers` to disable it. We already do this now.

spantaleev:
Is there a `main` image tag for this?

D4GU:
Mmm not really currently the workflow has not been accepted to build and push to ghcr in the official repository, https://github.com/commetchat/commet/pull/789. I'm using this self built one currently which has it https://github.com/D4GU/selfbuild-commet/pkgs/container/selfbuild-commet. For this it works correctly with the change.
Might make sense to maybe wait if they'll support it though. Not sure how often I'll maintain the container image.

luixxiul:
The official container image does not exist yet.

Issue #4281, opened 2025-04-27
"Insufficient capacity" when joining video room


INNKCake:
**Playbook Configuration**:

My `vars.yml` file looks like this:

```yaml
---
matrix_domain: server-name.org

matrix_homeserver_implementation: synapse

#Configure workers
matrix_synapse_workers_enabled: true
matrix_synapse_workers_preset: one-of-each
matrix_synapse_report_stats: false
matrix_homeserver_federation_enabled: true

#Configure element-web
matrix_client_element_default_country_code: "redacted"
matrix_client_element_themes_enabled: true
matrix_client_element_themes_repository_url: "https://github.com/aaronraimist/element-themes.git"

#Enable double-puppeting for bridges (https://docs.mau.fi/bridges/general/double-puppeting.html)
matrix_appservice_double_puppet_enabled: true

#Add admin webgui
matrix_synapse_admin_enabled: true

#Shared secret
matrix_synapse_ext_password_provider_shared_secret_auth_enabled: true
matrix_synapse_ext_password_provider_shared_secret_auth_shared_secret: "redacted"

#Enabled bridges
matrix_mautrix_discord_enabled: true
matrix_mautrix_discord_logging_level: 'debug'

matrix_mautrix_telegram_enabled: true
matrix_mautrix_telegram_api_id: 'redacted'
matrix_mautrix_telegram_api_hash: 'redacted'

matrix_element_call_enabled: true

matrix_synapse_enable_registration: true
matrix_synapse_registration_requires_token: true

matrix_homeserver_generic_secret_key: 'redacted'
matrix_playbook_reverse_proxy_type: playbook-managed-traefik
matrix_playbook_ssl_enabled: true
traefik_config_entrypoint_web_secure_enabled: false
traefik_container_web_host_bind_port: '127.0.0.1:81'
traefik_config_entrypoint_web_forwardedHeaders_insecure: true

matrix_playbook_public_matrix_federation_api_traefik_entrypoint_host_bind_port: '127.0.0.1:8449'
matrix_playbook_public_matrix_federation_api_traefik_entrypoint_config_http3_enabled: false
matrix_playbook_public_matrix_federation_api_traefik_entrypoint_config_custom:
  forwardedHeaders:
    insecure: true
  # trustedIPs: ['IP-ADDRESS-OF-YOUR-REVERSE-PROXY']

devture_systemd_docker_base_ipv6_enabled: true

postgres_connection_password: 'redacted'

[...]

PR #5086, opened 2026-03-26
Add reverse proxy configuration for Matrix RTC

thigg:
I added this because I had to dig it out of https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/4281


spantaleev:
Isn't this a better fit for: `examples/reverse-proxies/nginx`?

thigg:
You're right. I moved it to examples, but also mentioned the problem in the livekit document

PR #5090, opened 2026-03-27
support for MSC4439: PGP key URIs in .well-known

thetayloredman:
for details on the proposed (not yet accepted) MSC, see [msc4439](https://github.com/matrix-org/matrix-spec-proposals/pull/4439) ([rendered](https://github.com/thetayloredman/matrix-spec-proposals/blob/msc4439/proposals/4439-support-contact-encryption.md)).

I haven't messed with Ansible/Jinja2 much, so I'm not sure if this could be improved on. also a pointer as to documentation changes seen as fit would be appreciated!

### DIFF
--- a/roles/custom/matrix-static-files/templates/public/.well-known/matrix/support.j2
+++ b/roles/custom/matrix-static-files/templates/public/.well-known/matrix/support.j2
@@ -1,5 +1,15 @@
+{%- set ns = namespace(contacts_out=[]) -%}
+{%- for contact in matrix_static_files_file_matrix_support_property_m_contacts -%}
+{%- if contact.pgp_key is defined -%}
+{%- set contact_items = contact | dict2items | rejectattr('key', 'equalto', 'pgp_key') | list -%}
+{%- set contact_items = contact_items + [{'key': 'dev.zirco.msc4439.pgp_key', 'value': contact.pgp_key}] -%}
+{%- set ns.contacts_out = ns.contacts_out + [contact_items | items2dict] -%}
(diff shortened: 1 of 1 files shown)


spantaleev:
Thanks, but I'd rather not land this.

MSC4439 isn't accepted yet, and baking the `dev.zirco.` unstable prefix into the playbook would just have to be torn out once the MSC lands (likely with a different prefix or key name).

Also, this isn't needed. The existing template serializes each contact as-is via `to_json`, so any custom key in a contact entry passes through to the output. You can emulate MSC4439 today without any playbook change:

```yaml
matrix_static_files_file_matrix_support_property_m_contacts:
  - role: m.role.admin
    email_address: admin@example.com
    dev.zirco.msc4439.pgp_key: https://example.com/key.asc
```

On the implementation side: the key-rewrite logic (namespace + `dict2items`/`rejectattr`/`items2dict`) turns a one-line template into something fairly hard to follow, for a rename that users can do themselves in config.

Happy to revisit once the MSC is accepted and has a stable name. Closing.

thetayloredman:
@spantaleev All good, totally understand it. FWIW, the MSC process itself is built to flow in that way (implementing the unstable prefix, then once it merges dropping it). I'm maybe not familiar with how others use the playbook, but I thought adding the key that way so once it gets merged the user's vars didn't need changing?

The rewriting process was definitely not great looking though.

But anyway, no issue here, just needed to have an extra something as an implementation for the MSC process. Doesn't need to be merged, just thought I'd have it at least looked at :)

PR #5106, opened 2026-04-01
Update Continuwuity config and add options

DrKnees:
This PR adds the following config options for Continuwuity:

- ReCAPTCHA
- Allow encryption
- Allow room creation
- List of rooms to auto-join
- MSC4267 (forget forced on leave)
- Outgoing presence (disabled by default)
- MSC4143 Livekit discovery endpoint

The MSC4143 Livekit endpoint variable is also calculated from the playbook in matrix_servers (though there appears to be a bug in its implementation in Continuwuity currently: [link to issue](https://forgejo.ellis.link/continuwuation/continuwuity/issues/1602))

I also generally modernised the config template, including changing a number of examples for professionalism's sake.


spantaleev:
This is now merged. I did some fixups and reorganization.

I have not tested the final result, because I don't use this homeserver. So, let me know if there's any breakage by opening an issue, or better yet, a PR!

PR #5148, opened 2026-04-17
Make ansible-lint at least a bit more happy

mcepl:


spantaleev:
Thanks. Took the `defaults/main.yml` reformatting, the `matrix-static-files` `+` spacing fix, and the caddy example indent fix. Dropped the `galaxy/<role>` -> `<role>` renames in `setup.yml` and `jitsi_jvb.yml`. Ansible doesn't recurse into subdirectories of `roles/` by default and there's no `roles_path` in `ansible.cfg`, so those would break role resolution.

Merged to master as e0bc2be68 with you as author. Closing this.

mcepl:
Thank you for taking e0bc2be , but without https://github.com/openSUSE-Python/matrix-docker-ansible-deploy/tree/update_configuration I still don’t get `ansible-playbook` working. I don't understand.

spantaleev:
Why can't you get it working? Is it because of `convert_data`?

As far as I remember, removing `convert_data` breaks compatibility with older versions of Ansible. Having it shows some warnings on newer versions, but.. we need to tolerate those because of the older Ansible versions. At some point, `convert_data` will be dropped.

mcepl:
```
mitmanek~/p/matrix-docker-ansible-deploy (git: test_5148)$ ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start &|tee _log.txt
[ERROR]: the role 'galaxy/playbook_help' was not found in /home/matej/archiv/2026/projekty/matrix-docker-ansible-deploy/roles:/home/matej/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:/home/matej/archiv/2026/projekty/matrix-docker-ansible-deploy
Origin: /home/matej/archiv/2026/projekty/matrix-docker-ansible-deploy/setup.yml:8:7

6   roles:
7     # Most of the roles below are not distributed with the playbook, but downloaded separately using `ansible-galax...
8     - role: galaxy/playbook_help
        ^ column 7

mitmanek~/p/matrix-docker-ansible-deploy (git: test_5148)$ 
```
and similar.

(that’s, of course, AFTER I run `ansible-galaxy install -r requirements.yml`)

Issue #5186, opened 2026-04-30
Conditional restarts not working with forcepulled images?


FSG-Cat:
**Describe the bug**

Draupnir was not conditionally restarted after force pulled container image update.

**To Reproduce**
The relevant parts of my `vars.yml` file looks like this:

```yaml
matrix_bot_draupnir_enabled: true
matrix_bot_draupnir_version: "main"
matrix_bot_draupnir_container_image_force_pull: "{{ matrix_bot_draupnir_container_image.endswith ( 'main') }}"
```

**Expected behavior**
For the service to be conditionally restarted.

**Matrix Server:**
 - OS: Debian 12
 - Architecture AMD64

**Ansible:**
If your problem appears to be with Ansible, tell us:
Pipx installed ansible latest version as of writing as i ran pipx update-all before i had run the playbook update.

**Additional context**

I think appservice might not have been affected and its force pulled using the same logic targeting the same image as Draupnir for all is the same image as draupnir bot mode.

PR #5187, opened 2026-04-30
Modernise Draupnir Role ansible modules and update restart logic

FSG-Cat:
Move the draupnir role to use modern ansible docker modules so we get access to fancy features like buildx and smart pulls. No more force pull bullshit.

And also make the force restart logic that this commit breaks work again but much smarter. Since if we are pulling diffrent tags (Will happen a lot with test environments) then we dont need to force a restart unessecarily.

This also allows people who exclusively run bot mode to still benefit from conditional restarts even if targeting a moving tag like main or latest.

This PR is currently untested properly. Im publishing it as a PR early to gather feedback but please do note its not exactly my most pretty work currently as im using some hacks to allow the roles to be unaware of each others container pull success.

Edit: This PR is now tested on my dev system and it seems to work atleast when it comes to force restarting and not doing wasteful docker pulling when running this branch ontop of the branch for a upcomming draupnir feature. (The feature branch only contains new config keys. Its the Zero Touch Deployment branch found on my fork of this repo. That branch isnt ready and the draupnir PR its tied to isnt merged yet.)

### DIFF
--- a/collections/requirements.yml
+++ b/collections/requirements.yml
@@ -1,8 +1,4 @@
 ---
 collections:
-  # community.docker >= 3.6.0 is required for:
-  # - community.docker.docker_image_pull module (pull-only operations)
-  # - community.docker.docker_image_build module (build with BuildKit support)
   - name: community.general
   - name: community.docker
-    version: ">=3.6.0"
--- a/collections/requirements.yml
+++ b/collections/requirements.yml
@@ -1,4 +1,8 @@
 ---
 collections:
+  # community.docker >= 3.6.0 is required for:
+  # - community.docker.docker_image_pull module (pull-only operations)
+  # - community.docker.docker_image_build module (build with BuildKit support)
   - name: community.general
   - name: community.docker
+    version: ">=3.6.0"
--- a/docs/configuring-playbook-appservice-draupnir-for-all.md
+++ b/docs/configuring-playbook-appservice-draupnir-for-all.md
@@ -48,6 +48,14 @@ matrix_appservice_draupnir_for_all_enabled: true
 matrix_appservice_draupnir_for_all_config_adminRoom: "MANAGEMENT_ROOM_ALIAS_HERE"
 ```
 
+### Running Both Bot Mode and Appservice Mode
+
+When running both [Bot mode](./configuring-playbook-bot-draupnir.md) and appservice mode the playbook will force restart
+the bot if running a non release tag like `latest` or `main` or development builds.
+This is due to the conditional restart logic not being able to tell when a update happened correctly.
+
+Conditional restarts work perfectly for all tags when running only one of these two operating modes.
+
 ### Extending the configuration
 
 There are some additional things you may wish to configure about the component.
--- a/docs/configuring-playbook-bot-draupnir.md
+++ b/docs/configuring-playbook-bot-draupnir.md
@@ -82,6 +82,14 @@ matrix_bot_draupnir_login_native: true
 matrix_bot_draupnir_config_managementRoom: "MANAGEMENT_ROOM_ID_HERE"
 ```
 
+### Running Both Bot Mode and Appservice Mode
+
+When running both Bot mode and [appservice mode](./configuring-playbook-appservice-draupnir-for-all.md)(Draupnir for all) the
+playbook will force restart the bot if running a non release tag like `latest` or `main` or development builds.
+This is due to the conditional restart logic not being able to tell when a update happened correctly.
+
+Conditional restarts work perfectly for all tags when running only one of these two operating modes.
+
 ### Create and invite the bot to the management room
 
 Before proceeding to the next step, run the playbook with the following command to create the bot user.
--- a/roles/custom/matrix-appservice-draupnir-for-all/defaults/main.yml
+++ b/roles/custom/matrix-appservice-draupnir-for-all/defaults/main.yml
@@ -49,11 +49,16 @@ matrix_appservice_draupnir_for_all_systemd_wanted_services_list: []
 # Force restart tag is used to control if the tag that is used is rolling or not.
 # When both Draupnir roles are enabled, we only force restart if they are on the same
 # version string and that version string matches the moving tag.
-matrix_appservice_draupnir_for_all_force_restart_tag: "latest"
+matrix_appservice_draupnir_for_all_force_rolling_tag: "latest"
 
 # Force restart the service on all runs only when both roles are enabled, both roles
 # are using the same version string, and that version is the moving tag.
-matrix_appservice_draupnir_for_all_force_restart: "if {{ matrix_bot_draupnir_enabled | bool and matrix_appservice_draupnir_for_all_enabled | bool and matrix_bot_draupnir_version == matrix_appservice_draupnir_for_all_version and matrix_appservice_draupnir_for_all_version == matrix_appservice_draupnir_for_all_force_restart_tag }} then true else false end"
+matrix_appservice_draupnir_for_all_force_restart: "{{
+  matrix_bot_draupnir_enabled | bool and
+  matrix_appservice_draupnir_for_all_enabled | bool and
+  matrix_bot_draupnir_version == matrix_appservice_draupnir_for_all_version and
+  matrix_appservice_draupnir_for_all_version == matrix_appservice_draupnir_for_all_force_rolling_tag
+}}"
 
 # The room ID where people can use the bot. The bot has no access controls, so
 # anyone in this room can use the bot - secure your room!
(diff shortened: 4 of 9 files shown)


spantaleev:
> **Note:** This comment was drafted by Claude Opus 4.7 with substantial human direction and review from me. Not a fully automated review, but not purely human either. Treat it as a considered starting point.

The substance of this PR is a real bug fix. Conflating it with two unrelated rewrites is what I would push back on.

### Three changes are bundled here

| change | bug fix value | risk it introduces |
| --- | --- | --- |
| `force_restart` logic (replaces the d830c8827 heuristic) | ✅ this is the real fix | medium: current expression is broken Jinja, see below |
| switch to `docker_image_pull` / `docker_image_build` | ⚠️ tangential | low-but-not-zero: requires `community.docker >= 3.6.0` (Jan 2024); cuts off users on the system-bundled `ansible` package from ansible 8 (3.4.6) or ansible 9 (3.4.11) |
| version constraint added to `collections/requirements.yml` | ❌ no value here | almost no one in this project's documented install flow uses that file. `agru` explicitly does not support collections. Our `installing.md`, `quick-start.md`, `justfile` etc. only call `ansible-galaxy install -r requirements.yml`. So bumping a constraint there does not actually gate anything for most users; it only documents what the modules need. The real version requirement lives with whatever `community.docker` shipped alongside the user's `ansible` package |

### On the modernization itself

`community.docker.docker_image` is **not deprecated**. No \`deprecated_features\` entry in the changelog, no removal version announced, no runtime warning. The module survived \`4.0.0\` (Oct 2024) and \`5.0.0\` (Nov 2025), both of which removed other deprecated things. The docs carry a soft "we recommend the new modules" note and that is all.

That said, the new modules do have one nice property: BuildKit support in \`docker_image_build\`. For self-build users this is a real, if modest, win. The \`pull\` semantics are not actually different in a way that fixes #5186: \`docker_image\` with \`force_source: true, source: pull\` already reports \`changed: false\` when the local digest matches the registry. The dual-role bug is a coordination problem (role 2 sees role 1 already pulled), and \`docker_image_pull\` has the exact same problem. Which is why this PR has to introduce \`force_restart\` regardless of which module is used.

So the modernization is fine as a direction, but it does not unlock the bug fix. The bug fix unlocks itself.

### What I would suggest

Split into two or three PRs:

1. **PR for the bug.** Keep using \`community.docker.docker_image\`. Replace \`_force_pull\` with \`_force_restart\` and the rolling-tag logic. This is the substance, lands quickly, no version-floor concerns, no compatibility risk.
2. **(Optional) PR to modernize the module usage**, scoped to whatever roles you want to convert. Done in isolation, the conversation is "do we want to start migrating, and what is our floor for \`community.docker\`". That is a separate question with separate stakeholders.
3. **(Optional) PR to clean up \`collections/requirements.yml\`**, either by removing it (if our flow does not actually install from it), documenting it (if it is meant for some manual/CI use), or pinning it (if we want to make the floor explicit independently of what the modules need).

### Issues in the bug-fix portion that need fixing regardless

Ignoring the modernization, the new logic itself has problems:

1. **The \`force_restart\` expression is invalid Jinja2.**
   \`\`\`yaml
   matrix_..._force_restart: "if {{ ... }} then true else false end"
   \`\`\`
   \`if/then/else/end\` is not Jinja2 syntax. The inner \`{{ ... }}\` evaluates to the string \`"True"\` or \`"False"\`, and the variable ends up as a literal string like \`"if True then true else false end"\`. Then \`| bool\` of that string is \`False\`. Net effect: the new branch never fires. Fix:
   \`\`\`yaml
   matrix_..._force_restart: "{{ matrix_bot_draupnir_enabled | bool and matrix_appservice_draupnir_for_all_enabled | bool and matrix_bot_draupnir_version == matrix_appservice_draupnir_for_all_version and matrix_bot_draupnir_version == matrix_bot_draupnir_force_restart_tag }}"
   \`\`\`

2. **\`force_restart_tag\` defaults to \`latest\` only.** Users on \`:main\` or \`:edge\` hit the same #5186 coordination bug with no fix. Either widen the default (a list, or "treat any tag where both roles match as moving"), or document the limitation.

3. **The condition is too narrow on purpose, but that should be called out.** Old: any role on \`:latest\` always restarts. New: BOTH roles enabled AND same version AND that version equals \`force_restart_tag\`. Single-role users on a rolling tag get no force-restart and rely on \`pull_result.changed\` alone. The PR description argues this is a feature, and it is defensible, but it is a behavior change worth a one-line callout in the validate_config migration notice.

4. **(Only relevant if the modernization stays)** \`docker_image_build\` is gated on \`git_pull_results.changed\`. If the local image is missing (manual prune, fresh host with persisted git checkout, failed previous build) this skips the build entirely and the service will fail to start. The old behavior would have rebuilt. Drop \`git_pull_results.changed\` from the \`when:\` and let the module's own change-detection do its job.

### Naming nit

\`force_restart_tag\` reads ambiguously since "tag" can mean an Ansible tag too. \`_rolling_tag\` would read more clearly. Minor.

FSG-Cat:
Thanks for the detailed feedback. I went through each point and wanted to clarify intent and align on the path forward.

> Disclosure text synthesis assistance was used to compose this but i do aprove of the message and its just a refinement on a fully hand written draft that that i then had to hand perfect after help. So im ofc taking responsibility for the content.

### High-level decisions

- **Force restart behavior**  
  I agree that force-restarting should only be needed when a new image is pulled. The main reason I had it was that our conditional restart logic does not behave well with moving tags.

- **Move to the newer module**  
  I want this change mainly for correctness and performance with moving tags.  
  The current module’s force-pull path is slow and appears to re-pull on every run, which causes overly frequent restarts and significantly slows playbook runs.
  The newer module reports changes based on real upstream image updates, so we get cleaner changed-state signaling and only restart when something actually changed.

- **BuildKit requirement**  
  From a Draupnir perspective, BuildKit is not optional for us in practice. Our CI is BuildKit-based, and that is what we test and validate.  
  So right now, we do not have confidence in non-BuildKit self-build workflows.  
  If the older module is already using BuildKit internally and the docs are just unclear, then that is a fair point.

- **Minimum `community.docker.docker_image` version bump**  
  I consider this acceptable, especially since project policy already recommends using a reasonably recent Ansible version.  
  On the requirements file itself, I do not have any opinion and will follow your guidance.
  [Ansible v9.2.0 changelogs](https://github.com/ansible-community/ansible-build-data/blob/dde3809214cd8b32f85748e48e318bae4500da86/9/CHANGELOG-v9.rst#L5372) say that community.docker_image_pull is first included in this version. So if your ansible is newer than 2024-01-30 it has this submodule.

### Logic bug feedback

- **Force-restart expression**  
  If the expression is wrong, I trust that assessment. It would explain some of the inconsistent behavior I have seen while testing this branch.

- **Force-restart tag defaults**  
  That came from carrying over the previous force-pull logic. With the move to the newer module, force-pulling is no longer needed because moving tags are handled natively. Fixed tags like v3.0.0 are already handled through normal systemd service file change detection.

- **Narrow restart condition**  
  This narrow condition is intentional: restart should key off actual image change, not tag mutability alone.  
  Given the newer module’s change reporting, we can rely on real upstream updates to trigger restarts and avoid unnecessary service churn on runs where nothing changed. This fallback is only needed in the edge case where both modes are running against the same moving tag.

- **Image build logic**  
  Fair criticism. I have much less hands-on experience with image build logic, so I appreciate the correction there.

### Naming feedback

- **Naming nit**  
  Agreed, and thank you for calling it out.

### PR structure

I understand the concern about scope. If preferred, I am happy to split this into two separate PRs (or close this and open two fresh PRs), one per concern area.

FSG-Cat:
So i came back to this PR today and came to a bit of a realisation. The new restart logic is totally useless in my book without the new docker module.

That is because the current restart logic is tied to force pull being true. And we only remove forcepull as a variable by using the new docker module.

As for the build level stuff that is fair to split out in a totally separate PR in theory but i will fix it on this branch to make it easier for my self to split the branches later if that is the final call.

FSG-Cat:
Here is a version matrix of when the  supported distros for mdad got their native support for these new ansible features so ansible from the repo package manager wont be an issue.

Ubuntu gains support for a new enough ansible version in 25.04
Debian is Debian 13.
RHEL Family unclear due to no good sources that just list the ansible package.

Yes this does indeed mean we would have to run quite modern ansible but the playbook has no formal definition of how ancient your ansible is allowed to be before it cant be used and we do recommend having a version that is resonably fresh. And asking for your ansible version to be younger than 2 years old is not too extreme from my perspective. Especially when latest LTS for Ubuntu and Debian have a supported version as their default.

PR #5269, opened 2026-05-23
Update configuration

mcepl:
It seems we have still some unfinished business from #5148

### DIFF
--- a/jitsi_jvb.yml
+++ b/jitsi_jvb.yml
@@ -11,11 +11,11 @@
   become: true
 
   roles:
-    - role: galaxy/com.devture.ansible.role.playbook_help
-    - role: galaxy/com.devture.ansible.role.systemd_docker_base
+    - role: playbook_help
+    - role: systemd_docker_base
 
     - when: matrix_playbook_docker_installation_enabled | bool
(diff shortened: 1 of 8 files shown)


spantaleev:
The `convert_data` problem has been discussed here: https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/5148#issuecomment-4320378183

It may or may not be time to pull the plug on it.

spantaleev:
Thanks for surfacing the jitsi_jvb.yml breakage! Its role references had been stale since the roles were renamed in c0595d6e4 (2023). That part is now fixed in eb6328e5e99f49d42e96375a5c2d08494b78a377, keeping the `galaxy/` prefix, since this playbook installs roles into `roles/galaxy/` via `just roles` and resolves them by that path. Dropping the prefix would break that layout, even if it suits distro-packaged role locations.

As for the `convert_data=False` removals: that would break ansible-core versions older than 2.19 (there the template lookup auto-parses JSON, and the subsequent `from_json` fails on a dict), and we still aim to support older versions. It's worth revisiting when the playbook raises its minimum supported ansible-core, covering all 9 occurrences at once. Closing.

PR #5289, opened 2026-06-02
Add matrix-rustpush-bridge (iMessage)

jasonlaguidice:
This PR adds the [matrix-rustpush-bridge](https://github.com/jasonlaguidice/imessage) to the playbook.

This bridge is a mautrix-go bridgev2 bridge for rustpush - the back end for the OpenBubbles project. This bridge differs in a few key ways from the current mautrix-imessage bridge:

- Communicates directly with Apple's servers - does not need to run on a Mac
- Uses current mautrix-go framework
- Doesn't need wsproxy on the homeserver

This role points to a fork of the project which I have started. The bridge was initiated by [lrhodin](https://github.com/lrhodin) and has been updated by [mackid1993](https://github.com/mackid1993). The fork adds concurrency so multiple people can use the bridge at the same time, as well as other fixes and optimizations for this environment.

It's worth noting this bridge must write files to disk (a few, small XML files) per user. This is temporary 

This bridge is still a work in progress, there are likely bugs. I likely don't have an image ready for _every_ possible deployment scenario but I think this is in a good enough state (supporting x86_64 and arm64 linux deployments) to turn it over to the world.

I did have one weird issue - I could not get ansible to generate the displayname for puppets no matter what I did. I ended up making a [janky patch](https://github.com/jasonlaguidice/matrix-docker-ansible-deploy/blob/1e878d52ac871f87543d20d8af2ad23be72d3099/roles/custom/matrix-bridge-rustpush/tasks/setup_install.yml#L65-L83) for it to be sure it builds it anyway. Could you take a look and see what I might have done wrong? I couldn't figure it out. The LLM's couldn't figure it out. The way it is now works just fine but it's not the same as all the other bridges.


docs/configuring-playbook-bridge-rustpush.md
```diff
@@ -0,0 +1,102 @@
+<!--
+SPDX-FileCopyrightText: 2025 MDAD project contributors
```
luixxiul:
```suggestion
SPDX-FileCopyrightText: 2026 MDAD project contributors
```

docs/configuring-playbook-bridge-rustpush.md
```diff
@@ -0,0 +1,102 @@
...
+
```
luixxiul, line 4:
```suggestion
SPDX-FileCopyrightText: 2026 Jason LaGuidice

```

If you are fine to be mentioned, or course 🙂

jasonlaguidice: Of course, thank you!

I want to chat with the guys who made the bridge before I forked it to split it real quick before committing this as a non-draft PR - there was initially no interest in modifying the bridge to support multiple users thus the fork.

If they can be combined that's better, of course. But I wanted take one last (third) attempt at alignment.

roles/custom/matrix-bridge-rustpush/templates/config.yaml.j2.license
```diff
@@ -0,0 +1,3 @@
+SPDX-FileCopyrightText: 2025 MDAD project contributors
```
jasonlaguidice:
```suggestion
SPDX-FileCopyrightText: 2026 MDAD project contributors
SPDX-FileCopyrightText: 2026 Jason LaGuidice
```

jasonlaguidice:
Also realizing I forgot to add a link to the bridge in the other areas of the docs :) I'll get those in there and also update the attribution date across all the docs... tomorrow.

jasonlaguidice:
Scary - but should be ready for review for any brave souls who might like to help try this bridge.

roles/custom/matrix-bridge-rustpush/templates/config.yaml.j2
```diff
@@ -0,0 +1,208 @@
+#jinja2: lstrip_blocks: True
+# Network-specific config options (iMessage via RustPush)
+network:
+    # Displayname template for iMessage contacts.
+    # Available variables: {% raw %}{{.FirstName}}, {{.LastName}}, {{.Nickname}},
+    # {{.Phone}}, {{.Email}}, {{.ID}}{% endraw %}
```
bblacher: > I did have one weird issue - I could not get ansible to generate the displayname for puppets no matter what I did. I ended up making a [janky patch](https://github.com/jasonlaguidice/matrix-docker-ansible-deploy/blob/1e878d52ac871f87543d20d8af2ad23be72d3099/roles/custom/matrix-bridge-rustpush/tasks/setup_install.yml#L65-L83) for it to be sure it builds it anyway. Could you take a look and see what I might have done wrong? I couldn't figure it out. The LLM's couldn't figure it out. The way it is now works just fine but it's not the same as all the other bridges.

I think this is an issue I've encountered before, try removing these lines (can't test right now):

```suggestion
```

 This seems to be a bug (?) in ansible, somehow it seems to parse the `{% raw %}` from the comments.

bblacher (changes requested):
Maybe this fixes you issue.

roles/custom/matrix-bridge-rustpush/templates/config.yaml.j2
jasonlaguidice: I'll get it in and test!

spantaleev:
Thanks for working on this — happy to get a rustpush-based iMessage bridge into the playbook. The project being new and still having bugs is not a blocker for me. There are a few things to sort out first, though.

**1. Namespace overlap with the existing iMessage bridge (blocker)**

The defaults collide with the iMessage appservice already shipped by the `matrix-bridge-mautrix-wsproxy` role:

- bot username `imessagebot` — same as `matrix_mautrix_imessage_appservice_bot_username`
- puppet localpart `imessage_{{.}}` → namespace `@imessage_.+` — same
- `sender_localpart: _bot_imessagebot` — same

Both roles register these as **exclusive** namespaces, so a homeserver will reject the registrations if a user enables both. The override variables in the docs help, but I'd rather not rely on users knowing to set them — playbook components should be unique by default so the conflict can't arise.

Please change the defaults to a distinct namespace that doesn't overlap with the wsproxy iMessage bridge — e.g. a `rustpush`-based bot username and puppet localpart prefix. The `id: rustpush-bridge` registration already differs, which is good; the bot/puppet/sender identifiers need to follow.

**2. Modernize to current role conventions**

The role is modeled on an older template. Please align it with the current conventions — your own `matrix-bridge-steam` role is a good reference, since it's structured the same way and already does all of this:

- Rename `matrix_rustpush_bridge_docker_image*` / `..._docker_src_files_path` to `..._container_image*` / `..._container_src_files_path` (playbook-wide rename in `f0e0f5ec5`), and drop `..._docker_image_force_pull`.
- Use `community.docker.docker_image_pull` and `community.docker.docker_image_build` instead of `community.docker.docker_image` with `source:`/`force_source:`. The module switch and `force_pull` removal for the steam bridge are in `d9bb4cbdc` — same change applies here.
- Add the pull retry loop (`retries: {{ devture_playbook_help_container_retries_count }}`, `delay: {{ devture_playbook_help_container_retries_delay }}`, `until: ... is not failed`).
- For self-builds, switch the registry prefix to `localhost/` via the `..._container_image_registry_prefix` / `..._upstream` / `..._upstream_default` layering, rather than hardcoding `ghcr.io/`.

**3. Docs polish (minor)**

`docs/configuring-playbook-bridge-rustpush.md` uses raw HTML (`<hr>`, `<sup>`) and `*** ... ***` / `~~dragons~~`. Please match the plain-Markdown structure of the other bridge docs.

jasonlaguidice:
Huh! So weird that the docs ended up with HTML, especially because I hand-wrote that though I think I just used the Github direct editor to do it. I can't remember now.

Thanks so much for the pass through this! I probably need to reduce my project workload so I can spend a little more time on each one. I've been taking most of my time on the bridge side since it's in such a fragile state and I want to make it better. Testing is slow for me because I don't have a TON of iMessage contacts so the cycles take a little while.

Anyway I'll test this later this evening on my server - I need to migrate my bot user and localpart over. I'll ping you in the matrix channel once I've confirmed it's ready!

jasonlaguidice:
@spantaleev tested for the last couple of days as deployed on my homeserver and everything seems to be working well!

It's ready for another look.

docs/configuring-playbook-bridge-rustpush.md
```diff
@@ -0,0 +1,95 @@
...
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+# Setting up RustPush (iMessage) bridging (optional)
+
+> **Note:** This bridge is in early development and may have stability issues. It may not be desirable to deploy this to a large number of users. Your testing and feedback is appreciated.
+
+<sup>Refer the common guide for configuring mautrix bridges: [Setting up a Generic Mautrix Bridge](configuring-playbook-bridge-mautrix-bridges.md)</sup>
+
+The playbook can install and configure [rustpush bridge to iMessage](https://github.com/jasonlaguidice/imessage) for you using Apple's push notification service.
```
luixxiul:
```suggestion
The playbook can install and configure [RustPush bridge to iMessage](https://github.com/jasonlaguidice/imessage) for you using Apple's push notification service.
```

spantaleev:
This is now merged, and it's also announced on the changelog (via [this entry](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/CHANGELOG.md#support-for-bridging-to-imessage-via-rustpush)). Thanks again for the work on this!

A few follow-up notes:

**I removed the playbook's copy of your Dockerfile.** The role previously shipped its own `Dockerfile.j2` and templated it over the cloned source on self-build. I've dropped it, and the role now builds from your repo's own `Dockerfile`, like every other self-build role in the playbook (e.g. `matrix-bridge-steam`). This keeps the build recipe in one place (yours) instead of a copy here that drifts out of sync.

**Please consider tagging a new release.** The playbook pins `matrix_rustpush_bridge_version: v0.0.1`, so that tag's `Dockerfile` is what gets cloned and built for self-builders (and its prebuilt image is what the default path pulls). The v0.0.1 Dockerfile predates the HEIC conversion fix you added later on `master` ([ef214536](https://github.com/jasonlaguidice/imessage/commit/ef214536ed6cba92eb8c0842ac630332ffd17066), "Add correct libheif to docker image for HEIC conversion"), along with [other unreleased Dockerfile changes](https://github.com/jasonlaguidice/imessage/compare/v0.0.1...master) (11 commits ahead). Since HEIC is iMessage's main image format, both self-builders and users of the `v0.0.1` image currently lack correct HEIC decoding. If you tag a `v0.0.2`, the playbook can bump the pin and pick all of that up automatically.

**For reference, I'm closing a Renovate PR.** Renovate had opened #5358 to bump the playbook's copied Dockerfile base image to `ubuntu:26.04`. Now that the copy is gone, it's moot, so I've closed it. The base image is whatever your Dockerfile specifies.

jasonlaguidice:
Perfect thank you! Yeah I need to bump the version, I've been testing and am cautious because of the size and scope of the bridge (< 10k lines in client.go 😵‍💫) which I'm trying to pare back and simplify.

Ancillary but I plan to do monthly roll-ups from the main repo as well as implement my own fixes over time.

I'll double check my testing today on the bridge and bump the version!

PR #5315, opened 2026-06-13
Add Synology DSM support

cksit:
## What this does

Adds support for running the playbook on Synology DSM 7+.
Detection is automatic via `/etc/synoinfo.conf` — non-Synology hosts
are completely unaffected.

### User management

`setup_matrix_user.yml` is refactored to branch on platform:

- **`setup_matrix_user_synology.yml`** — uses `synouser`/`synogroup`
  (DSM-native CLI tools). `ansible.builtin.user` and `group` do not work
  correctly on Synology because DSM manages users outside of the standard
  Linux user database.
- **`setup_matrix_user_linux.yml`** — preserves the original behaviour for
  all other platforms.

### Synology prerequisites

- Pins `requests==2.31.0` on the remote host. Synology ships a newer version
  that breaks the Docker SDK with `Not supported URL scheme http+docker`.

### Boot-fix service (`matrix-synology-boot-fix`)

A `Type=oneshot` systemd service (`WantedBy=multi-user.target`,
`After=multi-user.target`) deployed to address two Synology boot-time quirks:

**1. `/volume1` shared mount propagation**

`bind-propagation=slave` mounts (used by matrix-synapse for its media store)
require `mount --make-shared /volume1` to run before the container is created.
Placing a custom service `Before=pkg-ContainerManager-dockerd.service` was
attempted but causes Container Manager to report a broken dependency chain and
prompt for repair on every boot, regardless of whether the custom service
succeeds or fails. Running it after Docker is already up (via this boot-fix
service) avoids that problem entirely.

**2. Services skipped at boot**

Synology's systemd drops services with multi-level dependency chains from the
boot-time activation queue (e.g. `matrix-traefik → matrix-container-socket-proxy
→ docker`). Affected services show as `inactive (dead)` after reboot despite
being enabled. The boot-fix script dynamically finds any enabled
`matrix-*.service` that is still inactive after boot and starts them, so no
hardcoded service list is needed.

### Documentation

- `docs/configuring-playbook-synology.md` — setup guide covering DSM GUI
  preparation, SSH, inventory, and `vars.yml` configuration
- `docs/README.md` — link added under "Other documentation pages"

## Tested on

- Synology DSM 7.2 (Virtual DSM)
- matrix-synapse v1.154.0
- Container Manager (Docker 20.10.x)

## Notes

All Synology-specific tasks are gated behind
`when: matrix_synoinfo_conf_stat.stat.exists | bool`.


cksit:
The CI failure appears to be a pre-existing infrastructure issue — mise is unable to fetch prek@0.4.4 from GitHub (403 Forbidden), unrelated to this PR's changes.

docs/configuring-playbook-synology.md
```diff
@@ -0,0 +1,162 @@
...
+
+**1. `/volume1` shared mount propagation**
+
+Docker requires `/volume1` to be mounted as shared (`mount --make-shared /volume1`) for container bind mounts with `bind-propagation=slave` to work correctly (used by matrix-synapse for its media store). On Synology, this cannot be inserted into the systemd chain before Container Manager starts — doing so causes Container Manager to detect a broken dependency and prompt for repair on every boot. The boot-fix service runs this command after Docker is already up, safely outside Container Manager's dependency chain.
+
+**2. Skipped services at boot**
+
+Synology's systemd drops services with multi-level dependency chains from the boot activation queue (e.g. `matrix-traefik → matrix-container-socket-proxy → docker`). These services show as `inactive (dead)` after reboot even though they are enabled. The boot-fix service scans for any enabled `matrix-*.service` that is still inactive after boot and starts them automatically.
+
+> **If you previously configured a Task Scheduler entry** (`Control Panel > Task Scheduler`) to run `mount --make-shared /volume1` at boot-up, you can remove it — the boot-fix service now handles this.
```
luixxiul:
```suggestion
1. `/volume1` shared mount propagation

    Docker requires `/volume1` to be mounted as shared (`mount --make-shared /volume1`) for container bind mounts with `bind-propagation=slave` to work correctly (used by matrix-synapse for its media store). On Synology, this cannot be inserted into the systemd chain before Container Manager starts — doing so causes Container Manager to detect a broken dependency and prompt for repair on every boot. The boot-fix service runs this command after Docker is already up, safely outside Container Manager's dependency chain.

2. Skipped services at boot

    Synology's systemd drops services with multi-level dependency chains from the boot activation queue (e.g. `matrix-traefik → matrix-container-socket-proxy → docker`). These services show as `inactive (dead)` after reboot even though they are enabled. The boot-fix service scans for any enabled `matrix-*.service` that is still inactive after boot and starts them automatically.

    > **If you previously configured a Task Scheduler entry** (`Control Panel > Task Scheduler`) to run `mount --make-shared /volume1` at boot-up, you can remove it — the boot-fix service now handles this.
```

spantaleev:
Thanks for this, and for the thorough write-up and the dedicated docs page. The detection approach is sound (Ansible reports `OtherLinux` on DSM, so keying off `/etc/synoinfo.conf` is the right call), and I appreciate that non-Synology hosts are left untouched. A few things I'd like reworked before this can go in.

**1. Give the detection a clean home and a named variable**

Right now the probe is registered as a side effect inside `setup_matrix_user.yml`, and the raw `matrix_synoinfo_conf_stat.stat.exists` is then reached into from `main.yml` and `setup_matrix_base.yml`. I'd prefer detection to happen once, in its own task file, and expose a single boolean that everything else consumes.

Could you add `roles/custom/matrix-base/tasks/detect_platform.yml`:

```yaml
- name: Detect Synology DSM
  ansible.builtin.stat:
    path: /etc/synoinfo.conf
  register: matrix_base_synoinfo_conf_stat
  when: matrix_base_host_is_synology is none

- name: Set matrix_base_host_is_synology from detection
  ansible.builtin.set_fact:
    matrix_base_host_is_synology: "{{ matrix_base_synoinfo_conf_stat.stat.exists }}"
  when: matrix_base_host_is_synology is none
```

with a default in `defaults/main.yml`:

```yaml
# Controls Synology DSM-specific handling. `null` means autodetect (via /etc/synoinfo.conf).
# Set to `true`/`false` to force.
matrix_base_host_is_synology: ~
```

and include it first in `main.yml` under the `always` tag, before the user-setup block (it needs to be `always`, because `setup_matrix_user.yml` is `always` and consumes it):

```yaml
- tags:
    - always
  block:
    - ansible.builtin.include_tasks: "{{ role_path }}/tasks/detect_platform.yml"
```

Then every other gate becomes `when: matrix_base_host_is_synology` instead of `when: matrix_synoinfo_conf_stat.stat.exists | bool`. This also makes the behavior overridable for anyone on an unusual host.

**2. The recursive chown needs justification (and ideally narrowing)**

`setup_matrix_base.yml` already sets ownership of `matrix_base_data_path` (by name, non-recursive). The new task re-chowns the same path recursively with numeric uid/gid, on every run.

Two questions:
- Did name-based ownership (`owner: matrix_user_name`) fail to apply on DSM? If so, it would be cleaner to make the *existing* "Ensure Matrix base paths exists" task use the numeric uid/gid on Synology, rather than adding a second pass.
- Is the `recurse: true` over the whole tree actually required (e.g. DSM re-owning files under the `/volume1` share), or was it belt-and-suspenders? Recursing the entire data path (which includes the Synapse media store) on every run can get very slow as the store grows. If it's needed, can it be scoped or made conditional rather than unconditional every run?

**3. The `requests` pin**

`requests==2.31.0` is a hard pin to an old version, installed at runtime via `ansible.builtin.pip` (a pattern we don't use anywhere else in the playbook). I understand the `http+docker` breakage with requests >= 2.32. Could this be a constraint (e.g. `requests<2.32`) rather than an exact pin, and exposed as a variable so it can be bumped without a code change? Also worth a one-line note on which interpreter it lands in.

**4. Minor**

- GID is fetched with an inline `python3 -c "import grp..."` while UID uses `id -u`. Could the GID use `getent group` (or `synogroup --get` parsing) for consistency and to drop the Python dependency?
- `synouser --add` creates the user with an empty password. Is that intentional / does DSM password policy ever reject it? A comment would help.
- The boot-fix only restarts units in state `inactive`, not `failed`. Intentional?

None of this is structural doubt about the feature, it's well isolated and the docs are great. Mainly I'd like the detection tidied up as above, and the chown / requests-pin decisions explained so we're confident they're right rather than incidental. Happy to help with any of it.

.gitignore
```diff
@@ -9,3 +9,5 @@
 # ignore roles pulled by ansible-galaxy
 /roles/galaxy/*
 !/roles/galaxy/.gitkeep
+myenv/
+bin/just
```
luixxiul: May I ask why they are necessary with this PR?

ansible.cfg
```diff
@@ -4,3 +4,6 @@ result_format = yaml
 
 [connection]
 pipelining = True
+
+[ssh_connection]
+transfer_method = piped
```
luixxiul: Ditto.

.gitignore
cksit: Sorry. Executed the wrong command and two local changes were pushed accidentally.

cksit:
@spantaleev Thanks again for the detailed review. All points have been addressed in the latest push — going through them one by one:

**1. Detection cleanup** - Done exactly as suggested. 

**2. Recursive chown**

To answer the two questions directly:

- _Did name-based ownership fail?_ No — `owner: matrix_user_name` works correctly for directly-touched paths on DSM. The issue is more specific: Synology assigns UIDs from its own internal counter via `synouser`. If the `matrix` user is ever deleted and recreated (e.g. after a DSM reinstall or reset), Synology assigns a new UID, leaving all existing files under `/volume1` with stale numeric ownership that the non-recursive name-based chown on top-level directories doesn't reach. 
    
- _Is `recurse: true` required on every run?_ It was — but no longer. We now stat `matrix_base_data_path` before the "Ensure Matrix base paths exists" task and only recurse when the pre-task UID or GID doesn't match the expected values. Normal runs skip the tree walk entirely. Additionally, we switched from Ansible's `file` module recurse (Python walking each entry) to `chown -R` via `ansible.builtin.command`, which is considerably faster on large trees.
    

**3. `requests` pin**

Changed to a `requests<2.32` constraint, exposed as `matrix_base_synology_requests_version_constraint` (so it can be bumped in `vars.yml` without touching role code), and the default comment notes it installs into `ansible_python_interpreter` on the remote host.

**4. Minor items**

- _GID lookup_: Replaced the Python one-liner with `synogroup --get <name>` parsed via Ansible's `regex_search` filter to extract the `[GID]` field. Note: `getent` is not available on Synology DSM, so that path wasn't viable.
    
- _Empty password_: This turned out to be a real issue — testing confirmed that `synouser --add` with an empty password allows login to the DSM web UI. Rather than just adding a comment, we now require `matrix_synology_user_password` to be set to a non-empty value (enforced with a `fail` task that gives a clear error message) and create the account with `expired=1`, which prevents login regardless of the password. If the user pre-creates the `matrix` account manually, the playbook leaves its settings untouched.
    
- _Boot-fix `failed` units_: Now handled. The script restarts matrix-*.service units in both `inactive` and `failed` states. Both indicate the service didn't come up at boot — `failed` typically means it attempted to start but Docker or the mount propagation wasn't ready yet, which is exactly what the boot-fix resolves before restarting them.

PR #5330, opened 2026-06-17
Gate c10y recaptcha config on it being configured.

### LINKED ISSUE #5329: Continuwuity role incorrectly sets invalid recaptcha values
**Describe the bug**
The Continuwuity role sets `matrix_continuwuity_config_recaptcha_site_key` and `matrix_continuwuity_config_recaptcha_private_site_key` to empty strings, but this causes Continuwuity to see the fields as present and enables recaptcha registration, but with an empty private key.  Continuwuity [only checks for the presence of this private key](https://forgejo.ellis.link/continuwuation/continuwuity/src/commit/0514491/src/api/client/account/register.rs#L440-L452) before enabling ReCAPTCHA registration flows, which causes the ReCAPTCHA flow to be enabled with an empty key value. Clients (Cinny at least, not Element Web) then try to fill in this captcha during registration, but fail due to that empty key.

**To Reproduce**

(I do not have a deployment myself, I am forwarding this bug report)

Do not set `matrix_continuwuity_config_recaptcha_private_site_key` and `matrix_continuwuity_config_recaptcha_private_site_key`

**Expected behavior**
The empty strings should not be present at all, as this triggers the bug - the config option should just be omitted entirely.

**Matrix Server:**
unknown

**Ansible:**
N/A

**Additional context**
I'm forwarding this report on behalf of a user who was encountering issues with registration that joined our support room (I have never actually used MDAD before, sorry if I'm using the wrong terminology). The offending lines are here:

https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/50408d699f914a9eace2fab8d2e66b585f74868b/roles/custom/matrix-continuwuity/defaults/main.yml#L193-L195

FSG-Cat:
Fixes: #5329 

Edit: Confidence is mid because i know i have a history of managing to screw up jinja expressions.

### DIFF
--- a/roles/custom/matrix-continuwuity/defaults/main.yml
+++ b/roles/custom/matrix-continuwuity/defaults/main.yml
@@ -1,5 +1,6 @@
 # SPDX-FileCopyrightText: 2025 MDAD project contributors
 # SPDX-FileCopyrightText: 2025 Slavi Pantaleev
+# SPDX-FileCopyrightText: 2026 Catalan Lover <catalanlover@protonmail.com>
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
@@ -191,6 +192,7 @@ matrix_continuwuity_config_turn_password: ''
 matrix_continuwuity_self_check_validate_certificates: true
 
 # If set, registration will require Google ReCAPTCHA verification.
+matrix_continuwuity_recaptcha_enabled: "{{ matrix_continuwuity_config_recaptcha_site_key != '' and matrix_continuwuity_config_recaptcha_private_site_key != '' }}"
 matrix_continuwuity_config_recaptcha_site_key: ''
 matrix_continuwuity_config_recaptcha_private_site_key: ''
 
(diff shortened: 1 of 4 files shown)


spantaleev:
Deriving an `_enabled` value in `defaults/main.yml` isn't how we do it. That file is the user-facing override surface, so computed state doesn't belong there (it belongs to `vars/main.yml` instead). Reworked before merging:

- Moved `matrix_continuwuity_recaptcha_enabled` into `vars/main.yml` as internal derived state (true only when both keys are set). This matches how Continuwuity itself works: no native toggle, presence of keys enables the flow.
- Replaced the partial-config `fail` task with a single consistency check that fails only when exactly one of the two keys is set.
- Template gating and the empty-key fix are unchanged.

Merged now. Thanks for the fix.

FSG-Cat:
Thanks for the cleanup. I wrote this as a quick and dirty PR honestly to help c10y out with this specific problem and it clearly was a bit too rushed.

PR #5346, opened 2026-06-21
Support additional container arguments for matrix-livekit-jwt

holliefae:
Hit an issue when testing Element Call with self-signed certificates where livekit wouldn't be able to reach the Matrix API due to untrusted certificates. This patch allows configuring JWT to accept additional container arguments which allowed the pass through of certificates (among other uses).

```
matrix_livekit_jwt_service_container_additional_arguments:
  - "--volume=<cert_path>/cert.pem:/ca.crt:ro"
```


spantaleev:
Thank you! In `4f00ad9bd`, I also added support for additional volumes for the livekit-jwt-service component. You can use it like this:

```yaml
matrix_livekit_jwt_service_container_additional_volumes_custom:
  - type: bind
    src: /path/on/host/cert.pem
    dst: /ca.crt
    options: readonly
```

This uses `--mount` under the hood.

PR #5363, opened 2026-06-25
matrix-rustpush-brigdge: Add renovate and bump version

jasonlaguidice:
I totally forgot to replace the commented line with one for renovate to notify when new versions were available. Apologies!

🙃


bblacher:
@jasonlaguidice Can't open a pr right now, but installing the bridge from the playbook currently fails because your tags are 0.0.1 and 0.0.2 rather than v0.0.1 and v0.0.2.

--> Your tags don't match what's specified here.

jasonlaguidice:
> @jasonlaguidice Can't open a pr right now, but installing the bridge from the playbook currently fails because your tags are 0.0.1 and 0.0.2 rather than v0.0.1 and v0.0.2.
> 
> --> Your tags don't match what's specified here.

Ugh sorry. I test with `latest` and got ahead of myself there. Fixed it up, there is now a [proper tag for this](https://github.com/jasonlaguidice/imessage/pkgs/container/imessage/978565692?tag=v0.0.2). Thanks for the heads-up.

Edit: Or so I thought I fixed it! Working still

Edit 2: Had to bump the tag commit as well, now formally all set.

PR #5456, opened 2026-07-22
Fix Synapse crash on empty experimental_features (rendered as null)

### LINKED ISSUE #5455: Bug: Synapse fails to start with `experimental_features: null` when no experimental features are enabled
### Bug: Synapse fails to start with `experimental_features: null` when no experimental features are enabled (Synapse ≥ 1.157)

**Describe the bug**

On a deployment that enables neither Matrix RTC (Element Call) nor Matrix
Authentication Service, upgrading Synapse to v1.157.0 makes it crash-loop on
startup:

File ".../synapse/config/server.py", line 971, in read_config
).get("msc4140_max_delayed_events_per_user", 100)
AttributeError: 'NoneType' object has no attribute 'get'

**Root cause**

`group_vars/matrix_servers` ties every experimental-feature toggle to services
that a minimal deployment doesn't run:

```yaml
matrix_synapse_experimental_features_msc3266_enabled: "{{ matrix_rtc_enabled }}"
matrix_synapse_experimental_features_msc4108_enabled: "{{ matrix_authentication_service_enabled and not matrix_authentication_service_migration_in_progress }}"
matrix_synapse_experimental_features_msc4140_enabled: "{{ matrix_rtc_enabled }}"
matrix_synapse_experimental_features_msc4143_enabled: "{{ matrix_rtc_enabled }}"
matrix_synapse_experimental_features_msc4222_enabled: "{{ matrix_rtc_enabled }}"
With matrix_rtc_enabled: false and MAS disabled, all toggles are false, so the
experimental_features: block in the generated homeserver.yaml has no children
and is written as:


experimental_features: null
Synapse ≥ 1.157 does config.get("experimental_features", {}).get("msc4140_max_delayed_events_per_user", 100).
Because the key is present but null (not absent), the {} default is not used,
.get() is called on None, and Synapse crashes. Synapse 1.154 tolerated it.

To reproduce

Deployment with matrix_rtc_enabled: false, matrix_authentication_service_enabled: false, and no experimental feature explicitly enabled.
Upgrade Synapse to v1.157.0 and restart it.
Expected behavior
The generated config should never contain experimental_features: null. It should
render experimental_features: {} (or omit the block) when nothing is enabled.

Workaround
Set in host_vars:


matrix_synapse_experimental_features_msc3266_enabled: true
Environment

matrix-docker-ansible-deploy: master (as of 2026-07-21, commit 37b1a36)
Synapse: v1.157.0
Matrix RTC / Element Call: disabled
Matrix Authentication Service: disabled
PostgreSQL 18, Docker on Linux

retroFWN:
Fix Synapse crash on empty experimental_features (rendered as null) when no experimental features are enabled

Fixes #5455

### DIFF
--- a/roles/custom/matrix-synapse/templates/synapse/homeserver.yaml.j2
+++ b/roles/custom/matrix-synapse/templates/synapse/homeserver.yaml.j2
@@ -3003,7 +3003,19 @@ matrix_authentication_service:
   secret: {{ matrix_synapse_matrix_authentication_service_secret | to_json }}
 {% endif %}
 
-experimental_features:
+{% set experimental_features_enabled =
+       matrix_synapse_experimental_features_msc2409_to_device_messages_enabled
+    or matrix_synapse_experimental_features_msc3202_transaction_extensions_enabled
+    or matrix_synapse_experimental_features_msc3266_enabled
+    or matrix_synapse_experimental_features_msc4108_enabled
+    or matrix_synapse_experimental_features_msc4140_enabled
+    or matrix_synapse_experimental_features_msc4143_enabled
+    or matrix_synapse_experimental_features_msc4222_enabled
+    or matrix_synapse_experimental_features_msc4306_enabled
+    or matrix_synapse_experimental_features_msc4354_enabled
+    or matrix_synapse_experimental_features_msc4429_enabled
+%}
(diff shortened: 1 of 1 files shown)


spantaleev:
Superseded by https://github.com/spantaleev/matrix-docker-ansible-deploy/commit/6b0a635454b90f5be3581d0c9097d54cc284112d

PR #5463, opened 2026-07-24
Update changelog version

fcnjd:
Changelog version is still on month May, effectively bypassing this feature. With this PR I'd like to update it to the current date from CHANGELOG.md.

### DIFF
--- a/examples/vars.yml
+++ b/examples/vars.yml
@@ -2,7 +2,7 @@
 # This variable acknowledges that you've reviewed breaking changes up to this version.
 # The playbook will fail if this is outdated, guiding you through what changed.
 # See the changelog: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/CHANGELOG.md
-matrix_playbook_migration_validated_version: v2026.05.18.0
+matrix_playbook_migration_validated_version: v2026.07.19.0
 
 # The bare domain name which represents your Matrix identity.
 # Matrix user IDs for your server will be of the form (`@alice:example.com`).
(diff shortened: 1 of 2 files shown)


spantaleev:
Bypassing which feature? Not every feature necessitates a bump to `matrix_playbook_migration_validated_version` and `matrix_playbook_migration_expected_version`.

The last breaking change was done on `v2026.05.18.0` and is about what's described here: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/CHANGELOG.md#2026-05-18

.. so that's why these variables are set to that version.

-----

Changelog entries after that are not major "silent" breakage that needs to be acknowledged by users, so they don't bump `matrix_playbook_migration_validated_version` and `matrix_playbook_migration_expected_version`.

luixxiul:
Not this one? https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/CHANGELOG.md#backward-compatibility-break-bridge-variables-have-been-renamed

spantaleev:
The playbook catches misnamed variables and reports this automatically, so we don't necessarily need to bump `matrix_playbook_migration_expected_version` for it.

We mostly (only?) bump this when there's a backward compatibility change which may silently slip through and you won't even know about it. The `matrix_playbook_migration_expected_version` and `matrix_playbook_migration_validated_version` mechanism is there to notify you, so you could potentially take action, etc.

For https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/CHANGELOG.md#backward-compatibility-break-bridge-variables-have-been-renamed , the playbook will [run validation tasks](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/45a0128b971ee0143284a770a114f74613d9ed5c/roles/custom/matrix_playbook_migration/tasks/validate_config.yml#L927-L946) and notice your variables are outdated and tell you what to do in any case.

fcnjd:
Ah, I see. thank you for clarifying. My understanding was that I could go into CHANGELOG.md, find the latest date, and set that as version. Needless to say, it makes more sense how it is.

PR #5513, opened 2026-08-12
feat: public client root redirection URL from enabled client hostnames

luneth:
We used to redirect matrix fnq > element fnq if element-web installed. Do the same if any other client is installed, default to none if multiple are installed.

### DIFF
--- a/group_vars/matrix_servers
+++ b/group_vars/matrix_servers
@@ -4924,7 +4924,36 @@ matrix_synapse_container_labels_traefik_compression_middleware_name: "{{ matrix_
 matrix_synapse_container_labels_matrix_labels_enabled: "{{ not matrix_synapse_workers_enabled }}"
 
 matrix_synapse_container_labels_public_client_root_redirection_enabled: "{{ matrix_synapse_container_labels_public_client_root_redirection_url != '' }}"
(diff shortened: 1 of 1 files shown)


spantaleev:
Thanks! This is a good idea.

I've implemented a variation of it in 5e2a2f804. The main difference is structural: the client list is built once into a `matrix_playbook_public_client_urls` variable, which all 5 homeserver implementations then share. The expression was duplicated across Synapse, Dendrite, Conduit, Continuwuity and Tuwunel, so patching only the Synapse copy would have left the other four redirecting to Element Web alone.

Two behavioural differences from your version. When several clients are enabled it redirects to the first one (Element Web is first, so existing setups keep their current target) instead of dropping the redirect — otherwise anyone running Element Web alongside a second client would silently lose it. And the URL is built from each client's own `_scheme`, `_hostname` and `_path_prefix`, so clients served under a subpath get linked correctly.

One thing to watch out for in the original: the scheme was prepended outside the conditional, so with zero or multiple clients the result was `https://` — non-empty, which keeps `..._redirection_enabled` true and redirects to a bare scheme.

Closing in favour of the above. Thanks again!

PR #5555, opened 2026-08-20
Add the restrict_room_creation module from intevation

pt1997:
I did not write the modules, but I tested the restrict_room_creation one and find it useful and would like to see it included :) https://heptapod.host/intevation/fac/synapse-modules/-/tree/master

### DIFF
--- a/roles/custom/matrix-synapse/defaults/main.yml
+++ b/roles/custom/matrix-synapse/defaults/main.yml
@@ -1636,6 +1636,10 @@ matrix_synapse_ext_spam_checker_mjolnir_antispam_config:
   ban_lists: "{{ matrix_synapse_ext_spam_checker_mjolnir_antispam_config_ban_lists }}"
   message_max_length: "{{ matrix_synapse_ext_spam_checker_mjolnir_antispam_config_message_max_length }}"
 
+# Enable this to activate the synapse-restrict-room-creation module.
+matrix_synapse_ext_synapse_restrict_room_creation_enabled: false
+matrix_synapse_ext_synapse_restrict_room_creation_git_repository_url: "https://heptapod.host/intevation/fac/synapse-modules.git"
+
 # Enable this to activate the synapse-http-antispam module.
 # See: github.com/maunium/synapse-http-antispam
 matrix_synapse_ext_synapse_http_antispam_enabled: false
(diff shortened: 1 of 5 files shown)


spantaleev:
Thanks for contributing this. The functionality looks useful, and your report that `RestrictRoomCreation` works is encouraging. Still, I think we should not merge this as-is.

The upstream project’s latest commit is from June 2024, while its last functional code change appears to be from April 2023. It may not be abandoned, but it has not kept up with current Synapse. This PR also enables `RestrictJoinRules`, which has several concrete problems:

- It uses private Synapse APIs instead of the public `is_user_admin()` and `get_room_state()` methods.
- Current Synapse expects a string in the internal `is_server_admin()` call, but the module passes a `UserID`.
- `has_join_rule` may be referenced before assignment.
- The callback also processes federated events, so join-rule changes from remote room administrators may be rejected because they are not administrators of the local homeserver.

Could you clarify whether you tested both classes, and with which Synapse version? Supporting only the tested `RestrictRoomCreation` class initially may be safer.

Other changes needed:

- Pin the repository to an exact commit instead of following `master`, because otherwise a future playbook run could silently download and execute changed third-party Python code inside Synapse. Renovate can track the branch commit later through its `git-refs` datasource.
- Follow the way we do Git repository checkouts for other Synapse extensions: correct existing repository ownership, specify `version:`, and use `force: "yes"`.
- Make `error_message_top_rooms` configurable instead of hard-coding the “in Home” message.
- Add a short documentation page, similar to the Synapse Simple Antispam documentation.
- Document that `is_direct` and `m.space.parent` come from the creation request, so this is not a strong security boundary.

It would also be helpful to report the compatibility problems upstream—the original author may simply not know that Synapse has changed. We are open to the feature, but need a pinned implementation that does not enable the currently problematic join-rules code.

pt1997:
> Could you clarify whether you tested both classes, and with which Synapse version? Supporting only the tested RestrictRoomCreation class initially may be safer.

I've tested the functionality provided by both classes on Synapse v1.159.0

RestrictRoomCreation works as expected, but the error message is not shown in Element, in Sable it is shown.
RestrictJoinRules also blocks everyone including Admins from changing the JoinRules / Access

I’ll try to address the other points as soon as I find the time :)

PR #5574, opened 2026-08-26
Add support for gomuks

mcepl:
Add [gomuks](https://github.com/gomuks/gomuks) ([docs](https://docs.mau.fi/gomuks/)) as an optional alternative web client.

gomuks is a bouncer-style Matrix client — Go backend that stores encryption keys and keeps syncing while the browser is closed + web frontend. Image: `dock.mau.dev/gomuks/gomuks:v0.2608.0`.

This is disabled by default (`matrix_client_gomuks_enabled: false`), so Element remains the default and the change is safe for existing installations.

### What's included

New role `matrix-client-gomuks` derived from `~/.config/containers/systemd/gomuks.container` and adapted to playbook conventions:

- single `/data` bind mount (`GOMUKS_ROOT=/data`), listens on `0.0.0.0:29325`
- Traefik labels, systemd service `matrix-client-gomuks.service` (`--cap-drop=ALL --read-only --tmpfs=/tmp`), `labels`/`env` support files, Docker network handling
- `config.yaml` templated from `matrix_homeserver_generic_secret_key`: bcrypt `password_hash` (with deterministic salt), `token_key`, `origin_patterns`, `logging` (stdout + `/data/logs/gomuks.log`), `matrix.*`/`media.*`/`push.*`
- VAPID: leave `matrix_client_gomuks_push_vapid_private_key` / `_public_key` empty to let gomuks generate on first start — role preserves the pair on later runs. Supplying a custom pair requires both keys and validates `^[A-Za-z0-9_-]{43}$` / `^B[A-Za-z0-9_-]{86}$`
- Auth: `matrix_client_gomuks_auth_username` (default `gomuks`) + `matrix_client_gomuks_auth_password` (required unless `matrix_client_gomuks_disable_auth: true` — then an inaccessible fallback password is used). `validate_config.yml` enforces this.

Hardening follow-up:

- disabling preserves `{{ matrix_client_gomuks_data_path }}` (stops/removes service only, no data wipe)
- pin container image, `0640` for `config.yaml`/`env`, `0750` for data dirs, correct `validate_config`/`self_check`/`setup_uninstall` behavior, SPDX headers, docs/changelog alignment

Wiring:

- `setup.yml`, `group_vars/matrix_servers` (`matrix_server_fqn_gomuks`, `matrix_playbook_public_client_urls`, `matrix_static_files` wiring)
- `roles/custom/matrix-base/defaults/main.yml`
- `docs/configuring-playbook-client-gomuks.md` (DNS, `vars.yml` example, `hostname`/`path_prefix`, `disable_auth`, VAPID, troubleshooting) + linked in `docs/README.md`
- `README.md` + `docs/container-images.md` + `CHANGELOG.md`

### Configuration

```yaml
matrix_client_gomuks_enabled: true

# protects the bouncer backend, separate from your Matrix password
matrix_client_gomuks_auth_password: "YOUR_STRONG_PASSWORD_HERE"

# optional
# matrix_client_gomuks_hostname: "{{ matrix_server_fqn_matrix }}"
# matrix_client_gomuks_path_prefix: /gomuks
# matrix_client_gomuks_disable_auth: true  # if behind authenticating reverse proxy
```

Requires DNS `CNAME` for `gomuks.example.com` → `matrix.example.com` (unless reusing `matrix.` domain). Available at `https://gomuks.example.com`.

### Testing

- `just install-all` / `ansible-playbook -i inventory/hosts setup.yml --tags=setup-all,start` with `matrix_client_gomuks_enabled: true` — service starts, Traefik routes `https://gomuks.example.com/`, self-check passes (200/302)
- login flow: basic-auth → Matrix login, keys persist in `/matrix/client-gomuks/data`, backend keeps syncing after browser close
- `matrix_client_gomuks_enabled: false` → service stopped/removed, data directory preserved, re-enable restores VAPID pair
- `matrix_client_gomuks_disable_auth: true` without password — no validation error, `disable_auth_because_i_want_my_account_to_be_hacked: true` rendered

### Checklist

- [x] Role follows `docs/style-guide.md` (`matrix_client_gomuks_*` prefix, `_auto`/`_custom` where applicable, validation in `validate_config.yml`)
- [x] Every new file has SPDX headers (REUSE)
- [x] Registered in `setup.yml`, `group_vars/matrix_servers`, `docs/README.md`, `README.md`, `docs/container-images.md`, `CHANGELOG.md`

Commits on this branch:
- `Add gomuks as optional alternative web client`
- `Harden gomuks configuration and lifecycle`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# 2026-08-26
+
+## Support for gomuks
+
+The playbook can now install [gomuks](docs/configuring-playbook-client-gomuks.md), a Matrix client with a Go backend and a web frontend. The backend stores encryption keys and continues syncing while the browser is closed.
+
+gomuks is disabled by default. Enabling it requires a separate password to protect the backend. Its database and encryption keys remain on the server if the component is disabled later.
+
+
 # 2026-08-20
 
 ## MatrixRTC transports are advertised in the client well-known again
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ Web clients for Matrix that you can host on your own domains.
 | [SchildiChat Web](https://schildi.chat/) | ❌ | Based on Element Web, with a more traditional instant messaging experience | [Link](docs/configuring-playbook-client-schildichat-web.md) |
 | [FluffyChat Web](https://fluffychat.im/) | ❌ | The cutest messenger in Matrix | [Link](docs/configuring-playbook-client-fluffychat-web.md) |
 | [Commet](https://github.com/commetchat/commet) | ❌ | Matrix web client | [Link](docs/configuring-playbook-client-commet.md) |
+| [gomuks](https://github.com/gomuks/gomuks) | ❌ | Bouncer-style Matrix client with web frontend | [Link](docs/configuring-playbook-client-gomuks.md) |
 
 ### Server Components
 
--- a/docs/configuring-playbook-client-gomuks.md
+++ b/docs/configuring-playbook-client-gomuks.md
@@ -1,0 +1,133 @@
+<!--
+SPDX-FileCopyrightText: 2026 Slavi Pantaleev
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+# Setting up gomuks (optional)
+
+The playbook can install and configure [gomuks](https://github.com/gomuks/gomuks) for you.
+
+gomuks is a Matrix client written in Go. The `gomuks` container in this playbook runs the gomuks backend which serves the web frontend. It acts as a bouncer — keeping your encryption keys and syncing in the background, even when the browser is closed. See the project's [documentation](https://docs.mau.fi/gomuks/) to learn what it does and why it might be useful to you.
+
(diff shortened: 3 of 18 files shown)


mcepl:
After one fix (already pushed here), seems to work just fine.

FSG-Cat:
unless Tulir has made changes to how gomuks works since i last checked in on this there is the flaw that gomuks is a strictly single user role. But that may be considered fine. I would argue there is the bigger problem that you have to put your password in ansible vault or plain in your config. Not an access token but the real credential for your prod account.

yes i know we already have this for bots but i would argue bots are different due to being system users.

mcepl:
Actually, there is a problem, because `gomuks` is apparently only single-user, which would be probably deadly for this PR.

mcepl:
> `{me}` Are there any plans to make gomuks multiuser (I am afraid this weakness could kill my effort at https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/5574).

> `{maintainer}` no; it's a client so it has to run on the user's hardware

PR #5583, opened 2026-09-16
Synapse: bump to v1.161.0

nevart:


spantaleev:
Closed in favor of https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/5593 .

Yours was earlier, but that one has more information as to why we're not merging this yet.

